PR #1, opened 2016-09-20
Project init!

JPLachance:
This is it, our simple lib of react component! Clone it, build it, beat it, I want your feedback!


docs/Index.tsx
```diff
@@ -0,0 +1,371 @@
+import {Model as BackboneModel} from 'backbone';
```
JPLachance: Yes, that file is a big blob of code. For now, we only need a place where to put code to test our components, but we will eventually created a structured project for component doc.

docs/index.html
```diff
@@ -0,0 +1,19 @@
...
+<html lang="en">
+<head>
+    <title>React Vapor</title>
+
+    <meta charset="UTF-8">
+</head>
+<body class="coveo-styleguide">
+<div class="header flex" style="min-height: 65px;">
+    <div class="header-section h1 p2 flex flex-center">
+        Will it blend?
```
wfortin: react-vapor as a title ?

package.json
```diff
@@ -1,19 +1,79 @@
...
   "repository": {
     "type": "git",
     "url": "git+https://github.com/JPLachance/react-vapor.git"
   },
+  "scripts": {
+    "postinstall": "typings install",
+    "start": "gulp docs && webpack-dev-server --devtool eval --inline --hot --progress --colors",
+    "build": "gulp",
+    "prepublish": "npm run build",
+    "add-file": "tsconfig . -i 2"
```
wfortin: Is that needed since you added this to the gulpfile ?

tslint.json
```diff
@@ -0,0 +1,116 @@
...
+    "whitespace": [
+      true,
+      "check-branch",
+      "check-decl",
+      "check-operator",
+      "check-separator",
+      "check-type"
+    ]
+  }
+}
```
wfortin: PrettyTypescript has its own TSLint file, so maybe we shouldn't have this.

gulpfile.js
```diff
@@ -1,3 +1,137 @@
...
+});
+//</editor-fold>
+
+//<editor-fold desc="Typescript d.ts generation">
+gulp.task('internalDefs', false, () => {
+  return require('dts-generator').default({
+    name: 'ReactVapor',
+    project: './',
+    baseDir: './src/',
+    out: 'dist/ReactVapor.d.ts',
```
JPLachance: Already done, hit F5!

lib/react-tether.d.ts
```diff
@@ -0,0 +1,59 @@
...
+    style?: Object;
+    onUpdate?: Function;
+    onRepositioned?: Function;
+    children?: [React.ReactNode, boolean | React.ReactNode];
+  }
+}
+
+declare module 'react-tether' {
+  export = TetherComponent;
+}
```
JPLachance: Yes, when I'll have a little bit of time for it!

tslint.json
JPLachance: Yes, but it requires contributors to know that and setup their IDE V.S. the default tslint file. Yes it's duplicated, but I think it's a good think to simplify devs life.

wfortin: Ok so it's the same config ?

JPLachance: Yes, I copy/pasted it.

docs/Index.tsx
```diff
@@ -0,0 +1,371 @@
...
+const onClickCancel = (modelCid: string, email: string, sendEmail: boolean) => {
+  return {
+    type: actions.CancelMemberEdition,
+    modelCid,
+    email,
+    sendEmail
+  };
+};
+
+const mapDispatchToProps = (dispatch, ownProps) => {
```
sroy3: If you want to type this, ownProps has `IMemberEditViewPropsConnected` as its type and dispatch has `(action: Redux.Action) => void` as its type.

docs/index.html
JPLachance: Done!

package.json
JPLachance: Removed!

docs/Index.tsx
JPLachance: Done! Thanks ;)

PR #2, opened 2016-09-22
Unit tests FTW

JPLachance:
Add unit tests and make the coverage report work on Typescript files!

Also fix a couple of issues I found and update outdated libs!
screen shot 2016-09-22 at 5 48 49 pm


package.json
```diff
@@ -23,38 +23,59 @@
...
     "prepublish": "npm run build"
   },
   "dependencies": {
     "react-tether": "0.5.2"
   },
   "peerDependencies": {
     "tether": "^1.3.7",
-    "react": "^15.3.1",
-    "react-dom": "^15.3.1"
+    "react": "^15.3.2",
```
pochretien: remove ^

tests/components/PopoverComponent.spec.tsx
```diff
@@ -0,0 +1,161 @@
...
+          >
+            <span id='PopoverToggle'>Toggle</span>
+            {false}
+          </PopoverComponent>
+        </div>,
+        {attachTo: document.getElementById('App')}
+      );
+    });
+
+    it('should not explode', () => {
```
pochretien: why this test? its a duplicate i think

wfortin:
All good !

package.json
JPLachance: https://nodejs.org/en/blog/npm/peer-dependencies/

So we don't want to force users of our library to use a fixed version of react, we only want to warn them on the fact they need to install it and they need to have at least the version specified there.

tests/components/PopoverComponent.spec.tsx
JPLachance: Like said in the describe: Document click handler without PopoverElement in DOM

In the test before, the popover had a tether element, but here it doesn't. That's a small difference in the code that we need to cover.

src/components/PopoverComponent.tsx
```diff
@@ -8,6 +8,15 @@ export interface IPopoverComponentProps extends TetherComponent.ITetherComponent
 }
 
 export class PopoverComponent extends React.Component<IPopoverComponentProps, any> {
+  static propTypes = _.extend(TetherComponent.propTypes, {
+    children: ({children}, propName, componentName) => {
+      if (_.isUndefined(children) || React.Children.count(children) != 2) {
+        return new Error(`${componentName} expects two children to use as target and element.` +
+          `Second child can either be a boolean or a ReactNode.`);
```
GermainBergeron: No need to end your string and put a# `+` with the string literals.

tests/components/PopoverComponent.spec.tsx
```diff
@@ -0,0 +1,161 @@
...
+      toggleOpenedTetherElement: toggleOpenedTetherElementSpy
+    };
+  });
+
+  it('should render without error', () => {
+    expect(() => {
+      shallow(
+        <PopoverComponent
+          {...popoverComponentProps}
+        >
```
GermainBergeron: IMO it looks weird with the `>` alone on its line

tsconfig.test.json
```diff
@@ -0,0 +1,32 @@
...
+    "typings/globals/marionette/index.d.ts",
+    "typings/globals/react-dom/index.d.ts",
+    "typings/globals/react-redux/index.d.ts",
+    "typings/globals/react-tether/index.d.ts",
+    "typings/globals/react/index.d.ts",
+    "typings/globals/redux/index.d.ts",
+    "typings/globals/tether/index.d.ts",
+    "typings/globals/underscore/index.d.ts",
+    "typings/globals/webpack-env/index.d.ts",
+    "typings/index.d.ts",
```
GermainBergeron: I think you should just have `"typings/index.d.ts"` instead of all the files in the `typings` folder

wfortin: +1

src/components/PopoverComponent.tsx
JPLachance: OK, but the formatting kinda suck...
```
  return new Error(`${componentName} expects two children to use as target and element. Second child can either be a boolean or a ReactNode.`);
```

tsconfig.test.json
JPLachance: I did the migration to TS 2.0 in another branch, no need for that anymore :)

tests/components/PopoverComponent.spec.tsx
JPLachance: Fixed!

PR #5, opened 2016-09-26
Add codecov support

wfortin:
Coverage file must be named coverage-final.json


package.json
```diff
@@ -57,6 +58,7 @@
     "@types/webpack": "1.12.34",
     "autoprefixer": "6.4.1",
     "backbone": "1.3.3",
+    "codecov": "^1.0.1",
```
JPLachance: Simply fix the version

wfortin:
Looks like codecov.io's service is degraded right now, so we can't test for sure the output. I'll put that on hold until we can test the service.

package.json
wfortin: done

PR #11, opened 2016-09-29
Make the PopoverComponent handle it's own state instead of passing a toggle method in props

JPLachance:
+ Add the mergeProps ReduxUtils method
* Make the PopoverComponent handle it's own state instead of passing a toggle method in props
* Adjust unit test
* Adjust the demo
+ Add the missing underscore peer dependency

### DIFF
--- a/docs/Index.tsx
+++ b/docs/Index.tsx
@@ -4,11 +4,14 @@ import * as React from 'react';
 import * as _ from 'underscore';
 import * as $ from 'jquery';
 import { Model as BackboneModel, Collection as BackboneCollection } from 'backbone';
-import { MouseEventHandler, FormEvent } from 'react';
+import { FormEvent } from 'react';
 import { render as ReactDOMRender } from 'react-dom';
 import { createStore } from 'redux';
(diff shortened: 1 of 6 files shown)


docs/Index.tsx
```diff
@@ -217,19 +198,19 @@ class MemberEditView extends React.Component<IMemberEditViewProps, IMemberEditVi
...
         }]}
-        toggleOpenedTetherElement={this.props.toggleOpenedTetherElement}
         >
-        <button className='btn' onClick={() => this.props.toggleOpenedTetherElement(!this.props.isOpen)}>
+        <button className='btn'>
           {this.props.memberModel.isNew() ? 'Add member' : this.props.memberModel.email}
         </button>
         {
-          this.props.isOpen && this.getPopoverContent()
+          this.getPopoverContent()
```
sroy3: This could be all on one line

JPLachance: I simplified it a little and removed the this.getPopoverContent method.. was not really useful anyway.

PR #28, opened 2016-10-16
Fix the Svg component for Edge and IE

JPLachance:
Since innerHTML isn't supported well on SVGElement, wrap the SVG in a span.

In IE, a Polyfill is required (could have been done), but the real problem is Edge. In Edge, <path .../> self closing tags are not well interpreted and the resulting SVG is broken. See http://codepen.io/JPLachance/pen/VKBoVV?editors=1010 (in Edge) for an example.


src/components/Svg.tsx
```diff
@@ -1,46 +1,47 @@
...
-    attibutes[attr.name] = attr.value;
-  }
+const setSvgClass = (svgString: string, svgClass: string): string => {
+  let parser = document.createElement('div');
+  parser.innerHTML = svgString;
 
-  return attibutes;
+  let svgElement: SVGElement = parser.children[0] as SVGElement;
+  svgElement.setAttribute('class', svgClass);
+  return svgElement.outerHTML; // Polyfill required here!
```
GermainBergeron: Why do you put the polyfill in another file? I think it would be better to have the fix directly here

src/utils/Polyfills.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * Some Polyfills required for cross browser compatibility.
+ */
+
+// IE does not support outerHTML property on SVGElement
+// http://stackoverflow.com/questions/12592417/outerhtml-of-an-svg-element
+Object.defineProperty(SVGElement.prototype, 'outerHTML', {
```
GermainBergeron: If you choose this solution, it will always override the outerHTML function of the SvgElements

src/components/Svg.tsx
JPLachance: I removed the polyfill finally.

src/utils/Polyfills.ts
JPLachance: Removed.

PR #54, opened 2016-11-15
Navigation components

sroy3:
screen shot 2016-11-15 at 10 32 31 am


src/components/loading/LoadingReducers.ts
```diff
@@ -0,0 +1,64 @@
...
+  switch (action.type) {
+    case LoadingActions.add:
+      return [
+        ...state,
+        loadingReducer(undefined, action)
+      ];
+    case LoadingActions.remove:
+      return _.reject(state, (l) => {
+        return l.id === action.payload.ids[0];
+      });
```
GermainBergeron: what is `l`? I think we should add the type and a better variable name

sroy3: Done

src/utils/ReduxUtils.ts
```diff
@@ -38,9 +44,12 @@ export interface IReduxAction<T> extends Redux.Action {
 
 export interface IReduxActionPayload {
```
JPLachance: What is this interface used for?

src/utils/ReduxUtils.ts
```diff
@@ -38,9 +44,12 @@ export interface IReduxAction<T> extends Redux.Action {
...
   id?: string;
-  filterText?: string;
+  ids?: string[];
   facet?: string;
   facetRow?: IFacet;
+  filterText?: string;
+  pageNb?: number;
```
JPLachance: Having a global IReduxActionPayload interface is clearly not where I would like to see this project go. I'm sorry, but each action should have it own "Payload" interface, not use a global one for the entire project. IMO, the "members example" is a quite good example on how we could strongly type each actions and their payload.

src/components/navigation/perPage/NavigationPerPageReducers.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+export const perPageInitialState: IPerPageState = {
+  id: undefined,
+  perPage: 10
+};
+
+export const perPageCompositeInitialState: IPerPageState[] = [];
+
+export const perPageReducer = (state: IPerPageState = perPageInitialState, action: IReduxAction<IReduxActionPayload>): IPerPageState => {
```
JPLachance: See https://github.com/coveo/react-vapor/blob/master/docs/members-example/actions/MemberEditionActions.ts#L27, the IMemberEditionActionsPayloads interface for an example of how I would handle this.

src/utils/ReduxUtils.ts
sroy3: Actually, reducers listen to all actions. I could for example listen to a loading event in my pagination reducer. Same goes for dispatch actions in mapDispatchToProps. Each action still does have its own payload.

docs/Reducers.ts
```diff
@@ -1,17 +1,29 @@
...
+import { loadingsReducer } from '../src/components/loading/LoadingReducers';
+import * as Redux from 'redux';
 
 export interface IReactVaporExampleState extends IReactVaporState {
   membersCompositeState: IMembersCompositeState;
+  lastAction: Redux.Action;
 }
 
-export const Reducers: Reducer<IReactVaporExampleState> = combineReducers<IReactVaporExampleState>({
+const lastAction = (state: IReactVaporExampleState = null, action: Redux.Action): Redux.Action => {
```
sroy3: In the navigation connected example, when we change page or change the per page number, it triggers the loading. I added a simple reducer to check on the last action and dispatch a removeLoading action after a few seconds when the last action is addLoading. It is only used in the context of the example and is not being used in the components themselves.

src/utils/ReduxUtils.ts
sroy3: It is being used to access all possible payload properties in reducers and mapDispatchToProps.

docs/Reducers.ts
JPLachance: OKay!

src/utils/ReduxUtils.ts
JPLachance: 👍

src/components/navigation/perPage/NavigationPerPageReducers.ts
JPLachance: If this interface is necessary for UTs, why not simply put it in UTs where you need it? For me the ReduxUtils.ts file is intended to regroup base interfaces/functions useful to connect components and use them.

sroy3: It is not simply for the UTs. It is being used by reducers and mapDispatchToProps as well.

sroy3: I moved the interfaces that were React Vapor specific out of ReduxUtils.

src/ReactVapor.ts
```diff
@@ -0,0 +1,26 @@
...
+export interface IReactVaporState {
+  lastUpdatedComposite?: ILastUpdatedState[];
+  facets?: IFacetState[];
+  filters?: IFilterState[];
+  perPageComposite?: IPerPageState[];
+  paginationComposite?: IPaginationState[];
+  loadings?: ILoadingState[];
+}
+
+export interface IReduxActionPayload {
```
JPLachance: I would also rename for IReactVaporActionsPayloads.

JPLachance: Yes no ? :)

sroy3: Oops sorry didn't see this. I think a mix of both would be better. How about `IReactVaporActionsPayload`? It covers all actions' payload, but it still is one payload. That is how I see it at least. I will change it in the next PR (nothing has been published yet, so it won't be a breaking change).

PR #56, opened 2016-11-23
Table components

sroy3:


docs/ReactVaporStore.ts
```diff
@@ -1,5 +1,5 @@
-import { createStore, Store } from 'redux';
+import * as Redux from 'redux';
```
JPLachance: Something was broken? Since we don't generate a d.ts file for the Doc I don't think this change was required :)

karma.entry.ts
```diff
@@ -3,7 +3,11 @@ import * as s from 'underscore.string';
 
 beforeEach(() => {
   if (!$('#App').length) {
-    $('body').append('<div id="App" class="coveo-styleguide"></div>');
+    $('body').append('<div id="App" class="coveo-styleguide">' +
+      '<table id="AppTable">' +
```
JPLachance: I don't think this should be in the karma.entry file. It should be append where required instead.

docs/ReactVaporStore.ts
sroy3: It actually is in the d.ts

karma.entry.ts
sroy3: You are right. Done.

docs/ReactVaporStore.ts
sroy3: Yes, I'll look into.

sroy3: I'm pretty sure it would work the way it was with a simple npm install, but when linking the module it doesn't. I think it is best the leave it as is (`import * as Redux`).

PR #60, opened 2016-12-12
Table sort

sbelzile-Coveo:
2016-12-12_10h44_17

### DIFF
--- a/src/components/tables/TableHeaderCell.tsx
+++ b/src/components/tables/TableHeaderCell.tsx
@@ -1,19 +1,11 @@
 import * as React from 'react';
 
-export interface ITableHeaderCellOwnProps extends React.ClassAttributes<TableHeaderCell> {
+export interface ITableHeaderCellProps extends React.ClassAttributes<TableHeaderCell> {
   title: string;
   className?: string;
-}
-
-export interface ITableHeaderStateProps {
   cellContent?: JSX.Element;
-}
-
-export interface ITableHeaderDispatchProps {
   onClick?: () => void;
 }
-
-export interface ITableHeaderCellProps extends ITableHeaderCellOwnProps, ITableHeaderStateProps, ITableHeaderDispatchProps { }
 
 export class TableHeaderCell extends React.Component<ITableHeaderCellProps, any> {
 
--- a/src/components/tables/TableSort.tsx
+++ b/src/components/tables/TableSort.tsx
@@ -11,10 +11,10 @@ export enum SortTypes {
 
 export interface ITableSortProps extends ITableHeaderCellProps {
   sortType: SortTypes;
-  isFixedHeader: boolean;
+  isFixedHeader?: boolean;
 }
 
-export abstract class TableSort implements ITableHeaderCellProps {
+export class TableSort implements ITableHeaderCellProps {
   title: string;
   className: string;
   cellContent: JSX.Element;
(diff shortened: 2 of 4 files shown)


src/components/tables/TableSort.tsx
```diff
@@ -0,0 +1,58 @@
...
+  Descending,
+  None
+}
+
+export interface ITableSortProps extends ITableHeaderCellProps {
+  sortType: SortTypes;
+  isFixedHeader: boolean;
+}
+
+export abstract class TableSort implements ITableHeaderCellProps {
```
sroy3: Why not create a React component with a render method instead?

src/components/tables/examples/TableHeaderExamples.tsx
```diff
@@ -12,9 +14,7 @@ export class TableHeaderExamples extends React.Component<any, any> {
       {
         title: 'Second col'
       },
-      {
-        title: 'Third col'
-      },
+      new SortExample(SortTypes.None, () => { }),
```
sroy3: I'd leave the example here just so we can easily compare a column with sorting from the other ones.

src/components/tables/TableSort.tsx
```diff
@@ -0,0 +1,58 @@
+// tslint:disable-next-line:no-unused-variable
```
sbelzile-Coveo: ??? Le comment c'est un `disable-next-line`.

sroy3: Oups, sorry. This works better than what we have then, cool.

src/components/tables/TableSort.tsx
sbelzile-Coveo: Mostly because I did not want to break the current framework: the class `TableHeader` takes an array of `ITableHeaderCellProps` and instantiates a cell with every item in this array.

I had 2 options: modify the `ITableHeaderProps` so that `columns` is now an array of TableHeaderCell or creating a class that computes the properties of the cells. 
The sort component content different when the table is fixed or not, the class of the cell is different if the sort type changes, meaning I could define my cells as follow:

    title: 'Name',
    className: 'admin-sort admin-sort-ascending', // need to know all class names or code something to convert it
    cellContent: <TableSort/>

or like this:

    new NameSort(SortTypes.Ascending);

The implementation on my side has some classes `SelectedSort`, `FieldNameSort`, with just a constructor which takes a `SortType` + onClick function as input and fixes the `title` by itself, `className` is computed by calling the base class.

Looks quite clean to me. Anything better to suggest?

src/components/tables/TableSort.tsx
```diff
@@ -0,0 +1,57 @@
...
+export class TableSort implements ITableHeaderCellProps {
+  title: string;
+  className: string;
+  cellContent: JSX.Element;
+  onClick: () => void;
+
+  constructor(props: ITableSortProps) {
+    this.title = props.title;
+    this.onClick = props.onClick;
+    this.className = props.className ? props.className + ' ' + this.computeClassName(props.sortType)
```
sroy3: We should use an array for the classes instead.

src/components/tables/TableHeaderCell.tsx
```diff
@@ -3,13 +3,24 @@ import * as React from 'react';
...
+      this.props.onClick();
+    }
+  }
+
   render() {
     return (
-      <th className={this.props.className || ''}>{this.props.title}</th>
+      <th className={this.props.className || ''} onClick={() => this.handleClick()}>
+        {this.props.title}
+        {this.props.cellContent}
```
sroy3: We can use `this.props.children` instead of {this.props.cellContent} and use it this way:
`<TableHeaderCell>{column.cellContent}</TableHeaderCell>`

src/components/tables/TableHeaderCell.tsx
```diff
@@ -3,13 +3,24 @@ import * as React from 'react';
...
+  private handleClick = () => {
+    if (this.props.onClick) {
+      this.props.onClick();
+    }
+  }
+
   render() {
     return (
-      <th className={this.props.className || ''}>{this.props.title}</th>
+      <th className={this.props.className || ''} onClick={() => this.handleClick()}>
```
sroy3: You can use default props to set className to the empty string if it is not set
```
static defaultProps: Partial< ITableHeaderCellProps > = {
    className: '',
  };
```

GermainBergeron:
Is this still needed? @sbelzile-Coveo

sbelzile-Coveo:
@GermainBergeron: I do not think of this as a good approach anymore. I will close it.

PR #61, opened 2017-01-11
+ Added a check to see if the tooltip element is still there before d…

sroy3:
…estroying it

### DIFF
--- a/src/components/tooltip/Tooltip.spec.tsx
+++ b/src/components/tooltip/Tooltip.spec.tsx
@@ -1,6 +1,6 @@
 import { shallow, mount, ReactWrapper } from 'enzyme';
 import { ITooltipProps, Tooltip } from './Tooltip';
-
+import * as $ from 'jquery';
 // Until Webpack provided plugins works with TS 2.0
 /* tslint:disable:no-unused-variable */
 import * as React from 'react';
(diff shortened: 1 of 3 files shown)


src/components/tooltip/Tooltip.spec.tsx
```diff
@@ -40,4 +40,36 @@ describe('<Tooltip>', () => {
...
+
+  it('should unmount without error when tooltip element is not present in the DOM', () => {
+    tooltipWrapper = mount(
+      <Tooltip {...tooltipProps}>
+        Hover me!
+      </Tooltip>,
+      { attachTo: document.getElementById('App') }
+    );
+
+    let tooltip = (tooltipWrapper.instance()as Tooltip)['tooltip'];
```
GermainBergeron: Is the formatter removing the space before the `as` ?

src/components/tooltip/Tooltip.spec.tsx
```diff
@@ -40,4 +40,36 @@ describe('<Tooltip>', () => {
...
+    }).not.toThrow();
+
+
+    tooltipWrapper = mount(
+      <Tooltip {...tooltipProps}>
+        Hover me!
+      </Tooltip>,
+      { attachTo: document.getElementById('App') }
+    );
+    (tooltipWrapper.instance()as Tooltip)['tooltip'] = null;
```
GermainBergeron: same for the space here

PR #62, opened 2017-01-16
+ add UserFeedback component with tests and doc

MichelML:


src/components/userFeedback/UserFeedback.tsx
```diff
@@ -0,0 +1,63 @@
...
+  constantClasses?: string;
+  displayOnShow?: string;
+}
+
+export interface IUserFeedbackStyle {
+  display: string;
+  className: string;
+}
+
+export const userFeedbackState = {
```
sroy3: Since this is an enum constant, we would capitalize it `UserFeedbackState`

src/components/userFeedback/UserFeedback.tsx
```diff
@@ -0,0 +1,63 @@
...
+  WARNING: 'WARNING',
+  ERROR: 'ERROR',
+};
+
+export class UserFeedback extends React.Component<IUserFeedbackProps, any> {
+
+  private getDefaultUserFeedbackStyle(): IUserFeedbackStyle {
+    return {
+      display: 'none',
+      className: this.props.constantClasses ? this.props.constantClasses + ' text-dark-grey' : 'text-dark-grey'
```
sroy3: Since `text-dark-grey` is used both in the if and the else, I would write it: `'text-dark-grey ' + (this.props.constantClasses ? this.props.constantClasses : '')`

src/components/userFeedback/UserFeedback.tsx
```diff
@@ -0,0 +1,63 @@
...
+  VALID: 'VALID',
+  WARNING: 'WARNING',
+  ERROR: 'ERROR',
+};
+
+export class UserFeedback extends React.Component<IUserFeedbackProps, any> {
+
+  private getDefaultUserFeedbackStyle(): IUserFeedbackStyle {
+    return {
+      display: 'none',
```
sroy3: I would use a class to toggle the display instead of using inline styles.

src/components/userFeedback/UserFeedback.tsx
```diff
@@ -0,0 +1,63 @@
...
+      case userFeedbackState.VALID:
+        return this.getDefaultUserFeedbackStyle();
+
+      case userFeedbackState.WARNING:
+        return extend(this.getDefaultUserFeedbackStyle(), { display: this.props.displayOnShow || 'block' });
+
+      case userFeedbackState.ERROR:
+        let newDisplayAndClassName: IUserFeedbackStyle = {
+          display: this.props.displayOnShow || 'block',
+          className: this.props.constantClasses ? this.props.constantClasses + ' text-red' : 'text-red'
```
sroy3: See comment above about `text-dark-grey`

src/components/userFeedback/UserFeedback.spec.tsx
```diff
@@ -0,0 +1,147 @@
...
+        expect(() => getShallowOutput(generateProps('hello', userFeedbackState.ERROR, 'm1', 'block'))).not.toThrow();
+      });
+    });
+
+    describe('text content', () => {
+      it('should contain the exact text passed to prop feedbackText', () => {
+        let testText = 'testing';
+        let wrongText: string;
+
+        expect(getShallowOutput(generateProps(testText, '', '')).text() === testText).toBe(true);
```
sroy3: You can also write `expect(getShallowOutput(generateProps(testText, '', '')).text()).toBe(testText);` which is a little more explicit.

Same comment for the other values expected below.

sroy3:
If you have a screenshot of your examples, you can post it here.

src/components/userFeedback/UserFeedback.tsx
MichelML: You mean like toggling a 'hidden' class like the one we already have? The advantage I see with inline style for the display is that you can choose which type of display you want (inline-block, block, etc.). If you use a class, you are either forced to choose one display type for your class, or you'll have to create a class for each display type. In one way I see it as a bit limiting, and on the other as a bit redundant. What are your thoughts?

docs/Index.tsx
```diff
@@ -89,6 +90,44 @@ class App extends React.Component<any, any> {
               </Tooltip>
             </div>
           </div>
+          <div className='form-group'>
```
MichelML: done! thanks steph!

src/components/userFeedback/UserFeedback.tsx
```diff
@@ -0,0 +1,63 @@
+/// <reference path="../../../types/react-tether/index.d.ts" /> Required to make dts-generator bundle react-tether definition file.
```
MichelML: done! thanks steph!

src/components/userFeedback/UserFeedback.tsx
MichelML: done! thanks steph!

src/components/userFeedback/UserFeedback.tsx
MichelML: done! thanks steph!

src/components/userFeedback/UserFeedback.tsx
MichelML: done! thanks steph!

src/components/userFeedback/UserFeedback.spec.tsx
MichelML: done! thanks steph!

src/components/userFeedback/UserFeedback.tsx
MichelML: I'll wait for your feedback before making change to this one

sroy3: If many display style are going to be used, I do see the advantage of using inline styles. I guess I'll have to see an example to understand why we want a component to be either display block or inline-block.

MichelML: screen shot 2017-01-16 at 3 18 58 pm

MichelML: userfeedback 2 is display:block, userfeedback 3 is display:inline-block

it made sens to me because I was not using tether, and I did not because I do not think it is possible in the context I am using the UserFeedback

MichelML: Does it make sens or I'm missing something? That could definitely be possible haha

sroy3: I think I'd rather have classes passed as props depending on what we want. Then we can style both differently if we want to. There should already be styles for that in the vapor repository. If not, you can add them.

MichelML: oh lord, display classes are already in vapor, my bad steph :), should have looked closer, will make the changes

sroy3: If we didn't already have vapor (and shared style between react and non-react components), your solution would have been nice though.

MichelML: thanks for the amazing feedback steph! just made the change

src/components/userFeedback/UserFeedback.spec.tsx
```diff
@@ -0,0 +1,143 @@
+import { shallow, ShallowWrapper } from 'enzyme';
```
JPLachance: Is there any reason why we would only use the shallow rendering here? In every other tests we are using the ReactWrapper and the mount function. Maybe we are doing it wrong! 😄

src/components/userFeedback/UserFeedback.spec.tsx
```diff
@@ -0,0 +1,143 @@
...
+        });
+
+        it('the display value should be the value of the displayOnShow prop', () => {
+          expect(getShallowOutput(generateProps('', UserFeedbackState.WARNING, 'inline-block')).hasClass('inline-block')).toBe(true);
+          expect(getShallowOutput(generateProps('', UserFeedbackState.WARNING, 'inline-block')).hasClass('hidden')).toBe(false);
+        });
+
+        it('the text should be dark-grey', () => {
+          let componentOnStateWarning = getShallowOutput(generateProps('', UserFeedbackState.WARNING, ''));
+          expect(componentOnStateWarning.hasClass(defaultTextColorClass));
```
JPLachance: Why not simply use TextColorClasses.default here instead of defaultTextColorClass? Same comment below.

src/components/userFeedback/UserFeedback.spec.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+describe('<UserFeedback>', () => {
+  const defaultTextColorClass: string = TextColorClasses.default;
+  const errorTextColorClass: string = TextColorClasses.error;
+
+  const generateProps = (feedbackText: string, state: string, extraClasses?: string, displayOnShow?: string): IUserFeedbackProps => {
+    return displayOnShow ? { feedbackText, state, extraClasses, displayOnShow } : { feedbackText, state, extraClasses };
+  };
+
+  const defaultProps: IUserFeedbackProps = generateProps('', '', '');
```
JPLachance: This const is only used once. I would simply call generateProps() like you do in other tests cases.

src/components/userFeedback/UserFeedback.spec.tsx
```diff
@@ -0,0 +1,143 @@
...
+  const defaultPropsWithDisplay: IUserFeedbackProps = generateProps('', '', '', 'block');
```
JPLachance: Same here.

src/components/userFeedback/UserFeedback.spec.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+    describe('text content', () => {
+      it('should contain the exact text passed to prop feedbackText', () => {
+        let testText = 'testing';
+        let wrongText: string;
+
+        expect(getShallowOutput(generateProps(testText, '', '')).text()).toBe(testText);
+
+        testText = 'hello this is a longer test string over here';
+        expect(getShallowOutput(generateProps(testText, '', '')).text()).toBe(testText);
```
JPLachance: It's kind of the exact same expect clause as the one above 😸

src/components/userFeedback/UserFeedback.spec.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+
+        testText = 'hello this is a test over here';
+        wrongText = 'hello this is the wrong text over here';
+        expect(getShallowOutput(generateProps(testText, '', '')).text()).not.toBe(wrongText);
```
JPLachance: Having a test case for the `toBe` then `not.toBe` is not very useful IMO.

src/components/userFeedback/UserFeedback.spec.tsx
```diff
@@ -0,0 +1,143 @@
...
+        });
+      });
+
+      describe('VALID state', () => {
+        it('should be invisible on state VALID without displayOnShow prop', () => {
+          expect(getShallowOutput(generateProps('', UserFeedbackState.VALID, '')).hasClass('hidden')).toBe(true);
+        });
+
+        it('should be invisible on state VALID, even with prop displayOnShow provided', () => {
+          expect(getShallowOutput(generateProps('', UserFeedbackState.VALID, '')).hasClass('hidden')).toBe(true);
```
JPLachance: _should be invisible on state VALID without displayOnShow prop_ and _should be invisible on state VALID, even with prop displayOnShow provided_ are doing the exact same thing. Are you sure you second test is valid?

src/components/userFeedback/UserFeedback.spec.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+      describe('WARNING state', () => {
+        it('should be visible (without prop displayOnShow)', () => {
+          expect(getShallowOutput(generateProps('', UserFeedbackState.WARNING, '')).hasClass('block')).toBe(true);
+          expect(getShallowOutput(generateProps('', UserFeedbackState.WARNING, '')).hasClass('hidden')).toBe(false);
+        });
+
+        it('the display value should be the value of the displayOnShow prop', () => {
+          expect(getShallowOutput(generateProps('', UserFeedbackState.WARNING, 'inline-block')).hasClass('inline-block')).toBe(true);
+          expect(getShallowOutput(generateProps('', UserFeedbackState.WARNING, 'inline-block')).hasClass('hidden')).toBe(false);
```
JPLachance: What if i'm stupid and I pass the class 'hidden' in the displayOnShow prop?

src/components/userFeedback/UserFeedback.spec.tsx
```diff
@@ -0,0 +1,143 @@
...
+        });
+      });
+
+      describe('WARNING state', () => {
+        it('should be visible (without prop displayOnShow)', () => {
+          expect(getShallowOutput(generateProps('', UserFeedbackState.WARNING, '')).hasClass('block')).toBe(true);
+          expect(getShallowOutput(generateProps('', UserFeedbackState.WARNING, '')).hasClass('hidden')).toBe(false);
+        });
+
+        it('the display value should be the value of the displayOnShow prop', () => {
```
JPLachance: We use to prefer when all tests sentences are starting with a "should", like `it('should set the displayOnShow class', ...)`.

src/components/userFeedback/UserFeedback.tsx
```diff
@@ -0,0 +1,68 @@
...
+      textColorClass: TextColorClasses.default,
+      extraClasses: this.props.extraClasses ? ' ' + this.props.extraClasses : ''
+    };
+  }
+
+  private adjustStyle(state: string): IUserFeedbackStyle {
+    switch (state) {
+      case UserFeedbackState.VALID:
+        return this.getDefaultUserFeedbackStyle();
+
```
JPLachance: `break;` statement is missing here.

src/components/userFeedback/UserFeedback.tsx
```diff
@@ -0,0 +1,68 @@
...
+      case UserFeedbackState.WARNING:
+        return extend(this.getDefaultUserFeedbackStyle(), { displayClass: this.props.displayOnShow || 'block' });
+
```
JPLachance: `break;` statement is missing here and below!

src/components/userFeedback/UserFeedback.tsx
```diff
@@ -0,0 +1,68 @@
+import * as React from 'react';
+import { extend } from 'underscore';
+
+export interface IUserFeedbackProps {
+  feedbackText: string;
+  state: string;
+  extraClasses?: string;
+  displayOnShow?: string;
```
JPLachance: displayOnShow vs displayClass, what's the difference? I think displayOnShow should be renamed for displayClass to fit with the IUserFeedbackStyle interface.

src/components/userFeedback/UserFeedback.tsx
```diff
@@ -0,0 +1,68 @@
...
+      default:
+        return this.getDefaultUserFeedbackStyle();
+    }
+  }
+
+  render() {
+    let style = this.adjustStyle(this.props.state);
+    return (
+      <div
+        className={`${style.textColorClass} ${style.displayClass}${style.extraClasses}`}>
```
JPLachance: I think it could be on the same line than the opening`<div` tag.

src/components/userFeedback/UserFeedbackExample.tsx
```diff
@@ -0,0 +1,47 @@
+import * as React from 'react';
+import { UserFeedback } from './UserFeedback';
+
+export class UserFeedbackExample extends React.Component<any, any> {
+  render() {
+    return (
```
JPLachance: @sroy3 Why is the coverage calculated on the example file?

src/components/userFeedback/UserFeedback.tsx
```diff
@@ -0,0 +1,68 @@
...
+        };
+
+        return extend(this.getDefaultUserFeedbackStyle(), newDisplayAndClassName);
+
+      default:
+        return this.getDefaultUserFeedbackStyle();
+    }
+  }
+
+  render() {
```
JPLachance: I would prefer to have public methods declared before private ones :)

src/components/userFeedback/UserFeedback.spec.tsx
MichelML: Both shallow or mount would be fine in this scenario. Shallow tests components 'as unit', and [its methods aren't indirectly asserting on behavior of child components](http://airbnb.io/enzyme/docs/api/shallow.html). Since there is only one html element in the component, it seems to me that both could have been utilized here.

MichelML: With multiple html elements nested inside a component, I think mount should be preferred most of the time, but that would still depends on your test case.

src/components/userFeedback/UserFeedback.spec.tsx
MichelML: Thanks JP. The negative test is to make sure that the positive tests were not false positive, would that make sense?

src/components/userFeedback/UserFeedback.spec.tsx
MichelML: Thanks JP. fixed!

src/components/userFeedback/UserFeedback.spec.tsx
MichelML: Thanks JP. fixed!

src/components/userFeedback/UserFeedback.spec.tsx
MichelML: Thanks JP. fixed! the displayOnShow prop has been added for the second test

src/components/userFeedback/UserFeedback.spec.tsx
MichelML: Thanks JP. fixed! the displayOnShow prop has been added for the second test

src/components/userFeedback/UserFeedback.spec.tsx
MichelML: Thanks JP. fixed!

src/components/userFeedback/UserFeedback.spec.tsx
MichelML: I think it should behave naturally and hide the class, we could add a test for that. Otherwise, could we throw an error with TypeScript at compile time in case the class is set as hidden, or is simply not a displayClass?

src/components/userFeedback/UserFeedback.spec.tsx
MichelML: Thanks JP. fixed!

src/components/userFeedback/UserFeedback.tsx
MichelML: I disagree. The displayOnShow is the display you want when the UserFeedback is visible, not always, whereas the displayClass is the final class that is added to the html element (which can be 'hidden' if the feedback state is VALID, or the value set on displayOnShow if the feedback state is WARNING or ERROR.

src/components/userFeedback/UserFeedback.tsx
MichelML: Break is not needed. Since the function only contains a switch statement and does not need to continue to evaluate the function after the switch statement, break is not needed here. The function returns as soon as it finds a case match.

src/components/userFeedback/UserFeedback.tsx
MichelML: Thanks JP. fixed!

src/components/userFeedback/UserFeedback.tsx
MichelML: Is it a team rule or only a personal preference? If a personal preference, would you like to make it a team rule?

src/components/userFeedback/UserFeedback.tsx
sroy3: break is not needed when we return something

src/components/userFeedback/UserFeedbackExample.tsx
sroy3: I think it is because it is not in a `example` folder

MichelML: just moved the example in a examples folder

src/components/userFeedback/UserFeedback.spec.tsx
JPLachance: Those tests are kind of obvious no? We use to make tests simple and make sure each code path are covered at least once and that side effects/results of those code paths are correct. Here you already covered that path and you already made sure that the result was correct. Therefore, do you think we should test it twice? :)

src/components/userFeedback/UserFeedback.tsx
JPLachance: Okay then!

src/components/userFeedback/UserFeedback.tsx
JPLachance: Sorry, I was sleeping...

src/components/userFeedback/UserFeedback.tsx
JPLachance: Sorry :P

src/components/userFeedback/UserFeedback.tsx
JPLachance: We are mostly following other languages ordering rules like Java or C# (http://stackoverflow.com/questions/150479/order-of-items-in-classes-fields-properties-constructors-methods), but we are sometime moving private methods near where they are used (http://stackoverflow.com/questions/4668218/are-there-any-java-method-ordering-conventions). In this case, clearly the render method would be the first one in the class.

src/components/userFeedback/UserFeedback.spec.tsx
JPLachance: I would simply add a test case.

src/components/userFeedback/UserFeedback.spec.tsx
MichelML: Yes? just kidding :D

src/components/userFeedback/UserFeedback.tsx
MichelML: Ok thanks for the additional details! Just made the change!

src/components/userFeedback/UserFeedback.spec.tsx
MichelML: a test case for what exactly? I'm not sure I understand what test case you would like :S

src/components/userFeedback/UserFeedback.spec.tsx
MichelML: Just removed it!

src/components/userFeedback/UserFeedback.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+export interface IUserFeedbackProps {
+  feedbackText: string;
+  state: string;
+  extraClasses?: string;
+  displayOnShow?: string;
+}
+
+export interface IUserFeedbackStyle {
+  displayClass: string;
```
GermainBergeron: `displayClass` is required but there's a default value in `getDefaultUserFeedbackStyle` ?

Is there a way that we could merge extraClasses and displayClass? It looks like the same thing to me

src/components/userFeedback/UserFeedback.spec.tsx
```diff
@@ -0,0 +1,134 @@
...
+/* tslint:enable:no-unused-variable */
+
+describe('<UserFeedback>', () => {
+  const generateProps = (feedbackText: string, state: string, extraClasses?: string, displayOnShow?: string): IUserFeedbackProps => {
+    return displayOnShow ? { feedbackText, state, extraClasses, displayOnShow } : { feedbackText, state, extraClasses };
+  };
+
+  const getShallowOutput = (props: IUserFeedbackProps): ShallowWrapper<IUserFeedbackProps, any> => {
+    return shallow(<UserFeedback {...props} />);
+  };
```
MichelML: thanks Germain! Done!

src/components/userFeedback/UserFeedback.spec.tsx
```diff
@@ -0,0 +1,134 @@
...
+      });
+
+      describe('WARNING state', () => {
+        it('should be visible (without prop displayOnShow)', () => {
+          expect(getShallowOutput(generateProps('', UserFeedbackState.WARNING, '')).hasClass('block')).toBe(true);
+          expect(getShallowOutput(generateProps('', UserFeedbackState.WARNING, '')).hasClass('hidden')).toBe(false);
+        });
+
+        it('should have a display class equal to the value of the displayOnShow prop', () => {
+          expect(getShallowOutput(generateProps('', UserFeedbackState.WARNING, 'inline-block')).hasClass('inline-block')).toBe(true);
```
MichelML: Hey Germain, I don't know if you meant like real enums like specified in the TS docs, but I could not comprehend how it would apply to our context. I think string literal types https://www.typescriptlang.org/docs/handbook/advanced-types.html would have been great but we can not use them in TS 1.7 (as JP pointed out to me). That being said I know Steph refered to enum as something like 

```
export const DisplayClass = {
  BLOCK: 'block',
  INLINE: 'inline',
  INLINE_BLOCK: 'inline-block',
  HIDDEN: 'hidden',
  TABLE: 'table',
  TABLE_CELL: 'table-cell'
};
``` 

which I implemented. Does that represent added value to you? I've put it in a new file called ComponentUtils.ts under the utils folder, all components can use this enum now. In jsadmin-v2 you would then do ReactVapor.DisplayClass.YOURCHOSENDISPLAYCLASS

src/components/userFeedback/UserFeedback.spec.tsx
```diff
@@ -0,0 +1,134 @@
...
+      });
+
+      it('should contain all classes passed through extraClasses, along with the text-[color] and display classes', () => {
+        let extraClass: string[] = ['onlyoneclass'];
+        let extraClasses: string[] = ['each', 'word', 'represent', 'a', 'class'];
+        let extraClassesRealLife: string[] = ['mt1', 'mb2'];
+        let testComponent: ShallowWrapper<IUserFeedbackProps, any> = getShallowOutput(generateProps('', '', extraClass.join(' ')));
+
+        const areClassesAddedToElement = (extraClasses: string[], shallowWrapper: ShallowWrapper<IUserFeedbackProps, any>): boolean => {
+          return extraClasses.reduce((hasClass: boolean, className: string) => !hasClass ? false : shallowWrapper.hasClass(className), true);
```
MichelML: Thanks Germain, done!

src/components/userFeedback/UserFeedback.tsx
MichelML: I'm not sure of what you mean here. IUserFeedbackStyle is the final style object that is passed to the view on render. It seems that it should be required to me, it is on hidden by default, because by default the state is valid. The displayClass than change if the state becomes WARNING or ERROR.  Does that make sens? I've changed the `getDefaultUserFeedbackStyle`to `getUserFeedbackStyleOnStateValid` in case that would be more explicit and/or comprehensible...hope it is :S

src/components/userFeedback/UserFeedback.tsx
```diff
@@ -0,0 +1,66 @@
...
+export const TextColorClasses = {
+  default: 'text-dark-grey',
+  error: 'text-red'
+};
+
+export class UserFeedback extends React.Component<IUserFeedbackProps, any> {
+  render() {
+    let style = this.adjustStyle(this.props.state);
+    return (
+      <div className={`${style.textColorClass} ${style.displayClass}${style.extraClasses}`}>
```
MichelML: It was not a mistake but I refactored these lines and some others to add one. I agree that it is less confusing. Thanks so much Germain.

src/components/userFeedback/UserFeedback.spec.tsx
GermainBergeron: Yes, that's exactly what I was hoping for!

We usually refer to those kind of dictionary when we use the term "enum"

PR #63, opened 2017-01-16
Adds possibility to add children in action bar

sbelzile-Coveo:
It was impossible to add a filterBox in the action bar.


src/components/actions/ActionBar.tsx
```diff
@@ -72,6 +72,7 @@ export class ActionBar extends React.Component<IActionBarProps, any> {
           {secondaryActionsView}
           {this.props.prompt}
         </div>
+        {this.props.children}
```
JPLachance: And it's working when children is undefined?

sbelzile-Coveo: Of course? You want an additional unit test?

PR #67, opened 2017-01-24
+ Added a table empty row component

sroy3:
screen shot 2017-01-24 at 8 20 32 am


sroy3:
@GermainBergeron detecting the colspan automatically would mean that the component knows of other components and we would need to pass it more props for it to calculate it.

PR #69, opened 2017-02-09
Date picker

sroy3:


src/components/datePicker/DatePickerReducers.ts
```diff
@@ -0,0 +1,111 @@
...
+  appliedUpperLimit: Date;
+}
+
+export const datePickerInitialState: IDatePickerState = {
+  id: undefined,
+  calendarId: undefined,
+  color: undefined,
+  isRange: false,
+  lowerLimit: new Date(new Date().setHours(0, 0, 0, 0)),
+  upperLimit: new Date(new Date().setHours(23, 59, 59, 999)),
```
GermainBergeron: I think the moment functions are easier to understand: `moment().startOf('day').toDate()` & `moment().endOf('day').toDate()`

src/components/datePicker/DatePickerReducers.ts
```diff
@@ -0,0 +1,111 @@
...
+  selected: '',
+  appliedLowerLimit: new Date(new Date().setHours(0, 0, 0, 0)),
+  appliedUpperLimit: new Date(new Date().setHours(23, 59, 59, 999))
```
GermainBergeron: same

src/components/datePicker/examples/DatePickerDropdownConnectedExamples.tsx
```diff
@@ -0,0 +1,53 @@
...
+            value: () => DateUtils.getRelativeDate(new Date(), { minutes: -30 }).toString() + DATES_SEPARATOR + new Date().toString()
+          },
+          {
+            label: 'Last hour',
+            value: () => DateUtils.getRelativeDate(new Date(), { hours: -1 }).toString() + DATES_SEPARATOR + new Date().toString()
+          },
+          {
+            label: 'Last day',
+            value: () => DateUtils.getRelativeDate(new Date(), { date: -1 }).toString() + DATES_SEPARATOR + new Date().toString()
+          }
```
GermainBergeron: Are those quick options always the same? You could extract them in a utils function maybe

src/utils/DateUtils.ts
```diff
@@ -0,0 +1,116 @@
...
+  static currentYear: number = DateUtils.currentDate.getFullYear();
+
+  static getPreviousYears(numberOfYears: number): string[] {
+    let years: string[] = [];
+
+    for (let year = DateUtils.currentYear - 1; year >= DateUtils.currentYear - numberOfYears; year--) {
+      years.push(year.toString());
+    }
+
+    return years.reverse();
```
GermainBergeron: What do you think about:
```
static getNextYears(numberOfYears: number): string[] {
  return _.range(DateUtils.currentYear, DateUtils.currentYear + numberOfYears).map(String);
}

static getPreviousYears(numberOfYears: number): string[] {
  return _.range(DateUtils.currentYear - numberOfYears, DateUtils.currentYear).map(String);
}
```

src/utils/DateUtils.ts
```diff
@@ -0,0 +1,116 @@
...
+    let year: number = startingDate.getFullYear() + dateModifiers.year;
+    let month: number = startingDate.getMonth() + dateModifiers.month;
+    let date: number = startingDate.getDate() + dateModifiers.date;
+    let hours: number = startingDate.getHours() + dateModifiers.hours;
+    let minutes: number = startingDate.getMinutes() + dateModifiers.minutes;
+    let seconds: number = startingDate.getSeconds() + dateModifiers.seconds;
+    let milliseconds: number = startingDate.getMilliseconds() + dateModifiers.milliseconds;
+
+    return new Date(year, month, date, hours, minutes, seconds, milliseconds);
+  }
```
GermainBergeron: In UA I uses moment().subtract(X, 'day|month|year|...') instead, what do you prefer?

http://momentjs.com/docs/#/manipulating/subtract/
http://momentjs.com/docs/#/manipulating/add/

src/components/calendar/tests/Calendar.spec.tsx
```diff
@@ -0,0 +1,330 @@
+import { shallow, mount, ReactWrapper } from 'enzyme';
+import { Calendar, ICalendarProps, DEFAULT_DAYS, DEFAULT_MONTHS, DEFAULT_YEARS } from '../Calendar';
+import { OptionsCycle, IOptionsCycleProps } from '../../optionsCycle/OptionsCycle';
+import { DateUtils } from '../../../utils/DateUtils';
+import { ICalendarDayProps, CalendarDay, IDay } from '../CalendarDay';
+import { IDatePickerState } from '../../datePicker/DatePickerReducers';
+import * as _ from 'underscore';
+import * as moment from 'moment';
+/* tslint:disable:no-unused-variable */
```
sroy3: Yes, done.

src/components/datePicker/DatePicker.tsx
```diff
@@ -0,0 +1,86 @@
...
+  withTime?: boolean;
+  hasSetToNowButton?: boolean;
+  upperLimit?: boolean;
+  date?: Date;
+  setToNowTooltip?: string;
+  isSelecting?: string;
+  color?: string;
+}
+
+export const DEFAULT_DATE_PICKER_COLOR: string = 'blue';
```
sroy3: Added the enum

src/components/datePicker/DatePicker.tsx
```diff
@@ -0,0 +1,86 @@
...
+      ? <SetToNowButton onClick={() => this.setToToday()} tooltip={this.props.setToNowTooltip} />
+      : null;
+
+    let inputClasses: string[] = [`border-${color}`];
+    if (this.props.isSelecting === DateLimits.upper && this.props.upperLimit
+      || this.props.isSelecting === DateLimits.lower && !this.props.upperLimit) {
+      inputClasses.push('picking-date');
+    } else if (this.dateInput && this.dateInput.value) {
+      inputClasses.push('date-picked');
+      inputClasses.push(`bg-${color}`);
```
sroy3: Done

src/components/datePicker/DatePickerReducers.ts
sroy3: You are right, much better this way

src/components/datePicker/examples/DatePickerDropdownConnectedExamples.tsx
sroy3: Done

src/utils/DateUtils.ts
sroy3: Excellent. Changed.

src/utils/DateUtils.ts
sroy3: I have been using it (add and subtract) elsewhere to simplify date manipulation. I have no idea why I did not think of using it here. I deleted the method and replaced it by subtract where used.

src/components/datePicker/DatePickerReducers.ts
```diff
@@ -0,0 +1,112 @@
...
+      });
+    case DatePickerActions.select:
+      if (state.id !== action.payload.id) {
+        return state;
+      }
+      return _.extend({}, state, {
+        selected: action.payload.limit
+      });
+    case DatePickerActions.apply:
+      if (state.id.indexOf(action.payload.id) !== 0) {
```
MichelML: why not if (state.id !== action.payload.id) here?

sroy3: Because we can apply all date pickers from a date picker box. Their id all start with the id of the box.

src/components/datePicker/DatePickerReducers.ts
```diff
@@ -0,0 +1,112 @@
...
+  lowerLimit: moment().startOf('day').toDate(),
+  upperLimit: moment().endOf('day').toDate(),
+  selected: '',
+  appliedLowerLimit: moment().startOf('day').toDate(),
+  appliedUpperLimit: moment().endOf('day').toDate()
+};
+export const datePickersInitialState: IDatePickerState[] = [];
+
+export const datePickerReducer = (state: IDatePickerState = datePickerInitialState,
+  action: IReduxAction<IReduxActionsPayload>): IDatePickerState => {
```
JPLachance: I thinks you could rework the IReduxActionsPayload interface and replace it with a union of all interfaces possible to use here, like I did in sources security identities. It would be better.

MichelML:
Ok, after testing the datepicker and the calendar in the docs, here are some thoughts and suggestions:

**Date Picker:**
The values entered in those inputs:
screen shot 2017-02-16 at 1 12 29 pm
are not properly validated (or the validation should appear in the ui) 
see https://drive.google.com/file/d/0B75i_S1A9mU2bHBWMTM1X29UUGc/view

suggestion: 
if the new value entered does not have the format YYYY-MM-DD HH:mm:ss , reset the input value to the value in the current state

extra: 
- if a new value of a start date input is entered and is valid but is OVER the current end date input value, set the end date input value to the start date input value
- if a new value of a end date input is entered and is valid but is BELOW the current start date input value, set the start date input value to the end date input value

**Calendar**
- suggestion: fixed width for months (as discussed)
- why is there sometimes 4 rows for the days, and sometimes 5? It was a bit frustrating when switching the months and years because the buttons move from their original place when the number of rows change. Suggest to fix it at 5 rows always if possible. 

**Calendar (with datepicker)**
- the calendar is not in sync with the datepicker right now (see https://drive.google.com/file/d/0B75i_S1A9mU2aDQ3YzlMLVNXdTg/view ) 
Suggestion:
I think this should be the case. More specifically, whenever there is a date interval change (a date range preset is clicked (last 10 seconds, etc), a new value is entered in the datetime inputs, or else, the calendar month and year should be set to the month and year of the start date. If there is multiple date ranges, the month and year in the calendar should be set to the month and year of the start date of the last date range modified.

- weird behavior: date range appearing in year 2007 and 2017? (see https://drive.google.com/file/d/0B75i_S1A9mU2cUxRRkdxR282Wnc/view)


Ceci etant dit, trop nice le date picker, huge job accomplished.

sroy3:
@MichelML 

I'll check for the fixed width for the months soon. It'll be in the vapor repo.
For the number of rows, it really depends on the months. I do not think it is best to add another row (which will only be filled by days from another month), but maybe UX thinks otherwise or has a better solution.

I'll apply your other comments before merging.

sroy3:
@MichelML I've applied your comments if you want to check before I merge.

src/components/optionsCycle/OptionsCycle.tsx
```diff
@@ -0,0 +1,81 @@
...
+    if (this.props.isInline) {
+      optionsCycleClasses.push('mod-inline');
+    }
+
+    return (
+      <div className={optionsCycleClasses.join(' ')}>
+        <button className='options-cycle-button previous-option' onClick={() => this.goToPreviousOption()}>
+          <Svg svgName='arrow-left-rounded' svgClass='icon fill-dark-blue' />
+        </button>
+        <span className='options-cycle-option'>{this.selectedOption}</span>
```
sroy3: As I have said before, it's simply a style change in Vapor. I will add it after. No need for a new class here.

src/components/datePicker/DatePickerReducers.ts
```diff
@@ -0,0 +1,112 @@
...
+        appliedLowerLimit: state.appliedLowerLimit,
+        appliedUpperLimit: state.appliedUpperLimit
+      };
+    case DatePickerActions.changeLowerLimit:
+      if (state.id !== action.payload.id) {
+        return state;
+      }
+      return _.extend({}, state, {
+        lowerLimit: action.payload.date
+      });
```
sroy3: I like that. Done. Although I will have to do it elsewhere later.

PR #70, opened 2017-02-10
Added a prop to be able to crop the item filter on the right or on the left

sroy3:
screen shot 2017-02-10 at 3 11 50 pm

### DIFF
--- a/src/components/actions/filters/ItemFilter.tsx
+++ b/src/components/actions/filters/ItemFilter.tsx
@@ -5,15 +5,29 @@ export interface IItemFilterProps extends React.ClassAttributes<ItemFilter> {
   label: string;
   item: string;
   onClear: () => void;
+  crop?: number;
 }
+
+export const ELLIPSIS: string = '...';
 
 export class ItemFilter extends React.Component<IItemFilterProps, any> {
 
   render() {
+    let itemFilter: string = this.props.item;
+    if (this.props.crop) {
+      let itemFilterLength: number = itemFilter.length;
+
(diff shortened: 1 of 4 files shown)


MichelML:
in what case would you use a left crop? are we planning to use it? that's nice though, just wondering in what circumstances it would be utilized 😄

sroy3:
@MichelML actually in the log browser, we will be cropping on the left to see the end of the uri.

src/components/actions/filters/ItemFilter.tsx
```diff
@@ -5,15 +5,29 @@ export interface IItemFilterProps extends React.ClassAttributes<ItemFilter> {
...
 export class ItemFilter extends React.Component<IItemFilterProps, any> {
 
   render() {
+    let itemFilter: string = this.props.item;
+    if (this.props.crop) {
+      let itemFilterLength: number = itemFilter.length;
+
+      if (itemFilterLength > Math.abs(this.props.crop)) {
+        itemFilter = this.props.crop > 0
+          ? itemFilter.substring(0, this.props.crop) + ELLIPSIS
```
pochretien: You can use s.truncate(itemFilter, this.props.crop) with underscore strings, just another way to add the ellipsis. We can add third param to custom the ellipsis if we want :) but default is ...

sroy3: I wish it did work, but it only truncates on one side. Better keep consistency between cropping right and cropping left and use substring.

PR #72, opened 2017-02-21
Cleaned up code for the date picker and the per page navigation

sroy3:
What changed: 

1. Value of the option picker is calculated on option click and not re-evaluated each time the state changes. I added the label in the state also to compare the options between them instead of using the value which can sometimes be different because of when it was calculated.

2. The per page navigation gets it's current per page and the page number from the state and not from its parent (currentPerPage was there twice and there were often conflicts were the selected per page number was not the right one).

3. Auto-selections of inputs on the date picker. We select the first one when opening the date picker and the second one when the first one have been selected.

4. Select the end of the day when clicking and selecting the upper limit. Feels more natural this way and fixes the problem of the lower and upper limit being the same when clicking on the same date.

### DIFF
--- a/src/components/datePicker/DatePickerReducers.ts
+++ b/src/components/datePicker/DatePickerReducers.ts
@@ -29,47 +29,72 @@ export const datePickerInitialState: IDatePickerState = {
 };
 export const datePickersInitialState: IDatePickerState[] = [];
 
+const addDatePicker = (state: IDatePickerState = datePickerInitialState,
+  action: IReduxAction<IReduxActionsPayload>): IDatePickerState => {
+  return {
+    id: action.payload.id,
+    calendarId: action.payload.calendarId,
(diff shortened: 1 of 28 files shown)


src/components/datePicker/DatePickerReducers.ts
```diff
@@ -29,47 +29,72 @@ export const datePickerInitialState: IDatePickerState = {
...
+  return state.id.indexOf(action.payload.id) !== 0
+    ? state
+    : _.extend({}, state, {
+      selected: '',
+      lowerLimit: state.appliedLowerLimit,
+      upperLimit: state.appliedUpperLimit
+    });
+};
+
 export const datePickerReducer = (state: IDatePickerState = datePickerInitialState,
```
JPLachance: Good job! It did the trick!

PR #73, opened 2017-02-21
Update all dependencies + WebPack 2!

JPLachance:
+ Add bitHound config file
+ Use Airbnb eslintrc config
* Update every outdated dependencies, fix what was broken, migrate to webpack 2
* Format all the code!
! Fix react-tether definition file following a breaking change…


JPLachance:
Fun fact, bitHound is telling us that some of our functions are too complex. We should fix that to make him approuve our future PRs.

Also, dependencies issues are caused by what seem's to be a bug (unable to retrive the license for `@types` dependencies).

types/react-tether/index.d.ts
```diff
@@ -57,5 +57,6 @@ declare namespace TetherComponent {
 }
 
 declare module 'react-tether' {
-  export = TetherComponent;
+  // tslint:disable-next-line:no-default-export
+  export default TetherComponent;
```
JPLachance: This breaks the generated denifition file. I'll fix it.

sroy3:
I do not really know how we can rework the reducers so that they aren't tagged as complex functions from bitHound. It probably has to do with the facts that there are many returns, which is expected.

src/components/calendar/tests/CalendarConnected.spec.tsx
```diff
@@ -8,18 +8,12 @@ import { CalendarConnected } from '../CalendarConnected';
...
-} from '../../datePicker/DatePickerActions';
+import { addDatePicker, selectDate, changeDatePickerUpperLimit, changeDatePickerLowerLimit } from '../../datePicker/DatePickerActions';
 import { addOptionPicker, changeOptionPicker } from '../../optionPicker/OptionPickerActions';
 import * as _ from 'underscore';
-import * as moment from 'moment';
 // tslint:disable-next-line:no-unused-variable
 import * as React from 'react';
-import { IDatePickerState } from '../../datePicker/DatePickerReducers';
+
 import moment = require('moment');
```
sroy3: I think `import * as moment from 'moment';` is better for uniformity, but maybe you have a better reason for this change.

JPLachance: If you check above, it was imported twice in the same file and flagged as an error by the compiler. Any clue why it was there twice and which import we should keep?

sroy3: Oh, you are right, sorry I thought you had add it. It probably is Webstorm that added `import moment = require('moment');` Like I said, I prefer `import * as moment from 'moment` since it keeps the same format we use for other imports. This is also how I import it everywhere else.

JPLachance: Fixed!

PR #74, opened 2017-02-23
Error rows modifications

sroy3:


gulpfile.js
```diff
@@ -146,7 +146,7 @@ gulp.task('test:remap', false, () =>
     })));
 
 gulp.task('test', 'Run all tests in PhantomJS and exit', (done) => {
-  runSequence('prettify:tests', 'test:single', 'test:remap', done);
+  runSequence('prettify:src', 'test:single', 'test:remap', done);
```
MichelML: is this wanted?

MichelML:
looks good to me!

gulpfile.js
sroy3: Yes, `'prettify:tests'` does not exist in the gulpfile and it is alright to not just prettify the test, but the src as well.

PR #78, opened 2017-03-16
Date picker validation

sroy3:


src/components/calendar/Calendar.tsx
```diff
@@ -11,6 +11,18 @@ import * as React from 'react';
...
+export const CalendarSelectionRuleType = {
+  all: 'ALL',
+  lower: 'LOWER',
+  upper: 'UPPER',
+  range: 'RANGE'
+};
+
+export interface ICalendarSelectionRule {
+  test: (date: Date, secondDate?: Date) => boolean;
+  isFor: 'ALL' | 'LOWER' | 'UPPER' | 'RANGE';
```
GermainBergeron: In the d.ts, is it typed as string? Because I install react-vapor in UA and it will probably break since i'm not in typescript 2

sroy3: I will change it to string and fallback on 'ALL' if it is something other than all, lower, upper or range.

sroy3: Done.

MichelML:
tested it, works great! good job steph

PR #95, opened 2017-03-23
Adds types field to .d.ts

sbelzile-Coveo:
https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html


sroy3:
There used to be a problem with some of our projects with this, but I tried it again and it seems to work now.

PR #105, opened 2017-03-28
Add possibility for the popover to behave as a modal

GermainBergeron:
Added UTs
Modified example

### DIFF
--- a/docs/members-example/views/MemberEditView.tsx
+++ b/docs/members-example/views/MemberEditView.tsx
@@ -80,7 +80,8 @@ export class MemberEditView extends React.Component<IMemberEditViewProps, IMembe
           pin: true
         }]}
         isOpen={this.props.isOpen}
-        onToggle={(isOpen: boolean) => this.props.toggleMemberOpen(isOpen)}>
+        onToggle={(isOpen: boolean) => this.props.toggleMemberOpen(isOpen)}
+        isModal={true}>
         <button type='button' className='btn'>
           {isNew ? 'Add member' : this.props.appliedEmail}
         </button>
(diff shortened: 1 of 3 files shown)


docs/members-example/views/MemberEditView.tsx
```diff
@@ -80,7 +80,8 @@ export class MemberEditView extends React.Component<IMemberEditViewProps, IMembe
           pin: true
         }]}
         isOpen={this.props.isOpen}
-        onToggle={(isOpen: boolean) => this.props.toggleMemberOpen(isOpen)}>
+        onToggle={(isOpen: boolean) => this.props.toggleMemberOpen(isOpen)}
+        isModal={true}>
```
sroy3: You pass boolean props just as you would on html elements `<Popover {...otherProps} isModal>`

src/components/popover/Popover.spec.tsx
```diff
@@ -178,6 +178,32 @@ describe('<Popover>', () => {
...
+        });
+
+        mountPopover(popoverProps);
+      });
+
+      it('should not close the popover on click tether element', () => {
+        // Using getElementById here since the Tether element is being rendered outside the popoverWrapper.
+        document.getElementById(popoverElementId).click();
+
+        expect(toggleOpenedSpy.calls.count()).toBe(0);
```
pochretien: I think we use now .not.toHaveBeenCalled()

src/components/popover/Popover.spec.tsx
```diff
@@ -178,6 +178,32 @@ describe('<Popover>', () => {
...
+      });
+
+      it('should close the popover when clicking outside Popover', () => {
+        document.getElementById('App').click();
+
+        expect(toggleOpenedSpy.calls.count()).toBe(1);
```
pochretien: toHaveBeenCalledTimes(1)

PR #111, opened 2017-03-29
Refactored the tooltip component to avoid conflicts with other tooltips

sroy3:
I will bump a major after this. Some props aren't available anymore. You can refer to this https://react-bootstrap.github.io/components.html#tooltips for what is available. Example: template does not exist anymore, but you can change the classes on the tooltip with the `bsStyle` prop.


src/components/tooltip/tests/Tooltip.spec.tsx
```diff
@@ -0,0 +1,76 @@
...
+describe('Tooltip', () => {
+  let tooltipWrapper: ReactWrapper<ITooltipProps, void>;
+  const TOOLTIP_PROPS: ITooltipProps = {
+    title: 'My test tooltip!'
+  };
+
+  describe('<Tooltip />', () => {
+    it('should render without error', () => {
+      expect(() => shallow(
+        <Tooltip {...TOOLTIP_PROPS}>
```
rousselm4: `<Tooltip {...TOOLTIP_PROPS}>
    Hover me!
</Tooltip>` is repeated 3 times in the next 3 tests. Can we extract it to a variable or constant?

PR #115, opened 2017-03-31
Long facet rows

sroy3:
screen shot 2017-03-30 at 8 48 26 pm


src/components/facets/FacetRow.tsx
```diff
@@ -8,9 +9,15 @@ export interface IFacetRowProps extends React.ClassAttributes<FacetRow> {
...
 }
 
+export const MAX_NAME_LENGTH: number = 25;
+
 export class FacetRow extends React.Component<IFacetRowProps, any> {
 
   render() {
+    const label: JSX.Element = this.props.facetRow.formattedName.length > MAX_NAME_LENGTH
+      ? <Tooltip title={this.props.facetRow.formattedName} placement='top'>{this.props.facetRow.formattedName}</Tooltip>
+      : <span>{this.props.facetRow.formattedName}</span>;
```
rousselm4: I don't know if it's pertinent, but adding a '...' a the end when the string is cut could be nice for the user.

sroy3: This could be done in css instead and would be more precise because 25 is a length that is more of an average than one that will work in every case. It depends mostly on the width of the facet.

PR #143, opened 2017-04-14
Hide navigation items instead of simply not rendering them and updated dependencies

sroy3:


MichelML:
So I don't approve blindly, why are we making this change exactly?

sroy3:
If the pagination is not rendered, the pagination state won't be initialized. Before parsing the state from the url in the admin we are waiting on some element state to be initialized before changing them. If the pagination state does not exist at first, we will not be able to change the state to the one saved in the url.

src/components/navigation/Navigation.tsx
```diff
@@ -34,48 +34,50 @@ export interface INavigationStateProps extends IReduxStatePossibleProps {
...
+    if (this.props.currentPerPage) {
+      perPageProps.currentPerPage = this.props.currentPerPage;
     }
+    const perPage: JSX.Element = this.props.withReduxState
+      ? <NavigationPerPageConnected id={this.props.id} loadingIds={this.props.loadingIds} {...perPageProps} />
+      : <NavigationPerPage onPerPageClick={this.props.onPerPageClick} {...perPageProps} />;
+    const perPageClass = this.props.perPageNumbers.length && this.props.totalEntries > this.props.perPageNumbers[0] ? '' : 'hidden';
 
     let navigationClasses: string = 'pagination-container' + (this.props.isLoading ? ' loading-view' : '');
     let loading: JSX.Element = this.props.withReduxState ? <LoadingConnected id={'loading-' + this.props.id} /> : <Loading />;
```
MichelML: i know this is out of scope but you could use const and template strings here

``` 
const navigationClasses: string = `pagination-container${this.props.isLoading ? ' loading-view' : ''}`;
const loading: JSX.Element = this.props.withReduxState ? <LoadingConnected id={`loading-${this.props.id}`} /> : <Loading />;

File: src/components/navigation/Navigation.tsx [hunk 2]
```diff
@@ -34,48 +34,50 @@ export interface INavigationStateProps extends IReduxStatePossibleProps {
...
-    }
+    const paginationProps: INavigationPaginationProps = {
+      totalPages: this.props.totalPages,
+      numberOfPagesToShow: this.props.numberOfPagesToShow,
+      previousLabel: this.props.previousLabel,
+      nextLabel: this.props.nextLabel,
+      hidePages: this.props.hidePages || !this.props.totalPages || this.props.totalPages === 1
+    };
+    const pagination: JSX.Element = this.props.withReduxState
+      ? <NavigationPaginationConnected id={'pagination-' + this.props.id} loadingIds={this.props.loadingIds} {...paginationProps} />
```
MichelML:
```
id={`pagintation-${this.props.id}`}
```

PR #216, opened 2017-06-08
Add BlankSlate and Button component

pochretien:
It's only a part of the examples
image
image


src/components/blankSlate/BlankSlate.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { Button } from '../button/Button';
+
+export interface IBlankSlateProps extends React.ClassAttributes<BlankSlate> {
+  title?: string;
+  description?: string;
+  buttons?: IBaseActionOptions[];
+  withModal?: boolean;
+}
+
+export class BlankSlate extends React.Component<IBlankSlateProps, any> {
```
sroy3: I know it is not that way in most of the other components, but you can use void as a type for the state (instead of any).

src/components/blankSlate/BlankSlate.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  static defaultProps: Partial<IBlankSlateProps> = {
+    title: '',
+    buttons: [],
+    withModal: false,
+  };
+
+  getDescriptionTemplate(): JSX.Element {
+    return this.props.description
+      ? (<p>{this.props.description}</p>)
```
sroy3: Parenthesis are unnecessary.

src/components/blankSlate/BlankSlate.tsx
```diff
@@ -0,0 +1,50 @@
...
+      : null;
+  }
+
+  getButtonsTemplate(): JSX.Element[] {
+    let buttonElements: Array<JSX.Element> = [];
+    _.each(this.props.buttons, (buttonProps: IBaseActionOptions) => {
```
sroy3: You could use map instead and return the result directly.

src/components/button/Button.tsx
```diff
@@ -0,0 +1,69 @@
...
+        <button className={buttonClass}
+          onClick={() => this.onClick()}
+          {...buttonAttrs}>
+          {this.props.name}
+        </button>);
+    }
+
+    const tooltipPlacement: string = this.props.tooltipPlacement || 'right';
+    return !_.isEmpty(this.props.tooltip)
+      ? <Tooltip title={this.props.tooltip}
```
sroy3: The Tooltip opening tag can be on one line. We break lines for each attribute only if the line is too long.

src/components/button/Button.tsx
```diff
@@ -0,0 +1,69 @@
...
+    }
+  }
+
+  getTemplate(buttonClass: string): JSX.Element {
+    let buttonElement: JSX.Element;
+
+    const disabled: boolean = !this.props.enabled;
+
+    if (this.props.link) {
+      const target = this.props.target ? this.props.target : '';
```
sroy3: You could add `target: ''` as a default prop.

src/components/button/Button.tsx
```diff
@@ -0,0 +1,69 @@
...
+      const rel = 'noopener noreferrer';
+      const buttonAttrs = { disabled, target, rel };
+      buttonElement = (
+        <a className={`${buttonClass} btn-container`}
+          href={this.props.link}
+          onClick={() => this.onClick()}
```
sroy3: You could add this to the `buttonAttrs`. `buttonAttrs` could be outside of the condition and contain only `onClick` and `disabled` at first, then you could add `target` and `rel` inside of your if.

src/components/button/Button.tsx
```diff
@@ -0,0 +1,69 @@
...
+      const buttonAttrs = { disabled };
+      buttonElement = (
+        <button className={buttonClass}
+          onClick={() => this.onClick()}
+          {...buttonAttrs}>
+          {this.props.name}
+        </button>);
+    }
+
+    const tooltipPlacement: string = this.props.tooltipPlacement || 'right';
```
sroy3: You could default the prop `tooltipPlacement` to right.

src/components/button/Button.tsx
```diff
@@ -0,0 +1,69 @@
...
+        placement={tooltipPlacement}>
+        <span>
+          {buttonElement}
+        </span>
+      </Tooltip>
+      : buttonElement;
+  }
+
+  render() {
+    let buttonClasses = ['btn'];
```
sroy3: This can be a const.

src/components/blankSlate/examples/BlankSlateExample.tsx
```diff
@@ -0,0 +1,59 @@
...
+          <label className='form-control-label'>Default BlankSlate</label>
+          <BlankSlate />
+        </div>
+        <div className='form-group'>
+          <label className='form-control-label'>BlankSlate with title</label>
+          <BlankSlate title='Title test' />
+        </div>
+        <div className='form-group'>
+          <label className='form-control-label'>BlankSlate with title</label>
+          <BlankSlate title='description test' />
```
GermainBergeron: It should probably be `BlankSlate with description` and description instead of title

src/components/button/Button.tsx
pochretien: I need a let for the concat

src/components/button/Button.tsx
```diff
@@ -0,0 +1,69 @@
...
+    if (this.props.primary) {
+      buttonClasses.push('mod-primary');
+    }
+    if (!this.props.enabled) {
```
GermainBergeron: You can check `if (disabled)` instead since you declared it above

src/components/button/Button.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  static defaultProps: Partial<IBaseActionOptions> = {
+    enabled: true,
+    name: '',
+    tooltip: '',
+    primary: false,
+  };
+
+  private onClick() {
+    if (this.props.onClick) {
```
GermainBergeron: I think we shouldn't call the onClick if the button is disabled

src/components/button/examples/ButtonExamples.tsx
```diff
@@ -0,0 +1,94 @@
...
+            <Button enabled={false} name='Disabled Button link with tooltip' tooltip='Tooltip test' link='http://perdu.com/' target='_blank' />
+          </div>
+        </div>
+        <div className='form-group'>
+          <label className='form-control-label'>Disabled Button link with tooltip placement top</label>
+          <div className='form-control'>
+            <Button enabled={false} name='Disabled Button link with tooltip top' tooltip='Tooltip test' tooltipPlacement='top' link='http://perdu.com/'
+              target='_blank' />
+          </div>
+        </div>
```
GermainBergeron: Can you add an example for a disabled button with a onClick?

src/components/button/Button.tsx
GermainBergeron: You can do `buttonClasses.push('state-disabled', 'disabled', 'text-medium-grey')` instead

src/components/button/examples/ButtonExamples.tsx
pochretien: done!

src/components/button/Button.tsx
pochretien: onClick remove on disabled

src/components/button/Button.tsx
pochretien: the variable disabled is declared in the function after

PR #218, opened 2017-06-12
remove example for chosen single select in index.tsx

pochretien:
add files to regroup all chosen select examples
add example witht 2000 options

image


src/components/chosen/examples/ChosenSelectExamples.tsx
```diff
@@ -0,0 +1,56 @@
+import * as React from 'react';
+import { ChosenSelect } from '../ChosenSelect';
+import * as _ from 'underscore';
+
+export class ChosenSelectExamples extends React.Component<any, any> {
+
+  getNumberOfOptions(number: number) {
+    return _.times(number, (index: number) => {
+      const indexShow = index + 1;
+      return <option value={indexShow}>Option {indexShow}</option>;
```
sroy3: Don't you need a unique key here?

PR #234, opened 2017-06-16
Add icon to blankslate

rousselm4:
Added a test, and an example for that. Now we can optionally add an icon in the top of the blankslate


src/components/blankSlate/BlankSlate.tsx
```diff
@@ -2,12 +2,16 @@ import * as React from 'react';
...
 import { IBaseActionOptions } from '../actions/Action';
 import { Button } from '../button/Button';
+import { Svg } from '../svg/Svg';
 
 export interface IBlankSlateProps extends React.ClassAttributes<BlankSlate> {
   title?: string;
   description?: string;
   buttons?: IBaseActionOptions[];
   withModal?: boolean;
+  classes?: string;
```
GermainBergeron: classes should probably be a `string[]`

PR #242, opened 2017-06-21
Rework date inputs

sroy3:
Rework of the date inputs to make sure there is no conflict between the selection of a date and the blur of an input


src/components/datePicker/DatePicker.tsx
```diff
@@ -80,8 +96,7 @@ export class DatePicker extends React.Component<IDatePickerProps, any> {
 
 
     const inputClasses: string[] = [`border-${this.props.color}`];
-    if (this.props.isSelecting === DateLimits.upper && this.props.upperLimit
-      || this.props.isSelecting === DateLimits.lower && !this.props.upperLimit) {
```
MichelML: ca va affecter la selection de dates en dehors des limites ca ou non?

sroy3: I have simply made an attribute out of it (`this.isPicked` initialized in `componentWillReceiveProps`) so it can also be used elsewhere.

sroy3:
@MichelML the issues you have in your videos are due to the weird validation I have set on this example. Trying the same thing on an other example should work flawlessly.

MichelML:
Great. Exact, these were only for the example in the dropdown

PR #295, opened 2017-07-24
fix the activeOption to be undefined if the filter return no options

pochretien:


MichelML:
tu game de poster un demo video ?

src/components/dropdownSearch/DropdownSearchReducers.tsx
```diff
@@ -44,13 +44,17 @@ export const getNextIndexPosition = (array: any[], item: any, key: number): numb
...
+export const getOptionsFiltered = (state: IDropdownSearchState, filterText?: string): IDropdownOption[] => {
   const currentFilterText: string = filterText || state.filterText;
-  return _.filter(state.options,
-    (option: IDropdownOption) => {
-      const value = option.displayValue || option.value;
-      return _.isEmpty(currentFilterText) || s.contains(value.toLowerCase(), (currentFilterText).toLowerCase());
-    });
+  if (_.isEmpty(currentFilterText)) {
+    return state.options;
+  } else {
```
sroy3: No need to wrap the return with a else here

src/components/dropdownSearch/tests/DropdownSearchReduces.spec.tsx
```diff
@@ -1,15 +1,15 @@
-import { IReduxAction } from '../../../utils/ReduxUtils';
-import { IDropdownOption } from '../DropdownSearch';
+import {IReduxAction} from '../../../utils/ReduxUtils';
```
sroy3: I think you need to run `gulp:prettify` so you do not break the format

PR #299, opened 2017-07-27
MultiSelect DropdownSearch component

dcarmichael1:


src/components/dropdownSearch/DropdownSearch.tsx
```diff
@@ -220,9 +232,9 @@ export class DropdownSearch extends React.Component<IDropdownSearchProps, void>
     }
   }
 
-  handleOnFilter(id: string, filterText: string) {
-    if (this.props.onFilterClick) {
-      this.props.onFilterClick(filterText);
+  handleOnFilterTextChange(id: string, filterText: string) {
```
sroy3: Why do you need the id here?

src/components/dropdownSearch/DropdownSearchActions.tsx
```diff
@@ -67,17 +85,73 @@ export const addDropdownSearch = (id: string, optionsDropdown: IDropdownOption[]
   },
 });
 
+export const addMultiSelectDropdownSearch = (id: string, optionsDropdown: IDropdownOption[] = []): IReduxAction<IOptionsDropdownSearchPayload> => ({
+  type: DropdownSearchActions.addMultiSelect,
+  payload: {
+    id,
+    optionsDropdown,
+    filterText: '',
```
sroy3: If you did reuse another action, setting `filterText` to an empty string would be alright, but in the case of a brand new action, I would simply reset `filterText` in the reducer instead.

src/components/dropdownSearch/DropdownSearchConnected.tsx
```diff
@@ -11,26 +11,29 @@ import {
...
 const mapStateToProps = (state: IReactVaporState, ownProps: IDropdownSearchProps): IDropdownSearchStateProps => {
   const dropdownSearch: IDropdownSearchState = _.findWhere(state.dropdownSearch, { id: ownProps.id });
-  const selectedOption = ownProps.defaultSelectedOption || defaultSelectedOption;
 
   if (dropdownSearch) {
     return {
-      isOpened: dropdownSearch.isOpened,
-      options: dropdownSearch.options,
-      selectedOption: dropdownSearch.selectedOption || selectedOption,
+      isOpened: dropdownSearch.isOpened || false,
```
sroy3: If you are able to move a few of these to the default props, that would make it cleaner here. Example: if you set `isOpened: false` as a default prop, you can simply write `isOpened: dropdownSearch.isOpened` here and whomever uses the component without Redux won't have to pass it all Redux props.

sroy3: Also, you wouldn't need your second return

src/components/dropdownSearch/DropdownSearchReducers.tsx
```diff
@@ -1,22 +1,28 @@
 import { IReduxAction } from '../../utils/ReduxUtils';
 import { IDropdownOption } from './DropdownSearch';
-import { DropdownSearchActions, IOptionsDropdownSearchPayload } from './DropdownSearchActions';
+import {DropdownSearchActions, IOptionsDropdownSearchPayload} from './DropdownSearchActions';
```
sroy3: You will need to run `gulp prettify` to clean up the formating

src/components/dropdownSearch/DropdownSearchReducers.tsx
```diff
@@ -1,22 +1,28 @@
...
+  selectedOptions?: FixedQueue<IDropdownOption>;
   selectedOption?: IDropdownOption;
   activeOption?: IDropdownOption;
   setFocusOnDropdownButton?: boolean;
 }
 
 export const defaultSelectedOption: IDropdownOption = {
-  value: 'Select an option',
+  value: UUID.generate(),
+  displayValue: 'Select an option',
```
sroy3: You have to make this available to the users so that they can localize it or change it according to their needs. This should be a prop on your top parent component with the default set to what you've set it here.

src/components/dropdownSearch/DropdownSearchReducers.tsx
```diff
@@ -44,9 +50,28 @@ export const getNextIndexPosition = (array: any[], item: any, key: number): numb
   return index;
 };
 
+export const removeSelectedOption = (state: IDropdownSearchState, displayValue: string): FixedQueue<IDropdownOption> => {
+  return new FixedQueue<IDropdownOption>(_.filter(state.selectedOptions.getQueue(),
+    (selectedOption: IDropdownOption) => {
```
sroy3: `(selectedOption: IDropdownOption) => selectedOption.displayValue !== displayValue;`

src/components/dropdownSearch/DropdownSearchReducers.tsx
```diff
@@ -44,9 +50,28 @@ export const getNextIndexPosition = (array: any[], item: any, key: number): numb
...
+      return selectedOption.displayValue != displayValue;
+    }
+  ));
+};
+
+export const addUniqueSelectedOption = (state: IDropdownSearchState, displayValue: string): FixedQueue<IDropdownOption> => {
+  return removeSelectedOption(state, displayValue).push({value: UUID.generate(), displayValue: displayValue});
+};
```
sroy3:
```
export const addUniqueSelectedOption = (state: IDropdownSearchState, displayValue: string): FixedQueue<IDropdownOption> 
  => removeSelectedOption(state, displayValue).push({value: UUID.generate(), displayValue: displayValue});
```

src/components/dropdownSearch/DropdownSearchReducers.tsx
```diff
@@ -44,9 +50,28 @@ export const getNextIndexPosition = (array: any[], item: any, key: number): numb
...
+
+export const getDisplayedOptions = (state: IDropdownSearchState) => {
+  return _.filter(state.options,
+    (option: IDropdownOption) => {
+      return _.findWhere(state.selectedOptions.getQueue(), {displayValue: option.displayValue}) == undefined;
+    });
```
sroy3: `(option: IDropdownOption) => _.findWhere(state.selectedOptions.getQueue(), {displayValue: option.displayValue}) === undefined;`

src/components/dropdownSearch/DropdownSearchReducers.tsx
```diff
@@ -111,11 +154,17 @@ export const dropdownSearchReducer = (state: IDropdownSearchState = dropdownSear
         return {
           ...state,
           id: action.payload.id,
-          selectedOption: state.activeOption,
           isOpened: false,
+          selectedOptions: state.selectedOptions.push(state.activeOption),
```
sroy3:
```
selectedOptions: [
  ...state.selectedOptions,
  _.extend({}, state.activeOption),
],
```
State is read-only and immutable (http://redux.js.org/docs/recipes/reducers/ImmutableUpdatePatterns.html)

src/components/dropdownSearch/DropdownSearchReducers.tsx
```diff
@@ -150,6 +199,22 @@ export const dropdownsSearchReducer = (state: IDropdownSearchState[] = dropdowns
...
+    case DropdownSearchActions.removeAllSelectedOptions:
+    case DropdownSearchActions.multiSelect:
+    case DropdownSearchActions.addCustomSelectedOption:
+    case DropdownSearchActions.onKeyDownMultiselect:
+    case DropdownSearchActions.removeSelectedOption:
+      return state.map((dropdownSearch: IDropdownSearchState) => {
+        if (dropdownSearch.id === action.payload.id) {
+          return multiSelectDropdownSearchReducer(dropdownSearch, action);
+        }
+        return dropdownSearch;
```
sroy3: You could use a ternary instead:
```
return dropdownSearch.id === action.payload.id
  ? multiSelectDropdownSearchReducer(dropdownSearch, action
  : dropdownSearch;
```

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearch.tsx
```diff
@@ -0,0 +1,39 @@
+import * as React from 'react';
+import { DropdownSearch } from '../DropdownSearch';
+import {MultiselectInput} from './MultiSelectInput';
+
+export class MultiSelectDropdownSearch extends DropdownSearch {
+
+  protected getNoOptions(): JSX.Element[] {
+    if (this.props.filterText.length > 0 && !this.props.selectedOptions.containsElementWithProperties({displayValue: this.props.filterText})) {
```
sroy3: no need for the `> 0`

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearch.tsx
```diff
@@ -0,0 +1,39 @@
...
+      return [
+        <li key='noResultDropdownSearch' onMouseDown={() => this.props.onCustomOptionClick(this.props.filterText)}>
+          <span>{`Create option for "${this.props.filterText}"`}</span>
```
sroy3: This needs to be fronted so it can be either localized or changed

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearch.tsx
```diff
@@ -0,0 +1,39 @@
...
+        </li>,
+      ];
+    }
+    return super.getNoOptions();
+  }
+
+  render() {
+    return (
+      <div className={this.getClasses()} style={this.getStyles()}>
+        <MultiselectInput selectedOptions={this.props.selectedOptions.getQueue()}
```
sroy3: Add a line break before selectedOptions and move your indentation way to the left.

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearch.tsx
```diff
@@ -0,0 +1,39 @@
...
+                          onRemoveAll={this.props.onRemoveAllSelectedOptions}
+                          onFilterTextChange={this.props.onFilterTextChange}
+                          onBlur={this.props.onBlur}
+                          onFocus={this.props.onFocus}
+                          onKeyDownFilterBox={(e: React.KeyboardEvent<HTMLInputElement>) => this.handleOnKeyDownFilterBox(e)}
+                          filterPlaceholder={this.props.filterPlaceholder}
+                          filterText={this.props.filterText}
+        />
+        <ul className='dropdown-menu'
+          ref={(input: HTMLUListElement) => { this.ulElement = input; }}
```
sroy3: Remove spaces before and after

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearchConnected.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { IDropdownSearchState } from './../DropdownSearchReducers';
+import { FixedQueue} from '../../../utils/FixedQueue';
+import { MultiSelectDropdownSearch } from './MultiSelectDropdownSearch';
+
+const mapStateToProps = (state: IReactVaporState, ownProps: IDropdownSearchProps): IDropdownSearchStateProps => {
+  const dropdownSearch: IDropdownSearchState = _.findWhere(state.dropdownSearch, { id: ownProps.id });
+
+  if (dropdownSearch) {
+    return {
+      isOpened: dropdownSearch.isOpened || false,
```
sroy3: Again try to move what comes after the pipes to the defaultProps and get rid of your second return

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearchConnected.tsx
```diff
@@ -0,0 +1,59 @@
...
+  return {
+    isOpened: false,
+    options: ownProps.defaultOptions || [],
+    displayedOptions: ownProps.displayedOptions || ownProps.options,
+    selectedOptions: new FixedQueue<IDropdownOption>(),
+    filterText: '',
+  };
+};
+
+const mapDispatchToProps = (dispatch: (action: IReduxAction<IReduxActionsPayload>) => void,
```
sroy3: This needs the type of what it return (`MultiSelectDropdownSearchDispatchProps`)

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearchReducer.tsx
```diff
@@ -0,0 +1,134 @@
...
+  IDropdownSearchState,
+  removeSelectedOption,
+} from '../DropdownSearchReducers';
+import { IReduxAction } from '../../../utils/ReduxUtils';
+import {FixedQueue} from '../../../utils/FixedQueue';
+import {IDropdownOption} from '../DropdownSearch';
+import {keyCode} from '../../../utils/InputUtils';
+
+export const multiSelectDropdownSearchReducer = (state: IDropdownSearchState = dropdownSearchInitialState,
+  action: IReduxAction<IOptionsDropdownSearchPayload>) => {
```
sroy3: You need a state interface and set it here as the return type

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearchReducer.tsx
```diff
@@ -0,0 +1,134 @@
...
+        filterText: action.payload.filterText,
+        isOpened: false,
+      };
+    case DropdownSearchActions.removeSelectedOption:
+      state.selectedOptions = removeSelectedOption(state, action.payload.selectedOptionValue);
+      state.displayedOptions = getDisplayedOptions(state);
+      return {
+        ...state,
+        id: action.payload.id,
+        selectedOptions: state.selectedOptions,
```
sroy3: For any other property that is not an object or an array, you could have simply removed this line since it is included in the `...state`, but since `selectedOptions` is an array, you will need to copy its content.
```
selectedOptions: [
  ...state.selectedOptions,
],
```
If the elements inside `selectedOptions` are objects, you will need to copy them too. If that is the case, I would recommend you only save their ids.

Same thing applies for `displayedOptions`

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearchReducer.tsx
```diff
@@ -0,0 +1,134 @@
...
+        displayedOptions: state.displayedOptions,
+      };
+    case DropdownSearchActions.multiSelect:
+      return {
+        ...state,
+        id: action.payload.id,
+        isOpened: true,
+        selectedOptions: state.selectedOptions.push(action.payload.addedSelectedOption),
```
sroy3: `selectedOptions` should be immutable.

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearchReducer.tsx
sroy3: It should also always be done whether it was changed or not.

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearchReducer.tsx
```diff
@@ -0,0 +1,134 @@
...
+        displayedOptions: state.displayedOptions,
+      };
+    case DropdownSearchActions.removeAllSelectedOptions:
+      state.selectedOptions = new FixedQueue<IDropdownOption>();
+      state.displayedOptions = getDisplayedOptions(state);
+      return {
+        ...state,
+        id: action.payload.id,
+        selectedOptions: state.selectedOptions,
+        displayedOptions: state.displayedOptions,
```
sroy3: You need to copy `selectedOptions` and `displayedOptions`

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearchReducer.tsx
```diff
@@ -0,0 +1,134 @@
...
+    case DropdownSearchActions.onKeyDownMultiselect:
+      const isFirstSelectedOption = action.payload.keyCode === keyCode.upArrow && state.activeOption === state.options[0];
+      const optionsFiltered = getOptionsFiltered(state);
+      if (action.payload.keyCode === keyCode.upArrow || action.payload.keyCode === keyCode.downArrow) {
+        return {
+          ...state,
+          isOpened: !isFirstSelectedOption,
+          activeOption: !isFirstSelectedOption ?
+            optionsFiltered[getNextIndexPosition(optionsFiltered, state.activeOption, action.payload.keyCode)] : undefined,
+          setFocusOnDropdownButton: isFirstSelectedOption,
```
sroy3: You need to copy `selectedOptions` and `displayedOptions`

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearchReducer.tsx
```diff
@@ -0,0 +1,134 @@
...
+        };
+      } else if ((action.payload.keyCode === keyCode.enter || action.payload.keyCode === keyCode.tab) && state.activeOption) {
+        return {
+          ...state,
+          id: action.payload.id,
+          isOpened: true,
+          selectedOptions: state.selectedOptions.push(state.activeOption),
```
sroy3: `selectedOptions` should be immutable

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearchReducer.tsx
```diff
@@ -0,0 +1,134 @@
...
+            selectedOptions: state.selectedOptions.removeLastElement(),
+            activeOption: undefined,
+            setFocusOnDropdownButton: true,
+            displayedOptions: getDisplayedOptions(state),
+          };
+        }
+      } else if (action.payload.keyCode === keyCode.escape) {
+        return {
+          ...state,
+          isOpened: false,
```
sroy3: You need to copy `selectedOptions` and `displayedOptions`

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearchReducer.tsx
```diff
@@ -0,0 +1,134 @@
...
+        };
+      } else if (action.payload.keyCode === -1) {
+        return {
+          ...state,
+          id: action.payload.id,
+          activeOption: undefined,
+          setFocusOnDropdownButton: false,
```
sroy3: You need to copy `selectedOptions` and `displayedOptions`

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearchReducer.tsx
```diff
@@ -0,0 +1,134 @@
...
+        };
+      }
+      return {
+        ...state,
+        id: action.payload.id,
+        activeOption: undefined,
+        isOpened: true,
+        setFocusOnDropdownButton: false,
```
sroy3: You need to copy `selectedOptions` and `displayedOptions`

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectInput.tsx
```diff
@@ -0,0 +1,94 @@
...
+      this.props.onKeyDownFilterBox(e);
+    }
+  }
+
+  private getSelectedOptionComponents(): JSX.Element[] {
+    const selectedOptionComponents: JSX.Element[] = [];
+
+    for (let selectedOption of this.props.selectedOptions) {
+      selectedOptionComponents.push(
+        <SelectedOption displayValue={selectedOption.displayValue}
```
sroy3: line break before `displayValue`

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectInput.tsx
```diff
@@ -0,0 +1,94 @@
...
+  private handleOnKeyDown(e: React.KeyboardEvent<HTMLInputElement>) {
+    if (this.props.onKeyDownFilterBox) {
+      this.props.onKeyDownFilterBox(e);
+    }
+  }
+
+  private getSelectedOptionComponents(): JSX.Element[] {
+    const selectedOptionComponents: JSX.Element[] = [];
+
+    for (let selectedOption of this.props.selectedOptions) {
```
sroy3: You can use a `_.map` instead and return its result

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectInput.tsx
```diff
@@ -0,0 +1,94 @@
...
+                        key={selectedOption.value}
+                        onRemoveClick={this.props.onRemoveClick}
+        />);
+    }
+
+    return selectedOptionComponents;
+  }
+
+  private getRemoveAllSelectedOptionsButton(): JSX.Element {
+    if (this.props.selectedOptions.length > 0) {
```
sroy3: No need for `> 0`

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectInput.tsx
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  render() {
+    return (
+      <div className='multiselect-input'>
+        <div className='selected-options-container'>
+          {this.getSelectedOptionComponents()}
+          <input
+            placeholder={this.props.filterPlaceholder}
+            onChange={(e) => this.handleInputChange(e)}
```
sroy3: e should be typed

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectInput.tsx
```diff
@@ -0,0 +1,94 @@
...
+            onBlur={() => this.handleOnBlur()}
+            onFocus={() => this.handleOnFocus()}
+            onKeyDown={(e) => this.handleOnKeyDown(e)}
```
sroy3: a should be typed

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectInput.tsx
```diff
@@ -0,0 +1,94 @@
...
+            autoFocus={true}
```
sroy3: Just like HTML boolean props, you can simply replace `autoFocus={true}` by `autoFocus`

src/components/dropdownSearch/MultiSelectDropdownSearch/tests/MultiSelectDropdownSearch.spec.tsx
```diff
@@ -0,0 +1,94 @@
...
+import {IDropdownOption, IDropdownSearchProps} from '../../DropdownSearch';
+import {FixedQueue} from '../../../../utils/FixedQueue';
+import {MultiSelectDropdownSearch} from '../MultiSelectDropdownSearch';
+
+describe('MultiSelectDropdownSearch', () => {
+  const id: string = UUID.generate();
+  const options = [
+    { value: 'test a', displayValue: 'test a' },
+    { value: 'test b', displayValue: 'test b' },
+    { value: 'test c', displayValue: 'test c' }];
```
sroy3: `];` should be on the other line. Do not forget the trailing comma

src/components/dropdownSearch/MultiSelectDropdownSearch/tests/MultiSelectDropdownSearch.spec.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  const props: IDropdownSearchProps = {
+    id,
+    modMenu: false,
```
sroy3: `options` should come before `modMenu`

src/components/dropdownSearch/MultiSelectDropdownSearch/tests/MultiSelectDropdownSearch.spec.tsx
```diff
@@ -0,0 +1,94 @@
...
+      it('should call custom option click function on mousedown when the value is not present in the selected options', () => {
+        const filterText: string = 'customValue';
+        const onCustomOptionClick = jasmine.createSpy('onCustomOptionClick');
+        multiSelectDropdownSearch.setProps({
+          filterText,
+          onCustomOptionClick,
+          isOpened: true,
+        });
+
+        multiSelectDropdownSearch.find('li').find('span').simulate('mouseDown');
```
sroy3: I am pretty sure you can do `.find('li span')`

src/components/dropdownSearch/MultiSelectDropdownSearch/tests/MultiSelectInput.spec.tsx
```diff
@@ -0,0 +1,80 @@
...
+      multiSelectInput.unmount();
+      multiSelectInput.detach();
+    });
+
+    describe('render', () => {
+      it('should render placeholder text', () => {
+        const filterPlaceholder = 'placeholdertext';
+        multiSelectInput.setProps({ filterPlaceholder });
+
+        expect(multiSelectInput.find(`input[placeholder="${filterPlaceholder}"]`).length).toBe(1);
```
sroy3: `[placeholder=${filterPlaceholder}]`

src/components/dropdownSearch/MultiSelectDropdownSearch/tests/MultiSelectInput.spec.tsx
```diff
@@ -0,0 +1,80 @@
...
+      });
+
+      it('should render filter text', () => {
+        const filterText = 'text';
+        multiSelectInput.setProps({ filterText });
+
+        expect(multiSelectInput.find(`input[value="${filterText}"]`).length).toBe(1);
```
sroy3: `[value=${filterText}]`

src/components/dropdownSearch/MultiSelectDropdownSearch/tests/MultiselectDropdownSearchReducer.spec.tsx
```diff
@@ -0,0 +1,241 @@
...
+
+    const options = [
+      {value: 'test 1', displayValue: 'test 1 display'},
+      {value: 'test 2', displayValue: 'test 2 display'},
+    ];
+
+    const oldState: IDropdownSearchState = {
+      id: 'new-dropdown-search',
+      isOpened: false,
+      options,
```
sroy3: this should be the first attribute

src/components/dropdownSearch/MultiSelectDropdownSearch/tests/SelectedOption.spec.tsx
```diff
@@ -0,0 +1,67 @@
...
+  describe('<SelectedOption />', () => {
+    let selectedOption: ReactWrapper<ISelectedOptionProps, any>;
+    let selectedOptionInstance: SelectedOption;
+
+    const renderSelectedOption = (props?: ISelectedOptionProps) => {
+      selectedOption = mount(
+        <SelectedOption {...props} />,
+        { attachTo: document.getElementById('App') },
+      );
+      selectedOptionInstance = selectedOption.instance() as SelectedOption;
```
sroy3: Is this needed?

sroy3:
Make sure you also address the drop in coverage.

gulpfile.js
```diff
@@ -115,6 +115,7 @@ gulp.task('cleanDefs', false, () =>
     .pipe(replace(/import.*$/gm, ''))
     .pipe(replace(/export =.+;$/gm, ''))
     .pipe(replace(/export default.+;$/gm, ''))
+    .pipe(replace(/: React.MouseEvent<[A-Za-z]+>/gm, ''))
```
MichelML: Actually it would cause a problem in some of our repos (v2) but the regexp is already at line 114 so you can remove it (I had the same problem)!

src/components/dropdownSearch/DropdownSearch.tsx
```diff
@@ -45,37 +47,48 @@ export interface IDropdownSearchDispatchProps {
...
-  private getSelectedOption(): JSX.Element {
-    return <span key={this.props.selectedOption.value}
-      className='dropdown-selected-value'
-      data-value={this.props.selectedOption.value}>
-      {this.props.selectedOption.displayValue || this.props.selectedOption.value}
-    </span>;
+  protected getSelectedOption(): JSX.Element[] {
+    if (this.props.selectedOptions.getQueue().length) {
+      return _.map(this.props.selectedOptions.getQueue(), (selectedOption: IDropdownOption) => {
+        return <span key={selectedOption.value}
```
MichelML:
```
return (
    <span 
       key={selectedOption.value}
       className='dropdown-selected-value'
       data-value={selectedOption.value}>
       {selectedOption.displayValue || selectedOption.value}
    </span>
);
```  
consider this suggestion for better indentation

src/components/dropdownSearch/DropdownSearchReducers.tsx
MichelML: @sroy3 you can already do it dropDownSearchConnected as a `defaultSelectedOption` that you can pass which will override this , unless i'm wrong but that's what I do with my use case

src/components/dropdownSearch/DropdownSearchReducers.tsx
MichelML: never mind it's a dropdown option i get it

src/utils/FixedQueue.ts
```diff
@@ -0,0 +1,55 @@
...
+    this.trimQueue();
+    return new FixedQueue<T>(this.queue, this.maxLength);
+  }
+
+  getFirstElement(): T {
+    return this.queue[0];
+  }
+
+  contains(element: T): boolean {
+    return this.queue.indexOf(element) != -1;
```
MichelML: !==

src/utils/FixedQueue.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  containsElementWithProperties(properties: any) {
+    return _.findWhere(this.queue, properties) != undefined;
```
MichelML: !==

src/utils/FixedQueue.ts
```diff
@@ -0,0 +1,55 @@
+import * as _ from 'underscore';
+
+export class FixedQueue<T> {
+
+  private queue: Array<T>;
+  private maxLength: number;
+
+  constructor(initialValues: Array<any> = [], maxLength: number = Number.MAX_VALUE) {
```
MichelML: Array\<T\> ?

src/utils/FixedQueue.ts
```diff
@@ -0,0 +1,55 @@
...
+    this.queue = initialValues;
+    this.trimQueue();
+  }
+
+  getQueue(): Array<T> {
+    return this.queue;
+  }
+
+  push(object: any): FixedQueue<T> {
+    this.queue.push(object);
```
MichelML: you could probably do this.queue = [...this.queue, object]; and avoid calling your constructor and just return the queue

src/utils/FixedQueue.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  removeLastElement(): FixedQueue<T> {
+    if (this.queue.length > 0) {
+      this.queue.pop();
+    }
+    return new FixedQueue<T>(this.queue, this.maxLength);
+  }
+
+  private trimQueue() {
```
MichelML: add a conditional to trim only if the array is too long otherwise it's not necessary

src/utils/FixedQueue.ts
MichelML:
```
this.queue = [...this.queue, object];
this.trimQueue();
return this.queue;
```

src/utils/FixedQueue.ts
```diff
@@ -0,0 +1,55 @@
+import * as _ from 'underscore';
```
MichelML: would definitely delete the methods  `contains`, `getFirstElement`, and `removeElement`

contains --> call get queue and use `_.contains`
getFirstElement --> call getQueue()[0]  (you already do this in your code)
removeElement --> error prone, won't work with objects or arrays, plus you don't use it in your code

your idea of this datastructure is very nice though 👍

MichelML:
just started trying the demo, quick notice...your autofocus makes the window scroll right to your component skipping a lot of components before it...i don't think we want that

src/components/dropdownSearch/DropdownSearchReducers.tsx
sroy3: But you still cannot change this display value. If it is later changed by some default passed as a prop from the user or a default prop, this should be an empty string until then.

gulpfile.js
dcarmichael1: I don't know how this came to be in my PR, but I removed it!

src/components/dropdownSearch/DropdownSearch.tsx
```diff
@@ -45,37 +47,48 @@ export interface IDropdownSearchDispatchProps {
...
   onKeyDownDropdownButton?: (keyCode: number) => void;
   onMouseEnterDropdown?: () => void;
+  onRemoveSelectedOption?: (value: string) => void;
+  onRemoveAllSelectedOptions?: () => void;
 }
 
 export interface IDropdownSearchProps extends IDropdownSearchOwnProps, IDropdownSearchStateProps, IDropdownSearchDispatchProps { }
 
-export class DropdownSearch extends React.Component<IDropdownSearchProps, void> {
+export class DropdownSearch extends React.Component<IDropdownSearchProps, any> {
```
dcarmichael1: Reverted

src/components/dropdownSearch/DropdownSearch.tsx
dcarmichael1: done

src/components/dropdownSearch/DropdownSearch.tsx
```diff
@@ -84,7 +97,7 @@ export class DropdownSearch extends React.Component<IDropdownSearchProps, void>
     )
       .map((option: IDropdownOption, index: number, options: IDropdownOption[]) => {
         const optionClasses = classNames({
-          'state-selected': option.value === this.props.selectedOption.value,
+          'state-selected': this.props.selectedOptions.getFirstElement() && option.value === this.props.selectedOptions.getQueue()[0].value,
```
dcarmichael1: done

src/components/dropdownSearch/DropdownSearch.tsx
```diff
@@ -110,16 +123,16 @@ export class DropdownSearch extends React.Component<IDropdownSearchProps, void>
...
         <span>{this.props.noResultText}</span>
       </li>,
     ];
   }
 
-  private getSvg(option: IDropdownOption): JSX.Element {
-    if (option.svg) {
+  protected getSvg(option: IDropdownOption): JSX.Element {
+    if (option && option.svg) {
       return <span key={option.svg.name}
```
dcarmichael1: done

src/components/dropdownSearch/DropdownSearch.tsx
dcarmichael1: I don't! Removed it

src/components/dropdownSearch/DropdownSearchActions.tsx
dcarmichael1: good point! done

src/components/dropdownSearch/DropdownSearchConnected.tsx
dcarmichael1: done, but I now return the default props if the id does not match any dropdown

src/components/dropdownSearch/DropdownSearchReducers.tsx
dcarmichael1: done

src/components/dropdownSearch/DropdownSearchReducers.tsx
dcarmichael1: Added a new prop to change the text

src/components/dropdownSearch/DropdownSearchReducers.tsx
dcarmichael1: The displayValue is more Reliable because in the case that we have a new custom option, it would only have a display value and a random value as a key

src/components/dropdownSearch/DropdownSearchReducers.tsx
dcarmichael1: I made the changes, but the arrow needs to be at the end of the previous line

src/components/dropdownSearch/DropdownSearchReducers.tsx
dcarmichael1: done

src/components/dropdownSearch/DropdownSearchReducers.tsx
dcarmichael1: You're right! but FixedOption.push returns a new object! So the previous state is not being changed

src/components/dropdownSearch/DropdownSearchReducers.tsx
dcarmichael1: done

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearch.tsx
dcarmichael1: done

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearch.tsx
dcarmichael1: done

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearch.tsx
dcarmichael1: done

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearch.tsx
dcarmichael1: done

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearchConnected.tsx
dcarmichael1: done

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearchConnected.tsx
dcarmichael1: done

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearchReducer.tsx
dcarmichael1: done

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearchReducer.tsx
dcarmichael1: done but with a twist maybe you could re-check it, but I create a new array in with util methods

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearchReducer.tsx
dcarmichael1: done

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearchReducer.tsx
dcarmichael1: as mentionned above, push creates a new object without changing the instance

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearchReducer.tsx
dcarmichael1: done

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectInput.tsx
dcarmichael1: done

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectInput.tsx
dcarmichael1: done

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectInput.tsx
dcarmichael1: done

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectInput.tsx
dcarmichael1: done

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectInput.tsx
dcarmichael1: done

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectInput.tsx
dcarmichael1: removed autoFocus

src/components/dropdownSearch/MultiSelectDropdownSearch/tests/MultiSelectDropdownSearch.spec.tsx
dcarmichael1: done

src/components/dropdownSearch/MultiSelectDropdownSearch/tests/MultiSelectDropdownSearch.spec.tsx
dcarmichael1: done

src/components/dropdownSearch/MultiSelectDropdownSearch/tests/MultiSelectDropdownSearch.spec.tsx
dcarmichael1: done

src/components/dropdownSearch/MultiSelectDropdownSearch/tests/MultiSelectInput.spec.tsx
dcarmichael1: we need the " for the selector to work

src/components/dropdownSearch/MultiSelectDropdownSearch/tests/MultiselectDropdownSearchReducer.spec.tsx
dcarmichael1: done

src/components/dropdownSearch/MultiSelectDropdownSearch/tests/SelectedOption.spec.tsx
dcarmichael1: This is to be coherent with how other component tests were implemented

src/components/dropdownSearch/DropdownSearchReducers.tsx
```diff
@@ -83,23 +125,24 @@ export const dropdownSearchReducer = (state: IDropdownSearchState = dropdownSear
...
         activeOption: undefined,
         setFocusOnDropdownButton: false,
       };
     case DropdownSearchActions.add:
       return {
+        ...state,
         id: action.payload.id,
         options: action.payload.optionsDropdown,
-        selectedOption: action.payload.selectedOption,
+        selectedOptions: new FixedQueue<IDropdownOption>([defaultSelectedOption], 1),
```
dcarmichael1: There is a default option variable in the reducer to push a default display value in the dropdown. That was how @pochretien implemented the first DropdownSearch

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearchReducer.tsx
```diff
@@ -0,0 +1,134 @@
...
+      };
+    case DropdownSearchActions.addCustomSelectedOption:
+      return {
+        ...state,
+        id: action.payload.id,
+        isOpened: true,
+        selectedOptions: addUniqueSelectedOption(state, action.payload.selectedOptionValue),
+        displayedOptions: getDisplayedOptions(state),
+      };
+    case DropdownSearchActions.onKeyDownMultiselect:
```
dcarmichael1: KeyCode must be in the payload otherwise how can we know what to do when the user presses a key? The other way would be to dispatch a specific action on a specific key but this would be handled in each component on key. The logic would either be in the reducer or in the component. We decided to put it in the reducer, but is it the wisest choice? I might go read on that.

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearchReducer.tsx
```diff
@@ -0,0 +1,134 @@
...
+          ...state,
+          id: action.payload.id,
+          isOpened: true,
+          selectedOptions: state.selectedOptions.push(state.activeOption),
+          activeOption: undefined,
+          filterText: '',
+          setFocusOnDropdownButton: true,
+          displayedOptions: getDisplayedOptions(state),
+        };
+      } else if (((action.payload.keyCode === keyCode.enter || action.payload.keyCode === keyCode.tab ) && !state.activeOption && state.filterText)) {
```
dcarmichael1: done

src/components/dropdownSearch/MultiSelectDropdownSearch/tests/MultiSelectDropdownSearchConnected.spec.tsx
```diff
@@ -0,0 +1,167 @@
...
+      wrapper.detach();
+    });
+
+    describe('mount and unmount', () => {
+
+      beforeEach(() => {
+        renderMultiSelectDropdownSearchConnected();
+      });
+
+      it('should call onMount prop when mounted', () => {
```
dcarmichael1: changed the test name

src/components/dropdownSearch/MultiSelectDropdownSearch/tests/MultiSelectDropdownSearchConnected.spec.tsx
```diff
@@ -0,0 +1,167 @@
...
+        expect(store.getState().dropdownSearch.length).toBe(0);
+      });
+    });
+
+    describe('mapDispatchToProps', () => {
+
+      beforeEach(() => {
+        renderMultiSelectDropdownSearchConnected();
+      });
+
```
dcarmichael1: that is to be consistent with how @pochretien implemented his Connected Components tests. It verifies if the component has all the method it needs when rendered.

src/components/dropdownSearch/MultiSelectDropdownSearch/tests/MultiSelectDropdownSearchConnected.spec.tsx
```diff
@@ -0,0 +1,167 @@
...
+      it('should toggle the close dropdown on blur', () => {
+        const dropdown = wrapper.find('.dropdown');
+
+        store.dispatch(closeDropdownSearch(id));
+
+        multiSelectDropdownSearchConnected.props().onBlur();
+
+        expect(dropdown.hasClass('open')).toBe(false, 'close the dropdown on blur');
```
dcarmichael1: removed it

src/utils/FixedQueue.ts
dcarmichael1: I would not because of a principle called "Tell don't ask". You might be interested by this Article: https://martinfowler.com/bliki/TellDontAsk.html

I want my data structure to handle all the logic that it needs to do instead of having a plain object that just contains data. The code is way easier to read that way!

src/utils/FixedQueue.ts
dcarmichael1: done

src/utils/FixedQueue.ts
dcarmichael1: I want to return a new instance of Fixed Queue on push to follow the redux immutable state

src/utils/FixedQueue.ts
dcarmichael1: done

src/utils/FixedQueue.ts
dcarmichael1: done

src/utils/FixedQueue.ts
dcarmichael1: good point, done

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearchReducer.tsx
MichelML: sorry little misunderstanding with the word extracting here , i meant referencing like in `const currentKeyCode = action.payload.keyCode;` i thought you could destructure the payload at first by doing `const {keyCode} = action.payload;` but that would have clashed with the keyCode variable already in the scope.

MichelML: the above suggestions were just for you to reduce the repetition in your code, if you apply them you would not have to write `actio.payload` and `keyCode.something` everywhere

MichelML: in short, i was not suggesting you move the KeyCode out of the payload :D.

src/components/dropdownSearch/MultiSelectDropdownSearch/tests/MultiselectDropdownSearchReducer.spec.tsx
```diff
@@ -0,0 +1,306 @@
...
+
+    it('should remove all from selectedOptions on "REMOVE_ALL_SELECTED_OPTIONS_MULTISELECT_DROPDOWN_SEARCH"', () => {
+      const action: IReduxAction<IOptionsDropdownSearchPayload> = {
+        type: DropdownSearchActions.removeAllSelectedOptions,
+        payload: defaultPayload,
+      };
+
+      const updatedState: IDropdownSearchState = multiSelectDropdownSearchReducer(oldState, action);
+
+      expect(updatedState.selectedOptions.getQueue.length).toBe(0);
```
MichelML: `getQueue()`

src/utils/FixedQueue.ts
MichelML: thanks for the article, interesting indeed. 

Trying to be a bit less academic and pragmatic here, the problem I see is that *you don't use your method `contains`  and `removeElement` anywhere at all*,  I suspect because you realized they don't work well with objects and arrays. So two choices here: 1) you reimplement the methods so they work with complex data structures, 2) you remove them along with their unit tests since you are not using them and sleep at night because we have everything we need to perform these actions in our stack anyways. 

For your `getFirstElement`, this is like saying you would need a method `getQueueLength()` to do `this.getQueue().length`...do we really want to abstract one liners like this all the time?

Here are other principles to reflect on:
1. Keep it simple stupid
2. Less code is better/Best code is no code
3. don't repeat yourself
4. https://softwareengineering.stackexchange.com/questions/202477/can-too-much-abstraction-be-bad

src/components/dropdownSearch/DropdownSearchReducers.tsx
MichelML: it returns a new FixedQueue object yes, but it keeps your first `instantiated this.queue` as a reference underneath all the time, in other words, you're not really creating a new queue, just a new reference to it.

MichelML: and if you have objects in your queue...those too will remain references and not new objects

That being said, I don't know how much side effects that would cause in your particular case.

src/components/dropdownSearch/DropdownSearchReducers.tsx
```diff
@@ -50,19 +50,16 @@ export const getNextIndexPosition = (array: any[], item: any, key: number): numb
   return index;
 };
 
-export const removeSelectedOption = (state: IDropdownSearchState, displayValue: string): FixedQueue<IDropdownOption> => {
-  return new FixedQueue<IDropdownOption>(_.filter(state.selectedOptions.getQueue(),
-    (selectedOption: IDropdownOption) => selectedOption.displayValue !== displayValue,
-  ));
-};
+export const removeSelectedOption = (state: IDropdownSearchState, displayValue: string): FixedQueue<IDropdownOption> =>
+  state.selectedOptions.removeElementsWithProperties({ displayValue, });
```
MichelML: no need for trailing comma if on one line

src/components/dropdownSearch/DropdownSearchReducers.tsx
```diff
@@ -50,19 +50,16 @@ export const getNextIndexPosition = (array: any[], item: any, key: number): numb
...
 
 export const addUniqueSelectedOption = (state: IDropdownSearchState, displayValue: string): FixedQueue<IDropdownOption> =>
-  removeSelectedOption(state, displayValue).push({ value: UUID.generate(), displayValue: displayValue });
+  removeSelectedOption(state, displayValue).push({ value: UUID.generate(), displayValue: displayValue, });
```
MichelML: same here, and you can use short syntax for displayValue

src/utils/FixedQueue.ts
```diff
@@ -33,13 +33,9 @@ export class FixedQueue<T> {
...
-  removeAtIndex(index: number) {
-    this.queue.splice(index, 1);
-  }
-
-  removeElement(element: any) {
-    const index = this.queue.indexOf(element);
-    this.removeAtIndex(index);
+  removeElementsWithProperties(properties: any): FixedQueue<T> {
+    const elementsWithoutProperties: Array<T> = _.reject(this.queue, properties);
+    return new FixedQueue<T> (elementsWithoutProperties);
```
MichelML: remove extra space before `(`

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearchReducer.tsx
```diff
@@ -86,15 +86,17 @@ export const multiSelectDropdownSearchReducer = (state: IDropdownSearchState = d
...
         };
       } else if ((_.contains([keyCode.enter, keyCode.tab], action.payload.keyCode) && !state.activeOption && state.filterText)) {
+        selectedOptions = addUniqueSelectedOption(state, state.filterText);
+        displayedOptions = getDisplayedOptions({ ...state, selectedOptions, });
         return {
           ...state,
           id: action.payload.id,
           isOpened: true,
-          selectedOptions: addUniqueSelectedOption(state, state.filterText),
+          selectedOptions: selectedOptions,
```
MichelML: you can use short syntax

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearchReducer.tsx
```diff
@@ -86,15 +86,17 @@ export const multiSelectDropdownSearchReducer = (state: IDropdownSearchState = d
...
           activeOption: undefined,
           filterText: '',
           setFocusOnDropdownButton: true,
-          displayedOptions: getDisplayedOptions(state),
+          displayedOptions: displayedOptions,
```
MichelML: same here

src/components/dropdownSearch/MultiSelectDropdownSearch/tests/SelectedOption.spec.tsx
sroy3: It is used to test the instance's methods, if you're not testing any, you can remove it.

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearchReducer.tsx
sroy3: It does not.
https://vincent.billey.me/pure-javascript-immutable-array/ "The mutative functions are push, pop, shift, unshift, sort, reverse, splice and delete (kind of)."

src/components/dropdownSearch/DropdownSearchReducers.tsx
sroy3: It's not about the side effects, it's about the state being read-only and immutable. You should be able to go backward and be able to see what the state was at any moment.

src/components/dropdownSearch/DropdownSearchReducers.tsx
sroy3: The default value should be an empty string and then you replace it with the new prop.

src/utils/FixedQueue.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  push(object: any): FixedQueue<T> {
+    const queue: Array<T> = [...this.queue, object];
+    return new FixedQueue<T>(queue, this.maxLength);
+  }
+
+  removeLastElement(): FixedQueue<T> {
+    const queue: Array<T> = [...this.queue];
+    if (this.queue.length > 0) {
+      queue.pop();
```
sroy3: You'll need to make this immutable as well.

src/components/dropdownSearch/DropdownSearchReducers.tsx
```diff
@@ -44,9 +50,21 @@ export const getNextIndexPosition = (array: any[], item: any, key: number): numb
   return index;
 };
 
-export const getOptionsFiltered = (state: IDropdownSearchState, filterText?: string) => {
-  const currentFilterText: string = filterText || state.filterText;
+export const removeSelectedOption = (state: IDropdownSearchState, displayValue: string): FixedQueue<IDropdownOption> =>
+  state.selectedOptions.removeElementsWithProperties({ displayValue });
+
+export const addUniqueSelectedOption = (state: IDropdownSearchState, displayValue: string): FixedQueue<IDropdownOption> =>
```
sroy3: If you have this with the unique / multiple distinction? What is the use of your `FixedQueue` util?

src/utils/FixedQueue.ts
sroy3: The more I look at it, the more I think it simply adds complexity for its use. @MichelML's arguments are excellent.

src/components/dropdownSearch/DropdownSearchReducers.tsx
dcarmichael1: Unique means that the option you add will be trimmed out if it already exists. It removes any instances of the same displayValue in the selectedOptions and then add the new one

dcarmichael1: Is it used only in multiselect to handle the fact that the user can manually enter a value that is already present

src/utils/FixedQueue.ts
dcarmichael1: I removed contains, removeElement, and replaced push by immutablePush

src/utils/FixedQueue.ts
```diff
@@ -0,0 +1,52 @@
...
+  immutablePush(object: any): FixedQueue<T> {
+    const queue: Array<T> = deepClone(this.queue);
+    queue.push(object);
+    return new FixedQueue<T>(queue, this.maxLength);
+  }
+
+  removeLastElement(): FixedQueue<T> {
+    const queue: Array<T> = deepClone(this.queue);
+    if (this.queue.length > 0) {
+      queue.pop();
```
sroy3: This needs to be immutable as well.

src/utils/FixedQueue.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  constructor(initialValues: Array<T> = [], maxLength: number = Number.MAX_VALUE) {
+    this.maxLength = maxLength;
+    this.queue = initialValues;
+    this.trimQueue();
+  }
+
+  immutablePush(object: any): FixedQueue<T> {
+    const queue: Array<T> = deepClone(this.queue);
+    queue.push(object);
```
sroy3: By sending `queue` to your new `FixedQueue`, you are still using the same reference. Honestly, this is getting a lot more complex than it needs to be. You are going through a maze of functions just to push an item into an array.

sroy3:
The immutability of your reducers still haven't been addressed either. For now you are only doing shallow copies of your state.

src/utils/FixedQueue.ts
dcarmichael1: How is it the same reference? queue is a clone of this.queue so it should be another reference?

dcarmichael1: I tested deepClone to be sure that it does not shallow copy, but copy each field of the object recursively. Same goes for arrays

sroy3: You are right, I somehow missed that the array was cloned as well. Still that is what I have been saying, a util class should simplify complex functions and what you are doing is complexifying simple function. It would be much cleaner and less error prone this way.

src/components/dropdownSearch/DropdownSearch.tsx
```diff
@@ -87,7 +102,8 @@ export class DropdownSearch extends React.Component<IDropdownSearchProps, void>
     )
       .map((option: IDropdownOption, index: number, options: IDropdownOption[]) => {
         const optionClasses = classNames({
```
rousselm4: Personnellement je sortirais `this.props.selectedOptions.getFirstElement()`

`firstElement = this.props.selectedOptions.getFirstElement();`

`'state-selected': firstElement && option.value === firstElement.value`

Rendu là ça coute pas plus chère de switcher l'ordre:

`'state-selected': firstElement && firstElement.value === option.value`

C'est de petits détails mais je trouve qu'on comprends plus facilement la logique ainsi

src/components/dropdownSearch/DropdownSearch.tsx
```diff
@@ -98,41 +114,45 @@ export class DropdownSearch extends React.Component<IDropdownSearchProps, void>
...
-            {this.getSvg(option)}
-            {valueToShow}
-          </span>
-        </li>;
+        return (
+          <li key={option.value}
+            className={liClasses}
+            title={value}>
+            <span className={optionClasses}
+              onMouseDown={(e: React.MouseEvent<HTMLSpanElement>) => this.handleOnOptionClick(e)} data-value={option.value}>
```
rousselm4: Je mettrais chaque props sur sa ligne

src/components/dropdownSearch/DropdownSearchReducers.tsx
```diff
@@ -76,46 +111,53 @@ export const dropdownSearchReducer = (state: IDropdownSearchState = dropdownSear
...
           ...state,
           id: action.payload.id,
-          selectedOption: state.activeOption,
           isOpened: false,
+          selectedOptions: state.selectedOptions.immutablePush(state.activeOption),
           activeOption: undefined,
+          filterText: '',
           setFocusOnDropdownButton: true,
         };
+      } else if (action.payload.keyCode === keyCode.escape) {
```
rousselm4: Tu répète souvent action.payload.keyCode. Je le sortirais dans une constante pour alleger le code

src/components/dropdownSearch/DropdownSearch.tsx
dcarmichael1: Le getFirstElement n'existe plus dans le refactor ahah :'D et j'ai corrigé le reste de tes commentaires dans le refactor! Merci! 👍

MichelML:
update coveo-styleguide to 1.16.x here too in package.json

src/components/dropdownSearch/DropdownSearchReducers.tsx
```diff
@@ -44,20 +48,88 @@ export const getNextIndexPosition = (array: any[], item: any, key: number): numb
...
+      selected: true,
+      custom: true,
+    });
+    return nextOptions;
+  }
+  return deepClone(options);
+};
+
+export const getDisplayedOptions = (options: IDropdownOption[]): IDropdownOption[] => {
+  return _.reject(options, (option) => {
```
MichelML: return _.reject(options, (option) => option.custom || option.hidden);

src/components/dropdownSearch/DropdownSearchReducers.tsx
```diff
@@ -44,20 +48,88 @@ export const getNextIndexPosition = (array: any[], item: any, key: number): numb
...
+    return option.custom || option.hidden;
+  });
+};
+
+export const getFilteredOptions = (state: IDropdownSearchState, filterText?: string) => {
+  const currentFilterText: string = filterText || state.filterText || '';
+  return _.filter(getDisplayedOptions(state.options),
     (option: IDropdownOption) => {
       const value = option.displayValue || option.value;
```
MichelML: const displayValue = option.displayValue || option.value; 

what you are handling here is the value displayed in the ui

MichelML:
tried the demo of the dropdown search and dropdown multiselect and both work well, as told to daniel it would be good to have a tool tip on the clear all selected options icon, other comments are mostly related to style

src/components/dropdownSearch/DropdownSearch.tsx
```diff
@@ -98,41 +124,45 @@ export class DropdownSearch extends React.Component<IDropdownSearchProps, void>
...
-  private getSvg(option: IDropdownOption): JSX.Element {
-    if (option.svg) {
-      return <span key={option.svg.name}
-        className='value-icon'>
-        <Svg {...option.svg} />
-      </span>;
+  protected getSvg(option: IDropdownOption): JSX.Element {
+    if (option && option.svg) {
+      return (
+        <span key={option.svg.name}
```
dcarmichael1: Curiously, I did not make any changes on this method, but it might have been moved around. I think PO made this in his original DropdownSearch

src/components/dropdownSearch/DropdownSearch.tsx
```diff
@@ -66,28 +73,47 @@ export class DropdownSearch extends React.Component<IDropdownSearchProps, void>
...
 
-  private getDropdownOptions(): JSX.Element[] {
-    const options = _.chain(this.props.options)
-      .filter(
-      (option: IDropdownOption) => {
+  protected getDropdownOptions(): JSX.Element[] {
+    const options = _.chain(this.getDisplayedOptions())
+      .filter((option: IDropdownOption) => {
         const value = option.displayValue || option.value;
         return _.isEmpty(this.props.filterText) || s.contains(value.toLowerCase(), this.props.filterText.toLowerCase());
```
dcarmichael1: done

src/components/dropdownSearch/DropdownSearch.tsx
dcarmichael1: done

src/components/dropdownSearch/DropdownSearchReducers.tsx
dcarmichael1: nice one-liner

src/components/dropdownSearch/DropdownSearchReducers.tsx
dcarmichael1: done

PR #302, opened 2017-08-01
Move pre commit to push

MichelML:
move pre commit to pre push to save time, and make the build fail if pretty typescript or the build itself fail

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -43,7 +43,7 @@ gulp.task('clean', 'Clean all', ['clean:dist', 'clean:docs', 'clean:tests']);
 // <editor-fold desc="Prettify">
 const prettify = (srcPaths, destPath) =>
   gulp.src(srcPaths)
-    .pipe(prettyTypescript())
+    .pipe(prettyTypescript({ emitError: true }))
     .pipe(gulp.dest(destPath));
 
 gulp.task('prettify:src', false, () => prettify(['src/**/*.ts', 'src/**/*.tsx'], 'src'));
--- a/package.json
+++ b/package.json
@@ -105,7 +105,7 @@
     "node-sass": "4.5.3",
     "postcss-loader": "2.0.6",
     "pre-commit": "1.2.2",
-    "pre-push": "^0.1.1",
+    "pre-push": "0.1.1",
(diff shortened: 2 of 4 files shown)


src/components/dropdownSearch/DropdownSearch.tsx
```diff
@@ -99,8 +99,8 @@ export class DropdownSearch extends React.Component<IDropdownSearchProps, void>
           : value;
 
         return <li key={option.value}
-                   className={liClasses}
-                   title={value}>
+          className={liClasses}
+          title={value}>
```
dcarmichael1: I changed this part on my PR! I also added Parenthesis around JSX object. You should revert those two lines because they are not related to this PR.

package.json
```diff
@@ -105,6 +105,7 @@
     "node-sass": "4.5.3",
     "postcss-loader": "2.0.6",
     "pre-commit": "1.2.2",
```
dcarmichael1: should we remove this dependency?

gulpfile.js
```diff
@@ -43,7 +43,7 @@ gulp.task('clean', 'Clean all', ['clean:dist', 'clean:docs', 'clean:tests']);
 // <editor-fold desc="Prettify">
 const prettify = (srcPaths, destPath) =>
   gulp.src(srcPaths)
-    .pipe(prettyTypescript())
+    .pipe(prettyTypescript({ emitError: true }))
```
dcarmichael1: what does that do?

MichelML: failing if finding errors

MichelML: https://github.com/coveo/pretty-typescript

package.json
MichelML: we could

gulpfile.js
dcarmichael1: Is there any user feedback so we now why we cannot push?

MichelML: yes

MichelML: example :

MichelML: screen shot 2017-08-01 at 1 40 15 pm

dcarmichael1: nice!

PR #311, opened 2017-08-03
dispatch close only when open

MichelML:
if you have many date pickers in a view, it triggers many CLOSE actions which cause the UI to be very very slow, we want to dispatch the action only if the datepicker is open

### DIFF
--- a/src/components/datePicker/DatePickerDropdown.tsx
+++ b/src/components/datePicker/DatePickerDropdown.tsx
@@ -77,7 +77,7 @@ export class DatePickerDropdown extends React.Component<IDatePickerDropdownProps
 
   private handleDocumentClick = (e: MouseEvent) => {
     let dropdown: HTMLDivElement = ReactDOM.findDOMNode<HTMLDivElement>(this.dropdown);
-    if (!dropdown.contains(e.target as Node)) {
+    if (!dropdown.contains(e.target as Node) && this.props.isOpened) {
       this.props.onDocumentClick();
       this.handleCancel();
     }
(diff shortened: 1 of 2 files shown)


src/components/datePicker/DatePickerDropdown.tsx
```diff
@@ -77,7 +77,7 @@ export class DatePickerDropdown extends React.Component<IDatePickerDropdownProps
 
   private handleDocumentClick = (e: MouseEvent) => {
     let dropdown: HTMLDivElement = ReactDOM.findDOMNode<HTMLDivElement>(this.dropdown);
-    if (!dropdown.contains(e.target as Node)) {
+    if (!dropdown.contains(e.target as Node) && this.props.isOpened) {
```
GermainBergeron: I don't know if we really care about performance but we should check for the for the isOpened before checking if the dropdown contains the target

src/components/datePicker/tests/DatePickerDropdown.spec.tsx
```diff
@@ -205,9 +205,28 @@ describe('Date picker', () => {
       expect(onDocumentClickSpy).toHaveBeenCalledTimes(1);
     });
 
+    it('should not trigger onDocumentClick dispatch on mount if prop onDocumentClick is set and isOpened is false', () => {
+      let onDocumentClickSpy = jasmine.createSpy('onDocumentClick');
+      let newDropdownProps = _.extend({}, DATE_PICKER_DROPDOWN_BASIC_PROPS, { onDocumentClick: onDocumentClickSpy, isOpened: false });
```
sroy3: Two const here

src/components/datePicker/DatePickerDropdown.tsx
MichelML: tested with multiple date pickers in a view and works really well

src/components/datePicker/tests/DatePickerDropdown.spec.tsx
MichelML: saw it after merged my bad :(

PR #312, opened 2017-08-04
render list elements only when dropdown is opened

MichelML:
- render list elements only when dropdown is opened 
- keep coverage to 100%  

this doesn't break anything and improves very very very significantly the speed of the rendering when multiple instances of these components are rendered into the DOM.


src/components/dropdownSearch/tests/DropdownSearch.spec.tsx
```diff
@@ -287,6 +291,34 @@ describe('DropdownSearch', () => {
         expect(dropdownSearch.find('.dropdown-toggle').prop('disabled')).toBe(true);
       });
```
MichelML: the following tests are not the best ones ever, although I had to write them to keep the coverage of 100%

src/components/datePicker/DatePickerDropdown.tsx
```diff
@@ -157,9 +157,13 @@ export class DatePickerDropdown extends React.Component<IDatePickerDropdownProps
         </footer>
       )
     };
-    let datePickerBox: JSX.Element = this.props.withReduxState
-      ? <DatePickerBox withReduxState id={this.props.id} {...datePickerBoxProps} />
-      : <DatePickerBox {...datePickerBoxProps} />;
+
+    let datePickerBox: JSX.Element = null;
```
vseguin: Have you tested thorougly that initializing this to null doesn't break elsewhere, especially since before it was always instantiated? Kind of nifty stuff

src/components/datePicker/tests/DatePickerDropdown.spec.tsx
```diff
@@ -56,7 +56,14 @@ describe('Date picker', () => {
       expect(datesSelectionBoxesProps).toEqual(DATE_PICKER_DROPDOWN_BASIC_PROPS.datesSelectionBoxes);
     });
 
-    it('should display a <DatePickerBox />', () => {
+    it('should not display a <DatePickerBox /> if it is not opened', () => {
+      expect(datePickerDropdown.find('DatePickerBox').length).toBe(0);
+    });
+
+    it('should display a <DatePickerBox /> if it is opened', () => {
+      let propsIsOpen: IDatePickerDropdownProps = _.extend({}, DATE_PICKER_DROPDOWN_BASIC_PROPS, { isOpened: true });
```
vseguin: Whole lot of nitpick, but the actual props is called isOpened, it should be the same in tests. (applies to the other ones too)

src/components/datePicker/tests/DatePickerDropdownConnected.spec.tsx
```diff
@@ -302,7 +309,10 @@ describe('Date picker', () => {
       expect(_.findWhere(store.getState().optionPickers, { id: pickerId }).selectedValue).toBe(newValue);
     });
 
-    it('should display a <DatePickerBox /> with a redux state a prop', () => {
+    it('should display a <DatePickerBox /> with a redux state a prop when the dropdown is opened', () => {
```
vseguin: This sentence doesn't make any sense :P

src/components/dropdownSearch/DropdownSearch.tsx
```diff
@@ -330,14 +331,20 @@ export class DropdownSearch extends React.Component<IDropdownSearchProps, void>
       width: this.props.width,
     };
 
-    return (
-      <div className={dropdownSearchClasses} style={dropdownSearchStyles}>
-        {this.getMainInput()}
+    const options: JSX.Element = !this.props.isOpened
+      ? null
```
vseguin: This is not very logical imho, would have done the inverse:

this.props.isOpened ? options : null

src/components/datePicker/DatePickerDropdown.tsx
MichelML: Thanks, I did, the only place it is used is in the JSX element that is returned by the method, we use null in JSX when we don't want any dom elements to be rendered on the page. Tried it in my use case as well, and others (e.g. log browser) and nothing breaks, it's just way faster to render.

src/components/dropdownSearch/DropdownSearch.tsx
MichelML: I agree that your suggestion is more readable, the logic here was simply to put the most common case first

PR #313, opened 2017-08-07
Refactoring in dropdown multiselect

dcarmichael1:
As discussed, I removed the fixed Queue and the state is all handled in the reducers. Now a dropdown state only contains one options array and IDropdownOption interface contains new values such as: selected, hidden and custom.


src/components/dropdownSearch/DropdownSearch.tsx
```diff
@@ -1,25 +1,25 @@
...
+import * as s from 'underscore.string';
 
 export interface IDropdownOption {
   svg?: ISvgProps;
   value: string;
   displayValue?: string;
   prefix?: string;
+  selected?: boolean;
+  custom?: boolean;
+  hidden?: boolean;
```
MichelML: quick notice, it seems like hidden and selected follows the same logic everywhere in your code, so it seems like duplication. Instead of using hidden where you use it, use !selected insteand, would it make sens ?

src/components/dropdownSearch/DropdownSearch.tsx
```diff
@@ -71,12 +71,25 @@ export class DropdownSearch extends React.Component<IDropdownSearchProps, void>
...
+    return _.findWhere(this.props.options, { selected: true });
+  }
+
+  protected getSelectedOptions(): IDropdownOption[] {
+    return _.where(this.props.options, { selected: true });
+  }
+
+  protected getDisplayedOptions(): IDropdownOption[] {
+    return _.reject(this.props.options, (option) => {
+      return option.custom || option.hidden;
```
MichelML: option.custom || !option.selected

src/components/dropdownSearch/DropdownSearch.tsx
dcarmichael1: No because in the old dropdown search that PO made, an option can be both selected and displayed. But in dropdown search multiselect, a selected option is hidden in the dropdown.

src/components/dropdownSearch/DropdownSearch.tsx
dcarmichael1: it would not work for the same reason as above

src/components/dropdownSearch/DropdownSearch.tsx
```diff
@@ -1,25 +1,25 @@
...
+  custom?: boolean;
+  hidden?: boolean;
 }
 
 export interface IDropdownSearchStateProps {
   isOpened?: boolean;
   filterText?: string;
   options?: IDropdownOption[];
-  displayedOptions?: IDropdownOption[];
-  selectedOptions?: FixedQueue<IDropdownOption>;
```
MichelML: no selectedOption in the state anymore ?

MichelML: nevermind

src/components/dropdownSearch/DropdownSearch.tsx
```diff
@@ -28,7 +28,7 @@ export interface IDropdownSearchOwnProps extends React.ClassAttributes<DropdownS
   id: string;
   modMenu?: boolean;
   defaultOptions?: IDropdownOption[];
-  defaultSelectedOptions?: FixedQueue<IDropdownOption>;
+  defaultSelectedOption?: IDropdownOption;
```
MichelML: do you really need defaultSelectedOption since you can pass a selected option in your defaultOptions ?

src/components/dropdownSearch/DropdownSearchReducers.tsx
```diff
@@ -4,17 +4,15 @@ import { DropdownSearchActions, IOptionsDropdownSearchPayload } from './Dropdown
 import * as _ from 'underscore';
 import * as s from 'underscore.string';
 import { keyCode } from '../../utils/InputUtils';
-import { FixedQueue } from '../../utils/FixedQueue';
 import { multiSelectDropdownSearchReducer } from './MultiSelectDropdownSearch/MultiSelectDropdownSearchReducer';
 import { UUID } from '../../utils/UUID';
+import { deepClone } from '../../utils/CloneUtils';
 
 export interface IDropdownSearchState {
```
MichelML: are IDropdownSearchStateProps and IDropdownSearchState interfaces supposed to be identical ? if so i would suggest keeping one of the two only.

src/components/dropdownSearch/DropdownSearch.tsx
dcarmichael1: This is to make the DropdownSearch have a fake option by default that says something like: "Select option". That was how the component PO made worked. Maybe will we need to make this as a text placeholder instead

src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearchReducer.tsx
```diff
@@ -3,14 +3,11 @@ import {
   addUniqueSelectedOption,
   dropdownSearchInitialState,
   dropdownSearchReducer,
-  getDisplayedOptions,
   getNextIndexPosition,
   getFilteredOptions,
-  IDropdownSearchState, removeSelectedOption,
+  IDropdownSearchState, removeSelectedOption, removeAllSelectedOption, multiSelectOption, removeLastSelectedOption,
 } from '../DropdownSearchReducers';
```
MichelML: put each imported entity on its own line

src/components/dropdownSearch/DropdownSearch.tsx
MichelML: ok 👍

src/components/dropdownSearch/DropdownSearch.tsx
MichelML: ok 👍

src/components/dropdownSearch/DropdownSearch.tsx
MichelML: make sure it's well tested so that if we remove it by error the uts will break

src/components/dropdownSearch/DropdownSearchReducers.tsx
```diff
@@ -50,27 +50,84 @@ export const getNextIndexPosition = (array: any[], item: any, key: number): numb
...
       const value = option.displayValue || option.value;
       return _.isEmpty(currentFilterText) || s.contains(value.toLowerCase(), (currentFilterText).toLowerCase());
     });
 };
 
+export const selectSingleOption = (options: IDropdownOption[], selectedOption: IDropdownOption): IDropdownOption[] => {
+  return _.map(options, (option: IDropdownOption) => {
+    const nextOption = deepClone(option);
+    nextOption.selected = option.value === selectedOption.value;
+    return nextOption;
```
rousselm4: très clean la solution à Michel. Je rajouterais d'enlever le return:
`return _.map(options, (option: IDropdownOption) => _.extend(deepClone(option), { selected: option.value === selectedOption.value }));`

src/components/dropdownSearch/DropdownSearchReducers.tsx
```diff
@@ -50,27 +50,84 @@ export const getNextIndexPosition = (array: any[], item: any, key: number): numb
   return index;
 };
 
-export const removeSelectedOption = (state: IDropdownSearchState, displayValue: string): FixedQueue<IDropdownOption> =>
-  state.selectedOptions.removeElementsWithProperties({ displayValue });
+export const removeSelectedOption = (options: IDropdownOption[], displayValue: string): IDropdownOption[] => {
```
dcarmichael1: That's better 👍

src/components/dropdownSearch/DropdownSearchReducers.tsx
```diff
@@ -50,27 +50,84 @@ export const getNextIndexPosition = (array: any[], item: any, key: number): numb
...
+      nextOptions.splice(nextOptions.indexOf(selectedOption), 1);
+    } else {
+      selectedOption.selected = false;
+      selectedOption.hidden = false;
+    }
+  }
+  return nextOptions;
+};
+
+export const removeLastSelectedOption = (options: IDropdownOption[]): IDropdownOption[] => {
```
dcarmichael1: done

src/components/dropdownSearch/DropdownSearchReducers.tsx
```diff
@@ -50,27 +50,84 @@ export const getNextIndexPosition = (array: any[], item: any, key: number): numb
...
+  const nextOptions: IDropdownOption[] = deepClone(options);
+  const lastSelectedOption: IDropdownOption = _.find(nextOptions.reverse(), { selected: true });
+  if (lastSelectedOption) {
+    return removeSelectedOption(options, lastSelectedOption.displayValue);
+  } else {
+    return nextOptions.reverse();
```
dcarmichael1: Your solution works better thanks!

src/components/dropdownSearch/DropdownSearchReducers.tsx
```diff
@@ -50,27 +50,84 @@ export const getNextIndexPosition = (array: any[], item: any, key: number): numb
...
+  }
+};
 
-export const addUniqueSelectedOption = (state: IDropdownSearchState, displayValue: string): FixedQueue<IDropdownOption> =>
-  removeSelectedOption(state, displayValue).immutablePush({ value: UUID.generate(), displayValue });
+export const removeAllSelectedOption = (options: IDropdownOption[]): IDropdownOption[] => {
+  return _.map(options, (option: IDropdownOption) => {
+    const nextOption = deepClone(option);
+    nextOption.selected = false;
+    nextOption.hidden = false;
```
dcarmichael1: Hidden is needed!

src/components/dropdownSearch/DropdownSearchReducers.tsx
```diff
@@ -50,27 +50,84 @@ export const getNextIndexPosition = (array: any[], item: any, key: number): numb
...
+export const getDisplayedOptions = (options: IDropdownOption[]): IDropdownOption[] => {
+  return _.reject(options, (option) => {
+    return option.custom || option.hidden;
+  });
 };
 
 export const getFilteredOptions = (state: IDropdownSearchState, filterText?: string) => {
   const currentFilterText: string = filterText || state.filterText;
-  return _.filter(getDisplayedOptions(state),
+  return _.filter(getDisplayedOptions(state.options),
```
dcarmichael1: This function does not alter the state so it's ok to not make it immutable I think. PO made this function to get the options that matches the filtertext, but that does not change the dropdown state

src/components/dropdownSearch/DropdownSearchReducers.tsx
dcarmichael1: thanks les boys!

src/components/dropdownSearch/DropdownSearchReducers.tsx
```diff
@@ -50,27 +50,84 @@ export const getNextIndexPosition = (array: any[], item: any, key: number): numb
...
+export const selectSingleOption = (options: IDropdownOption[], selectedOption: IDropdownOption): IDropdownOption[] => {
+  return _.map(options, (option: IDropdownOption) => {
+    const nextOption = deepClone(option);
+    nextOption.selected = option.value === selectedOption.value;
+    return nextOption;
+  });
+};
+
+export const multiSelectOption = (options: IDropdownOption[], selectedOption: IDropdownOption): IDropdownOption[] => {
+  return _.map(options, (option: IDropdownOption) => {
```
dcarmichael1: the difference is that selectSingle deselect all the other options.

dcarmichael1: The option needs to be hidden when multi-selected in this case so your proposition would not work here.

src/components/dropdownSearch/DropdownSearchReducers.tsx
```diff
@@ -127,8 +184,7 @@ export const dropdownSearchReducer = (state: IDropdownSearchState = dropdownSear
       return {
         ...state,
         id: action.payload.id,
-        options: action.payload.optionsDropdown,
-        selectedOptions: new FixedQueue<IDropdownOption>([defaultSelectedOption], 1),
+        options: action.payload.optionsDropdown ? action.payload.optionsDropdown.concat(defaultSelectedOption) : [defaultSelectedOption],
```
dcarmichael1: Good point... will investigate

src/components/dropdownSearch/DropdownSearch.tsx
dcarmichael1: Selected is now a props of IDropdownOption

src/components/dropdownSearch/DropdownSearchReducers.tsx
dcarmichael1: The renaming is done, but for the defaultSelectedOption I think we might answer this corner case another time if it becomes a real issue because for now it does not affect the multiselect at all.

PR #323, opened 2017-08-17
*fixes wrong class preventing action bar actions from being displayed

sbelzile-Coveo:


MichelML:
you have one test failing because we were using this class in a test, just change it to the new one in the test file.  

screen shot 2017-08-17 at 8 52 56 am

PR #334, opened 2017-08-28
fix tooltip footer related bug

MichelML:


MichelML:
demo when the build will complete http://rawgit.com/coveo/react-vapor/fix-tooltip-footer/docs/index.html

sroy3:
Is it normal that `docs/assets/bundle.js` is versioned? Isn't it supposed to be in the .gitignore? I doubt it's your change that did this though, but we might as well add it back.

MichelML:
on hover of the tooltip with a footer you should not see any error in the console

PR #336, opened 2017-09-05
Fix active option on non empty filtertext

MichelML:
how to test:  
Enter a non empty text in the input to have at least two options in the dropdown, press the down arrow once, the active option should be the second option in the filtered options.  

Try it for a default dropdown, and with the dropdown having the supportSingleCustomOption option

demo: https://rawgit.com/coveo/react-vapor/fix-active-option-on-non-empty-filtertext/docs/index.html


sroy3:
Are you able to fix the drop in coverage?

PR #337, opened 2017-09-06
Input improvements + example

sbelzile-Coveo:
+ Allow modifying input type
+ Allow classes on inputs inner input
+ Allow specifying valid and invalid messages
+ Adds example


src/components/input/Input.tsx
```diff
@@ -37,18 +40,22 @@ export class Input extends React.Component<IInputProps, any> {
   render() {
     const classes = ['input-wrapper'].concat(this.props.classes);
     const labelClasses = [].concat(this.props.labelClasses);
+    const innerInputClasses = [].concat(this.props.innerInputClasses);
```
sroy3: We have been using the package "classnNames" to generate classes for a little while. Might be a good idea to use it here.

src/components/input/Input.tsx
```diff
@@ -37,18 +40,22 @@ export class Input extends React.Component<IInputProps, any> {
...
 
     return (
       <div className={classes.join(' ')}>
         <input
-          type='text'
+          className={innerInputClasses.join(' ')}
+          type={this.props.type ? this.props.type : 'text'}
```
sroy3: You can set "text" as the default prop to type instead

src/components/input/examples/InputExamples.tsx
```diff
@@ -0,0 +1,51 @@
...
+    };
+  }
+
+  private onChange() {
+    const inputValue = this.innerInput.getInnerValue();
+    if (!inputValue) {
+      this.setInputClasses('invalid');
+    } else {
+      this.setInputClasses('valid');
+    }
```
sroy3: `this.setInputClasses(inputValue ? 'valid' : 'invalid');`

src/components/input/tests/Input.spec.tsx
```diff
@@ -31,6 +31,47 @@ describe('Input', () => {
       input.detach();
     });
 
+    it('should set inner input classes when specified', () => {
```
sroy3: All `let`s should be changed to `const`s

src/components/input/Input.tsx
```diff
@@ -1,19 +1,27 @@
...
   label?: string;
+  type?: string;
   classes?: string[];
+  innerInputClasses?: string[];
   labelClasses?: string[];
   value?: string;
   placeholder?: string;
   onChange?: (value: string) => void;
   onKeyUp?: (event: React.KeyboardEvent<HTMLInputElement>) => void;
-  validate?: (value: string) => string;
```
GermainBergeron: validate was unused?

src/components/input/tests/Input.spec.tsx
```diff
@@ -31,6 +31,47 @@ describe('Input', () => {
...
+      expect(innerLabel.prop('data-valid-message')).toBe(message);
+    });
+
+    it('should set inner label invalid message when specified', () => {
+      const message = 'salut';
+      const innerLabel = input.find('label').first();
+      expect(innerLabel.prop('data-invalid-message')).toBe(undefined);
+
+      input.setProps({ invalidMessage: message });
+      input.mount();
```
GermainBergeron: you can chain the calls and mount the input right after setting the props:

`input.setProps({ invalidMessage: message }).mount();`

PR #340, opened 2017-09-11
make options available in the payload on dropdown search close

MichelML:
I currently have a use case where i need the MultiselectDropdownSearch options in scope on a close action. I use the options to set the proper values in another section of the state of the application (cloud v2), but not the options themselves.


MichelML:
For info, this does not change anything in the behavior of the component, it just makes the options available in the action payload, in case an external application may want to "listen" on this action and use the options in question in the payload.

PR #341, opened 2017-09-13
add the box item

pochretien:
add uts for the box item
add examples for the box item
dont show the tooltip if the title is empty
add uts for the modif on the tooltip title


pochretien:
image
image

src/components/boxItem/BoxItem.tsx
```diff
@@ -0,0 +1,76 @@
...
+        'hidden': this.props.hidden,
+      },
+      this.props.classes);
+  }
+
+  protected getDropdownPrepend(prefix: string): JSX.Element {
+    if (prefix) {
+      return <span className='text-medium-grey mr1'>{prefix}</span>;
+    }
+    return null;
```
sroy3: `return prefix ? <span className='text-medium-grey mr1'>{prefix}</span> : null;`

src/components/boxItem/BoxItem.tsx
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  protected getSvg(svg: ISvgProps): JSX.Element {
+    if (svg) {
+      return (<Svg {..._.defaults(svg, { className: 'inline-flex mr1', svgClass: 'icon' }) } />);
+    }
+    return null;
```
sroy3: `return svg ? <Svg {..._.defaults(svg, { className: 'inline-flex mr1', svgClass: 'icon' }) } /> : null;`

src/components/boxItem/BoxItem.tsx
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  handleOnOptionClick = (e: React.MouseEvent<HTMLSpanElement>) => {
+    if (e.target) {
+      if (this.props.onOptionClick) {
```
sroy3: Merge the 2 if statements

src/components/boxItem/BoxItem.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  protected getSvg(svg: ISvgProps): JSX.Element {
+    if (svg) {
+      return (<Svg {..._.defaults(svg, { className: 'inline-flex mr1', svgClass: 'icon' }) } />);
+    }
+    return null;
+  }
+
+  handleOnOptionClick = (e: React.MouseEvent<HTMLSpanElement>) => {
+    if (e.target) {
```
sroy3: When is `e.target` not defined?

src/components/boxItem/tests/BoxItem.spec.tsx
```diff
@@ -0,0 +1,130 @@
...
+        { attachTo: document.getElementById('App') },
+      );
+    };
+
+    it('should render the display value', () => {
+      const displayValue: string = 'display value';
+      renderBoxItem({
+        displayValue,
+      });
+      expect(boxItemComponent.find('li').text()).toEqual(displayValue);
```
sroy3: `toBe` instead of `toEqual`

src/components/boxItem/BoxItem.tsx
pochretien: validation removed, I dont need the e, I use the this.props :)

src/components/itemBox/ItemBox.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+export class ItemBox extends React.Component<IItemBoxProps, void> {
+
+  static defaultProps: Partial<IItemBoxProps> = {
+    tooltip: {
+      title: '',
+    },
+  };
+
+  getClasses(): string {
```
GermainBergeron: private?

src/components/itemBox/ItemBox.tsx
```diff
@@ -0,0 +1,61 @@
...
+      this.props.onOptionClick(this.props);
+    }
+  }
+
+  render() {
+    return (
+      <Tooltip {...this.props.tooltip}>
+        <li
+          className={this.getClasses()}
+          onMouseDown={() => this.handleOnOptionClick()}
```
GermainBergeron: onClick instead of mouse down?

src/components/itemBox/examples/ItemBoxExamples.tsx
```diff
@@ -0,0 +1,99 @@
...
+      placement: 'bottom',
+      container: 'body',
+    };
+
+    const svg: ISvgProps = {
+      svgName: 'domain-google',
+      svgClass: 'icon',
+    };
+    const getSvgPrepend: IContentProps = { content: () => <Svg {...svg} />, classes: ['mr1'] };
+    const getSvgAppend: IContentProps = { content: () => <Svg {...svg} />, classes: ['ml1'] };
```
GermainBergeron: rename for `svgPrepend` and `svgAppend` as it's not a getter but a simple var

src/components/tooltip/Tooltip.tsx
```diff
@@ -81,12 +83,15 @@ export class Tooltip extends React.Component<ITooltipProps, any> {
...
+    if (this.props.title) {
+      return (
+        <OverlayTrigger {..._.omit(this.props, OVERLAY_PROPS_TO_OMIT) } overlay={tooltip}>
+          <span className={this.props.className}>
+            {this.props.children}
+          </span>
+        </OverlayTrigger>
+      );
+    }
+    return (<span>{this.props.children}</span>);
```
GermainBergeron: Still need the span? If you do, should you add the same className as the span in the overlay?

src/components/itemBox/ItemBox.tsx
pochretien: I got a problem with the onClick when we try to use it with other component like dropdown. Its a problem with the event order with the onBlur

src/components/tooltip/Tooltip.tsx
pochretien: span removed

src/components/content/Content.tsx
```diff
@@ -0,0 +1,24 @@
...
+  static defaultProps: Partial<IContentProps> = {
+    classes: [],
+  };
+
+  private getContent(): JSX.Element | string {
+    return _.isString(this.props.content) ? this.props.content : <this.props.content />;
+  }
+
+  render() {
+    return (<span className={this.props.classes.join(' ')}>
```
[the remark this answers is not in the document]
GermainBergeron: What don't you like about the package? I would use it everywhere in react-vapor to have only one way of doing things. The lib is quite nice when you need to set some class depending on some booleans or having a default class and still having the classes props.

If we don't use it we need to set the defaultProps and use a join, things that we then need to add tests for. The library is already tested which is great

PR #342, opened 2017-09-13
update dependencies

pochretien:


src/components/dropdownSearch/MultiSelectDropdownSearch/MultiSelectDropdownSearch.tsx
```diff
@@ -10,8 +10,6 @@ export class MultiSelectDropdownSearch extends DropdownSearch {
     createOptionText: 'Create option for ',
     isOpened: false,
     options: [],
-    displayedOptions: [],
-    selectedOptions: [],
```
sroy3: Is this part of another feature/fix or these are simply not used?

pochretien: Interface problem with typescript, fix incoming

PR #365, opened 2017-09-28
Add update to the dropdown search options on mount 

sroy3:
If there is a default selected option since it is currently being overwritten by the reducer, with the update on mount, we are now showing the custom option. This is a quick fix until the dropdown search is being refactored. The reducer should not initialized itself with options that should be customizable.


sroy3:
Strangely, the decrease in coverage comes from a file I haven't touch.

PR #366, opened 2017-09-29
Add flat select

pochretien:
image


src/components/flatSelect/FlatSelectReducers.ts
```diff
@@ -0,0 +1,54 @@
...
+export const flatSelectInitialState: IFlatSelectState = { id: undefined, selectedOption: undefined };
+export const flatSelectsInitialState: IFlatSelectState[] = [];
+
+export const flatSelectReducer = (state: IFlatSelectState = flatSelectInitialState,
+  action: (IReduxAction<IFlatSelectActionPayload>)): IFlatSelectState => {
+  switch (action.type) {
+    case FlatSelectActions.select:
+      if (state.id !== action.payload.id) {
+        return state;
+      }
```
GermainBergeron: The id check is duplicated in `flatSelectsReducer`. I don't know which one you should keep but I don't think you need both

src/components/flatSelect/examples/FlatSelectExamples.tsx
```diff
@@ -0,0 +1,129 @@
...
+        option: { content: 'Option 3' },
+      },
+    ];
+
+    const svg: ISvgProps = {
+      svgName: 'domain-google',
+      svgClass: 'icon',
+    };
+    const getPrepend: IContentProps = { content: () => <Svg {...svg} />, classes: ['mr1'] };
+    const getAppend: IContentProps = { content: () => <Svg {...svg} />, classes: ['ml1'] };
```
GermainBergeron: I think I already said this in another review but those two variables should be `const prepend` and `const append` since they are not getters

src/components/flatSelect/examples/FlatSelectExamples.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+    return (
+      <div className='mt2'>
+        <h1 className='text-blue mb1'>Flat Select List</h1>
+
+        <div className='form-group'>
+          <label className='form-control-label'>Default Flat Select</label>
+          <div className='form-control'>
+            <FlatSelectConnected {...{
+              id: UUID.generate(),
```
GermainBergeron: FlatSelectConnected should probably generate it's own id when added to the store (in your reducer)

src/components/flatSelect/FlatSelectOption.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export interface IFlatSelectOptionProps {
+  option: IContentProps;
+  classes?: string[];
+  prepend?: IContentProps;
+  append?: IContentProps;
+  tooltip?: ITooltipProps;
+  selected?: boolean;
+  onClick?: (option: IFlatSelectOptionProps) => void;
+}
```
GermainBergeron: no `id` or `value`? You get the value by the option which is the content?

src/components/flatSelect/tests/FlatSelectConnected.spec.tsx
```diff
@@ -0,0 +1,143 @@
...
+  describe('<FlatSelectConnected />', () => {
+    let wrapper: ReactWrapper<any, any>;
+    let flatSelect: ReactWrapper<IFlatSelectProps, void>;
+    let store: Store<IReactVaporState>;
+
+    let id: string = 'flatSelect';
+    let defaultOptions: IFlatSelectOptionProps[] = [{
+      option: { content: 'test' },
+    }, {
+      option: { content: 'test 1' },
```
GermainBergeron: Have you tried putting JSX or a class instead of a string in content? I'm not sure everything will go as expected

src/components/flatSelect/tests/FlatSelectOption.spec.tsx
```diff
@@ -0,0 +1,95 @@
+import { shallow, mount, ReactWrapper } from 'enzyme';
+import * as React from 'react';
+import * as _ from 'underscore';
+import { FlatSelectOption, IFlatSelectOptionProps } from '../FlatSelectOption';
+import { Content, IContentProps } from '../../content/Content';
+import { ISvgProps, Svg } from '../../svg/Svg';
+
+describe('FlatSelect', () => {
+  describe('<Checkbox />', () => {
```
GermainBergeron: checkbox?

src/components/flatSelect/tests/FlatSelectOption.spec.tsx
```diff
@@ -0,0 +1,95 @@
...
+    it('should render without errors', () => {
+      expect(() => {
+        shallow(
+          <FlatSelectOption option={{ content: 'test' }} />,
+        );
+      }).not.toThrow();
+    });
+  });
+
+  describe('<Checkbox />', () => {
```
GermainBergeron: Checkbox?

src/components/flatSelect/tests/FlatSelectOption.spec.tsx
```diff
@@ -0,0 +1,95 @@
...
+    let flatSelect: ReactWrapper<IFlatSelectOptionProps, any>;
+    let defaultOption: IContentProps = { content: 'test' };
+    let defaultProps: IFlatSelectOptionProps = {
+      option: defaultOption,
+    };
+    const svg: ISvgProps = {
+      svgName: 'domain-google',
+      svgClass: 'icon',
+    };
+    const getContent: IContentProps = { content: () => <Svg {...svg} />, classes: ['mr1'] };
```
GermainBergeron: `const content`

src/components/flatSelect/FlatSelectReducers.ts
```diff
@@ -0,0 +1,54 @@
...
+  selectedOption?: IFlatSelectOptionProps;
+}
+
+export const flatSelectInitialState: IFlatSelectState = { id: undefined, selectedOption: undefined };
+export const flatSelectsInitialState: IFlatSelectState[] = [];
+
+export const flatSelectReducer = (state: IFlatSelectState = flatSelectInitialState,
+  action: (IReduxAction<IFlatSelectActionPayload>)): IFlatSelectState => {
+  switch (action.type) {
+    case FlatSelectActions.select:
```
MichelML: you could use a ternary for this 

```
return state.id !== action.payload.id 
  ? state  
  : {...state, selectedOption: action.payload.selectedOption};  
```

src/components/flatSelect/FlatSelectReducers.ts
```diff
@@ -0,0 +1,54 @@
...
+      if (state.id !== action.payload.id) {
+        return state;
+      }
+
+      return {
+        id: state.id,
+        selectedOption: action.payload.selectedOption,
+      };
+    case FlatSelectActions.add:
+      return _.extend({}, action.payload);
```
MichelML: you don't need underscore and could do `return {...action.payload};`

src/components/flatSelect/examples/FlatSelectExamples.tsx
MichelML: I disagree, especially if you compose a component out of other components, it's way better to have ids set as prop if you want to know which component are linked together. Suppose you have a table with a filter, and want to do modification to the tableState based on a filterAction, you must know which id this filter has in the store to perform state modification only for this filter and not other ones, and I think you can only have that by having control over the ids of the components.

src/components/flatSelect/FlatSelect.tsx
```diff
@@ -0,0 +1,72 @@
...
+  optionPicker?: boolean;
+  onClick?: (option: IFlatSelectOptionProps) => void;
+}
+
+export interface IFlatSelectStateProps {
+  selectedOption?: IFlatSelectOptionProps;
+}
+
+export interface IFlatSelectDispatchProps {
+  onMount?: () => void;
```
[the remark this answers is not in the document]
GermainBergeron: It's only for consistency, it's the same thing otherwise

src/components/flatSelect/examples/FlatSelectExamples.tsx
GermainBergeron: You're absolutely right. Disregard my comment

src/Index.ts
```diff
@@ -30,6 +30,13 @@ export { FacetConnected } from './components/facets/FacetConnected';
 export { facetsReducer } from './components/facets/FacetReducers';
 export { changeFacet, emptyFacet, FacetActions } from './components/facets/FacetActions';
 
+// FlatSelect
+
```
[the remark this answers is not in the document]
pochretien: done!

src/components/flatSelect/FlatSelect.tsx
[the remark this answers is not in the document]
pochretien: Changed for onRender

src/components/flatSelect/FlatSelect.tsx
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  private handleOnOptionClick(option: IFlatSelectOptionProps) {
+    if (this.props.onOptionClick) {
+      this.props.onOptionClick(option);
+    }
+
+    if (this.props.onClick) {
+      this.props.onClick(option);
+    }
```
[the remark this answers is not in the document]
pochretien: the onClick can be used by the an external call if you need to do something when you the flat select change. the onOptionClick is used by the connected component. if a external call is sent in props. he will override the call used by the connected component. We need to front to call. onClick is defined in the OwnProps and onOptionClick in the DispatchProps

src/components/flatSelect/FlatSelect.tsx
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  private getOptions(): JSX.Element[] {
+    return _.map(this.props.options, (flatSelectOption: IFlatSelectOptionProps, index: number) => {
+      flatSelectOption.selected = this.props.selectedOption && this.props.selectedOption.option === flatSelectOption.option;
+      flatSelectOption.onClick = (option: IFlatSelectOptionProps) => this.handleOnOptionClick(option);
+
+      return <FlatSelectOption key={index} {...flatSelectOption} />;
```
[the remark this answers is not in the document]
pochretien: I dont have a value because the option is a <Content />

src/components/flatSelect/FlatSelectOption.tsx
pochretien: I add an id for the flatSelectOption

src/components/flatSelect/FlatSelectReducers.ts
pochretien: removed

src/components/flatSelect/tests/FlatSelectConnected.spec.tsx
pochretien: image
Work!

src/components/flatSelect/FlatSelectReducers.ts
pochretien: id validation removed in this reducer

src/components/flatSelect/FlatSelectReducers.ts
pochretien: I got a error ts caused by IReduxAction. I will keep the _.extend

PR #382, opened 2017-10-18
Added a description icon to the blankslate

sroy3:
screen shot 2017-10-18 at 11 22 26 am


sroy3:
Adding an icon there will break uniformity with the other blankslate. It is not needed at the moment.

PR #395, opened 2017-11-02
Link svg

pochretien:
add example for Svg
add new component to render a link svg with a tooltip
add example for this new component
add uts for the new component
move svg test in the right folder

image


sroy3:
I'm not a 100% sure, but I think you can use the component `LinkAction` to get the same result. You can also update if it's missing something.

src/components/svg/LinkSvg.tsx
```diff
@@ -0,0 +1,36 @@
+import * as React from 'react';
+import { ISvgProps, Svg } from './Svg';
+import * as classname from 'classnames';
+import { ITooltipProps, Tooltip } from '../tooltip/Tooltip';
+
+export interface ILink {
+  url: string;
+  target?: string;
+}
```
GermainBergeron: Is it usefull to have an object in the LinkSVGProps instead of two properties? The default is probably not deep (it won't set the target if you pass `{link: {url: 'testurl'}}` )

src/components/svg/LinkSvg.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+export class LinkSvg extends React.Component<ILinkSvgProps, {}> {
+  static defaultProps: Partial<ILinkSvgProps> = {
+    link: {
+      url: '',
+      target: '_blank',
+    },
+    linkClasses: [],
+    svg: { svgName: 'help', svgClass: `icon` },
```
GermainBergeron: same thing for the SVG, it should probably be 2 properties instead of an object

src/components/svg/tests/LinkSvg.spec.tsx
```diff
@@ -0,0 +1,104 @@
...
+          linkClasses: ['test', 'test1'],
+        },
+      );
+
+      const element = linkSvgComponent.find('a');
+      expect(element.hasClass('test')).toBe(true);
+      expect(element.hasClass('test1')).toBe(true);
+    });
+
+    it('should add custom class with linkClasses on <a>', () => {
```
GermainBergeron: twice the same test?

src/components/svg/LinkSvg.tsx
pochretien: I think for the SVG its better to use the Interface created by the `<SVG />`. This is why a added the svg option like an object

src/components/svg/LinkSvg.tsx
pochretien: I will add this 2 props in the ILinkSvgProps

src/components/svg/tests/LinkSvg.spec.tsx
pochretien: removed

pochretien:
@sroy3 I checked the link Action and to do the same thing, the problem I found is the class used for action style.  Without that. I can do the same result but I think the LinkAction is more specific for a action because of the style. I can use the LinkSVG after to build a linkAction. What do you think ?

src/components/svg/LinkSvg.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { ISvgProps, Svg } from './Svg';
+import * as classname from 'classnames';
+import { ITooltipProps, Tooltip } from '../tooltip/Tooltip';
+
+export interface ILinkSvgProps extends React.ClassAttributes<LinkSvg> {
+  url: string;
+  target?: string;
+  linkClasses?: string[];
+  svg?: ISvgProps;
+  tooltip?: ITooltipProps;
```
GermainBergeron: This is optional but always in the render? It works fine without any props?

pochretien: I modified the tooltip 1 month ago to render the children in a span if the prop title is empty or undefined to remove the ternary every time we want a tooltips or not to wrap the element

PR #399, opened 2017-11-08
export filter actions

MichelML:


GermainBergeron:
I wonder if we should export everything like we do in jsadmin-x projects

PR #400, opened 2017-11-08
allow the possibility to add an input max width

MichelML:
add with title feature

allow the possibility to add an input max width

screen shot 2017-11-08 at 3 29 59 pm 2

screen shot 2017-11-08 at 2 36 03 pm

### DIFF
--- a/src/components/filterBox/FilterBox.tsx
+++ b/src/components/filterBox/FilterBox.tsx
@@ -9,6 +9,8 @@ export interface IFilterBoxOwnProps extends React.ClassAttributes<FilterBox> {
   onBlur?: () => void;
   onKeyDown?: (e: React.KeyboardEvent<HTMLInputElement>) => void;
   isAutoFocus?: boolean;
+  maxWidth?: number;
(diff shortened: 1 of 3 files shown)


GermainBergeron:
Should the input have ellipsis ( `...` ) to show that it's cropped ?

MichelML:
@GermainBergeron I don't think so, you can navigate in it, that's an input

src/components/filterBox/FilterBox.tsx
```diff
@@ -79,8 +81,14 @@ export class FilterBox extends React.Component<IFilterBoxProps, any> {
     let filterPlaceholder = this.props.filterPlaceholder || FILTER_PLACEHOLDER;
     let filterBoxContainerClasses = classNames('filter-container', this.props.containerClasses);
 
+    const inputMaxWidth = { maxWidth: `${this.props.maxWidth}px` };
```
GermainBergeron: `const` should appear before let (maybe the `let` can be refactored to const?)

PR #401, opened 2017-11-09
add feature to toggle default classes for action bar and a tooltip on item filter

MichelML:
add feature to toggle default classes for action bar
add possibility to add a tooltip on item in item filter
add examples
add unit tests

screen shot 2017-11-09 at 12 45 39 pm

screen shot 2017-11-09 at 12 46 05 pm

### DIFF
--- a/src/components/actions/ActionBar.tsx
+++ b/src/components/actions/ActionBar.tsx
@@ -1,17 +1,21 @@
-import { IActionOptions } from './Action';
-import { IReduxStatePossibleProps } from '../../utils/ReduxUtils';
-import { PrimaryActionConnected } from './PrimaryActionConnected';
-import { PrimaryAction } from './PrimaryAction';
-import { SecondaryActionsConnected } from './SecondaryActionsConnected';
-import { SecondaryActions } from './SecondaryActions';
+import {IActionOptions} from './Action';
+import {ITooltipProps} from '../tooltip/Tooltip';
+import {IReduxStatePossibleProps} from '../../utils/ReduxUtils';
+import {PrimaryActionConnected} from './PrimaryActionConnected';
+import {PrimaryAction} from './PrimaryAction';
+import {SecondaryActionsConnected} from './SecondaryActionsConnected';
+import {SecondaryActions} from './SecondaryActions';
 import * as React from 'react';
 import * as _ from 'underscore';
-import { ItemFilter } from './filters/ItemFilter';
+import {ItemFilter} from './filters/ItemFilter';
 
 export interface IActionBarOwnProps extends React.ClassAttributes<ActionBar> {
   id?: string;
   itemFilterLabel?: string;
+  itemTooltipProps?: ITooltipProps;
   onClearItemFilter?: () => void;
+  extraContainerClasses?: string[];
+  removeDefaultContainerClasses?: boolean;
 }
 
 export interface IActionBarStateProps extends IReduxStatePossibleProps {
@@ -32,9 +36,10 @@ export interface IActionBarChildrenProps {
 }
 
 export interface IActionBarProps extends IActionBarOwnProps, IActionBarStateProps, IActionBarDispatchProps,
-  IActionBarChildrenProps { }
+  IActionBarChildrenProps {}
 
 export class ActionBar extends React.Component<IActionBarProps, any> {
+  static DEFAULT_ACTIONS_CONTAINER_CLASSES = ['coveo-table-actions-container', 'mod-cancel-header-padding', 'mod-border-bottom', 'mod-align-header'];
 
   private handleClear() {
     if (this.props.clearItemFilter) {
@@ -55,15 +60,16 @@ export class ActionBar extends React.Component<IActionBarProps, any> {
   }
 
   render() {
-    let itemFilter: JSX.Element = this.props.itemFilter
+    const itemFilter: JSX.Element = this.props.itemFilter
       ? <ItemFilter
         label={this.props.itemFilterLabel}
         item={this.props.itemFilter}
+        itemTooltipProps={this.props.itemTooltipProps}
         onClear={() => this.handleClear()} crop={this.props.itemFilterCropLength}
       />
       : null;
 
-    let primaryActions: JSX.Element[] = !this.props.prompt && _.map(this.props.actions, (action: IActionOptions, index: number): JSX.Element => {
+    const primaryActions: JSX.Element[] = !this.props.prompt && _.map(this.props.actions, (action: IActionOptions, index: number): JSX.Element => {
       if (action.primary) {
         let primaryAction = this.props.withReduxState
           ? <PrimaryActionConnected action={action} parentId={this.props.id} />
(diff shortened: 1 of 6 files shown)


src/components/actions/ActionBar.tsx
```diff
@@ -1,17 +1,21 @@
...
+import {ITooltipProps} from '../tooltip/Tooltip';
+import {IReduxStatePossibleProps} from '../../utils/ReduxUtils';
+import {PrimaryActionConnected} from './PrimaryActionConnected';
+import {PrimaryAction} from './PrimaryAction';
+import {SecondaryActionsConnected} from './SecondaryActionsConnected';
+import {SecondaryActions} from './SecondaryActions';
 import * as React from 'react';
 import * as _ from 'underscore';
-import { ItemFilter } from './filters/ItemFilter';
+import {ItemFilter} from './filters/ItemFilter';
```
sroy3: You should run the prettify task

src/components/actions/ActionBar.tsx
```diff
@@ -32,9 +36,10 @@ export interface IActionBarChildrenProps {
 }
 
 export interface IActionBarProps extends IActionBarOwnProps, IActionBarStateProps, IActionBarDispatchProps,
-  IActionBarChildrenProps { }
+  IActionBarChildrenProps {}
 
 export class ActionBar extends React.Component<IActionBarProps, any> {
+  static DEFAULT_ACTIONS_CONTAINER_CLASSES = ['coveo-table-actions-container', 'mod-cancel-header-padding', 'mod-border-bottom', 'mod-align-header'];
```
sroy3: I'd put each class on its own line

src/components/actions/ActionBar.tsx
```diff
@@ -85,16 +91,20 @@ export class ActionBar extends React.Component<IActionBarProps, any> {
...
       ? (<div className='coveo-table-actions'>
         {primaryActions}
         {secondaryActionsView}
         {this.props.prompt}
       </div>)
       : null;
 
+    const containerClasses = _.isUndefined(this.props.removeDefaultContainerClasses) || !this.props.removeDefaultContainerClasses
+      ? [...ActionBar.DEFAULT_ACTIONS_CONTAINER_CLASSES, ...(this.props.extraContainerClasses || [])].join(' ')
+      : (this.props.extraContainerClasses || []).join(' ');
```
sroy3: Don't we classnames to manage classes?

src/components/actions/examples/ActionBarExamples.tsx
```diff
@@ -36,6 +36,10 @@ export class ActionBarExamples extends React.Component<any, any> {
           <label className='form-control-label'>Action bar without actions</label>
           <ActionBar actions={actions} />
         </div>
+        <div className='form-group'>
+          <label className='form-control-label'>Action bar with default classes turned off and extra classes</label>
+          <ActionBar actions={actions} removeDefaultContainerClasses={true} extraContainerClasses={['coveo-table-actions-container', 'p2']}/>
```
sroy3: You don't have to set `removeDefaultContainerClasses` to true. You can just have the prop on or off if you want  it or not

src/components/actions/filters/ItemFilter.tsx
```diff
@@ -13,21 +15,25 @@ export const ELLIPSIS: string = '...';
...
-          ? itemFilter.substring(0, this.props.crop) + ELLIPSIS
-          : ELLIPSIS + itemFilter.substring(itemFilterLength + this.props.crop, itemFilterLength);
+        itemFilterText = this.props.crop > 0
+          ? itemFilterText.substring(0, this.props.crop) + ELLIPSIS
+          : ELLIPSIS + itemFilterText.substring(itemFilterLength + this.props.crop, itemFilterLength);
       }
     }
 
+    const itemFilter: JSX.Element = this.props.itemTooltipProps
+      ? <Tooltip {...this.props.itemTooltipProps}><span className='item-filter-item'>{itemFilterText}</span></Tooltip>
```
sroy3: I'd put the span on it's own line so it's easier to see the html

src/components/actions/filters/examples/ItemFilterConnectedExamples.tsx
```diff
@@ -43,6 +43,16 @@ export class ItemFilterConnectedExamples extends React.Component<any, any> {
             onClearItemFilter={() => alert('Item filter was cleared')}
           />
         </div>
+        <div className='form-group'>
+          <label className='form-control-label'>Action bar with an item filter cropped on the left with Redux state and tooltip options</label>
+          <ActionBarConnected
+            id={ACTION_BAR_ID + '2'}
```
sroy3: Template string

src/components/actions/ActionBar.tsx
MichelML: i started with it but was more verbose than this solution so reverted to that ⬆️ , having the default classes as a static variable also facilitates testing

src/components/actions/ActionBar.tsx
```diff
@@ -35,6 +39,12 @@ export interface IActionBarProps extends IActionBarOwnProps, IActionBarStateProp
   IActionBarChildrenProps { }
 
 export class ActionBar extends React.Component<IActionBarProps, any> {
+  static DEFAULT_ACTIONS_CONTAINER_CLASSES = [
```
vseguin: Minor detail but shouldn't this be private?

src/components/actions/ActionBar.tsx
pochretien: We can use a props classes and add a default props with DEFAULT_ACTIONS_CONTAINER_CLASSES. If the user want to override the class, he use the classes prob and by default we use the default.
I think its a lot of complexity to only manage classes with a boolean props and an array.

src/components/actions/filters/ItemFilter.tsx
pochretien: if the title in itemTooltipProps is empty or undefined, the Tooltip will wrap your item-filter with a span. I dont think you need this validation here. You can test but you can always use the Tooltip tag

src/components/actions/filters/ItemFilter.tsx
```diff
@@ -13,21 +15,27 @@ export const ELLIPSIS: string = '...';
...
-          ? itemFilter.substring(0, this.props.crop) + ELLIPSIS
-          : ELLIPSIS + itemFilter.substring(itemFilterLength + this.props.crop, itemFilterLength);
+        itemFilterText = this.props.crop > 0
+          ? itemFilterText.substring(0, this.props.crop) + ELLIPSIS
+          : ELLIPSIS + itemFilterText.substring(itemFilterLength + this.props.crop, itemFilterLength);
       }
     }
 
+    const itemFilter: JSX.Element = this.props.itemTooltipProps
+      ? <Tooltip {...this.props.itemTooltipProps}>
```
pochretien: if you itemTooltipProps title is empty or undefined. The child element will be wrap by a span tag. I dont think this condition is necessary. You can test but you probably can use <Tooltip/> in both case

src/components/actions/ActionBar.tsx
MichelML: done thx

src/components/actions/ActionBar.tsx
MichelML: the change was meant to be backward compatible, so implemented a new feature without removing the old way of using the component. It seems like what you describe would change the way it was priorly used.

MichelML: so it needs to be --> container classes by default --> use a custom mode if it does not work for you

MichelML: nvm i get what you mean, will see what I can do, might be simpler your way

PR #404, opened 2017-11-13
Added option to make toast undismissible

GermainBergeron:
Added example
Added unit tests

undimissible-toast

### DIFF
--- a/src/components/toast/Toast.tsx
+++ b/src/components/toast/Toast.tsx
@@ -10,6 +10,7 @@ export interface IToastProps {
   isOpened?: boolean;
   type?: string;
   dismiss?: number;
+  dismissible?: boolean;
   animate?: boolean;
   content?: ComponentContent;
   onRender?: () => void;
@@ -25,6 +26,10 @@ export const ToastType = {
 
 export class Toast extends React.Component<IToastProps, {}> {
   private timeout: number;
+
+  static defaultProps: Partial<IToastProps> = {
+    dismissible: true,
+  };
 
   componentWillMount() {
     if (this.props.onRender) {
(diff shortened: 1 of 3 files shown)


MichelML:
In what cases would you want such a behavior @GermainBergeron ?

GermainBergeron:
In some cases the error should be visible until the user fix the problem.

In my specific use case a user can add many cards to a dashboard and go over the size limit. To fix his problem he can resize a card, remove a card, etc. Until he fixes the situation the toast will stay visible (and he won't be able to save)

src/components/toast/examples/ToastExamples.tsx
```diff
@@ -56,6 +59,8 @@ export class ToastExamples extends React.Component<{}, IToastExamplesState> {
               </Toast>}
               {this.state.Timed &&
                 <Toast key='toast-4' id='toast-4' title='Timed Success!' dismiss={1000} onClose={() => this.setState({ Timed: false })} />}
+              {this.state.NonDismissible &&
+                <Toast key='toast-5' id='toast-5' title='Ohno! Something went wrong.' type={ToastType.Error} dismissible={false} content={ToastContentExample} />}
```
MichelML: almost shy to make this comment but `Ohno!` should be `Oh no!` ....

src/components/toast/Toast.tsx
```diff
@@ -45,7 +50,7 @@ export class Toast extends React.Component<IToastProps, {}> {
   }
 
   private setTimeout() {
-    if (this.props.dismiss > 0) {
+    if (this.props.dismissible !== false && this.props.dismiss > 0) {
```
sroy3: You don't need `!== false`

PR #407, opened 2017-11-17
display clear icon on filter through dispatch

MichelML:
https://drive.google.com/file/d/1wdlVYHWqAJFvMsYnW-O4Xsl2cBwzzbGz/view  

see demo, if you dispatch a `filterThrough` action from outside the filter with a non empty value (like in the video), you still want the clear icon to pop up, this was not the case previously. This does not break the FilterBox unconnected


pochretien:
Add a example if you can and :shipit:

MichelML:
@pochretien added uts, no need examples it's already there...this is just a fix of what should be the expected behavior

PR #408, opened 2017-11-17
Selected behavior for rows

MichelML:
Prior to this PR the selected behavior was handled by hand in sections we use the react-vapor rows. I suggest to handle the selected behavior through the rows state, which will solve a couple of problems linked to the log browser at the present moment and will standardize the behavior across places where we'll use them in the future.
 
demo https://drive.google.com/file/d/1Yc24baxWkhWeMFEaExvrUo2q-DB1s-PC/view


src/components/tables/tests/TableRowReducers.spec.ts
```diff
@@ -99,36 +99,124 @@ describe('Tables', () => {
...
+          type: TableRowActions.toggle,
+          payload: {
+            id: 'row1',
+            isCollapsible: false,
+          }
+        };
+        let collapsibleRowsState: ITableRowState[] = tableRowsReducer(oldState, action);
+
+        expect(collapsibleRowsState.length).toBe(oldState.length);
+        expect(collapsibleRowsState.filter(row => row.id === action.payload.id)[0].opened).toBeFalsy();
```
rousselm4: pourquoi pas `.toBe(false)` ?

src/components/tables/TableHeadingRowConnected.tsx
```diff
@@ -8,29 +8,24 @@ import * as React from 'react';
...
   };
 };
 
 const mapDispatchToProps = (dispatch: (action: IReduxAction<IReduxActionsPayload>) => void,
   ownProps: ITableHeadingRowOwnProps) => ({
     onClick: () => {
-      if (ownProps.isCollapsible) {
-        dispatch(toggleRow(ownProps.id));
-      }
+      dispatch(toggleRow(ownProps.id, ownProps.isCollapsible));
```
GermainBergeron: The row is always toggled even if not collapsible?

src/components/tables/TableRowReducers.ts
```diff
@@ -6,26 +6,30 @@ import * as _ from 'underscore';
...
         id: action.payload.id,
-        opened: false
+        opened: false,
+        selected: false,
       };
     case TableRowActions.toggle:
       if (state.id !== action.payload.id) {
         return _.extend({}, state, {
-          opened: false
+          opened: action.payload.isCollapsible && false,
```
GermainBergeron: This will always be false

src/components/tables/TableRowReducers.ts
```diff
@@ -6,26 +6,30 @@ import * as _ from 'underscore';
...
 export const tableRowReducer = (state: ITableRowState = tableRowInitialState, action: IReduxAction<IReduxActionsPayload>): ITableRowState => {
   switch (action.type) {
     case TableRowActions.add:
       return {
         id: action.payload.id,
-        opened: false
+        opened: false,
+        selected: false,
       };
     case TableRowActions.toggle:
```
GermainBergeron: Should the action be named `TableRowActions.select` ? It makes more sense IMHO since you're not always toggling it but always changing the `selected` attribute

src/components/tables/tests/TableRowReducers.spec.ts
```diff
@@ -99,36 +99,124 @@ describe('Tables', () => {
...
+        const action: IReduxAction<ITableRowActionPayload> = {
+          type: TableRowActions.toggle,
+          payload: {
+            id: 'row1',
+          }
+        };
+        let collapsibleRowsState: ITableRowState[] = tableRowsReducer(oldState, action);
+
+        expect(collapsibleRowsState.length).toBe(oldState.length);
+        expect(collapsibleRowsState.filter(row => row.id === action.payload.id)[0].opened).toBeUndefined();
```
GermainBergeron: I'd expect the value to be false instead of undefined

src/components/tables/tests/TableRowReducers.spec.ts
GermainBergeron: We usually use `toBe(false);` instead of falsy (which treats 0, '', etc as falsy values)

MichelML: yes this was the test I wanted to make here, but with your suggested changes `opened` will always be false, it could have been undefined with the current implementation.

src/components/tables/TableHeadingRowConnected.tsx
MichelML: I hesitated on this one, but I think it's good to have all the logic happening in one place (the state). Let me know what you think

GermainBergeron: After reading the rest of the PR I think that it's the name of the action that bothers me

src/components/tables/tests/TableRowReducers.spec.ts
GermainBergeron: I think it's easier to handle in the components if the value is either true or false and always defined

MichelML:
@rousselm4 @GermainBergeron all suggested changes applied

src/components/tables/TableHeadingRowConnected.tsx
MichelML: i've changed it for selectRow and SELECT_ROW and TableRowActions.select, added a check here as well but preserved the logic in the state too in case someone decides to dispatch a selectRow action from somewhere else

PR #410, opened 2017-11-20
Flat Select form component

sbelzile-Coveo:
flatselect


sbelzile-Coveo:
PR is useless since @pochretien coded a way better FlatSelect component. Closing PR.

PR #416, opened 2017-11-30
Add header component

pochretien:
screen shot 2017-11-30 at 5 50 03 pm
screen shot 2017-11-30 at 5 51 04 pm
screen shot 2017-11-30 at 5 51 12 pm
screen shot 2017-11-30 at 5 51 27 pm
screen shot 2017-11-30 at 5 51 40 pm


src/components/breadcrumbs/Breadcrumb.tsx
```diff
@@ -0,0 +1,36 @@
...
+export class Breadcrumb extends React.Component<IBreadcrumbProps, {}> {
+
+  static defaultProps: Partial<IBreadcrumbProps> = {
+    links: [],
+    defaultLinkPath: '',
+  };
+
+  private getLinks(): JSX.Element[] {
+    return _.map(this.props.links, (link: IBreadcrumbLinkProps) => {
+      return <BreadcrumbLink key={link.name} {..._.extend(link, { link: `${this.props.defaultLinkPath}${link.link}` }) } />;
```
sroy3: Brackets and return are not needed

src/components/headers/BasicHeader.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+export interface IBasicHeaderProps extends IHeaderWrapperProps, React.ClassAttributes<React.Component<any, any>> {
+  title: ITitleProps;
+}
+
+export class BasicHeader extends React.Component<IBasicHeaderProps, {}> {
+
+  static defaultProps: Partial<IBasicHeaderProps> = {
+    title: {
+      title: '',
```
sroy3: title inside of title, that seems a little weird, maybe we can change the name of one of them

src/components/headers/TabsHeader.tsx
```diff
@@ -0,0 +1,21 @@
...
+import * as _ from 'underscore';
+import { TabConnected } from '../tab/TabConnected';
+import { ITabProps } from '../tab/Tab';
+import { TabNavigation } from '../tab/TabNavigation';
+
+export interface ITabsHeaderProps extends React.ClassAttributes<React.Component<any, any>> {
+  tabs?: ITabProps[];
+}
+
+export class TabsHeader extends React.Component<ITabsHeaderProps, {}> {
```
sroy3: You could optimize and simply make a `const` instead:
```
export const TabsHeader = (props: ITabsHeaderProps) => {
  if (this.props.tabs) {
    return (
      <TabNavigation>
        {..._.map(this.props.tabs, (tab: ITabProps) => <TabConnected key={tab.id} {...tab} />)}
      </TabNavigation>);
    }
    return null;
  }
}
```

src/components/headers/tests/BasicHeader.spec.tsx
```diff
@@ -0,0 +1,47 @@
...
+    });
+
+    afterEach(() => {
+      basicHeaderComponent.unmount();
+      basicHeaderComponent.detach();
+    });
+
+    it('should render the default title', () => {
+      const titleComponent = basicHeaderComponent.find(Title);
+      expect(titleComponent.length).toEqual(1);
```
sroy3: Use `.toBe()` instead of `.toEqual()`.

sroy3: Practically everywhere as well.

MichelML:
screen shot 2017-12-01 at 11 24 57 am

src/components/breadcrumbs/tests/Breadcrumb.spec.tsx
```diff
@@ -0,0 +1,90 @@
...
+        { attachTo: document.getElementById('App') },
+      );
+    };
+
+    afterEach(() => {
+      breadcrumbComponent.unmount();
+      breadcrumbComponent.detach();
+    });
+
+    it('should render the BreadcrumbLink if the links has 1 element', () => {
```
MichelML: what if an empty array of links?

src/components/breadcrumbs/Breadcrumb.tsx
```diff
@@ -0,0 +1,36 @@
+import * as React from 'react';
+import * as _ from 'underscore';
+import { ITitleProps, Title } from '../title/Title';
+import { BreadcrumbLink, IBreadcrumbLinkProps } from './BreadcrumbLink';
+
+export interface IBreadcrumbProps extends React.ClassAttributes<Breadcrumb> {
+  links?: IBreadcrumbLinkProps[];
+  defaultLinkPath?: string;
```
MichelML: i wonder if an optional onClick call back could be nice here, suppose you want to send some analytics or any use case I can not think of.  @GermainBergeron thoughts?

src/components/breadcrumbs/examples/BreadcrumbsExamples.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+export class BreadcrumbsExamples extends React.Component<any, any> {
+
+  render() {
+    const documentationLink: ILinkSvgProps = {
+      url: 'https://www.google.ca',
+      target: '_blank',
+      svg: {
+        svgName: 'help',
+        svgClass: 'fill-orange icon mod-20',
```
MichelML: as we discussed `mod-20` is not applied on the icon, you could remove

pochretien: I dont want to remove it, its important to be apply on the icon. its the right size for the icon to look good. I think its caused by a change in vapor

src/components/breadcrumbs/examples/BreadcrumbsExamples.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+    const defaultTitle: ITitleProps = {
+      title: 'Pokemon',
+      prefix: 'Catch them all:',
+      documentationLink,
+    };
+
+    const link1: IBreadcrumbLinkProps = {
+      name: 'Pikachu',
+      link: '#',
```
MichelML: consider adding real paths here, `https://www.google.ca/?q=pokemon` `https://www.google.ca/?q=pokemon/squirtle` , it's nice to have demos that do something

src/components/breadcrumbs/tests/Breadcrumb.spec.tsx
pochretien: No BreadscrumbLink added.
UTs added for this use case

src/components/breadcrumbs/examples/BreadcrumbsExamples.tsx
MichelML: it could be nice to have an example with a defaultLinkPath, following the example above it would be `https://www.google.ca/?q=` as a default path link and your first breadscrum link would be `pikachu/` and your second one `squirtle`

MichelML: I know it won't be different in the examples, but it's a working example we can look at when about to use the component, they are examples but we should see it as documentation at the same time i think

src/components/headers/BasicHeader.tsx
MichelML: `text` ?

src/components/breadcrumbs/examples/BreadcrumbsExamples.tsx
pochretien: added

src/components/headers/BreadcrumbHeader.tsx
```diff
@@ -0,0 +1,28 @@
+import * as React from 'react';
+import { HeaderWrapper, IHeaderWrapperProps } from './HeaderWrapper';
+import { Breadcrumb, IBreadcrumbProps } from '../breadcrumbs/Breadcrumb';
+
+export interface IBasicHeaderProps extends IHeaderWrapperProps, React.ClassAttributes<React.Component<any, any>> {
```
MichelML: `IBreadcrumHeaderProps` instead here?

pochretien: Already fixed in my branch

src/components/headers/HeaderWrapper.tsx
```diff
@@ -0,0 +1,42 @@
...
+export interface IHeaderWrapperProps extends ITabsHeaderProps, React.ClassAttributes<React.Component<any, any>> {
+  description?: string;
+  actions?: IContentProps[];
+  classes?: string[];
+}
+
+export class HeaderWrapper extends React.Component<IHeaderWrapperProps, {}> {
+
+  private getClasses(): string {
+    return classNames('flex flex-center space-between mod-header-padding header-height pb2',
```
MichelML: consider putting each argument on its own line, code standards say it

pochretien: Its a string. why ?

src/components/breadcrumbs/examples/BreadcrumbsExamples.tsx
MichelML: the rule that is used in v2 is `.panel-header .coveo-icon-docs` could you consider implementing this one?

MichelML: nvm coveo-icon-docs is from v2 so does not make sens

MichelML:
fine by me overall 👍  looks really nice, good job @pochretien

PR #418, opened 2017-12-01
Table component

MichelML:
all uts added


src/components/loading/LoadingActions.ts
```diff
@@ -25,16 +26,20 @@ export const removeLoading = (id: string): IReduxAction<ILoadingActionPayload> =
...
   payload: {
-    ids
+    ids,
+    id,
   }
 });
 
-export const turnOffLoading = (ids: string[]): IReduxAction<ILoadingActionPayload> => ({
+// the second id is the id from which the action was triggered from
+export const turnOffLoading = (ids: string[], id?: string): IReduxAction<ILoadingActionPayload> => ({
```
GermainBergeron: Maybe you could change the parameters name to be clearer

`export const turnOffLoading = (loadingIDs: string[], triggeredFromId?: string)`

src/components/loading/LoadingTableConnected.tsx
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    return (
+      <table className='mod-collapsible-rows'>
+        {_.range(this.props.numberOfRows).map((num: number) => {
+          return (
+            <tbody key={`loading-row${num}`} className='loading-row'>
+              <tr>
+                <td colSpan={this.props.columnsPerRow}>
+                  <Loading />
```
GermainBergeron: Shouldn't it be one loading per table instead of one loading per row?

src/components/loading/LoadingTableConnected.tsx
```diff
@@ -0,0 +1,28 @@
+import * as React from 'react';
+import { LoadingConnected } from './LoadingConnected';
+import { Loading } from './Loading';
+import * as _ from 'underscore';
+
+export class LoadingTableConnected extends LoadingConnected {
```
GermainBergeron: I don't think we should extend other components

src/components/tables/Table.tsx
```diff
@@ -0,0 +1,337 @@
...
+export interface IData {
+  id: string;
+  [attribute: string]: any;
+};
+
+export interface ITableRowData {
+  [id: string]: {
+    id: string;
+    [attribute: string]: any;
+  };
```
GermainBergeron:
```
export interface ITableRowData {
    [id: string]: IData;
}
```

src/components/tables/Table.tsx
```diff
@@ -0,0 +1,337 @@
...
+};
+
+export type IAttributeFormatter = (attributeValue: any, attributeName?: string) => JSXRenderable;
+export type IAttributeNameFormatter = (attributeName: string) => string;
+
+export interface ITableHeadingAttribute {
+  attributeName: string;
+  titleFormatter: (attributeName: string) => string;
```
GermainBergeron: You should probably use a `IAttributeNameFormatter` here too

src/components/tables/Table.tsx
```diff
@@ -0,0 +1,337 @@
...
+  sort?: boolean;
+  sortByMethod?: (attributeValue: any) => string;
```
GermainBergeron: `IAttributeNameFormatter`?

src/components/tables/Table.tsx
```diff
@@ -0,0 +1,337 @@
...
+
+export interface ITableStateProps {
+  tableState?: ITableState;
+};
+
+export interface ITableProps extends ITableOwnProps, ITableStateProps, Partial<ITableDispatchProps> { }
+
+export class Table extends React.Component<ITableProps, any> {
+  componentDidMount() {
+    if (this.props.onDidMount) {
```
GermainBergeron: Everywhere else we use onRender and onDestroy

src/components/tables/Table.tsx
```diff
@@ -0,0 +1,337 @@
...
+      this.props.onDidMount();
+    }
+  }
+
+  componentWillReceiveProps(nextProps: ITableProps) {
+    const { tableState } = this.props;
+
+    if (this.hasTableStateChanged(tableState, nextProps.tableState)) {
+      const shouldResetPage =
+        tableState.page === nextProps.tableState.page
```
GermainBergeron: if the page is the same we should reset page?

src/components/tables/Table.tsx
```diff
@@ -0,0 +1,337 @@
...
+      || currentTableState.sortState.order !== nextTableState.sortState.order
+      || (_.isEmpty(currentTableState.predicates) && !_.isEmpty(nextTableState.predicates))
+      || _.some(
+        currentTableState.predicates,
+        (attributeValue: any, attributeName: string) => attributeValue !== nextTableState.predicates[attributeName],
+      )
+    );
+  }
+
+  buildLoadingTable(): JSX.Element {
```
GermainBergeron: I think that all of your buildX methods should be private

src/components/tables/Table.tsx
```diff
@@ -0,0 +1,337 @@
...
+        id={getTableChildComponentId(this.props.id, TableChildComponent.FILTER)}
+        key={getTableChildComponentId(this.props.id, TableChildComponent.FILTER)} />
+      : null;
+
+    const predicatesConnected: JSX.Element[] = actionBar && predicates
+      ? predicates.map((predicate: ITablePredicate, i: number) => {
+        const predicateId = `${getTableChildComponentId(this.props.id, TableChildComponent.PREDICATE)}-${predicate.attributeName}`;
+        const containerClasses = i ? ['ml1'] : [''];
+        return (
+          <DropdownSearchConnected
```
GermainBergeron: I'm confused, why would we use a DropdownSearch if we turn off the search?

sroy3:
I'm slowly making my way through all of this. Demo looks great in general. two things I've noticed:

* I wouldn't add loading on every row, but simply over the whole table like we do elsewhere.
* The first column should be aligned with the per page navigation

screen shot 2017-12-04 at 9 49 03 am

src/components/tables/Table.tsx
```diff
@@ -0,0 +1,337 @@
...
+  render() {
+    return (
+      <div>
+        {this.buildActionBar()}
+        <table className='mod-collapsible-rows'>
+          {this.buildTableHeader()}
+          {this.buildTableBody()}
+        </table>
+        {this.buildLoadingTable()}
+        {this.buildBlankSlate()}
```
GermainBergeron: shouldn't the blankslate and loading be in the table?

sroy3:
Collapsible row toggles should align with filter and page navigation

screen shot 2017-12-04 at 9 52 43 am

src/components/dropdownSearch/DropdownSearchReducers.tsx
```diff
@@ -238,7 +238,14 @@ export const dropdownSearchReducer = (state: IDropdownSearchState = dropdownSear
...
         ? selectSingleOption(state.options, action.payload.addedSelectedOption)
-        : removeCustomOptions(selectSingleOption(deselectAllOptions(state.options, true), action.payload.addedSelectedOption), state.supportSingleCustomOption, false);
+        : removeCustomOptions(
+          selectSingleOption(
+            deselectAllOptions(state.options, true),
+            action.payload.addedSelectedOption
+          ),
+          state.supportSingleCustomOption,
+          false
+        );
```
GermainBergeron: I think it easier to grasp which parameter is for which function when formated like this:
```
: removeCustomOptions(
  selectSingleOption(deselectAllOptions(state.options, true), action.payload.addedSelectedOption),
  state.supportSingleCustomOption,
  false
);
```

src/components/navigation/perPage/NavigationPerPageConnected.tsx
```diff
@@ -34,7 +34,7 @@ const mapDispatchToProps = (dispatch: (action: IReduxAction<IReduxActionsPayload
     onPerPageClick: (perPageNb: number, oldPerPageNb: number, currentPage: number) => {
       dispatch(turnOnLoading(ownProps.loadingIds));
       dispatch(changePerPage(ownProps.id, perPageNb));
-      dispatch(changePage(`pagination-${ownProps.id}`, Math.floor(currentPage * oldPerPageNb / perPageNb)));
+      dispatch(changePage(`pagination-${ownProps.id}`, 0));
```
sroy3: Any particular reason why this was changed?

src/components/tables/TableActions.ts
```diff
@@ -0,0 +1,62 @@
...
+  initialTableData: ITableData,
+  initialPerPage: number,
+): IReduxAction<ITableActionPayload> => ({
+  type: TableActions.add,
+  payload: { id, initialTableData, initialPerPage },
+});
+
+export const removeTable = (
+  id: string,
+): IReduxAction<ITableActionPayload> => ({
```
sroy3: I think this fits on one line

src/components/tables/TableActions.ts
```diff
@@ -0,0 +1,62 @@
...
+export enum MODIFY_SATE_TABLE { MODIFY_SATE_TABLE = 'MODIFY_STATE_TABLE' };
+
+export type ITableStateModifyer = (state: ITableState) => ITableState;
+
+export const TableActions = {
+  add: 'ADD_TABLE',
+  remove: 'REMOVE_TABLE',
+  inError: 'IN_ERROR_TABLE',
+  toggleLock: 'TOGGLE_LOCK_TABLE',
+  modifyState: MODIFY_SATE_TABLE.MODIFY_SATE_TABLE,
```
sroy3: Why the enum for this one? If it is necessary, I would change its name.

src/components/tables/TableActions.ts
```diff
@@ -0,0 +1,62 @@
+import { IData } from './Table';
+import { ITableState, ITableData } from './TableReducers';
+import { IReduxAction } from '../../utils/ReduxUtils';
+
+export enum MODIFY_SATE_TABLE { MODIFY_SATE_TABLE = 'MODIFY_STATE_TABLE' };
+
+export type ITableStateModifyer = (state: ITableState) => ITableState;
```
sroy3: ITableStateModifier

src/components/tables/TableConnected.tsx
```diff
@@ -0,0 +1,79 @@
...
+  onUnmount: () => void;
+  onModifyData: (shouldResetPage: boolean) => void;
+  onRowClick: (actions: IActionOptions[]) => void;
+  onPredicateOptionClick: (predicateId: string, option: IDropdownOption) => void;
+}
+
+const mapStateToProps = (state: IReactVaporState, ownProps: ITableOwnProps) => {
+  const tableState: ITableState = state.tables[ownProps.id];
+
+  return { tableState: { ...tableState } };
```
sroy3: You can simply return the state without having to copy it into a new object. The reducers will make sure the state is immutable on the other end.

src/components/tables/TableConnected.tsx
```diff
@@ -0,0 +1,79 @@
...
+};
+
+const mapDispatchToProps = (
+  dispatch: (action: IReduxAction<IReduxActionsPayload> | ThunkAction<any, any, any>) => void,
+  ownProps: ITableOwnProps,
+): ITableDispatchProps => ({
+  onDidMount: () => {
+    let initialPerPage = ownProps.navigation && ownProps.navigation.perPageNumbers && ownProps.navigation.perPageNumbers[0];
+    initialPerPage = !_.isUndefined(initialPerPage) ? initialPerPage : PER_PAGE_NUMBERS[0];
```
sroy3: You could have default props instead.

src/components/tables/TableHeader.tsx
```diff
@@ -1,17 +1,31 @@
...
+        : TableHeaderCell;
+
+      return (
+        <TableHeaderCellClass
+          key={`th-${column.id || index}`}
+          id={column.id}
+          title={column.title}
+          className={column.className}
+          tableRefForSort={column.tableRefForSort}
+          onClickCallback={column.onClickCallback}
```
sroy3: You could replace the last 5 lines with `{...column}`

src/components/tables/TableHeaderCell.tsx
```diff
@@ -1,15 +1,77 @@
...
+      'admin-sort': tableCellHasSort,
+      'admin-sort-ascending': this.props.sorted === TableSortingOrder.ASCENDING,
+      'admin-sort-descending': this.props.sorted === TableSortingOrder.DESCENDING,
+    });
+
     return (
-      <th className={this.props.className || ''}>{this.props.title}</th>
+      <th
+        className={headerCellClasses}
+        onClick={(e: React.MouseEvent<any>) => this.handleClick(e)}>
```
sroy3: `HTMLTableHeaderCellElement` instead of `any`

src/components/tables/TableHeaderCell.tsx
```diff
@@ -1,15 +1,77 @@
...
-export interface ITableHeaderCellProps extends React.ClassAttributes<TableHeaderCell> {
+export interface ITableHeaderCellOwnProps extends React.ClassAttributes<TableHeaderCell> {
   title: string;
+  id?: string;
+  tableRefForSort?: {
+    tableId: string;
+    attributeToSort: string;
+  };
   className?: string;
+  onClickCallback?: (e: React.MouseEvent<any>) => void;
```
sroy3: `HTMLTableHeaderCellElement` instead of `any`

src/components/tables/TableHeaderCellConnected.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { IReactVaporState, IReduxActionsPayload } from '../../ReactVapor';
+import { IReduxAction, ReduxUtils } from '../../utils/ReduxUtils';
+import * as React from 'react';
+import { connect } from 'react-redux';
+
+const mapStateToProps = (state: IReactVaporState, ownProps: ITableHeaderCellOwnProps) => {
+  const headerCell: ITableHeaderCellState = state.tableHeaderCells[ownProps.id];
+
+  return {
+    sorted: headerCell && headerCell.sorted || TableSortingOrder.UNSORTED,
```
sroy3: I would add `TableSortingOrder.UNSORTED` as a default prop instead

src/components/tables/TableHeadingRow.tsx
```diff
@@ -23,10 +26,14 @@ export interface ITableHeadingRowProps extends ITableHeadingRowOwnProps, ITableH
 
 export class TableHeadingRow extends React.Component<ITableHeadingRowProps, any> {
 
-  private handleClick = () => {
+  private handleClick = (e: React.MouseEvent<any>) => {
```
sroy3: `HTMLTableRowElement` instead of `any`

src/components/tables/TableHeadingRow.tsx
```diff
@@ -42,15 +49,17 @@ export class TableHeadingRow extends React.Component<ITableHeadingRowProps, any>
...
     const rowClasses = classNames({
       'heading-row': this.props.isCollapsible,
       'selected': this.props.selected,
       'opened': this.props.opened,
     });
 
-    return (
-      <tr className={rowClasses} onClick={() => this.handleClick()}>
+    return this.props.hide ? null : (
+      <tr className={rowClasses} onClick={(e: React.MouseEvent<any>) => this.handleClick(e)}>
```
sroy3: `HTMLTableRowElement` instead of `any`

src/components/tables/TableThunkActionCreators.ts
```diff
@@ -0,0 +1,151 @@
...
+  );
+  dispatch(turnOnLoading(getTableLoadingIds(tableOwnProps.id), tableOwnProps.id));
+};
+
+export const dispatchPostTableStateModification = (tableOwnProps: ITableOwnProps, dispatch: any) => {
+  dispatch(turnOffLoading(getTableLoadingIds(tableOwnProps.id), tableOwnProps.id));
+  dispatch(changeLastUpdated(getTableChildComponentId(tableOwnProps.id, TableChildComponent.LAST_UPDATED)));
+};
+
+export const defaultTableStateModifyer = (
```
sroy3: defaultTableStateModifier

src/components/tables/TableThunkActionCreators.ts
```diff
@@ -0,0 +1,151 @@
...
+    if (tableState.filter) {
+      const filterDefault = (dataId: string): boolean => {
+        let shouldKeep = false;
+
+        tableOwnProps.headingAttributes.forEach((headingAttribute: ITableHeadingAttribute) => {
+          const { attributeName, attributeFormatter } = headingAttribute;
+          const attributeValue = tableDataById[dataId][attributeName];
+          const attributeValueToUse = attributeFormatter ? attributeFormatter(attributeValue) : attributeValue;
+          shouldKeep =
+            shouldKeep
```
sroy3: You should wrap the content of your loop with `if (!shouldKeep) {` so that it does not compute everything every time if `shouldKeep` is set to true on the first time. You could also change it for a `for (const headingAttribute in  tableOwnProps.headingAttributes) {` and use a `break` once we set `shouldKeep` to true.

src/components/tables/TableThunkActionCreators.ts
```diff
@@ -0,0 +1,151 @@
...
+            || attributeValueToUse
+              .toString()
+              .toLowerCase()
+              .indexOf(tableState.filter.toLowerCase()) > -1;
```
sroy3: You could use underscore.string here.

src/components/tables/TableThunkActionCreators.ts
```diff
@@ -0,0 +1,151 @@
...
+        displayedIds: nextDisplayedIds,
+        totalEntries,
+        totalPages,
+      },
+      page: shouldResetPage ? 0 : tableState.page,
+    };
+  };
+};
+
+export const defaultTableStateModifyerThunk = (tableOwnProps: ITableOwnProps, shouldResetPage: boolean) => {
```
sroy3: `defaultTableStateModifierThunk`

src/components/tables/TableThunkActionCreators.ts
```diff
@@ -0,0 +1,151 @@
...
+  return (dispatch: any) => {
+    dispatchPreTableStateModification(tableOwnProps, dispatch);
+    dispatch(modifyState(tableOwnProps.id, defaultTableStateModifyer(tableOwnProps, shouldResetPage)));
+    dispatchPostTableStateModification(tableOwnProps, dispatch);
+  };
+};
+
+export const serverTableStateModifyerThunk = (tableOwnProps: ITableOwnProps, shouldResetPage: boolean) => {
```
sroy3: `serverTableStateModifierThunk`

src/components/tables/TableThunkActionCreators.ts
```diff
@@ -0,0 +1,151 @@
...
+  return (dispatch: any, getState: () => { [globalStateProp: string]: any; tables: ITablesState; }) => {
+    dispatchPreTableStateModification(tableOwnProps, dispatch);
+    $.get(tableOwnProps.serverMode.url(getState().tables[tableOwnProps.id], tableOwnProps))
```
sroy3: I'm not quite sure this is the kind of thing that should be in React Vapor

src/components/tables/TableUtils.ts
```diff
@@ -0,0 +1,21 @@
+import { TableSortingOrder, TableChildComponent } from './TableConstants';
+import * as _ from 'underscore';
+
+const {
+  ASCENDING,
+  DESCENDING,
+  UNSORTED,
+  } = TableSortingOrder;
```
sroy3: Line starting with a space

src/components/tables/TableUtils.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const getNextTableSortingOrder = (sortedState: TableSortingOrder): TableSortingOrder =>
+  _.contains([UNSORTED, DESCENDING], sortedState)
+    ? ASCENDING
+    : DESCENDING;
```
sroy3: I think this fits on a line

src/components/tables/TableThunkActionCreators.ts
GermainBergeron: It's a good use case for `_.some` instead of forEach

package.json
```diff
@@ -98,6 +99,7 @@
     "karma-source-map-support": "1.2.0",
     "karma-sourcemap-loader": "0.3.7",
     "karma-webpack": "2.0.5",
+    "lorem-ipsum": "1.0.4",
```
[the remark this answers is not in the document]
MichelML: yes! have a better idea for filling tables with data?

src/components/dropdownSearch/DropdownSearch.tsx
```diff
@@ -330,6 +345,26 @@ export class DropdownSearch extends React.Component<IDropdownSearchProps, {}> {
     if (this.props.onKeyDownDropdownButton) {
       this.props.onKeyDownDropdownButton(e.keyCode);
     }
+
+    if (this.props.turnOffSearch &&
+      (_.contains([keyCode.enter, keyCode.tab], e.keyCode)
+        || (e.keyCode === keyCode.upArrow && this.props.activeOption === this.props.options[0]))) {
```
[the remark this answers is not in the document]
MichelML: yes lines 349-363 are an exact copy paste of the lines you mentioned except for the `this.props.turnOffSearch` condition in the if statement

src/components/dropdownSearch/DropdownSearchReducers.tsx
MichelML: i think the problem with this is the nesting of functions three levels deep. I'll move selectSingleOption in a const and add the const as the first argument of removeCustomOptions

src/components/filterBox/FilterBoxConnected.tsx
```diff
@@ -12,6 +12,11 @@ import { connect } from 'react-redux';
 import * as _ from 'underscore';
 import * as React from 'react';
 
+const debouncedFilterThrough = _.debounce(
+  (dispatch: (action: IReduxAction<IReduxActionsPayload>) => void, id: string, filterText: string) => dispatch(filterThrough(id, filterText)),
```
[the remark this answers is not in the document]
MichelML: thanks i like that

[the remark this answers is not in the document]
MichelML: we could eventually consider typing dispatch at any everywhere, it's kind of a reserved keyword and we all know what it does.

src/components/lastUpdated/examples/LastUpdatedConnectedExamples.tsx
```diff
@@ -25,14 +25,14 @@ export class LastUpdatedConnectedExamples extends React.Component<ILastUpdateCon
   componentDidMount() {
     setInterval(() => {
       this.props.onRefresh();
-    }, 5000);
+    }, 10000);
```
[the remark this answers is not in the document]
MichelML: it's kind of annoying when you're trying to debug something

src/components/filterBox/FilterBoxConnected.tsx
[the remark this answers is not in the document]
MichelML: or making a generic type for it `(action: IReduxAction<any>) => void` as @sroy3 suggested

src/components/loading/Loading.tsx
```diff
@@ -2,6 +2,10 @@ import * as React from 'react';
 
 export interface ILoadingOwnProps extends React.ClassAttributes<Loading> {
   id?: string;
+  hide?: boolean;
+  // loading table props
+  columnsPerRow?: number;
+  numberOfRows?: number;
```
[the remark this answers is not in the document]
MichelML: yes i'll change that for what we see in the `Fields` section or the Log browser , I've looked too quickly and I did not think we had something standard for that, but we have.

src/components/loading/LoadingActions.ts
MichelML: thanks I like that

src/components/loading/LoadingTableConnected.tsx
MichelML: as said earlier, this will be removed, it's ugly anyways :D

src/components/navigation/perPage/NavigationPerPageConnected.tsx
MichelML: I thought this was more consistent with other behaviors coming from other components (onFilter resets the page to 0, onSort resets the page to 0, onPredicate resets the page to 0)...after we discussed i'll probably revert to the old behavior somehow.

src/components/tables/Table.tsx
```diff
@@ -0,0 +1,337 @@
...
+};
+
+export interface ITableOwnProps extends React.ClassAttributes<Table> {
+  id: string;
+  initialTableData: ITableData;
+  headingAttributes: ITableHeadingAttribute[];
+  collapsibleFormatter?: (tableRowData: ITableRowData, props: ITableProps) => JSXRenderable;
+  actionBar?: IActionBarProps;
+  getActions?: (rowData?: ITableRowData, props?: ITableProps) => IActionOptions[];
+  blankSlates: {
```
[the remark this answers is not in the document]
MichelML: you mean you would like optional props instead (blankSlateDefault) (blankSlateNoResultsOnAction) (blankSlateOnError)?  i don't hate the idea

src/components/tables/Table.tsx
MichelML: i would tend to say it's good to define a dispatch prop that reflects where you use it afterward, here i initialize the table state in componentDidMount because i want its children to be in the state before it being in the state (and they do mount before the parent). https://github.com/facebook/react/blob/v15.0.1/docs/docs/ref-03-component-specs.md#mounting-componentdidmount

src/components/tables/Table.tsx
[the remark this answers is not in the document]
GermainBergeron: Yes

src/components/tables/Table.tsx
MichelML: because our `Dropdown` is pretty custom IIRC and the DropdownSearch currently has everything we need for this case except minor tweaks i've made to it...open to other ideas.

MichelML: but I agree it's not instinctive though.

src/components/tables/Table.tsx
MichelML: loadingTable won't be there anymore, blankSlate could go anywhere

src/components/tables/TableConnected.tsx
MichelML: thanks

src/components/tables/TableConnected.tsx
MichelML: definitely thanks

src/components/tables/TableThunkActionCreators.ts
MichelML: @GermainBergeron had doubts too, we could mirror https://react-table.js.org/#/story/server-side-data and just add an `onFetchData` that would be triggered where` this.props.onModifyData` is currently triggered. A bit like @GermainBergeron suggested: if you pass a `onFetchData` prop (which i think would be a dispatchable thunk action creator), the this.props.onModifyData won't be triggered and onFetchData will be dispatched instead, and thus the user of the component will have to trigger the state modification inside onFetchData.  If we implement this, I think customMode and serverMode could be removed and onFetchData will replace both.

MichelML: it could be more generic, like `onManualModifyData`

MichelML:
@sroy3 i think you're right and also think the table should be a wrapper. I don't see this as a major change, will see how it goes when i'll try to implement it. It could solve some problems I had actually.

src/components/tables/TableActions.ts
MichelML: it's not, will remove

src/components/tables/TableThunkActionCreators.ts
MichelML: definitely `some` is the thing to use here, fyi @sroy3 `shouldKeep = true || thisIsNotEvaluatedOrComputedSinceJSisLazylyEvaluated` only compute `shouldKeep = true` , the `for (const ...` would have been a good solution too but `some` is cleaner.

pochretien:
The number of row dont change for the server table view
In the network tab, the call look great with the right param 👍 
image

pochretien:
Minor detail but the space between predicates and the filter box should be 30px
image

pochretien:
To handle the style for the deleted row. We need to add the class `mod-deleting` on the tag table. To do that, the table component should listent on the TriggerActionConnected to add or remove the class when the confirm prompt is shown in the action bar.
image

src/components/blankSlate/BlankSlate.tsx
```diff
@@ -12,6 +12,7 @@ export interface IBlankSlateProps extends React.ClassAttributes<BlankSlate> {
   classes?: string[];
   svgName?: string;
   svgClass?: string;
+  hide?: string;
```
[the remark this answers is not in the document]
MichelML: this is true only for non connected components, so you're right in this case, however with the way we do things, we must `render`, even if it means to render `null`,  connected components for them to be in the store

src/components/dropdownSearch/DropdownPrepend.tsx
```diff
@@ -0,0 +1,11 @@
+// tslint:disable-next-line:no-unused-variable
+import * as React from 'react';
+
+export interface IDropdownPrependProps {
+  prepend: string;
```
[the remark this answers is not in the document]
MichelML: will add that 👍

src/components/dropdownSearch/DropdownSearch.tsx
```diff
@@ -44,6 +47,7 @@ export interface IDropdownSearchOwnProps extends React.ClassAttributes<DropdownS
   onMountCallBack?: () => void;
   onClickCallBack?: () => void;
   supportSingleCustomOption?: boolean;
+  turnOffSearch?: boolean;
```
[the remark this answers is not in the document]
MichelML: won't be there anymore, will implement what @GermainBergeron suggested, the dropdown search won't be a dropdown search under a certain threshold

src/components/loading/Loading.tsx
```diff
@@ -2,6 +2,10 @@ import * as React from 'react';
 
 export interface ILoadingOwnProps extends React.ClassAttributes<Loading> {
   id?: string;
+  hide?: boolean;
```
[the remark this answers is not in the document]
MichelML: this needs to be there sadly, unless you have another idea, for the reason explained above:

```
this is true only for non connected components, so you're right in this case, however with the way we do things, we must render, even if it means to render null, connected components for them to be in the store
```

src/components/loading/LoadingActions.ts
```diff
@@ -9,6 +9,7 @@ export const LoadingActions = {
 
 export interface ILoadingActionPayload {
   ids: string[];
+  id?: string;
```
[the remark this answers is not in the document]
MichelML: see germain's comment about that

src/components/tables/Table.tsx
```diff
@@ -0,0 +1,337 @@
...
+      hide={!this.props.tableState.isLoading}
+      columnsPerRow={this.props.headingAttributes.length + 1}
+      numberOfRows={
+        this.props.tableState.data && this.props.tableState.data.displayedIds.length
+        || this.props.tableState.perPage
+      } />;
+  }
+
+  buildLoadingNavigation(): JSX.Element {
+    return <LoadingConnected
```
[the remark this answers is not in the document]
MichelML: restating the reason why it's not possible here:

> this is true only for non connected components, so you're right in this case, however with the way we do things, we must render, even if it means to render null, connected components for them to be in the store

src/components/tables/TableConnected.tsx
MichelML: don't think it suits here finally

src/components/tables/TableConnected.tsx
```diff
@@ -22,6 +21,8 @@ import { IFilterState } from '../filterBox/FilterBoxReducers';
 import { ITableHeaderCellState } from './TableHeaderCellReducers';
 import { IDropdownSearchState } from '../dropdownSearch/DropdownSearchReducers';
 import { contains } from 'underscore.string';
+import { Dispatch } from '../../utils/ReduxUtils';
+
```
sroy3: Empty line

src/components/dropdownSearch/DropdownSearch.tsx
```diff
@@ -305,31 +315,56 @@ export class DropdownSearch extends React.Component<IDropdownSearchProps, {}> {
...
+      || e.keyCode === keyCode.tab
+      || (e.keyCode == keyCode.upArrow && this.props.activeOption === this.props.options[0]);
+  }
+
+  private handleOnOptionClickOnKeyDown(e: React.KeyboardEvent<HTMLInputElement>) {
+    // prevent onClick because an enter key on an input trigger an click event. OnClick re open the dropdown.
+    // prevent the tab event to select the option and change the focus on next element.
+    // We only want to close the dropdown and get focus on him.
+    // prevent the double focus on input dropdown when key up when the active option
+    // is the first option of the list filter by the filterText
```
sroy3: Since this is a quite long comment, I would use a multi-line comment instead. Also, a few things to increase readability:

* `prevent onClick because pressing the enter key`?
* `triggers a click event`
* `event and OnClick re-opens the dropdown`
* `and get focus on it`
* `on input dropdown on key up`

src/components/loading/Loading.tsx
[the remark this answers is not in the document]
sroy3: This can be done in the parent component

src/components/dropdownSearch/DropdownSearch.tsx
MichelML: this was a comment made by @pochretien , @pochretien want to reformulate it? I don't want to modify your first intent erroneously

pochretien: its fine for me the rework for the comment 👍

src/components/tables/Table.tsx
```diff
@@ -0,0 +1,360 @@
...
+  private buildActionBar(): JSX.Element {
+    const { actionBar, filter, predicates } = this.props;
+
+    const filterBoxConnected: JSX.Element = actionBar && filter
+      ? (
+        <div className='coveo-table-actions'>
+          <FilterBoxConnected
+            {...filter}
+            id={getTableChildComponentId(this.props.id, TableChildComponent.FILTER)}
+            key={getTableChildComponentId(this.props.id, TableChildComponent.FILTER)} />
```
sroy3: Close tag on next line

src/components/tables/Table.tsx
```diff
@@ -0,0 +1,360 @@
...
+    const headerClass = classNames(
+      'mod-no-border-top',
+      { 'mod-deactivate-pointer': !!this.props.tableCompositeState.isLoading }
+    );
+
+    return (
+      <TableHeader
+        headerClass={headerClass}
+        columns={[...tableHeaderCells, { title: '' }]}
+        connectCell />
```
sroy3: Close tag on next line

src/components/tables/Table.tsx
```diff
@@ -0,0 +1,360 @@
...
+
+  private buildTableHeadingRowContent(
+    attributeValue: any,
+    attributeName: string,
+    tableCoordinate: string,
+    attributeFormatter?: IAttributeFormatter,
+  ): JSXRenderable {
+    return attributeFormatter
+      ? <td key={tableCoordinate}>{attributeFormatter(attributeValue, attributeName)}</td>
+      : <td key={tableCoordinate}>{convertUndefinedAndNullToEmptyString(attributeValue)}</td>;
```
sroy3: I would rather have the ternary inside for the content instead of the whole tag

src/components/tables/Table.tsx
```diff
@@ -0,0 +1,360 @@
...
+        id={getTableChildComponentId(this.props.id, TableChildComponent.NAVIGATION)}
+        loadingIds={[getTableChildComponentId(this.props.id, TableChildComponent.LOADING_NAVIGATION)]}
+      />
+    ) : null;
+  }
+
+  private buildLastUpdated(): JSX.Element {
+    return <LastUpdatedConnected
+      label={this.props.lastUpdatedLabel}
+      id={getTableChildComponentId(this.props.id, TableChildComponent.LAST_UPDATED)} />;
```
sroy3: Close tag on next line

src/components/tables/Table.tsx
```diff
@@ -0,0 +1,360 @@
...
+                <Loading />
+              </td>
+            </tr>
+          </tbody>
+          {this.buildTableHeader()}
+          {this.buildTableBody()}
+        </table>
+        {this.buildBlankSlate()}
+        {this.buildNavigation()}
+        {this.buildLastUpdated()}
```
sroy3: This file is getting very big. How about we move all of the `buildSomething` as const with props in other files and passing the needed props? This way the file will be smaller and we will already have the logic if we want to do something similar elsewhere.

src/components/tables/TableConnected.tsx
```diff
@@ -0,0 +1,115 @@
...
+import { addTable, removeTable } from './TableActions';
+import { IReactVaporState } from '../../ReactVapor';
+import { ReduxUtils } from '../../utils/ReduxUtils';
+import * as React from 'react';
+import { connect } from 'react-redux';
+import * as _ from 'underscore';
+import { IActionOptions } from '../actions/Action';
+import {
+  defaultTableStateModifierThunk,
+} from './TableThunkActionCreators';
```
sroy3: This does not fit on one line?

src/utils/ReduxUtils.ts
```diff
@@ -2,6 +2,9 @@ import * as ReactRedux from 'react-redux';
 import * as Redux from 'redux';
 import { extend } from 'underscore';
 
+export type Dispatch = (action: IReduxAction<any> | ThunkAction) => void;
+export type ThunkAction = (dispatch: Dispatch, getState?: () => any) => void;
```
sroy3: That is very nice. It looks much cleaner elsewhere.

src/components/filterBox/FilterBox.tsx
```diff
@@ -100,7 +100,6 @@ export class FilterBox extends React.Component<IFilterBoxProps, any> {
           onBlur={() => this.handleOnBlur()}
           onFocus={(e: React.FocusEvent<any>) => { this.placeCursorAtEndOfInputValue(e); }}
           onKeyDown={(e) => this.handleOnKeyDown(e)}
-          value={this.props.filterText}
```
GermainBergeron: This is a wanted change?

MichelML: it is, I know I solved a problem with that change but i don't remember what it was exactly. I'll let you know if it comes to mind.

src/components/dropdownSearch/DropdownSearch.tsx
```diff
@@ -201,22 +207,26 @@ export class DropdownSearch extends React.Component<IDropdownSearchProps, {}> {
...
+      : null;
   }
 
   protected getMainInput(): JSX.Element {
     const selectedOption: IDropdownOption = _.findWhere(this.props.options, { selected: true });
     const filterPlaceHolder: string = selectedOption && (selectedOption.displayValue || selectedOption.value)
       || this.props.filterPlaceholder;
 
-    if (this.props.isOpened) {
+    if (this.props.isOpened && this.props.options.length > this.props.searchThresold) {
```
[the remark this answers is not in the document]
MichelML: I assume you meant `supportSingleCustomOption` right?

src/ReactVapor.ts
```diff
@@ -66,4 +73,16 @@ export interface IReduxActionsPayload {
   item?: string;
   label?: string;
   selected?: string;
+  tableId?: string;
+  isInError?: boolean;
+  attributeToFormat?: string;
+  attributeToSort?: string;
+  initialTableData?: ITableData;
+  initialPerPage?: number;
+  TableStateModifier?: ITableStateModifier;
```
pochretien: tableStateModifier

src/components/actions/ActionBarReducers.ts
```diff
@@ -18,15 +20,23 @@ export const actionBarReducer = (state: IActionBarState = actionBarInitialState,
...
+        ? state
+        : { ...state, actions: action.payload.actions };
     case ActionBarActions.add:
       return {
         id: action.payload.id,
-        actions: []
+        actions: [],
+        isLoading: false,
       };
+    case LoadingActions.turnOn:
```
pochretien: I try to figure out why we need to sent ids in the payload instead of the state array of each action bar if the id is one of then ?

src/components/tables/Table.tsx
```diff
@@ -0,0 +1,173 @@
...
+  getActions?: (rowData?: ITableRowData, props?: ITableProps) => IActionOptions[];
+  blankSlateDefault: IBlankSlateProps;
+  blankSlateNoResultsOnAction?: IBlankSlateProps;
+  blankSlateOnError?: IBlankSlateProps;
+  filter?: IFilterBoxProps;
+  filterMethod?: (attributeValue: any, props: ITableOwnProps) => boolean;
+  predicates?: ITablePredicate[];
+  navigation?: INavigationChildrenProps;
+  lastUpdatedLabel?: string;
+  manual?: (
```
pochretien: Nice!

src/components/filterBox/FilterBox.tsx
MichelML: since the filterbox change is now debounced, it was causing problems

src/components/actions/ActionBarReducers.ts
MichelML: we should discuss this in person, I don't understand what you mean

sroy3: @pochretien loading is started by outside components. We usually have a constant with the id of what we want to turn on/off when the page is loading.

src/components/dropdownSearch/DropdownSearch.tsx
```diff
@@ -254,6 +266,10 @@ export class DropdownSearch extends React.Component<IDropdownSearchProps, {}> {
     return (elTop >= boxTop) && (elBottom <= boxBottom);
   }
 
+  private isSearchOn(): boolean {
+    return this.props.options.length > this.props.searchThresold;
```
sroy3: I've made the comment previously, but haven't had any feedback on it. What about when length is smaller than the threshold, but we can add a new option?

Also can you explain what this does? It is not quite clear when looking at the code.

src/components/loading/Loading.tsx
```diff
@@ -26,12 +26,14 @@ export class Loading extends React.Component<ILoadingProps, any> {
   }
 
   render() {
-    return (
+    const spinner: JSX.Element = (
```
sroy3: Why this change?

src/components/navigation/perPage/tests/NavigationPerPage.spec.tsx
```diff
@@ -146,7 +146,20 @@ describe('NavigationPerPage', () => {
...
       navigationPerPageInstance['handleClick'].call(navigationPerPageInstance, expectedPerPage);
 
-      expect(newProps.onPerPageClick).toHaveBeenCalled();
+      expect(newProps.onPerPageClick).toHaveBeenCalledTimes(1);
+    });
+
+    it('should not call onPerPageClick prop if perPage is identical to currentPerPage', () => {
+      let newProps: INavigationPerPageProps = _.extend({}, NAVIGATION_PER_PAGE_BASIC_PROPS,
+        { onPerPageClick: jasmine.createSpy('onPerPageClick'), currentPerPage: 10 });
+      let expectedPerPage: number = 10;
```
sroy3: `const` instead of `let`

src/components/tables/Table.tsx
```diff
@@ -0,0 +1,173 @@
...
+  componentDidUpdate() {
+    if (this.updateCountForLoadingBehavior < 2) {
+      this.updateCountForLoadingBehavior += 1;
+    } else {
+      /**
+       *  A count is utilized for a clean loading behavior during the initial load of the table
+       *  The first count occurs after mount
+       *  The second count occurs before the real data has loaded in the table
+       *  The initial load is than completed on the third update after the real data has loaded
+       */
```
sroy3: You could probably compare previous props and current props instead to figure out if the loading should turn off. The second time the component updates doesn't necessarily means it was triggered by the data being passed as props. Any state change or other prop could trigger an update.

src/components/tables/TableConnected.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  return {
+    tableCompositeState: {
+      id: tableState.id,
+      data: tableState.data,
+      isInError: tableState.isInError,
+      isLoading: tableState.isLoading,
+      filter: filterState && filterState.filterText || '',
+      page: paginationState && paginationState.pageNb || 0,
+      perPage: perPageState && perPageState.perPage || DEFAULT_TABLE_PER_PAGE,
```
sroy3: `DEFAULT_TABLE_PER_PAGE` is already in Table's default props. You can remove it from here. `0` and `''` should also be default props.

src/components/tables/TableHeader.tsx
```diff
@@ -1,17 +1,27 @@
 import { ITableHeaderCellProps, TableHeaderCell } from './TableHeaderCell';
+import { TableHeaderCellConnected } from './TableHeaderCellConnected';
 import * as React from 'react';
 import * as _ from 'underscore';
 
 export interface ITableHeaderProps extends React.ClassAttributes<TableHeader> {
   columns: ITableHeaderCellProps[];
   headerClass?: string;
+  connectCell?: true;
```
sroy3: I've created `IReduxStatePossibleProps` from which you can extend instead. It's a simple interface:
```
export interface IReduxStatePossibleProps {
  withReduxState?: boolean;
}
```
It's exactly the same, but it's might be better to use the same thing everywhere. Also I don't think there is a `TableHeaderConnected`, so this won't apply here, but usually I let the connected component set this value to true or false.

src/components/tables/TableReducers.ts
```diff
@@ -0,0 +1,138 @@
...
+        [action.payload.id]: {
+          ...tableInitialState,
+          id: action.payload.id,
+          data: action.payload.initialTableData,
+          perPageId: getTableChildComponentId(action.payload.id, TableChildComponent.PER_PAGE),
+          paginationId: getTableChildComponentId(action.payload.id, TableChildComponent.PAGINATION),
+          filterId: getTableChildComponentId(action.payload.id, TableChildComponent.FILTER),
+          predicateIds: action.payload.predicates.map((predicate: ITablePredicate) =>
+            `${getTableChildComponentId(action.payload.id, TableChildComponent.PREDICATE)}${predicate.attributeName}`),
+        },
```
sroy3: This should be done by the `tableReducer`.

src/components/tables/TableUtils.ts
```diff
@@ -0,0 +1,59 @@
...
+  }, DEFAULT_TABLE_DATA.byId);
+};
+
+export const convertDataByIdToCollection = (dataById: ITableRowData, keepIdAttribute = true): Collection => {
+  return _.values(dataById).map((data: IData) => {
+    const model = {
+      ...data,
+      id: keepIdAttribute && (data.ORIGINAL_MODEL_ID_BEFORE_TRANSFORMATION || data.id),
+    };
+    delete (model as IData).ORIGINAL_MODEL_ID_BEFORE_TRANSFORMATION;
```
sroy3: Why not just use the `IData` instead of creating the type `Model`?

src/components/tables/examples/TableRowExamples.tsx
```diff
@@ -33,7 +33,7 @@ export class TableRowExamples extends React.Component<any, any> {
               <TableHeadingRow isCollapsible={true} opened={true}>
                 {firstRowCols}
               </TableHeadingRow>
-              <TableCollapsibleRow id='second-row' nbColumns={4} opened={true}>
+              <TableCollapsibleRow nbColumns={4} opened={true} id='second-row'>
```
sroy3: `opened={true}` can simply be `opened`

src/components/tables/table-children/TableChildBlankSlate.tsx
```diff
@@ -0,0 +1,34 @@
...
+    blankSlateOnError,
+    tableCompositeState,
+    initialTableData,
+  } = props;
+  const tableData = tableCompositeState.data || initialTableData;
+
+  let blankSlatePropsToUse: IBlankSlateProps;
+
+  if (tableData.displayedIds.length
+    || tableCompositeState.isLoading) {
```
sroy3: I think this can be on a single line.

src/components/dropdownSearch/DropdownSearch.tsx
MichelML: Sorry not having telling you, but if you look at the code and UTs, this is handled now @sroy3 .

src/components/loading/Loading.tsx
MichelML: changed

src/components/calendar/tests/Calendar.spec.tsx
```diff
@@ -293,11 +293,11 @@ describe('Calendar', () => {
           isFor: CalendarSelectionRuleType.all
         },
         {
-          test: (date: Date) => date.getDay() !== 6, // You cannot start your selection on a Saturday
+          test: (date: Date) => date.getDay() !== 6, // You cannot start your selection with Saturday
           isFor: CalendarSelectionRuleType.lower
         },
         {
-          test: (date: Date) => date.getDay() !== 0, // You cannot end your selection on a Sunday
+          test: (date: Date) => date.getDay() !== 0, // You cannot end your selection witn Sunday
```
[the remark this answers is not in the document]
MichelML: reverted, tried something else with the rules before the final change and forgot to revert

src/components/tables/TableUtils.ts
MichelML: because IData has a mandatory id, `Model` has not. but made a change not to force typing

src/components/tables/TableHeader.tsx
MichelML: i'm not sure of what you want me to do here, I have an idea, but I think the current implementation is good enough. Like you said, `withReduxState` is not applied on the TableHeader here, and if I put it at the TableHeaderCell level, I would have to add a kind of TableHeaderCellFactory component to add the `withReduxState` logic in it, and return the connected TableHeaderCell or unconnected one depending on the props passed to it.

sroy3: `ITableHeaderProps` should extend `IReduxStatePossibleProps` so that all components that can be toggled from not connected to connected all work the same way.

src/components/tables/table-children/TableChildBlankSlate.tsx
MichelML: changed

src/components/tables/examples/TableRowExamples.tsx
MichelML: changed

src/components/calendar/tests/Calendar.spec.tsx
```diff
@@ -422,59 +422,64 @@ describe('Calendar', () => {
         expect(day.isSelectable).toBe(true);
       });
 
-      it('should return day isSelectable if the day is not a Saturday and selecting lower limit', () => {
-        let otherDay: IDay = _.extend({}, DAY, { date: moment().endOf('week').add(1, 'day') });
-        let saturday: IDay = _.extend({}, DAY, { date: moment().endOf('week') });
-        let selectionLowerLimit: IDatePickerState = _.extend({}, CALENDAR_SELECTION, { selected: DateLimits.lower });
+      describe('saturday and sunday rule', () => {
```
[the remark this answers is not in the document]
MichelML: thanks

src/components/tables/TableConnected.tsx
MichelML: moved in default props, does not affect UTs

src/components/tables/TableHeader.tsx
MichelML: changed, let me know what you think

src/components/navigation/perPage/tests/NavigationPerPage.spec.tsx
MichelML: changed

package.json
```diff
@@ -52,8 +52,10 @@
     "@types/chosen-js": "1.6.1",
     "@types/classnames": "2.2.3",
     "@types/enzyme": "2.8.7",
+    "@types/faker": "^4.1.2",
```
GermainBergeron: set specific version

package.json
```diff
@@ -77,6 +79,7 @@
     "eslint-plugin-import": "2.8.0",
     "eslint-plugin-jsx-a11y": "6.0.2",
     "eslint-plugin-react": "7.4.0",
+    "faker": "^4.1.0",
```
GermainBergeron: set specific version

src/components/actions/tests/ActionBar.spec.tsx
```diff
@@ -164,5 +164,16 @@ describe('Actions', () => {
         });
       });
     });
+
+    describe('loading behavior', () => {
+      it('should not have the mod-deactivate-pointer class if the action bar is not loading', () => {
+        expect(actionBar.find('div').first().hasClass('mod-deactivate-pointer')).toBe(false);
```
GermainBergeron: here we should explicitly set the isLoading to false to avoid confusion if we change the default props

src/components/actions/tests/actionBarsReducers.spec.ts
```diff
@@ -67,90 +69,92 @@ describe('Actions', () => {
...
+            actions: undefined
+          }
+        ];
+      });
+
+      it('should return the old state without the PromptState with the prompt id when the action is "REMOVE_ACTION_BAR', () => {
+        const action: IReduxAction<IActionBarPayload> = {
+          type: ActionBarActions.remove,
+          payload: {
+            id: 'some-action-bar'
```
GermainBergeron: `id: oldState[1].id` will avoid issue on refactoring (same thing for the next ids, you should use the value from the array)

src/components/dropdownSearch/DropdownSearch.tsx
```diff
@@ -326,10 +347,27 @@ export class DropdownSearch extends React.Component<IDropdownSearchProps, {}> {
     }
   }
 
+  handleOnKeyDownFilterBox(e: React.KeyboardEvent<HTMLInputElement>) {
```
GermainBergeron: Should probably be private (handleOnKeyDownDropdownButton too)

src/components/dropdownSearch/DropdownPrepend.tsx
```diff
@@ -0,0 +1,11 @@
+// tslint:disable-next-line:no-unused-variable
+import * as React from 'react';
+
+export interface IDropdownPrependProps {
+  prepend: JSX.Element | string;
+  key?: string;
+  children?: any;
+}
+
+export const DropdownPrepend = (props: IDropdownPrependProps): JSX.Element =>
```
GermainBergeron: Can you use a `<Content/>` instead? It looks quite similar

src/components/dropdownSearch/tests/DropdownSearch.spec.tsx
```diff
@@ -230,14 +250,35 @@ describe('DropdownSearch', () => {
...
+      it('should show the filterBox if the dropdown is open and search is on', () => {
+        renderDropdownSearch(_.extend({}, ownProps, {
+          isOpened: true,
+        }));
+
+        expect(dropdownSearch.find(FilterBox).length).toBe(1);
+      });
+
+      it('should show the button if the dropdown is open and search is off and supportSingleCustomOption is true', () => {
+        const infiniteSerchThresold = 1000000000000;
```
GermainBergeron: `const infiniteSearchThreshold = Infinity;` (use infinity and you had two typos)

src/components/dropdownSearch/tests/DropdownSearch.spec.tsx
```diff
@@ -230,14 +250,35 @@ describe('DropdownSearch', () => {
...
           isOpened: true,
+          searchThresold: infiniteSerchThresold,
+          supportSingleCustomOption: true,
         }));
 
         expect(dropdownSearch.find(FilterBox).length).toBe(1);
       });
 
+      it('should show the button if the dropdown is open and search is off and supportSingleCustomOption is false', () => {
+        const infiniteSerchThresold = 1000000000000;
```
GermainBergeron: same thing here

src/components/tables/Table.tsx
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  componentWillUnmount() {
+    if (this.props.onUnmount) {
+      this.props.onUnmount();
+    }
+  }
+
+  private hasTableCompositeStateChanged(currentTableCompositeState: ITableCompositeState, nextTableCompositeState: ITableCompositeState): boolean {
+    return !!currentTableCompositeState && (
```
rousselm4: 3 points:

1. Je sortirais la petite fonction pour checker si les predicates on changé, ça aiderait juste à voir ce que ça fait en 1 coup d'oeil, ça ferait une moins grosse charge...

2. Je remplacerais les isEmpty par des comparaison de lenght. Enfait je suis même pas de comprendre pourquoi on fait un isEmpty... Mais d'après moi comparer la length devrait être suffisant.. Corrige-moi si je me trompe :)

3. Je sais pas si c'est mieux, mais tu pourrais checker voir si ça serait pas plus claire de mettre la chaine de OR dans un some. Je te fais un exemple des 3 modifications applique les si tu crois que c'est plus claire :)

```
const predicatesChanged =  _.some(currentTableCompositeState.predicates,
    (attributeValue: any, attributeName: string) => attributeValue !== nextTableCompositeState.predicates[attributeName],
);
return !!currentTableCompositeState && _.some([
    currentTableCompositeState.filter !== nextTableCompositeState.filter,
    currentTableCompositeState.perPage !== nextTableCompositeState.perPage,
    currentTableCompositeState.page !== nextTableCompositeState.page,
    currentTableCompositeState.sortState.attribute !== nextTableCompositeState.sortState.attribute,
    currentTableCompositeState.sortState.order !== nextTableCompositeState.sortState.order,
    currentTableCompositeState.predicates.length !== nextTableCompositeState.predicates.length,
    predicatesChanged,
], (changed: boolean) => changed);
```

src/components/navigation/perPage/tests/NavigationPerPage.spec.tsx
```diff
@@ -146,7 +146,20 @@ describe('NavigationPerPage', () => {
...
-      expect(newProps.onPerPageClick).toHaveBeenCalled();
+      expect(newProps.onPerPageClick).toHaveBeenCalledTimes(1);
+    });
+
+    it('should not call onPerPageClick prop if perPage is identical to currentPerPage', () => {
+      const newProps: INavigationPerPageProps = _.extend({}, NAVIGATION_PER_PAGE_BASIC_PROPS,
+        { onPerPageClick: jasmine.createSpy('onPerPageClick'), currentPerPage: 10 });
+      const expectedPerPage: number = 10;
+
+      expect(() => navigationPerPageInstance['handleClick'].call(navigationPerPageInstance, expectedPerPage)).not.toThrow();
```
GermainBergeron: Can you call `() => navigationPerPageInstance.handleClick(expectedPerPage)` instead?

src/components/navigation/perPage/tests/NavigationPerPage.spec.tsx
```diff
@@ -146,7 +146,20 @@ describe('NavigationPerPage', () => {
...
+
+      navigationPerPage.setProps(newProps);
+      // two clicks should call the function once
+      navigationPerPageInstance['handleClick'].call(navigationPerPageInstance, expectedPerPage);
```
GermainBergeron: here too

src/components/tables/TableBase.tsx
```diff
@@ -0,0 +1,124 @@
...
+      : null;
+  }
+
+  buildNavigation(): JSX.Element {
+    const { perPage, pagination, styles } = this.props;
+
+    if (!perPage && !pagination) {
+      return null;
+    }
+
```
rousselm4: un space de trop 😬

src/components/tables/TableRowWrapper.tsx
```diff
@@ -0,0 +1,5 @@
+import * as React from 'react';
+
+export const TableRowWrapper = (props?: any): JSX.Element => (
```
rousselm4: J'aime ben ça quand c'est des fonctions au lieux de classes, ça fait très simple, c'est nice. Pour le any tu peux pas mettre autre chose?

src/components/dropdownSearch/tests/DropdownSearch.spec.tsx
MichelML: not a fan of Infinity because of it's "specialness" , such as Math.ceil(1/Infinity) === 0 to be true, which may be mathematically correct, but leaves it behaving differently than other regular numbers.

src/components/navigation/perPage/tests/NavigationPerPage.spec.tsx
MichelML: this is like an `as any` by passing for private methods, i copy pasted it from other places in the file

src/components/navigation/perPage/tests/NavigationPerPage.spec.tsx
MichelML: see previous comment

src/components/tables/Table.tsx
```diff
@@ -0,0 +1,173 @@
...
+import { TableChildHeader } from './table-children/TableChildHeader';
+import { TableChildLoadingRow } from './table-children/TableChildLoadingRow';
+import { TableChildBlankSlate } from './table-children/TableChildBlankSlate';
+import { TableChildNavigation } from './table-children/TableChildNavigation';
+import { TableChildLastUpdated } from './table-children/TableChildLastUpdated';
+import { TableChildBody } from './table-children/TableChildBody';
+
+export interface IData {
+  id: string;
+  [attribute: string]: any;
```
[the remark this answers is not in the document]
MichelML: consider receiving a JSON object from the server, the data could be anything

src/components/tables/Table.tsx
```diff
@@ -0,0 +1,173 @@
...
+};
+
+export interface ITableRowData {
+  [id: string]: IData;
+};
+
+export type IAttributeFormatter = (attributeValue: any, attributeName?: string) => JSXRenderable;
```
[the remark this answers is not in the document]
MichelML: let me know if you have an idea

src/components/tables/Table.tsx
MichelML: i could refactor it in many ways but i think the current implementation is good enough, we can talk about it live if you want

src/components/dropdownSearch/DropdownPrepend.tsx
MichelML: done

src/components/dropdownSearch/DropdownPrepend.tsx
```diff
@@ -0,0 +1,11 @@
+// tslint:disable-next-line:no-unused-variable
+import * as React from 'react';
+
+export interface IDropdownPrependProps {
+  prepend: JSX.Element | string;
+  key?: string;
+  children?: any;
```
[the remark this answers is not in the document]
MichelML: deleted the file

src/components/dropdownSearch/DropdownSearch.tsx
MichelML: refactored methods accessibility in the whole file

src/components/dropdownSearch/tests/DropdownSearch.spec.tsx
MichelML: had a specific case for the pagination calculation in TableDataModifier.ts where Infinity was causing problems and not 1000000000000

src/components/navigation/perPage/tests/NavigationPerPage.spec.tsx
GermainBergeron: In this case i'd prefer the `as any`: `() => (navigationPerPageInstance as any).handleClick(expectedPerPage)` its easier to understand in my opinion

src/components/tables/tests/TableHeaderCellReducers.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+    const unrelatedAction = { type: 'any', payload: {} };
+    expect(tableHeaderCellsReducer(undefined, unrelatedAction)).toEqual(tableHeaderCellsInitialState);
+  });
+
+  it('should return the same state if the action is unrelated and the state is defined for tableHeaderCellReducer', () => {
+    const unrelatedAction = { type: 'any', payload: {} };
+    expect(tableHeaderCellReducer(testHeaderCell, unrelatedAction)).toEqual(testHeaderCell);
+  });
+
+  it(`should return the state with the new header cell in it on ${TableHeaderCellActions.add}`, () => {
```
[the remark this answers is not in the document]
GermainBergeron: Same thing for the other `TableHeaderCellActions.X` tests

src/components/tables/tests/TableReducers.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    expect(newState).toEqual(tablesInitialState);
+  });
+
+  it('should return the old state when the action is not defined', () => {
+    expect(tablesReducer(tablesInitialState, genericAction)).toBe(tablesInitialState);
+  });
+
+  it('should return the old state with a new table in it on add table', () => {
+    const tableId = 'super-table';
```
GermainBergeron: You should define the id once instead of in each test

src/components/tables/tests/TableTestCommon.tsx
```diff
@@ -0,0 +1,110 @@
...
+  initialTableData: DEFAULT_TABLE_DATA,
+  tableCompositeState: { sortState: {}, data: DEFAULT_TABLE_DATA } as any,
+};
+
+export const predictableData: IData = {
+  id: 'funkyUserName',
+  email: 'awesome@awesome.com',
+  avatar: 'whatever',
+  userName: 'funkyUserName',
+  password: 'lakjhdslfksh98323qasd',
```
GermainBergeron: 🙀

src/components/tables/tests/TableUtils.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+describe('TableUtils', () => {
+  describe('getNextTableSortingOrder', () => {
+    const {
+      ASCENDING,
+      DESCENDING,
+      UNSORTED,
+    } = TableSortingOrder;
+
+    it('should return ASCENDING if the previous order was UNSORTED', () => {
+      expect(getNextTableSortingOrder(UNSORTED)).toBe(ASCENDING);
```
GermainBergeron: Quick question, is it possible to change the default sort from UNSORTED? In UA tables if we sort on metrics, we sort descending by default (e.g., to get the values with most clicks first)

MichelML: you could initialize the table with `initialTableData` sorted in the way you want, but the icon in the table header would not be a reflection of this state for now

src/components/tables/Table.tsx
```diff
@@ -0,0 +1,172 @@
...
+      },
+    );
+
+    return (
+      <div className={classNames('table-container', this.props.tableContainerClasses)}>
+        <TableChildActionBar {...this.props} />
+        <table className={tableClasses}>
+          <TableChildLoadingRow {...this.props} isInitialLoad={this.isInitialLoad} />
+          <TableChildHeader {...this.props} />
+          {TableChildBody(this.props)}
```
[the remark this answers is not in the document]
MichelML: yes it returns an array of JSX.Element

src/components/tables/TableUtils.ts
```diff
@@ -0,0 +1,61 @@
...
+export const getTableLoadingIds = (tableId: string): string[] => ([
+  tableId,
+  getTableChildComponentId(tableId, TableChildComponent.ACTION_BAR),
+  getTableChildComponentId(tableId, TableChildComponent.LOADING_NAVIGATION),
+]);
+
+export type Model = {
+  [key: string]: any;
+};
+export type Collection = Model[];
```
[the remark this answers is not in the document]
MichelML: imho models and collections are an abstract concepts way beyond Backbone

watch before and after, imho this this is cleaner naming them:  

```
export const convertInitialCollectionToDataById = (collection: Collection, attributeNameForId: string): ITableRowData => {
  return collection.reduce((finalData: ITableRowData, model: Model) => {
    return {
      ...finalData,
      [model[attributeNameForId]]: {
        ...model,
        id: model[attributeNameForId],
        ORIGINAL_MODEL_ID_BEFORE_TRANSFORMATION: model.id,
      }
    };
  }, DEFAULT_TABLE_DATA.byId);
};

export const convertDataByIdToCollection = (dataById: ITableRowData, keepIdAttribute = true): Collection => {
  return _.values(dataById).map((data: IData) => {
    const model: Model = {
      ...data,
      id: keepIdAttribute && (data.ORIGINAL_MODEL_ID_BEFORE_TRANSFORMATION || data.id),
    };
    delete model.ORIGINAL_MODEL_ID_BEFORE_TRANSFORMATION;

    return model;
  });
};
```

```
export const convertInitialCollectionToDataById = (collection: ({[key: string]: any})[], attributeNameForId: string): ITableRowData => {
  return collection.reduce((finalData: ITableRowData, model: {[key: string]: any}) => {
    return {
      ...finalData,
      [model[attributeNameForId]]: {
        ...model,
        id: model[attributeNameForId],
        ORIGINAL_MODEL_ID_BEFORE_TRANSFORMATION: model.id,
      }
    };
  }, DEFAULT_TABLE_DATA.byId);
};

export const convertDataByIdToCollection = (dataById: ITableRowData, keepIdAttribute = true): ({[key: string]: any})[] => {
  return _.values(dataById).map((data: IData) => {
    const model: {[key: string]: any} = {
      ...data,
      id: keepIdAttribute && (data.ORIGINAL_MODEL_ID_BEFORE_TRANSFORMATION || data.id),
    };
    delete model.ORIGINAL_MODEL_ID_BEFORE_TRANSFORMATION;

    return model;
  });
};
```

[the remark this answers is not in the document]
MichelML: removed anyway

src/components/tables/tests/TableChildActionBar.spec.tsx
```diff
@@ -0,0 +1,140 @@
...
+        expect(tableActionBar.find(ActionBarConnected).length).toBe(1);
+        expect(tableActionBar.find(ActionBarConnected).find(FilterBoxConnected).length).toBe(1);
+      });
+
+      it('should render with an action bar and a predicate inside it if there is an actionBar prop and one predicate', () => {
+        const tableActionBar = mountComponentWithProps({
+          ...tablePropsMock,
+          actionBar: true,
+          predicates: [
+            { attributeName: 'email', attributeNameFormatter: _.identity, props: {} },
```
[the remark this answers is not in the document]
MichelML: i've made a change so you only have to pass the dropdown options to it in the props, and it humanizes the attributeName if you haven't passed an attributeNameFormatter. However in V2 you should always pass an attributeNameFormatter for localization purposes.

src/components/tables/TableReducers.ts
MichelML: done

src/components/tables/TableUtils.ts
MichelML: removed Collection and Model types

src/components/tables/table-children/TableChildBody.tsx
```diff
@@ -0,0 +1,62 @@
...
+    const tableRowWrapperClasses = classNames({ 'table-body-loading': !!props.tableCompositeState.isLoading });
+
+    return (
+      <TableRowWrapper key={rowWrapperId} className={tableRowWrapperClasses}>
+        <TableHeadingRowConnected
+          id={headingAndCollapsibleId}
+          key={headingAndCollapsibleId}
+          tableId={props.id}
+          isCollapsible={!!collapsibleData}
+          onClickCallback={(e: React.MouseEvent<any>) => {
```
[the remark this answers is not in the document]
MichelML: done

src/components/tables/TableHeadingRowConnected.tsx
```diff
@@ -16,18 +16,17 @@ const mapStateToProps = (state: IReactVaporState, ownProps: ITableHeadingRowOwnP
...
-  });
+const mapDispatchToProps = (dispatch: IDispatch, ownProps: ITableHeadingRowOwnProps) => ({
+  onClick: () => {
+    dispatch(selectRow(ownProps.id, ownProps.isCollapsible, ownProps.tableId));
+  },
+  onRender: () => {
+    dispatch(addRow(ownProps.id, ownProps.tableId));
+  },
+  onDestroy: () => {
+    dispatch(removeRow(ownProps.id));
```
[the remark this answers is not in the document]
MichelML: done

src/components/navigation/pagination/tests/NavigationPaginationReducers.spec.ts
```diff
@@ -143,5 +145,38 @@ describe('Reducers', () => {
 
       expect(paginationState[0]).toEqual(jasmine.objectContaining(newState));
     });
+
+    it('should set the page number at 0 for the table state modification action if the table id is in the pagination id and shouldResetPage is true', () => {
+      const tableId = 'table';
+      const shouldResetPage = true;
+      const oldState: IPaginationState = {
+        id: `pagination${tableId}`,
+        pageNb: 22
```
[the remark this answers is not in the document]
MichelML: done

src/components/dropdownSearch/DropdownSearchConnected.tsx
```diff
@@ -9,7 +9,7 @@ import {
   removeDropdownSearch,
   selectOptionDropdownSearch,
   toggleDropdownSearch,
-  updateActiveOptionDropdownSearch, updateOptionsDropdownSearch,
+  updateActiveOptionDropdownSearch, updateOptionsDropdownSearch, closeDropdownSearch,
```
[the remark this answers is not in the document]
MichelML: done

src/components/navigation/pagination/tests/NavigationPagination.spec.tsx
```diff
@@ -96,6 +96,10 @@ describe('NavigationPagination', () => {
         navigationPagination.find('.selectable').first().simulate('click');
         expect(clickSpy.calls.count()).toBe(1);
 
+        // Number button (does not call again since already on this page)
+        navigationPagination.find('.flat-select-option').first().simulate('click');
+        expect(clickSpy.calls.count()).toBe(1);
```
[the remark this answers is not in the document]
MichelML: done

PR #430, opened 2017-12-22
handle padding depending on context and add padding top

MichelML:
screen shot 2017-12-22 at 3 26 07 pm
screen shot 2017-12-22 at 3 25 50 pm
screen shot 2017-12-22 at 3 26 17 pm

tough to see in the image, but the first one does not have enough padding on top and is different compare to other headers

### DIFF
--- a/src/components/headers/HeaderWrapper.tsx
+++ b/src/components/headers/HeaderWrapper.tsx
@@ -13,12 +13,19 @@ export interface IHeaderWrapperProps extends ITabsHeaderProps, React.ClassAttrib
 export class HeaderWrapper extends React.Component<IHeaderWrapperProps, {}> {
 
   private getClasses(): string {
-    return classNames('flex',
-      'flex-center space-between',
+    return classNames(
+      'flex',
+      'flex-center',
+      'space-between',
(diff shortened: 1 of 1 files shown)


src/components/headers/HeaderWrapper.tsx
```diff
@@ -13,12 +13,19 @@ export interface IHeaderWrapperProps extends ITabsHeaderProps, React.ClassAttrib
...
+      'flex-center',
+      'space-between',
       'mod-header-padding',
       'header-height',
-      'pb2',
-      this.props.classes);
+      {
+        'pt1': !!this.props.tabs,
+        'pb2': !!this.props.tabs,
+        'p1': !this.props.tabs,
```
GermainBergeron: This will add padding left and right too which will be overriden by `mod-header-padding`. I'd prefer `pl1` and `pr1`

MichelML: nice catch, i actually need pt1 and pb1

MichelML: fixed

PR #438, opened 2018-01-08
use IData instead of ITableRowData where appropriate

MichelML:
```
	interface IData {
	    id: string;
	    [attribute: string]: any;
	}
```

```
	interface ITableRowData {
	    [id: string]: IData;
	}
```


MichelML:
yes @sroy3 , `ITableRowData` is simply one level of abstraction above compared to `IData`, see the main description i've just edited.

PR #441, opened 2018-01-09
fix type to avoid complaints from ts

MichelML:


MichelML:
i think it would solve this, which i get in ml right now
```
ERROR in ./src/ml-models/MLModelsTable.tsx
[tsl] ERROR in /Users/mmlapointe/cloud_frontend/jsadmin-machine-learning/src/ml-models/MLModelsTable.tsx(47,9)
      TS2605: JSX element type 'Component<ITableProps, ComponentState>' is not a constructor function for JSX elements.
  Types of property 'state' are incompatible.
    Type 'void | Readonly<{}>' is not assignable to type 'Readonly<any>'.
      Type 'void' is not assignable to type 'Readonly<any>'.
```

PR #446, opened 2018-01-12
Add slider

pochretien:
screen shot 2018-01-12 at 11 08 49 am


MichelML:
@pochretien you have a live demo at https://coveo.github.io/react-vapor/add-slider/ (but the vapor style is not there yet obviously) ... made some adjustments to the live-demo script because it was breaking your build

src/components/slider/Slider.tsx
```diff
@@ -0,0 +1,29 @@
+import * as React from 'react';
+import * as classNames from 'classnames';
+import RCSlider, { createSliderWithTooltip, SliderProps } from 'rc-slider';
+
+export interface ISliderProps {
+  slider?: SliderProps;
+  classes?: string[];
```
sbelzile-Coveo: Since we use `classnames` everywhere, I don't really see the point of classes properties being arrays. `className="foo foo2"` takes less space than `className={["foo", "foo2"]}`. Furthermore, these props could be named `className` to keep the same conventions as in React.

sroy3:
You haven't exported your component in the Index file.

src/components/slider/Slider.tsx
sroy3: We use classnames internally, but not necessarily externally. It might be shorter without, but I think the array is cleaner and is easier to interact with. Changing the prop name for `className` is a good idea though.

src/Index.ts
```diff
@@ -218,3 +218,6 @@ export { ToastContainerConnected } from './components/toast/ToastContainerConnec
 export { BreadcrumbHeader } from './components/headers/BreadcrumbHeader';
 export { HeaderWrapper } from './components/headers/HeaderWrapper';
 export { BasicHeader } from './components/headers/BasicHeader';
+
+// Slider
+export { Slider } from './components/slider/Slider';
```
GermainBergeron: You should probably export `ISliderProps` too

PR #448, opened 2018-01-14
print the demo url in travis logs

MichelML:
Since the live demo script is nom in a `after_script` the demo url doesn't show in the build logs. We need to echo it at the `script` level to be able to retrieve it in the logs. I added something to retrieve it in the commits seeable in the gh-pages branch https://github.com/coveo/react-vapor/tree/gh-pages (see besides the `print-live-demo` folder)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,8 +20,10 @@ cache:
     - node_modules
 
 script:
- - 'if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then npm run docs && node create-live-demo.js; fi' # run this line on pull request builds only
  - npm run test
 
 after_success:
  - npm run report-coverage
+
+after_script:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -21,6 +21,7 @@ cache:
 
 script:
  - npm run test
+ - 'if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then echo "Your live demo url, if everything went well, should be https://coveo.github.io/react-vapor/$(TRAVIS_PULL_REQUEST_BRANCH)/"; fi' # run this line on pull request builds only
```
sroy3: Any reason we aren't using `== "true"` instead of `!= "false"`?

MichelML: IIRC I copied this from the travis docs so i just did not bother trying `== "true"`

PR #463, opened 2018-01-24
Collapsible Container - Connected with UTs

MichelML:
video: https://drive.google.com/file/d/1vgA5992nNksyDjkCm3RSrUGgFpRgBw8M/view
image:
screen shot 2018-01-24 at 2 38 29 pm  

live demo (the style is not good for the icons, but this is just in the live demo): https://coveo.github.io/react-vapor/collapsiblecontainer/


src/components/collapsibleContainer/CollapsibleContainerConnected.spec.tsx
```diff
@@ -0,0 +1,142 @@
...
+      <Provider store={store}>
+        <CollapsibleContainerConnected {...props}>
+          dummy children
+        </CollapsibleContainerConnected>
+      </Provider>,
+      { attachTo: document.getElementById('App') },
+    );
+  };
+
+  it('should get what to do onMount', () => {
```
sbelzile-Coveo: Le nom du test devrait être qqchose comme `should not throw onMount`. Même chose pour les 2 tests d'après.

src/utils/ClassNameUtils.ts
```diff
@@ -0,0 +1,9 @@
+import { IClassName } from './ClassNameUtils';
```
sbelzile-Coveo: Nécéssaire ce self-import?

src/components/collapsibleContainer/CollapsibleContainer.tsx
```diff
@@ -0,0 +1,124 @@
...
+  onUnmount?: () => void;
+  onToggleExpandedState?: (currentExpandedState: boolean) => void;
+}
+
+export interface ICollapsibleContainerProps extends
+  ICollapsibleContainerOwnProps,
+  ICollapsibleContainerStateProps,
+  ICollapsibleContainerDispatchProps { }
+
+export class CollapsibleContainer extends React.Component<ICollapsibleContainerProps, {}> {
```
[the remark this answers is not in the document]
MichelML: fixed

src/components/collapsibleContainer/CollapsibleContainerConnected.spec.tsx
MichelML: fixed

src/utils/ClassNameUtils.ts
MichelML: fixed

src/components/collapsibleContainer/CollapsibleContainer.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+export class CollapsibleContainer extends React.Component<ICollapsibleContainerProps> {
+  private toggleCollapsibleBodyOnExpandedStateChange(nextProps: ICollapsibleContainerProps) {
+    const collapsibleBodySelector = `#${this.props.id} .collapsible-body`;
+    const animationOptions = { duration: 350 };
+
+    if (this.props.expanded && !nextProps.expanded) {
+      $(collapsibleBodySelector).slideUp(animationOptions);
+    } else if (!this.props.expanded && nextProps.expanded) {
+      $(collapsibleBodySelector).slideDown(animationOptions);
```
sroy3: Let's try to use jquery only when absolutely necessary. You can easily use css transitions here. You can also use React's transitions https://reactjs.org/docs/animation.html.

src/components/collapsibleContainer/CollapsibleContainer.tsx
```diff
@@ -0,0 +1,124 @@
...
+    }
+  }
+
+  private getCollapsibleHeaderClass(): string {
+    return classNames(
+      'collapsible-header btn',
+      { active: this.props.expanded },
```
sroy3: I think `active: ...` should be on its own line just like `display-block` below, or the opposite.

src/components/collapsibleContainer/CollapsibleContainer.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+    const collapsibleHeaderIcon = this.props.informationUrl
+      ? <LinkSvg url={this.props.informationUrl} target='_blank' tooltip={tooltipProps} svg={svgProps} />
+      : <Tooltip {...tooltipProps}><Svg {...svgProps} /></Tooltip>;
+
+    return <span className='round-contextual-help'>{collapsibleHeaderIcon}</span>;
+  }
+
+  componentWillReceiveProps(nextProps: ICollapsibleContainerProps) {
+    this.toggleCollapsibleBodyOnExpandedStateChange(nextProps);
```
sroy3: You might want to look if the expanded prop has changed here. Else `this.toggleCollapsibleBodyOnExpandedStateChange(nextProps);` will be called with every prop change.

MichelML: i do it in the function itself...that would be duplicate logic

src/components/collapsibleContainer/CollapsibleContainer.tsx
MichelML: I did try with css and nothing worked as what we have in backbone/marionette...i guess you arrived at the same conclusion for the table collapsible rows where you used jquery as well? I would like you to consider leaving it as is here...this is the exact same behavior we have everywhere in the anim

MichelML: We could log another story for that too, i don't see it as a priority as of now. I can do that if you want.

src/components/collapsibleContainer/CollapsibleContainer.tsx
MichelML: fixed

src/components/collapsibleContainer/CollapsibleContainer.tsx
sroy3: `toggleCollapsibleBodyOnExpandedStateChange` is still called every time though. You could remove it from the function and bring it here instead.

MichelML: if i move it here, the same code will run in both cases...might not be here anyway after the jquery thing

src/components/collapsibleContainer/CollapsibleContainerReducers.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+      const setExpandedToTrueAction = setExpandedCollapsibleContainer('a different id', true);
+      const newState = collapsibleContainersReducer(oldState, setExpandedToTrueAction);
+
+      expect(newState[0].expanded).toBeUndefined();
+
+      const setExpandedToFalseAction = setExpandedCollapsibleContainer('a different id', false);
+      const secondNewState = collapsibleContainersReducer(oldState, setExpandedToFalseAction);
+
+      expect(secondNewState[0].expanded).toBeUndefined();
+    });
```
GermainBergeron: You should probably also validate that your actions aren't modifying the oldstate

MichelML:
@sroy3 https://drive.google.com/file/d/1kbEAJRdsOyOJhUsbKcOSI1c7ASUN_AmT/view ... a slideIn, slideDown

i started from here https://github.com/frankwallis/react-slidedown/tree/master/lib and reimplemented it with the new API of the lib you suggested

src/components/collapsibleContainer/CollapsibleContainerReducers.spec.ts
MichelML: somewhat done depending on what you meant...you can check the last test

src/components/collapsibleContainer/CollapsibleContainer.tsx
```diff
@@ -0,0 +1,111 @@
...
+    return (
+      <div id={this.props.id} className={classNames('collapsible', this.props.className)}>
+        <button
+          type='button'
+          className={this.getCollapsibleHeaderClass()}
+          onClick={(e: React.MouseEvent<HTMLInputElement>) => this.props.onToggleExpandedState(this.props.expanded)}>
+          {this.props.title}
+          {this.getCollapsibleHeaderIcon()}
+        </button>
+        <SlideY in={this.props.expanded} timeout={350}>
```
GermainBergeron: Really nice, I like it!

PR #465, opened 2018-01-26
Add markup to allow for disabled row

sroy3:
screen shot 2018-01-26 at 12 43 19 pm


src/components/tables/table-children/TableChildBody.tsx
```diff
@@ -1,70 +1,90 @@
...
 import { JSXRenderable } from '../../../utils/JSXUtils';
+import { IActionOptions } from '../../actions/Action';
 
-export const TableChildBody = (props: ITableProps): JSX.Element[] => {
-  const tableData = props.tableCompositeState.data || props.initialTableData;
-  return tableData.displayedIds.map((id: string, yPosition: number): JSX.Element => {
-    const rowData: IData = tableData.byId[id];
-    const rowWrapperId = `${getTableChildComponentId(props.id, TableChildComponent.TABLE_ROW_WRAPPER)}${rowData.id}`;
-    const headingAndCollapsibleId = `${getTableChildComponentId(props.id, TableChildComponent.TABLE_HEADING_ROW)}${rowData.id}`;
+export interface IInheritedFromTableProps {
```
GermainBergeron: Is this interface used by something else than the table? I'd expect the name to be ITableBodyInheritedFromTableProps or something similar if that's not the case

If it's used by something else (e.g., the table action bar) I'd expect it to be somewhere more generic

src/components/tables/table-children/TableChildBody.tsx
```diff
@@ -1,70 +1,90 @@
...
+      ? attributeFormatter(props.rowData[attributeName], attributeName)
+      : convertUndefinedAndNullToEmptyString(props.rowData[attributeName]);
 
-    const tableRowWrapperClasses = classNames({ 'table-body-loading': !!props.tableCompositeState.isLoading });
+    return (<td key={tableCoordinate}>
+      <div className='wrapper'>{headingRowContent}</div>
+    </td>);
+  });
+
+  const collapsibleRowKey = `${getTableChildComponentId(props.tableId, TableChildComponent.TABLE_COLLAPSIBLE_ROW)}${props.rowData.id}`;
```
GermainBergeron: `const collapsibleRowKey = getTableChildComponentId(props.tableId, TableChildComponent.TABLE_COLLAPSIBLE_ROW) + props.rowData.id;`

I don't see the value of the literal string here

src/components/tables/Table.tsx
```diff
@@ -161,12 +159,47 @@ export class Table extends React.Component<ITableProps, {}> {
...
+    const tableData = this.props.tableCompositeState.data || this.props.initialTableData;
+
+    return tableData.displayedIds.map((id: string, yPosition: number): JSX.Element => {
+      const rowData: IData = tableData.byId[id];
+
+      return (<TableChildBody
+        key={id}
+        tableId={this.props.id}
+        rowData={rowData}
+        yPosition={yPosition}
```
GermainBergeron: Why do we need yPosition? Can't we use rowData.id for the key? (It was probably like this before). It seems to be use to generate the key of the row https://github.com/coveo/react-vapor/pull/465/files#diff-cfa237717e086f207d2c932bb4d33246R34

src/components/tables/tests/TableChildBody.spec.tsx
```diff
@@ -0,0 +1,145 @@
+import * as React from 'react';
```
GermainBergeron: Only renamed?

src/components/tables/table-children/TableChildBody.tsx
sroy3: I moved the loop out of the file and into the `Table` file. I do not need this line anymore (I've removed it on my side, but haven't pushed yet).

src/components/tables/tests/TableChildBody.spec.tsx
sroy3: The props used to be the whole table props, I edited `TableChildBody` so that it is simpler to test. So all tests have changed to only get what they need as props. And I've added a few more.

src/components/tables/Table.tsx
sroy3: You're right. Removed and simplified the key.

src/components/tables/Table.tsx
```diff
@@ -161,12 +159,47 @@ export class Table extends React.Component<ITableProps, {}> {
...
-          {TableChildBody(this.props)}
+          {this.getTableBody()}
         </table>
         <TableChildBlankSlate {...this.props} isInitialLoad={this.isInitialLoad} />
         <TableChildNavigation {...this.props} />
         <TableChildLastUpdated {...this.props} />
       </div>
     );
   }
+
```
MichelML: the following two methods should go above render https://github.com/coveo/tsjs/tree/master/react_redux#ordering

sroy3: Didn't we say that private functions should come after public ones?

src/components/tables/Table.tsx
MichelML: IIRC it's true that it was use for the key of the cells of the row...might not be needed if you use the row id with the x position of the cell or something

src/components/tables/Table.tsx
MichelML: i guess these two rules are clashing with each other than...we could consider tossing out the `private` concept out of react-vapor...anyways it's not major you can leave it like this

src/components/tables/table-children/TableChildBody.tsx
```diff
@@ -1,70 +1,84 @@
...
-          id={headingAndCollapsibleId}
-          key={collapsibleRowKey}
-          nbColumns={props.headingAttributes.length + TOGGLE_ARROW_CELL_COUNT}>
-          {collapsibleData}
-        </TableCollapsibleRowConnected>
-      )
-      : null;
+    return (<td key={`cell-${xPosition}`}>
+      <div className='wrapper'>{headingRowContent}</div>
+    </td>);
```
MichelML: i think it's quite unanimous that this style is better:

```jsx
return (
    <td key={`cell-${xPosition}`}>
        <div className='wrapper'>{headingRowContent}</div>
   </td>
);
```

src/components/tables/Table.tsx
```diff
@@ -161,12 +159,46 @@ export class Table extends React.Component<ITableProps, {}> {
...
+    );
+  }
+
+  private getTableBody() {
+    const tableData = this.props.tableCompositeState.data || this.props.initialTableData;
+
+    return tableData.displayedIds.map((id: string, yPosition: number): JSX.Element => {
+      const rowData: IData = tableData.byId[id];
+
+      return (<TableChildBody
```
MichelML: put TableChildBody on its own line

src/components/tables/examples/TableExamples.tsx
```diff
@@ -7,7 +7,7 @@ import * as _ from 'underscore';
 import { ITableData, ITableState, ITablesState, ITableCompositeState } from '../TableReducers';
```
MichelML: Can you add an example with a disabled row?

sroy3:
@MichelML there is nothing to override. It simply uses attributes that could possibly already be on the rowData. We already have something binded to the state of the row and that's exactly what we are changing here to set the disabled class here.

MichelML:
@sroy3 do you have a code example for me to see what you are doing in your current use case? It's like you assume that a `disabled` prop from the data you receive from the backend will always serve as a property to know if the row is disable or enable.  

We could question the implementation of the table which make it a bit confusing to have rowData in the `tables` reducer and also information about the rows in the `rows` reducer. I still think the `disabled` logic should start from the state of the rows (in the `rows` property of the state).

sroy3:
@MichelML it isn't part of the state, it is part of the data. I'm not changing it on the front end, it's when the data changes on fetch that the disable/enable properties change. For it to be in the state, we'd have to parse each row and to trigger an action for each one of them so that we could set them to disabled or enabled, while never changing this property manually. Only the data would change this state, which isn't really a state in the first place because we never change it.

MichelML:
@sroy3 can you add an example, so we make sure we are talking about the same thing?

src/components/tables/examples/TableExamples.tsx
sroy3: I have added an example

PR #466, opened 2018-01-30
Input connected + add possibility to set a label from the input props

MichelML:
@sbelzile-Coveo this doesn't break the examples in the docs, but you tell me if this could potentially break something on your side.

demo: https://coveo.github.io/react-vapor/input-connected/ , search for `Inputs list` in the page for quick access


src/animations/SlideY.spec.tsx
```diff
@@ -3,49 +3,49 @@ import { shallow } from 'enzyme';
 import { SlideY, onEntering, onExiting } from './SlideY';
```
MichelML: ignore this file, this is out of scope

src/components/input/InputConnected.tsx
```diff
@@ -0,0 +1,33 @@
...
+    value: input && input.value,
+    disabled: input && input.disabled,
+  };
+};
+
+const mapDispatchToProps = (dispatch: IDispatch, ownProps: IInputOwnProps): IInputDispatchProps => ({
+  onRender: (value: string = '', valid = true, disabled = false) => {
+    const validOnMount = ownProps.validateOnMount
+      && ownProps.validate
+      && ownProps.validate(value);
```
sroy3: Aren't you already getting this result from the `valid` parameter? You do the same thing on line 60 of `Input.tsx` where you call `onRender`.

src/components/input/InputConnected.tsx
```diff
@@ -0,0 +1,33 @@
...
+    dispatch(addInput(ownProps.id, value, validOnMount, disabled));
+  },
+  onDestroy: () => dispatch(removeInput(ownProps.id)),
+  onChange: (value: string) => {
+    const validOnChange = ownProps.validateOnChange
+      && ownProps.validate
+      && ownProps.validate(value);
```
sroy3: You should move this to the `Input.tsx` file like you have for the `validOnMount`. This way it can still be used when not using the connected component.

src/components/input/tests/InputConnected.spec.tsx
```diff
@@ -0,0 +1,164 @@
...
+  const mountComponentWithProps = (props: IInputProps) => mount(
+    <Provider store={store}>
+      <InputConnected {...props} />
+    </Provider>,
+    { attachTo: document.getElementById('App') },
+  );
+
+  it('should mount without errors in various props scenarios', () => {
+    expect(() => {
+      mountComponentWithProps({});
```
sroy3: You could set `{}` as the default and send nothing here.

MichelML:
@sbelzile-Coveo this is somewhat good news...let me know if ever you discover something more major

src/components/datePicker/tests/DatesSelection.spec.tsx
```diff
@@ -222,7 +222,7 @@ describe('Date picker', () => {
       }).not.toThrow();
 
       datesSelection.setProps({ onBlur: onBlurSpy });
-      (datesSelection.find('input').get(0) as React.HTMLAttributes<HTMLButtonElement>).value = DateUtils.getDateWithTimeString(new Date());
+      (datesSelection.find('input').get(0) as any).value = DateUtils.getDateWithTimeString(new Date());
```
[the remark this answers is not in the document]
MichelML:
```
      TS2339: Property 'value' does not exist on type 'HTMLAttributes<HTMLInputElement>'.
```

tried a couple of unsuccessful other types before using any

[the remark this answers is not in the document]
sroy3: Shouldn't it be `HTMLInputElement`?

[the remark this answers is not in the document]
MichelML: tried it, same error

MichelML:
@sroy3 for the validateOnMount, there could be a potential case in ML, we'll have a new feature which is to give a name to a Machine Learning Model, initially all machine learning models won't have a name, so we may prompt them to add one the first time they arrive in their machine learning model configuration in the ML UI.

src/components/input/InputConnected.tsx
MichelML: done 👍

src/components/input/tests/InputConnected.spec.tsx
MichelML: done 👍

src/components/input/InputConnected.tsx
MichelML: done 👍 , initially i wanted to put them in InputConnected only but you surely have a good point here

src/components/datePicker/tests/DatesSelection.spec.tsx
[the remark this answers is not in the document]
sroy3: I think it is because it is a `HTMLInputElement` and not a `React.HTMLAttributes<HTMLInputElement>`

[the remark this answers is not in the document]
MichelML: Tried it too, but says you can't convert a ReactElement to an HTMLInputElement, tried with ReactElement<HTMLInputElement> too but it's not working, there's always something.

PR #468, opened 2018-02-02
Modal composite

sroy3:
Quicker way to create a modal

Demo: https://coveo.github.io/react-vapor/modal-composite/

### DIFF
--- a/src/components/modal/ModalComposite.tsx
+++ b/src/components/modal/ModalComposite.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import * as _ from 'underscore';
 import { IModalProps, Modal } from './Modal';
-import {IModalHeaderProps, ModalHeader } from './ModalHeader';
+import { IModalHeaderProps, ModalHeader } from './ModalHeader';
 import { IModalFooterProps, ModalFooter } from './ModalFooter';
 import { IModalBackdropProps, ModalBackdrop } from './ModalBackdrop';
 import { IReduxStatePossibleProps } from '../../utils/ReduxUtils';
@@ -74,11 +74,15 @@ export class ModalComposite extends React.Component<IModalCompositeProps, {}> {
   }
 
   private getModalBody() {
-    return <ModalBody classes={this.props.modalBodyClasses}>{this.props.modalBodyChildren}</ModalBody>;
+    return this.props.modalBodyChildren
+      ? <ModalBody classes={this.props.modalBodyClasses}>{this.props.modalBodyChildren}</ModalBody>
+      : null;
   }
 
   private getModalFooter() {
-    return <ModalFooter classes={this.props.modalFooterClasses}>{this.props.modalFooterChildren}</ModalFooter>;
+    return this.props.modalFooterChildren
+      ? <ModalFooter classes={this.props.modalFooterClasses}>{this.props.modalFooterChildren}</ModalFooter>
+      : null;
   }
 
   private getModalBackdrop() {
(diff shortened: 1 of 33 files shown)


MichelML:
Demo: https://coveo.github.io/react-vapor/modal-composite/

src/components/modal/ModalComposite.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { IReduxStatePossibleProps } from '../../utils/ReduxUtils';
+import { ModalBackdropConnected } from './ModalBackdropConnected';
+import { ModalHeaderConnected } from './ModalHeaderConnected';
+import { JSXRenderable } from '../../utils/JSXUtils';
+import { ModalConnected } from './ModalConnected';
+import { ModalBody } from './ModalBody';
+
+export interface IModalCompositeOwnProps extends IModalProps, IModalHeaderProps, IModalFooterProps, IModalBackdropProps {
+  modalHeaderChildren?: JSXRenderable;
+  modalHeaderClasses?: string[];
```
MichelML: as-tu considerer mettre `IClassName` ici ? j'ai ajoute ca dans une PR recemment.

src/components/modal/ModalComposite.tsx
```diff
@@ -0,0 +1,106 @@
...
+  render() {
+    return (
+      <div>
+        {this.getModal([this.getModalHeader(), this.getModalBody(), this.getModalFooter()])}
+        {this.getModalBackdrop()}
+      </div>
+    );
+  }
+
+  private getModal(children: JSXRenderable) {
```
MichelML: je pense que tu pourrais faire:

```
 private getModal(...children: JSXRenderable[]) {
```  
et dans getModal tu pourrais juste passer les children sans array. c'est minor ceci dit

```
{this.getModal(this.getModalHeader(), this.getModalBody(), this.getModalFooter())}
```

src/components/modal/ModalComposite.tsx
sroy3: I could probably change it to use it. For now the components underneath do not use classnames. They are built manually.

src/components/modal/ModalCompositeConnected.tsx
```diff
@@ -0,0 +1,13 @@
+import { connect } from 'react-redux';
+import { ReduxUtils } from '../../utils/ReduxUtils';
+import { IModalCompositeDispatchProps, IModalCompositeProps, IModalCompositeStateProps, ModalComposite } from './ModalComposite';
+
+
```
MichelML: extra space

src/components/modal/ModalComposite.tsx
```diff
@@ -0,0 +1,106 @@
...
+      >
+        {children}
+      </Modal>;
+  }
+
+  private getModalHeader() {
+    const basicProps: IModalHeaderProps = {
+      id: this.props.id,
+      title: this.props.title,
+      classes: this.props.modalHeaderClasses,
```
MichelML: would it be possible to add a tooltip icon besides the title as of now? If not, could you add it?

sroy3: I think it is something you send to the ModalHeader component by using its children, but I'll look into it to make sure.

MichelML: Ok cool thanks, I think even if it's at the children level, I wonder if we could not have a prop `documentation link` and some tooltip props which would generate the doc icon  the proprer way...otherwise this is something we'll have to do over and over manually i think

MichelML: if you don't need it for your use case though fine by me we can do it when the time comes

sroy3: That is exactly the case actually. Just like we can see in the simple ModalExamples. I have also added an example in the ModalCompositeExamples just to make sure.

PR #471, opened 2018-02-07
add datePicker as an option to the table component

ThibodeauJF:
Demo 
https://coveo.github.io/react-vapor/feature-table-datepicker/

Screenshot
screen shot 2018-02-06 at 3 42 51 pm

You can now easily plug the datePicker to the ReactVapor's table. See examples for implementation. 😁 

Edit: On date range for the datepicker is supported. Single date selection would be a possible future feature.


src/components/tables/TableConnected.tsx
```diff
@@ -28,6 +29,9 @@ const mapStateToProps = (state: IReactVaporState, ownProps: ITableOwnProps): ITa
   const perPageState: IPerPageState = tableState && _.findWhere(state.perPageComposite, { id: tableState.perPageId });
   const tableHeaderCellState: ITableHeaderCellState = tableState && state.tableHeaderCells[tableState.tableHeaderCellId];
   const predicateStates: IDropdownSearchState[] = tableState && _.reject(state.dropdownSearch, (dropdownSearch: IDropdownSearchState) => !contains(dropdownSearch.id, ownProps.id)) || [];
+  const datePickerState: IDatePickerState = tableState && _.find(state.datePickers, (datePicker: IDatePickerState) => {
```
gdostie: Unless I'm mistaken, I think you could use `const datePickerState: IDatePickerState = tableState && _.findWhere(state.datePickers, { id: tableState.datePickerId });` here.

src/components/tables/TableDataModifier.ts
```diff
@@ -79,6 +80,24 @@ export const applyFilterOnDisplayedIds = (
...
+  nextDisplayedIds: string[],
+  tableDataById: ITableRowData,
+  tableCompositeState: ITableCompositeState,
+  tableOwnProps: ITableOwnProps,
+): string[] => {
+  const { from, to } = tableCompositeState;
+  const { datePicker } = tableOwnProps;
+  if (from && to && datePicker && datePicker.attributeValue) {
+    nextDisplayedIds = nextDisplayedIds.filter((dataId: string): boolean => {
+      const rowDate = moment(tableDataById[dataId][datePicker.attributeValue]).toDate();
```
gdostie: I think you can use `return moment(tableDataById[dataId][datePicker.attributeValue]).isBetween(from, to);`

src/components/tables/table-children/TableChildBlankSlate.tsx
```diff
@@ -28,7 +28,8 @@ export const TableChildBlankSlate = (props: ITableChildBlankSlateProps): JSX.Ele
   if (tableCompositeState.isInError) {
     blankSlatePropsToUse = blankSlateOnError || blankSlateDefault;
   } else if (tableCompositeState.filter
-    || _.some(tableCompositeState.predicates, (value: any) => !_.isUndefined(value) && value !== TABLE_PREDICATE_DEFAULT_VALUE)) {
+    || _.some(tableCompositeState.predicates, (value: any) => !_.isUndefined(value) && value !== TABLE_PREDICATE_DEFAULT_VALUE)
+    || tableCompositeState.from || tableCompositeState.to) {
```
gdostie: I would do a carriage return before the second `||`

src/components/tables/TableDataModifier.ts
ThibodeauJF: I didn't know that, thanks!

src/components/tables/TableConnected.tsx
ThibodeauJF: It doesn't work because children of tableChildComponents have a different id example `react-vapor-table124datepicker` for the table `react-vapor-table124`. I could parse out `datepicker` before `findWhere` and it would work! Cleaner also

gdostie: Oh okay, I wasn't aware of that. Would the following work then?
``_.findWhere(state.datePickers, { id: `${tableState.datePickerId}datepicker` })``
IMHO its cleaner and simpler to add to an existing string, than to parse it.

src/components/tables/Table.tsx
```diff
@@ -51,6 +52,11 @@ export interface ITablePredicate {
   attributeNameFormatter?: IAttributeNameOrValueFormatter;
 }
 
+export interface ITableDatePicker extends IDatePickerDropdownProps {
+  // If the table is not manual, tells which attribute should be used to filter dates
+  attributeValue?: string;
```
GermainBergeron: I'm wondering if this attribute should be in `IDatePickerDropdownProps` ?

src/components/tables/TableConnected.tsx
GermainBergeron: Can you use `getTableChildComponentId(tableState.datePickerId, TableChildComponent.DATEPICKER)` to get the component id?

sroy3: You should remove the curly braces and the return also.

ThibodeauJF: What works is this: `const datePickerState: IDatePickerState = tableState && _.findWhere(state.datePickers, { id: `${tableState.datePickerId}-date-range` });`

I think there is a reason why indexOf is used as `-date-range` might not always be there, or it might be something else

Same method is also here : https://github.com/coveo/react-vapor/blob/master/src/components/datePicker/DatePickerDropdownConnected.tsx#L24

src/components/tables/Table.tsx
ThibodeauJF: Good point!

src/components/tables/table-children/TableChildBlankSlate.tsx
ThibodeauJF: Should I put `|| (tableCompositeState.from && tableCompositeState.to))`? since it's an both or none kind of situation

gdostie: I'd say we should leave it with the OR, because maybe in the future the datepicker could apply only one the limits.

src/components/tables/Table.tsx
MichelML: should be `attributeName` and not `attributeValue` right?

MichelML:
@ThibodeauJF an example with enough rows to have some pagination and perpage would be nice to make sure it behaves well with multiple pages too

src/components/tables/tests/TableChildActionBar.spec.tsx
```diff
@@ -78,6 +86,21 @@ describe('<TableChildActionBar />', () => {
...
+        expect(() => {
+          mountComponentWithProps({
+            ...tablePropsMock,
+            actionBar: true,
+            filter: true,
+            predicates: [
+              { attributeName: 'email', attributeNameFormatter: _.identity, props: {} },
+              { attributeName: 'userName', attributeNameFormatter: _.identity, props: {} },
+            ],
+            datePicker: { datesSelectionBoxes: SELECTION_BOXES, attributeValue: 'date' },
```
MichelML: I like that...could you had this complex example in your example so we can try it?

src/components/tables/TableConnected.tsx
MichelML: @ThibodeauJF are you saying that the datepicker you retrieve doesn't always have the same id? if it doesn't that I understand your strategy, otherwise if you always retrieve ``${tableState.datePickerId}-date-range}`` i would consider making a special case in `getTableChildComponentId` for it where the id of TableChildComponent.DATEPICKER will be `${tableId}${TableChildComponent.DATEPICKER}-date-range` . If it's not always that I would add a comment explaining the situation that the id is not always the same (which is a bit weird actually).

MichelML: in otherwise if you can, you should make it so that you can user `_.findWhere(state.datePickers, { id: tableState.datePickerId });` since the only use case of `tableState.datePickerId` seems to be here

src/components/tables/TableDataModifier.ts
MichelML: TIL

src/components/tables/TableConnected.tsx
ThibodeauJF: If you want to add a datePicker with no date range (single date), the Id will look like this: `datepicker-react-vapor-table5-select-date`. Right now both are possible with the feature (props `isLinkedToDateRange` and `datesSelectionBoxes` with single `quickOptions` and `isRange` to false)

ThibodeauJF: I would add a comment that explains this

ThibodeauJF:
Fixes applied 👍

MichelML:
Very nice @ThibodeauJF , you just have one UT failing still. Just DM me when fixed. The build logs are also weird though I am trying to figure out why.

src/components/tables/TableDataModifier.ts
```diff
@@ -79,6 +80,23 @@ export const applyFilterOnDisplayedIds = (
...
+  tableDataById: ITableRowData,
+  tableCompositeState: ITableCompositeState,
+  tableOwnProps: ITableOwnProps,
+): string[] => {
+  const { from, to } = tableCompositeState;
+  const { datePicker } = tableOwnProps;
+  if (from && to && datePicker && datePicker.attributeName) {
+    nextDisplayedIds = nextDisplayedIds.filter((dataId: string): boolean => {
+      return moment(tableDataById[dataId][datePicker.attributeName]).isBetween(from, to);
+    });
```
sroy3:
```
nextDisplayedIds = nextDisplayedIds.filter((dataId: string): boolean =>
    moment(tableDataById[dataId][datePicker.attributeName]).isBetween(from, to));
```

PR #474, opened 2018-02-12
Extract and export a TableCompositeState "builder"

ThibodeauJF:
I needed to be able to build the Table composite state from outside of ReactVapor to use the manual version of the Table in more complex ways (fetch from outside for example). We decided to extract the function from `mapStateToProps`.  

I changed the API endpoint for the TableExamples because it was broken.

**POSSIBLE BREAKING CHANGE** the `dispatchPreTableStateModification ` and `dispatchPostTableStateModification` are called with the table ID directly instead of the whole `ownProps`

### DIFF
--- a/src/components/tables/TableConnected.tsx
+++ b/src/components/tables/TableConnected.tsx
@@ -22,40 +22,48 @@ import { ITableHeaderCellState } from './TableHeaderCellReducers';
 import { ITableCompositeState, ITableState } from './TableReducers';
 import { getTableChildComponentId } from './TableUtils';
 
-const mapStateToProps = (state: IReactVaporState, ownProps: ITableOwnProps): ITableCompositeStateProps => {
-  const tableState: ITableState = state.tables[ownProps.id] || {} as ITableState;
-  const filterState: IFilterState = tableState && _.findWhere(state.filters, { id: tableState.filterId });
-  const paginationState: IPaginationState = tableState && _.findWhere(state.paginationComposite, { id: tableState.paginationId });
-  const perPageState: IPerPageState = tableState && _.findWhere(state.perPageComposite, { id: tableState.perPageId });
+export interface IReactVaporStateExtended extends IReactVaporState {
+  [additionalState: string]: any;
+}
+
+export const getTableCompositeState = (state: IReactVaporStateExtended, id: string): ITableCompositeState => {
+  const tableState: ITableState = state.tables[id] || {} as ITableState;
+  const filterState: IFilterState = tableState && _.findWhere(state.filters, {id: tableState.filterId});
+  const paginationState: IPaginationState = tableState && _.findWhere(state.paginationComposite, {id: tableState.paginationId});
+  const perPageState: IPerPageState = tableState && _.findWhere(state.perPageComposite, {id: tableState.perPageId});
   const tableHeaderCellState: ITableHeaderCellState = tableState && state.tableHeaderCells[tableState.tableHeaderCellId];
-  const predicateStates: IDropdownSearchState[] = tableState && _.reject(state.dropdownSearch, (dropdownSearch: IDropdownSearchState) => !contains(dropdownSearch.id, ownProps.id)) || [];
-  const datePickerState: IDatePickerState = tableState && _.findWhere(state.datePickers, { id: tableState.datePickerRangeId });
+  const predicateStates: IDropdownSearchState[] = tableState && _.reject(state.dropdownSearch, (dropdownSearch: IDropdownSearchState) => !contains(dropdownSearch.id, id)) || [];
+  const datePickerState: IDatePickerState = tableState && _.findWhere(state.datePickers, {id: tableState.datePickerRangeId});
 
   return {
-    tableCompositeState: {
-      id: tableState.id,
-      data: tableState.data,
-      isInError: tableState.isInError,
-      isLoading: tableState.isLoading,
-      filter: filterState && filterState.filterText,
-      page: paginationState && paginationState.pageNb,
-      perPage: perPageState && perPageState.perPage,
-      sortState: {
-        attribute: tableHeaderCellState && tableHeaderCellState.attributeToSort,
-        order: tableHeaderCellState && tableHeaderCellState.sorted,
-      },
-      predicates: predicateStates.reduce((currentPredicates, nextPredicate: IDropdownSearchState) => {
-        // the attribute name is stored in the id of the dropdownSearch
-        const attributeName = nextPredicate.id.split(getTableChildComponentId(ownProps.id, TableChildComponent.PREDICATE))[1];
-        const selectedOption = _.findWhere(nextPredicate.options, { selected: true });
-        return {
-          ...currentPredicates,
-          [attributeName]: selectedOption && selectedOption.value || TABLE_PREDICATE_DEFAULT_VALUE,
-        };
-      }, {}),
-      from: datePickerState && datePickerState.appliedLowerLimit,
-      to: datePickerState && datePickerState.appliedUpperLimit,
-    } as ITableCompositeState,
+    id: tableState.id,
+    data: tableState.data,
+    isInError: tableState.isInError,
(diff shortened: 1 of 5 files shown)


src/components/tables/TableConnected.tsx
```diff
@@ -22,40 +22,48 @@ import { ITableHeaderCellState } from './TableHeaderCellReducers';
 import { ITableCompositeState, ITableState } from './TableReducers';
 import { getTableChildComponentId } from './TableUtils';
 
-const mapStateToProps = (state: IReactVaporState, ownProps: ITableOwnProps): ITableCompositeStateProps => {
-  const tableState: ITableState = state.tables[ownProps.id] || {} as ITableState;
-  const filterState: IFilterState = tableState && _.findWhere(state.filters, { id: tableState.filterId });
-  const paginationState: IPaginationState = tableState && _.findWhere(state.paginationComposite, { id: tableState.paginationId });
-  const perPageState: IPerPageState = tableState && _.findWhere(state.perPageComposite, { id: tableState.perPageId });
+export interface IReactVaporStateExtended extends IReactVaporState {
+  [additionalState: string]: any;
```
ThibodeauJF: open to a better way to do this...

src/components/tables/examples/TableExamples.tsx
```diff
@@ -94,8 +94,8 @@ const buildNewTableStateManually = (data: any, currentState: ITableState, tableC
 const manualModeThunk = (tableOwnProps: ITableOwnProps, shouldResetPage: boolean, tableCompositeState: ITableCompositeState): IThunkAction => {
   return (dispatch: IDispatch, getState: () => { [globalStateProp: string]: any; tables: ITablesState; }) => {
     const currentTableState = getState().tables[tableOwnProps.id];
-    dispatchPreTableStateModification(tableOwnProps, dispatch);
-    $.get('https://raw.githubusercontent.com/toddmotto/public-apis/master/json/entries.json')
+    dispatchPreTableStateModification(tableOwnProps.id, dispatch);
+    $.get('https://jsonplaceholder.typicode.com/comments')
```
ThibodeauJF: Result looks like this
screen shot 2018-02-12 at 2 58 27 pm

src/components/tables/TableConnected.tsx
```diff
@@ -22,40 +22,48 @@ import { ITableHeaderCellState } from './TableHeaderCellReducers';
...
+}
+
+export const getTableCompositeState = (state: IReactVaporStateExtended, id: string): ITableCompositeState => {
+  const tableState: ITableState = state.tables[id] || {} as ITableState;
+  const filterState: IFilterState = tableState && _.findWhere(state.filters, {id: tableState.filterId});
```
pochretien: I think if the state.filters is undefined, we can use the `store.getState().filters.findWhere((filter) => filters.id === tableState.id + 'filter').filterText`

src/components/tables/TableConnected.tsx
GermainBergeron: Is `additionalState` used? I wonder if you could simply use `IReactVaporState` and in other projects make the `UAState` extends `IReactVaporState`.

In this case you'll be able to call the function with any store that extends the react-vapor one and the function will use only know properties

src/components/tables/examples/TableExamples.tsx
```diff
@@ -94,8 +94,8 @@ const buildNewTableStateManually = (data: any, currentState: ITableState, tableC
 const manualModeThunk = (tableOwnProps: ITableOwnProps, shouldResetPage: boolean, tableCompositeState: ITableCompositeState): IThunkAction => {
```
GermainBergeron: Are the arguments here still defined?

src/components/tables/TableConnected.tsx
```diff
@@ -22,40 +22,48 @@ import { ITableHeaderCellState } from './TableHeaderCellReducers';
...
-const mapStateToProps = (state: IReactVaporState, ownProps: ITableOwnProps): ITableCompositeStateProps => {
-  const tableState: ITableState = state.tables[ownProps.id] || {} as ITableState;
-  const filterState: IFilterState = tableState && _.findWhere(state.filters, { id: tableState.filterId });
-  const paginationState: IPaginationState = tableState && _.findWhere(state.paginationComposite, { id: tableState.paginationId });
-  const perPageState: IPerPageState = tableState && _.findWhere(state.perPageComposite, { id: tableState.perPageId });
+export interface IReactVaporStateExtended extends IReactVaporState {
+  [additionalState: string]: any;
+}
+
+export const getTableCompositeState = (state: IReactVaporStateExtended, id: string): ITableCompositeState => {
```
pochretien: u need to export you getter in index.ts

ThibodeauJF: I didn't do it correctly?

src/components/tables/examples/TableExamples.tsx
ThibodeauJF: It's using the `defaultTableStateModifier` which uses many of the `tableOwnProps`. But the `dispatchPreTableStateModification` and `dispatchPostTableStateModification` are called with the id only. The `ownProps` are still being passed in the arguments but they are not necessary if you want to update the state.

ThibodeauJF: For `additionalState`, won't Typescript complain when I pass a state like `UAState` that will contain substates not defined in `ReactVaporState`?

src/components/tables/TableConnected.tsx
ThibodeauJF: I put my other comment in the wrong place:
Won't Typescript complain if I pass a state like `UAState` that contains a lot of substates that are not defined in `ReactVaporState`?

src/components/tables/TableConnected.tsx
GermainBergeron: Yes you did, forget this comment ;)

src/components/tables/TableConnected.tsx
GermainBergeron: If UAState extends IReactVaporState it will be ok because every properties that the method use exists in UAState

src/components/tables/examples/TableExamples.tsx
GermainBergeron: Perfect!

src/components/tables/TableConnected.tsx
ThibodeauJF: Works perfectly, Thanks!

src/components/tables/TableConnected.tsx
ThibodeauJF: I don't really understand.. It's allright if it's undefined the filter won't show up in the table if desired.

GermainBergeron: I don't think we need to change the filter here, state will always be defined

src/components/tables/TableDataModifier.ts
```diff
@@ -13,20 +13,20 @@ import { ITableCompositeState, ITableState } from './TableReducers';
 import { unselectAllRows } from './TableRowActions';
 import { getTableChildComponentId, getTableLoadingIds } from './TableUtils';
 
-export const dispatchPreTableStateModification = (tableOwnProps: ITableOwnProps, dispatch: IDispatch) => {
-  dispatch(unselectAllRows(tableOwnProps.id));
+export const dispatchPreTableStateModification = (id: string, dispatch: IDispatch) => {
```
MichelML: can you consider naming `id` for `tableId` here? Since there are a lot of components interrelated/moving together in the table it's good to remain explicit in this context imho.

src/components/tables/examples/TableExamples.tsx
MichelML: thanks for that

ThibodeauJF:
Fixed!

src/components/tables/TableConnected.tsx
pochretien: oups my bad, good job

PR #479, opened 2018-02-14
Add isPrompt prop that adds the style in the modalBackdrop

ThibodeauJF:
An unwanted style was enforced on all `ModalComposite` because of the `displayFor` prop


ThibodeauJF:
@MichelML done 👍

PR #480, opened 2018-02-14
move undefined getActions at the table level

MichelML:
to prevent this:  
screen shot 2018-02-14 at 3 11 32 pm


src/components/tables/table-children/TableChildBody.tsx
```diff
@@ -67,7 +67,7 @@ export const TableChildBody = (props: ITableChildBodyProps): JSX.Element => {
         isCollapsible={!!collapsibleData}
         onClickCallback={() => {
           if (props.onRowClick) {
-            props.onRowClick(props.getActions && props.getActions(props.rowData));
+            props.onRowClick(props.getActions(props.rowData));
```
MichelML: props.getActions is always defined at this level but not at the table level

PR #482, opened 2018-02-15
DatepickerDropdown's clear selection button

gdostie:
- Added a `Clear` button to the datepicker dropdown. It is rendered by setting `isClearable` prop to true.
- Datepicker can now be initially unselected by setting `initiallyUnselected` prop to true.
- Improved calendar and date pickers behavior in general.

Demo: https://coveo.github.io/react-vapor/datepicker-clear-selection/ (then search for "clearable")


src/components/calendar/tests/Calendar.spec.tsx
```diff
@@ -85,6 +85,7 @@ describe('Calendar', () => {
         'avril',
         'mai',
         'juin',
+        'juillet',
```
ThibodeauJF: hehe good catch! 😄

ThibodeauJF:
What do you think if we used `moment.js` for the `DEFAULT_MONTHS` (`months()`) and `DEFAULT_DAYS` (`weekdaysShort()`) in `Calendar.tsx`. We could even add the locale as a prop and moment would translate it for us. This would mean that we wouldn't have to rewrite the strings for months and days in other repos.

ThibodeauJF:
If there is a use case where a clearable DatepickerDropdown is used with a Table, I think it would be great to have an example to see them interact together! 😄

gdostie:
I agree with you @ThibodeauJF, since we use momentJs, I say we should leverage it as much as we can. It is very much less error prone and more versatile then using the plain native Date object. Good call :smile: 

Regarding the table use case, for now the table we use isn't the react-vapor one, but I don't see why we couldn't add the following props to the DatePickerDropdownConnected component in the existing table with date picker example: `isClearable` and `initiallyUnselected`.

package.json
```diff
@@ -26,7 +26,7 @@
     "build": "gulp",
     "pretest": "gulp clean:tests",
     "test": "node --max-old-space-size=2048 node_modules/karma/bin/karma start",
-    "test:watch": "node --max-old-space-size=2048 node_modules/karma/bin/karma start --singleRun false",
+    "test:watch": "node --max-old-space-size=4096 node_modules/karma/bin/karma start --no-single-run",
```
MichelML: really necessary?

gulpfile.js
```diff
@@ -37,15 +37,15 @@ gulp.task('clean:tests', false, (done) => {
 });
 
 gulp.task('clean', 'Clean all', ['clean:dist', 'clean:docs', 'clean:tests'], (done) => {
```
MichelML: in theory we should have indentation 4 spaces but tsjs doesn't seem to autofix js files for that rule :( , don't do anything for now with this

MichelML:
Your drop of coverage is mainly due to the untested `onClear` i think, if you add a UT to test this you'll be good I think

src/components/calendar/CalendarConnected.tsx
```diff
@@ -38,25 +38,28 @@ const mapStateToProps = (state: IReactVaporState, ownProps: ICalendarOwnProps):
...
           // this will cause the selected lower limit date to display in the calendar right after selection of the lower limit date
-          if (!ownProps.isLinkedToDateRange && ownProps.isLinkedToDateRange !== undefined) {
+          if (!_.isUndefined(ownProps.isLinkedToDateRange) && !ownProps.isLinkedToDateRange) {
             dispatch(changeDatePickerUpperLimit(pickerId, value));
+            dispatch(selectDate(pickerId, DateLimits.lower));
           }
         }
+      } else {
+        dispatch(resetDatePickers(pickerId));
       }
```
MichelML: not too much a fan of nested ifs, i know it's debatable but, here's my suggestion:

```
if (value && isUpperLimit) {
    dispatch(changeDatePickerUpperLimit(pickerId, moment(value).endOf('day').toDate()));
} else if (value) {
    dispatch(changeDatePickerLowerLimit(pickerId, value));

    // mirror upper limit to lower limit if not linked with a date range
    // this will cause the selected lower limit date to display in the calendar right after selection of the lower limit date
    if (!_.isUndefined(ownProps.isLinkedToDateRange) && !ownProps.isLinkedToDateRange) {
        dispatch(changeDatePickerUpperLimit(pickerId, value));
        dispatch(selectDate(pickerId, DateLimits.lower));
    }
} else {
    dispatch(resetDatePickers(pickerId));
}
```

package.json
gdostie: No, I was testing some settings because I had a maximum call stack exceeded error in some UT. I forgot to remove it. Thanks

src/components/calendar/CalendarConnected.tsx
gdostie: Personnaly, I don't have a preference so I'll apply your solution, thanks.

src/components/datePicker/DatePickerReducers.ts
```diff
@@ -36,27 +38,37 @@ export const datePickerInitialState: IDatePickerState = {
 export const datePickersInitialState: IDatePickerState[] = [];
 
 const addDatePicker = (state: IDatePickerState, action: IReduxAction<IAddDatePickerPayload>): IDatePickerState => {
+  const mayBeNull = (d: Date) => action.payload.initiallyUnselected ? null : d;
```
MichelML: you should move this function in `FalsyValuesUtils.ts` and add quick UTs for it

src/components/datePicker/DatesSelectionConnected.tsx
```diff
@@ -29,10 +29,10 @@ const mapStateToProps = (state: IReactVaporState, ownProps: IDatesSelectionOwnPr
   return {
     lowerLimit: item ? item.lowerLimit : new Date(),
     upperLimit: item ? item.upperLimit : new Date(),
-    inputLowerLimit: item && item.inputLowerLimit,
-    inputUpperLimit: item && item.inputUpperLimit,
-    quickOption: optionPicker && optionPicker.selectedValue,
-    isSelecting: item && item.selected,
+    inputLowerLimit: item ? item.inputLowerLimit : new Date(),
+    inputUpperLimit: item ? item.inputUpperLimit : new Date(),
+    quickOption: optionPicker ? optionPicker.selectedValue : '',
```
MichelML: why not undefined anymore?

src/components/datePicker/DatesSelectionConnected.tsx
```diff
@@ -29,10 +29,10 @@ const mapStateToProps = (state: IReactVaporState, ownProps: IDatesSelectionOwnPr
...
+    isSelecting: item ? item.selected : '',
```
MichelML: why not undefined anymore?

src/components/datePicker/SetToNowButton.tsx
```diff
@@ -13,7 +13,7 @@ export class SetToNowButton extends React.Component<ISetToNowProps, any> {
 
   render() {
     return (
-      <Tooltip title={this.props.tooltip || SET_TO_NOW_DEFAULT_TOOLTIP}>
+      <Tooltip title={this.props.tooltip || SET_TO_NOW_DEFAULT_TOOLTIP} placement={'top'}>
```
MichelML: `placement='top'`

src/components/datePicker/tests/DatePickerDropdownConnected.spec.tsx
```diff
@@ -116,6 +116,12 @@ describe('Date picker', () => {
       expect(onDocumentClickProp).toBeDefined();
     });
 
+    it('should get what to do on clear as a prop', () => {
+      const onClearProp = datePickerDropdown.props().onClear;
+
+      expect(onClearProp).toBeDefined();
+    });
```
MichelML: I think if you just add the line 

```
expect(() => onClearProp()).not.toThrow();  
```  

your coverage will increase

src/components/datePicker/DatesSelectionConnected.tsx
gdostie: I had an error in the terminal while running UTs. I stumbled accross this error and this is the way to fix it. (see https://github.com/facebook/react/issues/6895)

src/components/datePicker/tests/DatePickerReducers.spec.ts
```diff
@@ -181,24 +183,33 @@ describe('Date picker', () => {
...
+        _.extend({}, BASE_DATE_PICKER_STATE),
+        _.extend({}, BASE_DATE_PICKER_STATE, { id: 'other-id' }),
+      ];
+      const action: IReduxAction<IDatePickerPayload> = {
+        type: DatePickerActions.clear,
+        payload: {
+          id: 'some-date-picker',
+        },
+      };
+      const datePickersState: IDatePickerState[] = datePickersReducer(oldState, action);
```
MichelML: if you name datePickersState to `newState` i think your test will be more readable overall

src/components/datePicker/DatesSelectionConnected.tsx
gdostie: See other comment

MichelML:
`Date picker dropdown with a range limit of 3 days Redux state` j'ai pogne un state weirdo (infinite updates) sur cet exemple la ou ca l'arretait pas d'updater mais je pense que y'a des `unrealistic` rules sur ce picker la IIRC

gdostie:
@MichelML if somehow you're able to reproduce, I'll be glad to adress this.

MichelML:
@gdostie i can, i'll post this very soon

MichelML:
@gdostie not sure I can finally, this is a date picker with very special rules anyway, probably a case we'll never have. I would suggest we ignore this for now. My comments are really minor outside of that, I think you can add some tests to make sure we don't reduce our coverage %. Once done i'll happily approve

src/components/calendar/CalendarConnected.tsx
ThibodeauJF: Fans of early return?

src/components/datePicker/DatePickerReducers.ts
gdostie: I don't think this could be useful elsewhere because the criteria used to nullify the value is `action.payload.initiallyUnselected` which is the only place where the payload contains such information.

If I generalize the function by passing the criteria as a parameter, then it just falls down to using a basic ternary operation. I made this function only to avoid repeating the same ternary over and over again.

src/components/calendar/Calendar.tsx
```diff
@@ -179,7 +172,8 @@ export class Calendar extends React.Component<ICalendarProps, any> {
       isInline: true,
     };
 
-    const startingYear: number = this.props.startingYear || this.props.years.indexOf(DateUtils.currentYear.toString());
+    const startingYearIndex: number = this.props.years.indexOf(DateUtils.currentYear.toString());
+    const startingYear: number = this.props.startingYear || (startingYearIndex > 0 ? startingYearIndex : Math.floor(this.props.years.length / 2));
```
sroy3: What if the index is 0?

src/components/datePicker/DatePickerReducers.ts
```diff
@@ -103,6 +109,17 @@ const resetDates = (state: IDatePickerState, action: IReduxAction<IReduxActionsP
     });
 };
 
+const clearSelection = (state: IDatePickerState, action: IReduxAction<IReduxActionsPayload>): IDatePickerState => {
+  return state.id.indexOf(action.payload.id) !== 0 || !state.isClearable
+    ? state : _.extend({}, state, {
```
sroy3: The second branch should be on its own line.

src/components/datePicker/DatePickerReducers.ts
```diff
@@ -103,6 +109,17 @@ const resetDates = (state: IDatePickerState, action: IReduxAction<IReduxActionsP
     });
 };
 
+const clearSelection = (state: IDatePickerState, action: IReduxAction<IReduxActionsPayload>): IDatePickerState => {
+  return state.id.indexOf(action.payload.id) !== 0 || !state.isClearable
```
sroy3: What if the index is 0?

src/components/datePicker/DatesSelectionConnected.tsx
sroy3: You should add these to the default props instead

src/components/calendar/Calendar.tsx
gdostie: Good catch

src/components/datePicker/DatePickerReducers.ts
gdostie: If the index is zero then it will fall into the second branch of the ternary which is what we want.

sroy3: I'm not sure I understand. It will never apply if the item is the first one in the array? I think the order of the array might always change. Did you mean to check if `index !== -1`? If not, you might want to check for another property that will make sure you target the right items.

gdostie: `state.id` is not an array, it is a string. The logic here is that if the state id doesn't begin with the payload id, then the state is not modified.

This allows to match partial ids like so:
* `MyDatePickerDropdown`
* `MyDatePickerDropdown-calendar`

gdostie:
@sroy3 

> You should add these to the default props instead

In the DatesSelectionConnected default props or DateSelection default props? Tried both and UTs are failing if I remove the ternaries in mapStatesToProps

src/components/datePicker/DatePickerReducers.ts
sroy3: Sorry about that, I get it now. Thanks.

PR #487, opened 2018-02-21
implement visible disabled actions

MichelML:
we need this for a use case in ML, where offline models won't have access to an action, and we want to give context to the user through a tooltip on the action ("You can not open this model because it is currently inactive") . 

demo: https://coveo.github.io/react-vapor/unhidden-disabled-actions/  

search for `Action bar with actions`

### DIFF
--- a/src/components/actions/ActionBarConnected.tsx
+++ b/src/components/actions/ActionBarConnected.tsx
@@ -25,7 +25,7 @@ const mapStateToProps = (state: IReactVaporState, ownProps: IActionBarOwnProps):
 
   return {
     withReduxState: true,
-    actions: actionBar && actionBar.actions ? _.filter(actionBar.actions, (action: IActionOptions) => _.result(action, 'enabled')) : [],
+    actions: actionBar && actionBar.actions ? _.filter(actionBar.actions, (action: IActionOptions) => action.enabled || action.hideDisabled === false) : [],
     isLoading: actionBar && actionBar.isLoading,
     prompt: prompt && prompt.options ?
       <div className='prompt'>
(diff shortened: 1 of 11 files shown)


src/components/actions/ActionBarConnected.tsx
```diff
@@ -25,7 +25,7 @@ const mapStateToProps = (state: IReactVaporState, ownProps: IActionBarOwnProps):
 
   return {
     withReduxState: true,
-    actions: actionBar && actionBar.actions ? _.filter(actionBar.actions, (action: IActionOptions) => _.result(action, 'enabled')) : [],
+    actions: actionBar && actionBar.actions ? _.filter(actionBar.actions, (action: IActionOptions) => action.enabled || action.hideDisabled === false) : [],
```
ThibodeauJF: Is the ` === false` really necessary? I don't think it's really used elsewhere in the repo.

src/components/actions/tests/ActionBarConnected.spec.tsx
```diff
@@ -1,35 +1,49 @@
-import { mount, ReactWrapper } from 'enzyme';
```
ThibodeauJF: Is the removal of the bracket spacing wanted?
I got this problem too when I was lintfixing 😄

src/components/actions/ActionBarConnected.tsx
MichelML: I had a problem in the UTs when I used `!action.hideDisabled` in the ActionBarConnected.spec.tsx like the defaultProps were not applied. This solved the problem.

src/components/actions/tests/ActionBarConnected.spec.tsx
MichelML: removing bracket spacing is where we want to go, I'll put some time in the near future to see how we can solve this through tsjs.

PR #489, opened 2018-02-22
Added a callback when closing the modal

sroy3:
Mostly used to clear inputs when closing.

### DIFF
--- a/src/components/modal/Modal.tsx
+++ b/src/components/modal/Modal.tsx
@@ -28,6 +28,9 @@ export class Modal extends React.Component<IModalProps, {}> {
   }
 
   componentWillUnmount() {
+    if (this.props.closeCallback) {
+      this.props.closeCallback();
+    }
     if (this.props.onDestroy) {
       this.props.onDestroy();
     }
--- a/src/components/modal/Modal.tsx
+++ b/src/components/modal/Modal.tsx
@@ -5,6 +5,7 @@ import { IClassName } from '../../utils/ClassNameUtils';
 export interface IModalOwnProps {
   id?: string;
   classes?: IClassName;
+  closeCallback?: () => void;
 }
 
 export interface IModalStateProps {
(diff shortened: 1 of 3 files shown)


MichelML:
Could it be a `onDestroyCallback` on `Modal.tsx` instead ? I think it would look a bit more like a pattern we have elsewhere. The `Modal.tsx` class would be:

```
import * as classNames from 'classnames';
import { IClassName } from '../../utils/ClassNameUtils';

export interface IModalOwnProps {
  id?: string;
  classes?: IClassName;
  onDestroyCallback?: () => void;
}

export interface IModalStateProps {
  isOpened?: boolean;
}

export interface IModalDispatchProps {
  onDestroy?: () => void;
  onRender?: () => void;
}

export interface IModalProps extends IModalOwnProps, IModalStateProps, IModalDispatchProps { }

export class Modal extends React.Component<IModalProps, {}> {

  componentWillMount() {
    if (this.props.onRender) {
      this.props.onRender();
    }
  }

  componentWillUnmount() {
    if (this.props.onDestroy) {
      this.props.onDestroy();
    }

    if (this.props.onDestroyCallback) {
      this.props.onDestroyCallback();
    }
  }

  render() {
    const classes = classNames('modal-container', this.props.classes, {
      'opened': this.props.isOpened,
    });

    return (
      <div className={classes}>
        <div className='modal-content'>
          {this.props.children}
        </div>
      </div>
    );
  }
}  
```

sroy3:
@MichelML you can close the modal without unmounting it. Actually the modal is always there when you can toggle its state.

MichelML:
@sroy3 ok I get it :). I wonder if it would be a good use case for a thunk action since it is directly correlated to the closeModal action. Something like:

```
const closeModalWithCallback = (id: string, callback: (state?: any) => void) =>
  (dispatch: IDispatch, getState: () => any) => {dispatch(closeModal(id)); callback(getState());};
```

that you could dispatch when you want to close with side effects (callback)

sroy3:
There is also a not-connected version of the modal, so a Redux solution won't work.

sroy3:
@gdostie I first thought it'd be a problem, but since it's on `componentWillUnmount` and not `componentDidUnmount`, I think it'd make sense to add it. I don't need it for my current use case, but we might want to do other things when the modal is closing/unmounting like cancelling a call to the server. Good idea, I'll add it.

PR #490, opened 2018-02-23
Use ReactVapor's Checkbox in facets instead of input

ThibodeauJF:
Reuse one of ReactVapor's own component.
Before, clicking on the checkbox didn't toggle the change in the facets, only clicking the label worked.
Looks like this change will not be breaking behaviour outside of ReactVapor, but I'm still curious.

To test, go checkout connected facets in the demo and click on the checkbox.

DEMO https://coveo.github.io/react-vapor/update-facets-checkboxes/

### DIFF
--- a/src/components/facets/FacetRow.tsx
+++ b/src/components/facets/FacetRow.tsx
@@ -1,4 +1,6 @@
 import * as React from 'react';
+import { Checkbox } from '../checkbox/Checkbox';
+import { Label } from '../input/Label';
 import { Tooltip } from '../tooltip/Tooltip';
 import { IFacet } from './Facet';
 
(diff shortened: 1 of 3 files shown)


src/components/facets/FacetRow.tsx
```diff
@@ -20,17 +22,13 @@ export class FacetRow extends React.Component<IFacetRowProps, any> {
...
-          />
-          <button type='button'></button>
-          <span className='label'>{label}</span>
-        </label>
+        <Checkbox
+          name={this.props.facetRow.name}
+          classes={['facet-value-label']}
+          checked={this.props.isChecked}
+          onClick={() => this.props.onToggleFacet(this.props.facetRow)}>
+          <Label classes={['label']}>{label}</Label>
```
GermainBergeron: Shouldn't the classes be `['facet-value-label']`?

PR #492, opened 2018-02-26
Show the display value instead of the value

sroy3:


src/components/dropdownSearch/MultiSelectDropdownSearch/tests/SelectedOption.spec.tsx
```diff
@@ -32,10 +32,10 @@ describe('SelectedOption', () => {
 
     describe('render', () => {
       it('should render the display value in the selectedOption', () => {
-        const value: string = 'displayTest';
-        selectedOption.setProps({ value });
+        const label: string = 'displayTest';
+        selectedOption.setProps({label });
```
GermainBergeron: you have a trailling space after `label`

MichelML: it's the linter i think

sroy3: Fixed

PR #495, opened 2018-02-28
fix sort method

MichelML:
previously the sort was applied to the raw data value only. It sorted by the raw value when in fact the value was formatted in the UI to another value, so the values were not sorted correctly.  On top of that, if you added a custom sortByMethod to fix this, the sort was made on the id of the data and not on the attribute value, so it was not helping.  

Now, if you have a formatted value in the ui, you can use the sortByMethod to use the formatted value you want for a proper sort.

### DIFF
--- a/src/components/dropdownSearch/MultiSelectDropdownSearch/tests/SelectedOption.spec.tsx
+++ b/src/components/dropdownSearch/MultiSelectDropdownSearch/tests/SelectedOption.spec.tsx
@@ -9,6 +9,7 @@ describe('SelectedOption', () => {
   const props: ISelectedOptionProps = {
     key,
     value: 'test',
+    label: '',
   };
 
   describe('<SelectedOption />', () => {
--- a/src/components/tables/TableDataModifier.ts
+++ b/src/components/tables/TableDataModifier.ts
@@ -1,17 +1,17 @@
 import * as moment from 'moment';
 import * as _ from 'underscore';
-import { contains } from 'underscore.string';
-import { convertUndefinedAndNullToEmptyString } from '../../utils/FalsyValuesUtils';
-import { IDispatch } from '../../utils/ReduxUtils';
-import { addActionsToActionBar } from '../actions/ActionBarActions';
-import { changeLastUpdated } from '../lastUpdated/LastUpdatedActions';
-import { turnOffLoading, turnOnLoading } from '../loading/LoadingActions';
-import { ITableHeadingAttribute, ITableOwnProps, ITableRowData } from './Table';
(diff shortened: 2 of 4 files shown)


src/components/dropdownSearch/MultiSelectDropdownSearch/tests/SelectedOption.spec.tsx
```diff
@@ -9,6 +9,7 @@ describe('SelectedOption', () => {
   const props: ISelectedOptionProps = {
     key,
     value: 'test',
+    label: '',
```
MichelML: builds were failing because of this one ⬆️

src/components/tables/TableDataModifier.ts
```diff
@@ -103,16 +103,27 @@ export const applySortOnDisplayedIds = (
...
+  const {sortState} = tableCompositeState;
   if (sortState && sortState.order !== TableSortingOrder.UNSORTED && !_.isUndefined(sortState.attribute)) {
-    const defaultSortBy = (displayedId: string) => {
-      const cleanAttributeValue = convertUndefinedAndNullToEmptyString(tableDataById[displayedId][sortState.attribute]);
-      return cleanAttributeValue.toString().toLowerCase();
-    };
-    const headingAttributeToSort = _.findWhere(tableOwnProps.headingAttributes, { attributeName: sortState.attribute });
-    const sortByMethod = headingAttributeToSort && headingAttributeToSort.sortByMethod || defaultSortBy;
-
-    nextDisplayedIds = _.sortBy(nextDisplayedIds, sortByMethod);
```
MichelML: this is the core change below

ThibodeauJF:
Is there a plan to change all the bracket spacing in one separate PR?

I was a little confused by the name sortByMethod, it's more of a attributeSortFormatter, not a method that actually sorts. Not really an issue tough 😄 

Code looks good! 👍

MichelML:
@ThibodeauJF I agree, the intent is to mirror what you would actually pass to the underscore `sortBy` function.

src/components/tables/TableDataModifier.ts
```diff
@@ -103,16 +103,27 @@ export const applySortOnDisplayedIds = (
   tableCompositeState: ITableCompositeState,
```
gdostie: If there's a quick win to have less function params I would do it since it looks quite tricky to call the function. Not a major issue, just refactor proposition to keep in mind.

MichelML: yeah this has been made in the context of a very big PR, there is surely room for improvement in these methods

PR #498, opened 2018-03-02
upgrade tsjs

MichelML:
much much better with https://github.com/coveo/react-vapor/pull/498/files?w=1

it's actually almost only white-spaces changes


MichelML:
well the two commands in tsjs are clashing, will have to review this

PR #501, opened 2018-03-02
Upgraded to webpack 4

sroy3:


package.json
```diff
@@ -20,10 +20,12 @@
     "url": "git+https://github.com/coveo/react-vapor.git"
   },
   "scripts": {
-    "start": "gulp docs && webpack-dev-server --devtool eval --inline --hot --progress --colors",
-    "docs": "gulp docs",
+    "start": "npm run docs && webpack-dev-server --mode development --config webpack.config.js --devtool eval --inline --hot --progress --colors",
+    "docs": "webpack --config webpack.config.js",
```
MichelML: `webpack` implique qu'on aille webpack 4 installer globalement sur notre ordi, je propose d'essayer d'utiliser `npx webpack --config webpack.config.js` https://www.npmjs.com/package/npx

MichelML: ou `node node_modules/pathToWebpackCommand`

MichelML:
tout semble beau dans le demo https://coveo.github.io/react-vapor/upgrade-webpack/

package.json
sroy3: I'll check, but I did not upgrade my local version of webpack for it to be updated, simply upgrade the version in package.json. What you're saying makes sense, but I'm wondering why it works anyway.

MichelML: You are right! TIL https://docs.npmjs.com/cli/run-script#description --> "Any binaries provided by locally-installed dependencies can be used without the node_modules/.bin prefix." This is very good to know :)

webpack.config.prod.minify.js
```diff
@@ -0,0 +1,7 @@
+const prodConfig = require('./webpack.config.prod');
+
+prodConfig.output.filename = 'react-vapor.min.js';
+prodConfig.optimization = {};
+prodConfig.optimization.minimize = true;
```
GermainBergeron:
```
prodConfig.output.filename = 'react-vapor.min.js';
prodConfig.optimization = {minimize: true};
```

PR #507, opened 2018-03-05
Add count to facets + other props (tooltipName, maxTooltipNameLength, maxRowsToShow)

ThibodeauJF:
Demo: https://coveo.github.io/react-vapor/add-count-to-facets/

**Adds the follow props to the facet**
```
maxRowsToShow?: number; // Number of rows until there is no  "more" button
maxTooltipNameLength?: number; // Number of characters until there is no tooltip
```

**And for the facetRow**
```
tooltipName?: string; // What is shown in the tooltip, optional and by default its the formattedName
count?: string; // a string because it let's the formatting part outside the lib, shows a label at the right of the row
```


ThibodeauJF:
Added a little fix

src/components/facets/FacetConnected.tsx
```diff
@@ -1,18 +1,19 @@
...
+import {IReactVaporState, IReduxActionsPayload} from '../../ReactVapor';
+import {IReduxAction, ReduxUtils} from '../../utils/ReduxUtils';
+import {Facet, IFacet, IFacetDispatchProps, IFacetOwnProps, IFacetProps, IFacetStateProps} from './Facet';
+import {addFacet, changeFacet, emptyFacet, removeFacet} from './FacetActions';
+import {IFacetState} from './FacetReducers';
 
 const mapStateToProps = (state: IReactVaporState, ownProps: IFacetOwnProps): IFacetStateProps => {
-  const item: IFacetState = _.findWhere(state.facets, { facet: ownProps.facet.name });
+  const item: IFacetState = _.findWhere(state.facets, {facet: ownProps.facet.name});
+  const selectedFacetRows = item ? ownProps.facetRows.filter((row) => _.findWhere(item.selected, {name: row.name})) : [];
```
ThibodeauJF: Gets the `selectedFacetRows` from the props using the name as an index. This way, if an info (for now the `count`) is modified from the `facetRows` prop, changes will be passed to the `selectedFacetRows`, before it was not the case.

src/components/facets/FacetRow.tsx
```diff
@@ -28,6 +36,7 @@ export class FacetRow extends React.Component<IFacetRowProps, any> {
           checked={this.props.isChecked}
           onClick={() => this.props.onToggleFacet(this.props.facetRow)}>
           <Label classes={['label']}>{label}</Label>
+          {!_.isUndefined(this.props.facetRow.count) ? <span className='facet-value-count'>{this.props.facetRow.count}</span> : null}
```
gdostie: This is more of a personal preference, but I would move the ternary logic into a `const` above the return. That's just what I've seen elsewhere in the repo.

src/components/facets/Facet.tsx
```diff
@@ -1,17 +1,19 @@
...
+import {FacetMoreRows} from './FacetMoreRows';
+import {FacetMoreRowsConnected} from './FacetMoreRowsConnected';
+import {FacetMoreToggle} from './FacetMoreToggle';
+import {FacetMoreToggleConnected} from './FacetMoreToggleConnected';
+import {FacetRow} from './FacetRow';
 
 export interface IFacet {
   name: string;
   formattedName: string;
+  tooltipName?: string;
```
gdostie: IMHO `tooltipName` should be renamed to `tooltipLabel` or something similar, because _"name"_ could confuse someone into thinking this refers to the `name` attribute of the html tag used for the tooltip.

ThibodeauJF:
Fixes applie

PR #514, opened 2018-03-07
Make it possible to send the default options after a fetch

sroy3:


src/components/dropdownSearch/DropdownSearch.tsx
```diff
@@ -91,7 +91,8 @@ export class DropdownSearch extends React.Component<IDropdownSearchProps, {}> {
   }
 
   protected getDisplayedOptions(): IDropdownOption[] {
-    return _.reject(this.props.options, (option) => {
+    const optionsToUses = this.props.options.length ? this.props.options : this.props.defaultOptions;
```
GermainBergeron: Is there already a test for both cases?

sroy3:
I will rework this and make another PR.

PR #518, opened 2018-03-08
Update options

sroy3:


MichelML:
@sroy3 can you explain briefly why you need this?

sroy3:
@MichelML my options aren't fetched yet on the first render. I think it's much cleaner to send my options than updating the options outside of react vapor. It's a little sad that this component does not react to new most new props being sent to it.

PR #535, opened 2018-03-13
add the props to have the small actions in the ActionBar component

pochretien:
image

much better without spaces for review https://github.com/coveo/react-vapor/pull/535/files?w=1

### DIFF
--- a/src/components/actions/ActionBar.tsx
+++ b/src/components/actions/ActionBar.tsx
@@ -24,7 +24,7 @@ export interface IActionBarOwnProps extends React.ClassAttributes<ActionBar> {
   onClearItemFilter?: () => void;
   extraContainerClasses?: string[];
   removeDefaultContainerClasses?: boolean;
-  withsSmallActions?: boolean;
+  withSmallActions?: boolean;
 }
 
 export interface IActionBarStateProps extends IReduxStatePossibleProps {
@@ -54,7 +54,7 @@ export interface IActionBarProps extends
 export class ActionBar extends React.Component<IActionBarProps, any> {
   static defaultProps: Partial<IActionBarOwnProps> = {
     extraContainerClasses: [],
-    withsSmallActions: false,
+    withSmallActions: false,
   };
 
   private handleClear() {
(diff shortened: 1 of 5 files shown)


src/components/actions/ActionBar.tsx
```diff
@@ -24,6 +24,7 @@ export interface IActionBarOwnProps extends React.ClassAttributes<ActionBar> {
   onClearItemFilter?: () => void;
   extraContainerClasses?: string[];
   removeDefaultContainerClasses?: boolean;
+  withsSmallActions?: boolean;
```
MichelML: withsSmallActions --> withSmallActions

src/components/actions/examples/ActionBarExamples.tsx
```diff
@@ -64,6 +64,10 @@ export class ActionBarExamples extends React.Component<any, any> {
           <label className='form-control-label'>Action bar with default classes turned off and extra classes</label>
           <ActionBar actions={actions} removeDefaultContainerClasses extraContainerClasses={['coveo-table-actions-container', 'p2']} />
         </div>
+        <div className='form-group'>
+          <label className='form-control-label'>Action bar with small actions</label>
+          <ActionBar actions={actions} withsSmallActions={true} />
```
MichelML: i don't know if it is in the style guide, but we usually omit the ={true} for boolean prop, you can only put `withSmallActions` and it does the same

src/components/actions/ActionBar.tsx
```diff
@@ -53,6 +54,7 @@ export interface IActionBarProps extends
 export class ActionBar extends React.Component<IActionBarProps, any> {
   static defaultProps: Partial<IActionBarOwnProps> = {
     extraContainerClasses: [],
+    withsSmallActions: false,
```
sroy3: You don't have to define it. `true` or `undefined` is good enough.

pochretien: I think its better to handle a boolean always defined instant of an undefined value. And more, we can see faster the default state for an actionBar rendered

PR #537, opened 2018-03-15
Json editor

sroy3:
Demo: https://coveo.github.io/react-vapor/json-editor/
Scroll to the bottom to look at the 3 examples


docs/Index.tsx
```diff
@@ -1,78 +1,82 @@
...
-import { TableRowConnectedExamples } from '../src/components/tables/examples/TableRowConnectedExamples';
-import { TableRowExamples } from '../src/components/tables/examples/TableRowExamples';
-import { TitleExamples } from '../src/components/title/examples/TitleExamples';
-import { ToastConnectedExamples } from '../src/components/toast/examples/ToastConnectedExamples';
-import { ToastExamples } from '../src/components/toast/examples/ToastExamples';
-import { TooltipExamples } from '../src/components/tooltip/examples/TooltipExamples';
-import { UserFeedbackExample } from '../src/components/userFeedback/examples/UserFeedbackExample';
-import { MembersExample } from './members-example/MembersExample';
-import { ReactVaporStore } from './ReactVaporStore';
+import {render as ReactDOMRender} from 'react-dom';
```
gdostie: Did we change the lintfix for this? IIRC everywhere else in react-vapor there is a space before and after the import brackets.

sroy3: pretty-typescript used to do this for us. lintfix does not and if it did, it would remove the spaces as well. To get it back to what it used to be, I would have to do it by hand. I do not mind doing it, but I thought that since this is what's in our code style, I would leave it as is so it would slowly become standard in the repo.

ThibodeauJF:
Nice! Would it be possible to add a clearHistory props? I had the case where the CodeEditor's was empty in the beginning and the content was fetched later, so when the user did a Ctrl+Z in the editor he was shown the original empty JSON which was not desirable. The clearHistory method already exist in CodeMirror https://codemirror.net/doc/manual.html.

sroy3:
@ThibodeauJF added. Thanks for the tip.

PR #538, opened 2018-03-16
replace help icon with info icon where appropriate

MichelML:
the info icon is not looking very good on the labeled values right now, i'll see what i can do in vapor for this

### DIFF
--- a/src/components/collapsibleContainer/CollapsibleContainer.tsx
+++ b/src/components/collapsibleContainer/CollapsibleContainer.tsx
@@ -57,7 +57,7 @@ export class CollapsibleContainer extends React.Component<ICollapsibleContainerP
 
   private getSvgClass(): string {
     return classNames(
-      'icon mod-lg',
+      'icon mod-lg ml1',
       {
         'fill-orange': !!this.props.informationUrl,
         'fill-medium-grey': !this.props.informationUrl,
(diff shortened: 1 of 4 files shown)


src/components/collapsibleContainer/CollapsibleContainer.tsx
```diff
@@ -71,7 +71,7 @@ export class CollapsibleContainer extends React.Component<ICollapsibleContainerP
     }
 
     const tooltipProps = { placement: 'right', ...this.props.informationTooltip };
-    const svgProps = { svgName: 'help', svgClass: this.getSvgClass() };
+    const svgProps = { svgName: this.props.informationUrl ? 'help' : 'info', svgClass: this.getSvgClass() };
```
sroy3: It was `help-no-fill-18` before, will this change anything?

MichelML: you can have a look by searching collapsiblecontainer here https://coveo.github.io/react-vapor/info-instead-of-help/ , let me know what you think  

looks fine i wonder if we should add something to white-fill the inside of the icon

gdostie: Personally, I prefer when its not white filled. IMHO transparency looks better

sroy3: I understand the difference now. The background of the icon is wanted. That was a request from the UX team, so we shouldn't even use `help-no-fill`. The background color is also the right one.

MichelML: cool thanks guys

PR #547, opened 2018-03-19
Partially bolded string

MichelML:
screen shot 2018-03-19 at 3 13 49 pm


src/components/partial-string-match/PartialStringMatch.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+export interface PartialStringMatchProps {
+    wholeString: string;
+    partialMatch?: string;
+    caseInsensitive?: boolean;
+}
+
+export const PartialStringMatch = (props: PartialStringMatchProps): JSX.Element => {
+    const wholeString = props.wholeString || '';
+    const partialMatch = props.partialMatch || '';
```
sroy3: You should use defaultProps instead of testing the props

src/components/partial-string-match/PartialStringMatch.tsx
```diff
@@ -0,0 +1,33 @@
...
+    const regExp = new RegExp(partialMatch, props.caseInsensitive && 'i');
+    const splitPlaceholder = '%%split%%';
+
+    let hasMatch = false;
+    let realMatch: string;
+    let finalString: string | string[] = wholeString.replace(regExp, (match: string) => {
```
sroy3: You could name it something else instead of using a let and casting to both string and array of string. Only the one on line 27 is used anyway,

GermainBergeron:
I have something similar in UA but it replaces all links with `a` tags, from my understanding your code only replace the first match? You might want to took a look at `NoteCardView` and `CardUtils.urlify`

src/components/partial-string-match/PartialStringMatch.spec.tsx
```diff
@@ -0,0 +1,43 @@
+import {shallow} from 'enzyme';
+import * as React from 'react';
+import {PartialStringMatch} from './PartialStringMatch';
+
+describe('PartialStringMatch', () => {
+    const testString = 'test-string';
+
+    it('should render an empty span if the wholeString is a falsy value', () => {
+        [undefined, null, ''].forEach((falsyVal) => {
+            expect(shallow(<PartialStringMatch wholeString={falsyVal} />).find('span').text())
```
GermainBergeron: Can you use `text()` on the component instead of finding the span first?

AKA `.text()` instead of `.find('span').text()`

MichelML: I did it the way you mention at some point, but I reverted telling myself I wanted to test everything is wrapped in a span as well

MichelML:
@GermainBergeron for the use case I have i only need to highlight the first match, I could add it as an option if you really want that.

src/components/partial-string-match/PartialStringMatch.tsx
MichelML: both are used but I'll refactor as suggested 👍

MichelML:
@GermainBergeron changed applied as discussed. See image:  
  
screen shot 2018-03-19 at 6 21 34 pm
  
I tried to replace it in DropdownSearch but without success thus far, I wouldn't touch this beast if you don't mind :D .

src/components/partial-string-match/PartialStringMatch.spec.tsx
```diff
@@ -0,0 +1,95 @@
...
+                .toBe('');
+        });
+    });
+
+    it('should render the wholeString unchanged if partialMatch is undefined', () => {
+        expect(
+            shallow(<PartialStringMatch wholeString={testString} />)
+                .find('span')
+                .prop('dangerouslySetInnerHTML')
+                .__html,
```
sroy3: You can probably just check `.html()` instead of `.prop('dangerouslySetInnerHTML').__html`, no?

MichelML: good point, will do before merging

MichelML: finally the `html()` function includes the `<span></span>` wrapper which we don't want here. I'll leave it like that.

PR #549, opened 2018-03-20
Feature flippable card

gdostie:
Implemented a generic `Flippable` component. 

Click on the Flippable to flip it, and click outside to unflip it.


src/components/flippable/Flippable.tsx
```diff
@@ -0,0 +1,115 @@
+import * as classNames from 'classnames';
+import * as React from 'react';
+import * as ReactDOM from 'react-dom';
+
+export interface IFlippableOwnProps {
+    id?: string;
+    front?: React.ReactNode;
+    back?: React.ReactNode;
```
sroy3: Shouldn't front and back be mandatory? Else you'll be displaying `undefined` below.

gdostie: Well, if undefined, the divs for the front and/or the back sides will simply be rendered empty, which won't cause any error. However, it doesn't make that much sense to render an empty flippable. I don't really know if there is a use case for this.

gdostie:
Demo: https://coveo.github.io/react-vapor/feature-flippable-card/

gdostie:
Not quite sure why travis is failing :thinking:

sroy3:
@gdostie you are import underscore in `FlippableReducers.spec.ts` and not using it. Build is failing on warnings.

src/components/flippable/Flippable.tsx
ThibodeauJF: It would be safer to just add default props of empty nodes.

src/components/flippable/FlippableConnected.tsx
```diff
@@ -0,0 +1,30 @@
...
+import {IFlippableState} from './FlippableReducers';
+
+const mapStateToProps = (state: IReactVaporState, ownProps: IFlippableOwnProps): IFlippableStateProps => {
+    const flippable: IFlippableState = _.findWhere(state.flippables, {id: ownProps.id});
+    return {
+        isFlipped: flippable && flippable.flipped,
+    };
+};
+
+const mapDispatchToProps = (dispatch: (action: IReduxAction<IReduxActionsPayload>) => void, ownProps: IFlippableOwnProps): IFlippableDispatchProps => ({
```
ThibodeauJF: This line is getting a little long IMO

GermainBergeron: You can use `dispatch: IDispatch` instead of `dispatch: (action: IReduxAction<IReduxActionsPayload>) => void`

src/Index.ts
```diff
@@ -253,3 +253,18 @@ export {SlideY} from './animations/SlideY';
```
GermainBergeron: This file should be deleted, the file Index.ts on the root automatically export everything

src/components/flippable/FlippableReducers.ts
```diff
@@ -0,0 +1,74 @@
...
+    flipped: state.flipped,
+});
+
+const flip = (state: IFlippableState, action: IReduxAction<IFlippablePayload>): IFlippableState => {
+    return state.id !== action.payload.id ? state : _.extend({}, state, {
+        flipped: true,
+    });
+};
+
+const unflip = (state: IFlippableState, action: IReduxAction<IFlippablePayload>): IFlippableState => {
```
GermainBergeron: Instead of having flip and unflip doing the same thing you could have only one function with a `flipped` parameter (either true or false)

src/components/flippable/exemples/FlippableExamplesCommon.ts
```diff
@@ -0,0 +1,4 @@
+export const sizedDiv = {
```
GermainBergeron: Do you really need a file for this?

src/components/flippable/FlippableReducers.ts
gdostie: I'm not against it

src/Index.ts
gdostie: Nice, it'll be my pleasure to kill that file.

src/components/flippable/exemples/FlippableExamplesCommon.ts
gdostie: No

PR #552, opened 2018-03-23
Search bar

MichelML:
demo incoming


docs/ReactVaporStore.ts
```diff
@@ -1,6 +1,20 @@
...
+
 export const ReactVaporStore: Redux.Store<IReactVaporExampleState> =
-    Redux.createStore<IReactVaporExampleState>(Reducers, (window as any).devToolsExtension && (window as any).devToolsExtension.call(), Redux.applyMiddleware(thunk));
+    Redux.createStore<IReactVaporExampleState>(
+        Reducers,
+        (window as any).devToolsExtension && (window as any).devToolsExtension.call(),
+        Redux.applyMiddleware(
+            thunk,
+            // uncomment the line below for the redux logs to show in your chrome console
+            // createLogger(),
```
GermainBergeron: I think we can use the `process.env.NODE_ENV` and only add the createLogger in `development` (like we do in UA)

src/components/input/Input.tsx
```diff
@@ -111,39 +120,52 @@ export class Input extends React.Component<IInputProps, any> {
...
                 {this.props.children}
             </div>
         );
     }
+
+    render() {
+        return this.props.rawInput
+            ? this.getRawInput()
+            : this.getDefaultInput();
+    }
```
GermainBergeron: ~~What's the difference between getting the RawInput and getting the DefaultInput if you don't have `children`, `labelTitle` and `labelProps` ?~~

It might be worth it to create a `RawInput` component, and use it in the Input and elsewhere where you need it. In your current solution you're ignoring a lot of props and it seems counter intuitive to me

docs/ReactVaporStore.ts
MichelML: I actually commented it because it crates lots of logs and it seems to make the UI slower...which can be annoying/useful depending on your context

src/components/input/Input.tsx
MichelML: I currently disagree with your proposition but maybe if you give me more details on how you see the implementation i'll change my mind.  
  
I need most of the business logic already implemented by `Input` and `InputConnected`... i just need the view layer to change a little and I think it is not an anti-pattern.  Take https://github.com/buildo/react-placeholder for example, if you pass a placeholder type it will render a certain way, but if you pass a customPlaceholder it will be rendered another way, and some props won't be considered in this context too. 
  
I'll stop here and wait for your explanation to make sure I understand what you want correctly.

docs/ReactVaporStore.ts
ThibodeauJF: As someone who always goes and disable it, commenting it by default is welcomed 👍

src/components/input/Input.tsx
ThibodeauJF: Maybe it is just the naming that is confusing. I am more used to a label + Input being called a Field, while an input for me is by default "raw". But IMO the simplest implementation is often the best and what we have here is pretty simple, maybe make the naming clearer if possible. (or just a comment saying what raw means)

MichelML: I can't really rename the Input class though as I think it is used in many places outside react-vapor now, I'll add a comment for for the rawInput prop

sroy3: You could create `DefaultInput` and `RawInput` and `Input` would simply be the boolean that renders either one or the other. This way it will be easier to know the `RawInputProps`, but you'll be able to get everything you need from `InputConnected`. It will also be useful if someone ever needs the raw input being non-connected.

MichelML: added a comment for rawInput 
screen shot 2018-03-26 at 10 01 33 am
  
but will also dig the Raw/Default input thing

ThibodeauJF: I like this

MichelML:
So, i started doing the DefaultInput RawInput thing to rearrive at the conclusion that it seems to me a too complex solution for a simple problem. Again, DefaultInput and RawInput share all the same business logic, creating two components and two business logics when only the ui of the component changes seem overly complex for the problem at hand to me.  

I however refactored `Input` so there seems to be a clearer seperation of concern between the Input with Label and Container and the raw Input. If even you need a raw Input unconnected (steph's point), it's pretty simple you just have to do `<Input raw />` and that's it.

MichelML:
Spoke to Germain about his current Autocomplete use case and we concluded it made more sense to align my component with his, which won't use `Input` . Update coming soon

MichelML:
done @GermainBergeron 👍

src/components/searchBar/SearchBarReducers.tsx
```diff
@@ -0,0 +1,47 @@
...
+                    disabled: action.payload.disabled,
+                },
+            ];
+        case SearchBarActions.remove:
+            return _.reject(state, (searchBar: ISearchBarState) => searchBar.id === action.payload.id);
+        case SearchBarActions.toggleSearching:
+        case SearchBarActions.toggleDisabled:
+        case SearchBarActions.setValue:
+            return state.map((searchBarState: ISearchBarState) => searchBarState.id === action.payload.id
+                ? {...searchBarState, ...action.payload}
```
GermainBergeron: neat trick here

PR #555, opened 2018-03-24
Color bar

MichelML:
screen shot 2018-03-23 at 8 43 45 pm


src/components/colorBar/ColorBar.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+    getAdjustedWidthPerColor(): KeyValue<number> {
+        const nonZeroWidthColors = _.omit(this.props.widthPerColor, (width: number) => !width);
+        const totalColoredWidth = _.reduce(nonZeroWidthColors, (totalWidth: number, width: number) => width + totalWidth, 0);
+
+        if (totalColoredWidth === 0) {
+            return {transparent: 100};
+        } else if (totalColoredWidth < 100) {
+            return {...nonZeroWidthColors, transparent: 100 - totalColoredWidth};
+        } else if (totalColoredWidth === 100) {
```
GermainBergeron: You can remove this `else if` since the else cover this case

PR #557, opened 2018-03-26
Split multiline

sroy3:
I also paved the way for CSS Modules in this PR. New style can now be added directly in React Vapor instead of creating a PR in Vapor first.

Demo: https://coveo.github.io/react-vapor/split-multiline/ Scroll to the bottom to see the examples. ~~There is currently some style missing, it'll work once https://github.com/coveo/vapor/pull/479 is merged.~~ Style is now up to date.


.editorconfig
```diff
@@ -1,10 +1,10 @@
 root=true
```
MichelML: thanks for that

src/components/multilineInput/SplitMultilineInput.tsx
```diff
@@ -0,0 +1,167 @@
...
+                    const deleteButton: JSX.Element = inputIndex + 1 === inputLength ? <DeleteInputAction onClick={() => this.removeLine(index)} /> : null;
+                    return (
+                        <Input
+                            classes={styles.input}
+                            value={values[labelId]}
+                            placeholder={input.placeholder}
+                            validate={input.validation ? (value: any) => input.validation(value) : undefined}
+                            key={labelId + inputIndex}
+                            onChange={(value?: string, valid?: boolean) => this.changeValue(value, valid, index, labelId)}
+                        >
```
MichelML: pretty much the same but you have the `labelTitle` and `labelProps` directly on the Input component too, you could create your label from there, but it's not major

src/components/multilineInput/SplitMultilineInput.tsx
```diff
@@ -0,0 +1,167 @@
...
+                    {inputs}
+                </li>
+            );
+        });
+    }
+
+    private getNewInput() {
+        const inputRefs: Input[] = [];
+        const inputs: JSX.Element[] = _.map(this.props.inputs, (input: ISplitInput, inputIndex: number) => {
+            const addButton: JSX.Element = inputIndex + 1 === this.props.inputs.length ? <AddInputAction onClick={() => this.addLine(inputRefs)} /> : null;
```
MichelML: you could maybe add it inline with:  
   
```
{inputIndex === _.lastIndexOf(this.props.inputs)} 
    ? <AddInputAction onClick={() => this.addLine(inputRefs)} />
    : null}   
```  
you could remove the return and brackets too

src/components/multilineInput/SplitMultilineInput.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+            inError = inError || input.validation && !input.validation(value);
+            values[input.id] = value;
+        });
+
+        if (!inError) {
+            this.setState({
+                values: [
+                    ...this.state.values,
+                    values,
```
MichelML: i would have expected a destructuring of `values` here too, maybe rename it to avoid confusion?

MichelML:
lgtm, i like the way we use it as a end user, your examples in SplitMultilineExamples.tsx look very clean

src/components/multilineInput/styles/SplitMultilineInput.css
```diff
@@ -0,0 +1,22 @@
...
+.input {
+  flex: auto;
+  margin-right: 2em;
+  margin-bottom: 1em;
+}
+
+.input:last-child {
+  margin-right: 0;
+  padding-right: 2em;
+}
```
GermainBergeron: Does it add a style tag to the header?

I wonder how the `:last-child` work

src/components/multilineInput/styles/SplitMultilineInput.css
```diff
@@ -0,0 +1,22 @@
+.inputBlock {
```
GermainBergeron: Is it possible to use sass instead of css? This way we could import variables from vapor when needed

src/components/multilineInput/styles/SplitMultilineInput.css
```diff
@@ -0,0 +1,22 @@
...
+  margin-top: 0;
+}
+
+.inputLine {
+  display: flex;
+  justify-content: space-between;
+  position: relative;
```
GermainBergeron: I'm wondering if we should use `className="flex space-between relative pb1"` instead defining the style here. I'm not against the css-modules, I'm just wondering what's the best in this case

src/components/multilineInput/styles/SplitMultilineInput.css
sroy3: We lose a lot of the CSS Modules' power when using sass instead. But it is possible to import sass variables.

GermainBergeron: Good to know!

src/components/multilineInput/styles/SplitMultilineInput.css
sroy3: https://github.com/css-modules/css-modules/blob/master/docs/pseudo-class-selectors.md

All pseudo classes and html tag work as long as they have another class you can import.

sroy3: There is a new style tag, but I'm not sure if it is always there. As in, I don't know if it only gets added when the code is being executed.

sroy3: There is also https://github.com/gajus/react-css-modules but I do not see any mention of it not always adding the style tag either.

PR #558, opened 2018-03-27
Feature autocomplete input

GermainBergeron:
demo https://coveo.github.io/react-vapor/feature-autocomplete-input/


GermainBergeron:
I've been trying to get the missing line to be covered by a test but no success... I might let it slip

src/components/autocomplete/AutocompleteConnected.tsx
```diff
@@ -0,0 +1,173 @@
...
+    const visibleLength = _.filter(itemsWithHidden, (item: IItemBoxProps) => !item.hidden && !item.disabled).length;
+    const mod = (x: number, n: number) => (x % n + n) % n; // mod is a modulo function that works with negative numbers
+    const visibleItems = _.map(itemsWithHidden, (item: IItemBoxProps): IItemBoxProps => {
+        let active = false;
+        if (!item.hidden && !item.disabled) {
+            active = mod(activeIndex, visibleLength) === index;
+            index++;
+        }
+        return {...item, active};
+    });
```
sroy3: I don't think this (and `itemsWithHidden`) should be here. It is not part of the state. It should be done in the component itself.

src/components/autocomplete/AutocompleteReducers.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    switch (action.type) {
+        case AutocompleteActions.add:
+            return {
+                id: action.payload.id,
+                open: state.open,
+                value: state.value,
+                active: state.active,
+            };
```
sroy3:
```
return {
  ...state,
  id: action.payload.id,
}
```

src/components/itemBox/ItemBox.tsx
```diff
@@ -24,7 +26,31 @@ export class ItemBox extends React.Component<IItemBoxProps, any> {
         tooltip: {
             title: '',
         },
+        highlight: '',
     };
+    private el: any;
+
+    componentDidUpdate(prevProps: IItemBoxProps, prevState: IItemBoxProps) {
+        if (prevProps.active !== this.props.active && this.props.active) {
```
sroy3: Can't this be simplified as `if (!this.props.active && this.props.active) {`?

src/components/autocomplete/AutocompleteConnected.tsx
```diff
@@ -0,0 +1,173 @@
...
+import {ListBoxConnected} from '../listBox/ListBoxConnected';
+import {addAutocomplete, removeAutocomplete, setAutocompleteActive, setAutocompleteValue, toggleAutocomplete} from './AutocompleteActions';
+import {IAutocompleteState} from './AutocompleteReducers';
+
+const defaultMatchFilter = (filterValue: string, item: IItemBoxProps) => {
+    if (filterValue === '') {
+        return true;
+    }
+
+    const regex = new RegExp(filterValue, 'gi');
```
MichelML: do you really need to match everywhere at this level or just to know it matches would do? if the latter is ok you could remove the `g`.

src/components/autocomplete/AutocompleteConnected.tsx
```diff
@@ -0,0 +1,173 @@
...
+                        onKeyUp={(e: React.KeyboardEvent<HTMLInputElement>) => this.onKeyUp(e)}
+                        onKeyDown={(e: React.KeyboardEvent<HTMLInputElement>) => this.onKeyDown(e)}
+                        onChange={(e: React.ChangeEvent<HTMLInputElement>) => this.props.onChange(e.target.value)}
+                        value={this.props.value}
+                        required
+                    />
+                    {this.props.children}
+                </div>
+                <div className={dropdownClasses} ref={(ref: HTMLDivElement) => this.menu = ref}>
+                    <ListBoxConnected id={this.props.id} classes={['relative']} items={this.props.visibleItems} highlight={this.props.value} />
```
ThibodeauJF: Is the autocomplete behaviour will always be to push the content down? Would it be worth it to add the option for it to be floating like the dropdown?

src/components/itemBox/ItemBox.tsx
ThibodeauJF: did you mean `if (!this.prevProps.active && this.props.active)`

gdostie: Isn't `!this.props.active && this.props.active` always gonna be false?

src/components/autocomplete/AutocompleteConnected.tsx
gdostie: I agree, and I think the default behaviour should be absolutly positionned.

src/components/autocomplete/AutocompleteReducers.ts
MichelML: or destructure both  
  
``` 
return {...state, ...action.payload};  
```

src/components/autocomplete/AutocompleteReducers.ts
```diff
@@ -0,0 +1,70 @@
...
+        case AutocompleteActions.toggle:
+            const open = !_.isUndefined(action.payload.open) ? action.payload.open : !state.open;
+            // if open was sent in the dispatch use the value. Otherwise toggle the property
+            return {...state, open, active: undefined};
+        case AutocompleteActions.setValue:
+            return {...state, open: true, value: action.payload.value};
```
MichelML: why does open become true on `setValue` ?

src/components/autocomplete/AutocompleteReducers.ts
```diff
@@ -0,0 +1,70 @@
...
+        case AutocompleteActions.setActive:
+            let active = state.active + action.payload.diff;
+            if (_.isUndefined(state.active)) {
+                active = action.payload.diff === 1 ? 0 : -1;
+            }
```
MichelML: very much like your idea for the `diff` logic, good idea

sroy3:
You could also use `levenshtein` from underscore string to get similar values like we used for the Lightbox component in common.

src/components/autocomplete/AutocompleteConnected.tsx
```diff
@@ -0,0 +1,173 @@
...
+export interface IAutocompleteProps extends IAutocompleteOwnProps, IAutocompleteStateProps, IAutocompleteDispatchProps {}
+
+const mapStateToProps = (state: IReactVaporState, ownProps: IAutocompleteOwnProps): IAutocompleteStateProps => {
+    const autocomplete: IAutocompleteState = _.findWhere(state.autocompletes, {id: ownProps.id});
+    const listbox = _.findWhere(state.listBoxes, {id: ownProps.id});
+    const defaultValue = listbox && listbox.selected && listbox.selected.length ? listbox.selected[0] : '';
+    const value = autocomplete && autocomplete.value || defaultValue;
+
+    const itemsWithHidden = _.map(ownProps.items, (item: IItemBoxProps): IItemBoxProps => {
+        const visible = _.isUndefined(ownProps.matchFilter)
```
MichelML: you could use !_.isFunction(ownProps.matchFilter) , but it's minor

src/components/autocomplete/tests/AutocompleteConnected.spec.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+        it('should filter the visible values with custom function', () => {
+            const spy = jasmine.createSpy('matchFilter').and.callFake(() => false);
+            mountAutocomplete([
+                {value: 'a'},
+                {value: 'b'},
+            ], {
+                    // Take the value twice and make it uppercase
+                    matchFilter: spy,
+                });
```
MichelML: weird indent here

MichelML:
kapture 2018-03-28 at 10 47 24  
  
When I select an option by pressing Enter, i think it should hide all the options just like when I click on an option. Both should have a similar behavior.   
  
kapture 2018-03-28 at 10 50 10
When there is no items, I think it should just hide the listbox, not to display "No items".

src/components/autocomplete/AutocompleteConnected.tsx
MichelML: This is handled by PartialStringMatch

src/components/itemBox/ItemBox.tsx
```diff
@@ -48,11 +74,12 @@ export class ItemBox extends React.Component<IItemBoxProps, any> {
         return (
             <Tooltip {...this.props.tooltip}>
                 <li
+                    ref={(li) => this.el = li}
                     className={this.getClasses()}
                     onClick={() => this.handleOnOptionClick()}
                     data-value={this.props.value}>
                     {this.props.prepend ? <Content {...this.props.prepend} /> : null}
-                    {this.props.displayValue || this.props.value}
+                    <PartialStringMatch wholeString={this.props.displayValue || this.props.value} partialMatch={this.props.highlight} caseInsensitive />
```
MichelML: I wonder if `caseInsensitive={true}` should not be the default in PartialStringMatch 🤔

src/components/autocomplete/AutocompleteConnected.tsx
GermainBergeron: Yeah I only need to know if there's a match, the value of the match is useless to me

src/components/autocomplete/AutocompleteReducers.ts
GermainBergeron: I want to open the dropdown when the user type in the input

GermainBergeron: I added a props an `open` in the payload to make it more obvious

src/components/autocomplete/AutocompleteConnected.tsx
GermainBergeron: I'm 100% against moving the logic, but I think it makes more sense to put it here since it re-use the pattern of the single select component

GermainBergeron:
@sroy3 I didn't add the leveinstein distance in this PR since I only wanted to put a default "filter" functionnality, it's easy to add a custom match function tho

@gdostie I added a SlideY component to make the dropdown animate on open, good call! 🥇

PR #560, opened 2018-03-27
Json editor errors

sroy3:


src/components/editor/JSONEditor.tsx
```diff
@@ -1,16 +1,30 @@
 import 'codemirror/mode/javascript/javascript';
+
```
GermainBergeron: I thought the line break was to separate the import from different modules and our relative imports

src/components/editor/JSONEditor.tsx
```diff
@@ -25,26 +39,68 @@ export class JSONEditor extends React.Component<IJSONEditorProps, {}> {
...
+            </div>
         );
     }
 
+    private getValidationDetails(): JSX.Element {
+        return this.state.isInError
+            ? (
+                <div className='input-validation-error-details'>
+                    <span className='input-validation-error-icon'>
+                        <Svg svgName='message-alert' svgClass='icon fill-white' />
```
GermainBergeron: You probably have a svg inside a span inside a span, IIRC you can use className on the Svg instead of wrapping it

gdostie: I refactored `<Svg />` recently and I confirm you can use className directly on the Svg

src/components/editor/JSONEditor.tsx
sroy3: VSCode also does that. I guess a general import like that is another category.

PR #569, opened 2018-04-03
text area connected

MichelML:
https://coveo.github.io/react-vapor/text-area-connected/


src/components/textarea/TextArea.tsx
```diff
@@ -0,0 +1,83 @@
...
+    render() {
+        return (
+            <textarea
+                {...this.props.additionalAttributes}
+                id={this.props.id}
+                disabled={this.props.disabled}
+                className={this.props.className}
+                value={this.props.value}
+                onChange={(e: React.ChangeEvent<HTMLTextAreaElement>) => this.props.onChange(e)}
+            >
```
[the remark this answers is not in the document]
MichelML: changed thanks, forgot to revert after some changes

MichelML:
@sroy3 I will check this up. I don't need it for my current use case that's why I implemented it this way.

src/ReactVapor.ts
```diff
@@ -64,6 +65,7 @@ export interface IReactVaporState {
     collapsibleContainers?: ICollapsibleContainerState[];
     searchBars: ISearchBarState[];
     flippables?: IFlippableState[];
+    textAreas: ITextAreaState[];
```
gdostie: shouldn't this be optional like the others?

src/ReactVapor.ts
```diff
@@ -64,6 +65,7 @@ export interface IReactVaporState {
     collapsibleContainers?: ICollapsibleContainerState[];
     searchBars: ISearchBarState[];
```
gdostie: same here, shouldn't this be optional like the others?

src/components/textarea/TextArea.spec.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+        it('should set textarea id when specified', () => {
+            expect(textArea.prop('id')).toBe('textarea-id');
+        });
+
+        it('should set className when specified', () => {
+            const className = 'a-class';
+            expect(textArea.hasClass(className)).toBe(false);
+
+            wrapper.setProps({className}).mount();
```
gdostie: Not sure of what I'm saying here, but I think the call to mount after setProps is unecessary (here and other places in the file)

MichelML:
@sroy3 I added `TextAreaLabel`. Looks pretty clean to me but would like to know your thoughts and am open for changes. It also feels more modular this way to me than using the props to adjust the markup of TextArea depending on whether you want the vapor style or not.

src/components/textarea/TextArea.spec.tsx
```diff
@@ -0,0 +1,187 @@
...
+            let store: Store<IReactVaporState>;
+            let textAreaProps: ITextAreaProps;
+
+            const mountComponentWithProps = (props: ITextAreaProps) => mount(
+                <Provider store={store}>
+                    <TextAreaConnected {...props} />
+                </Provider>,
+                {attachTo: document.getElementById('App')},
+            );
+
```
gdostie: Since you often use the same logic with findWhere, you could extract this in a helper function:
```typescript
const getTextAreaStateFromId = (id: string): ITextAreaState => findWhere(store.getState().textAreas, {id});
```
And use it in your tests like so:
```typescript
getTextAreaStateFromId(textAreaProps.id);
```

src/components/textarea/TextAreaReducers.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+                });
+            });
+        });
+
+        it('should return the old state with one less TextArea when the action is TextAreaActions.remove', () => {
+            const action = removeTextArea(oldState[0].id);
+            const newState = textAreasReducer(oldState, action);
+
+            expect(newState.length).toBe(oldState.length - 1);
+            expect(newState.filter((textArea) => textArea.id === oldState[0].id).length).toBe(0);
```
gdostie: Either type `textArea` or remove parenthesis

src/ReactVapor.ts
MichelML: good catch

src/components/textarea/TextArea.spec.tsx
MichelML: removed where useless

src/components/textarea/TextArea.spec.tsx
MichelML: done

ThibodeauJF:
Is there a plan to add the possibility to make the textarea "autosizable"? 
In the same manner of https://github.com/andreypopp/react-textarea-autosize or https://github.com/jackmoore/autosize

I'm sure it would be useful!

MichelML:
tried some quick wins for the autosize (https://github.com/jackmoore/autosize) but was not working when you were deleting text , added a comment and we'll do it as needs be

PR #578, opened 2018-04-10
Add step="any" when input type is number

ThibodeauJF:
One liner, prevents value to disappear when entering decimals


sroy3:
No UTs?

ThibodeauJF:
Added UTs, Sadly it isn't a one liner anymore 😢

PR #579, opened 2018-04-11
Fix tbody tag in simple table

pochretien:


src/components/tables/Table.tsx
```diff
@@ -57,17 +57,22 @@ export interface ITableOwnProps extends React.ClassAttributes<Table>, ITableBody
...
+    tableClasses?: string[];
+    tableBodyClasses?: string[];
     initialTableData?: ITableData;
     actionBar?: true | IActionBarProps;
     blankSlateNoResultsOnAction?: IBlankSlateProps;
     blankSlateOnError?: IBlankSlateProps;
     datePicker?: IDatePickerDropdownProps;
     filter?: true | IFilterBoxProps;
-    filterMethod?: (attributeValue: any, props: ITableOwnProps) => boolean;
+    filterMethod?: (attributeValue: any, props: ITableOwnProps, filterValue: string) => boolean;
```
pochretien: give the param to check with the filter value if we want to add a custom filterMethod

pochretien:
Improvement:
Filter can be use with an empty column without error
Add the param to remove the lastUpdate component
Add the param to set the fixed header classes
set the tbody only 1 time if we use the simple table without collapsible
Fix the value returned in the filter method to handle undefined value with a custom row
Remove the last column empty with a simple table without collapsible

PR #588, opened 2018-04-16
render the blankslate in the table

pochretien:
image


ThibodeauJF:
kapture 2018-04-16 at 16 53 22
We want that hover effect?

PR #594, opened 2018-04-18
close modal on escape

sroy3:


src/components/modal/ModalBackdrop.tsx
```diff
@@ -39,4 +57,12 @@ export class ModalBackdrop extends React.Component<IModalBackdropProps, {}> {
             </div>
         );
     }
+
+    private onKeyDown = (e: KeyboardEvent) => {
+        if (this.canClose && e.code === 'Escape') {
```
GermainBergeron: should be `e.keyCode === keyCode. escape`

src/components/modal/ModalReducers.ts
```diff
@@ -69,3 +69,19 @@ export const modalsReducer = (
...
+    switch (action.type) {
+        case ModalAction.openModal:
+            return [
+                ...state,
+                action.payload.id,
+            ];
+        case ModalAction.closeModal:
+            return _.without(state, action.payload.id);
+        case ModalAction.closeModals:
+            return _.without(state, ...action.payload.ids);
```
GermainBergeron: Instead of having two reducers you could modify your action to always send an array

src/components/modal/tests/ModalBackdrop.spec.tsx
```diff
@@ -68,5 +67,43 @@ describe('ModalBackdrop', () => {
...
+
+        it('should call handleClick when user hits escape and the modal is the last one opened', () => {
+            jasmine.clock().install();
+            const handleClickSpy = spyOn<any>(modalBackdrop.instance(), 'handleClick');
+
+            modalBackdrop.setProps({lastOpened: true});
+            jasmine.clock().tick(5);
+
+            const event = document.createEvent('Event');
+            (event as any).code = 'Escape';
```
GermainBergeron: event. keyCode = keyCode.escape

src/components/modal/ModalBackdrop.tsx
GermainBergeron: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code#Browser_compatibility

sroy3: I will use key instead (https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) since keyCode is deprecated and cannot even be used in the tests.

PR #601, opened 2018-04-20
Sticky action bar

sroy3:
Since it depends too much on the environment in which the table and action bar are integrated, there is no real implementation of the sticky actions in react vapor. These change are for adding the width of the bar when it is a dynamic width, such as in the log and activity browsers, and for adding the "fixed-actions" class when scrolling pass a certain threshold. Both the scroll value and the scroll threshold need to be passed down.


src/components/actions/ActionBar.tsx
```diff
@@ -27,13 +27,16 @@ export interface IActionBarOwnProps extends React.ClassAttributes<ActionBar> {
     removeDefaultContainerClasses?: boolean;
     withSmallActions?: boolean;
     prefixContent?: IContentProps;
+    width?: number;
```
gdostie: Isn't `width` supposed to be a string in order to be able to specify the css units (px, %, vw, ...) ?

sroy3: Here, it'll always be `px` since it's the width calculated. I wish I didn't have to use this, but the width is not predictable at all.

sroy3:
@MichelML Not really, because like you said our demo page isn't built for that at all. It would look kind of weird to make the actions sticky when there is something else below the tables. I could show a live demo another project maybe.

PR #609, opened 2018-04-24
* Fix modal for real

vseguin:
Using the good key help 🤦‍♂️ 

@sroy3  we still have a problem though... if we have multiple modals opened, they _all_ close when pressing X. Pressing escape works correctly though. I'm not sure how to fix this one...

EDIT: I have a tentative fix in there, feel free to review properly.

### DIFF
--- a/src/components/modal/ModalHeader.tsx
+++ b/src/components/modal/ModalHeader.tsx
@@ -31,7 +31,7 @@ export class ModalHeader extends React.Component<IModalHeaderProps, {}> {
         let closeComponent: JSX.Element = null;
         if (this.props.onClose) {
             closeComponent = (
-                <span className='small-close' onClick={() => {this.close();}}>
+                <span className='small-close' onClick={(e) => {this.close();}}>
                     <Svg svgName='close' className='icon mod-lg fill-pure-white' />
                 </span>
             );
--- a/src/components/modal/ModalHeader.tsx
+++ b/src/components/modal/ModalHeader.tsx
@@ -1,5 +1,6 @@
 import * as classNames from 'classnames';
 import * as React from 'react';
+import * as _ from 'underscore';
 import {IClassName} from '../../utils/ClassNameUtils';
 import {Svg} from '../svg/Svg';
 
@@ -9,7 +10,9 @@ export interface IModalHeaderOwnProps {
     classes?: IClassName;
 }
 
-export interface IModalHeaderStateProps {}
+export interface IModalHeaderStateProps {
+    lastOpened?: boolean;
+}
 
 export interface IModalHeaderDispatchProps {
     onClose?: () => void;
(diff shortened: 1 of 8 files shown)


src/components/modal/ModalHeader.tsx
```diff
@@ -18,9 +21,14 @@ export interface IModalHeaderDispatchProps {
 export interface IModalHeaderProps extends IModalHeaderOwnProps, IModalHeaderStateProps, IModalHeaderDispatchProps {}
 
 export class ModalHeader extends React.Component<IModalHeaderProps, {}> {
+    private canClose: boolean = true;
+
+    componentDidUpdate() {
+        this.canClose = _.isUndefined(this.props.lastOpened) ? true : this.props.lastOpened;
```
GermainBergeron: What do you think of this syntax?: `this.canClose = _.result(this.props, 'lastOpened', true);`

vseguin: I like it! Will change.

src/components/modal/ModalHeaderConnected.tsx
```diff
@@ -1,9 +1,20 @@
 import {connect} from 'react-redux';
+import * as _ from 'underscore';
+import {IReactVaporState} from '../../ReactVapor';
 import {IDispatch, ReduxUtils} from '../../utils/ReduxUtils';
 import {closeModal} from './ModalActions';
-import {IModalHeaderDispatchProps, IModalHeaderOwnProps, IModalHeaderProps, ModalHeader} from './ModalHeader';
+import {
+    IModalHeaderDispatchProps, IModalHeaderOwnProps, IModalHeaderProps, IModalHeaderStateProps,
+    ModalHeader,
```
sroy3: Optimize imports

sroy3:
Demo: https://coveo.github.io/react-vapor/fix-modal-for-real/

sroy3:
It does not seem to work on the "Modal Composite Connected with another modal inside" example. It does not work using the X and the close button of the second modal.

vseguin:
@sroy3 Was it working before? I never saw it working actually...

src/components/modal/ModalHeaderConnected.tsx
```diff
@@ -1,9 +1,20 @@
...
+    IModalHeaderDispatchProps, IModalHeaderOwnProps, IModalHeaderProps, IModalHeaderStateProps,
+    ModalHeader,
+} from './ModalHeader';
 
-const mapStateToProps = () => ({});
+const mapStateToProps = (state: IReactVaporState, ownProps: IModalHeaderOwnProps): IModalHeaderStateProps => {
+    const lastOpenedModal = _.last(state.openModals);
+
+    return {
+        lastOpened: ownProps.id === lastOpenedModal,
```
MichelML: you could just name this `canClose` if this is the only condition for it to close, or just use `this.props.lastOpened`   
  
```  
if (this.canClose && this.props.onClose) {  
```

src/components/modal/tests/ModalHeaderConnected.spec.tsx
```diff
@@ -31,6 +31,7 @@ describe('ModalHeader', () => {
                     <ModalHeaderConnected
                         id={id}
                         title={title}
+                        lastOpened={true}
```
MichelML: if this is a boolean prop in react, only putting `lastOpened` means the same as `lastOpened={true}`

sroy3:
You have 3 tests failing

PR #613, opened 2018-04-25
+ Add close timeout prop in modal

vseguin:
This will allow to specify a timeout before executing the closeCallback, ensuring a user can wait for the fade animation to be done to perform any cleanup.

### DIFF
--- a/src/components/modal/Modal.tsx
+++ b/src/components/modal/Modal.tsx
@@ -6,6 +6,7 @@ export interface IModalOwnProps {
     id?: string;
     classes?: IClassName;
     closeCallback?: () => void;
+    closeTimeout?: number;
 }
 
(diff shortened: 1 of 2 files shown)


src/components/modal/Modal.tsx
```diff
@@ -38,6 +39,14 @@ export class Modal extends React.Component<IModalProps, {}> {
 
     componentWillReceiveProps(nextProps: IModalProps) {
         if (this.props.isOpened && !nextProps.isOpened && this.props.closeCallback) {
+            this.closeModal();
+        }
+    }
+
+    private closeModal() {
```
sroy3: You could move the `if(this.props.closeCallback) {` condition here.

PR #616, opened 2018-04-26
Checkbox clicked

sroy3:


src/components/checkbox/Checkbox.tsx
```diff
@@ -40,6 +41,10 @@ export class Checkbox extends React.Component<ICheckboxProps> {
             if (this.props.handleOnClick) {
                 this.props.handleOnClick(this.props.checked);
             }
+            this.isClicked = true;
+
+            // Useful if the component is re-rendered so that the click event is not sent to the re-rendered component
+            setTimeout(() => this.isClicked = false, 1);
```
MichelML: are you sure this is a problem ? @GermainBergeron made a fix for the checkboxes recently in common, there was a listener in common listening on all `.coveo-checkbox` which was causing side effects on the react-vapor checkboxes. Could your version of common just be unupdated?

PR #620, opened 2018-05-01
Fix modals stacking

gdostie:
Nested modals were glitchy and often needed css positionning corrections to make them display properly. I suggest we get rid of that pain by simply not rendering them in a nested fashion and by specifying "manually" its layer level. This layering already exists in vapor because the same technique is used for straigth pipe back bone modals: we render them all at the same DOM level as sibblings.

Before:
```jsx
<Modal>
    First level modal
    ...
    <Modal>
        Nested modal
        ...
    </Modal>
</Modal>
```

After:
```jsx
<Modal>
    First level modal (layer = 1 by default) 
    ...
</Modal>
<Modal layer={2} >
    Nested modal.
    ...
</Modal>
```
modals_fix


gdostie:
@vseguin I've just tried it and it works well. The bug is actually more because we open a nested react-modal without specifying its layer inside a backbone modal which specifies automatically a layer.

PR #623, opened 2018-05-02
! Fix table display when collapsable rows are mixed normal rows

gdostie:
Before:
image

After:
image


gdostie:
Breaks something else, will have to rework solution.

PR #633, opened 2018-05-10
Fix Date Picker Duplicated Labels

GermainBergeron:
Added condition when the picker is a range to avoid duplicated labels
Added unit tests

date-range


src/components/datePicker/tests/DatePickerDropdown.spec.tsx
```diff
@@ -1,4 +1,5 @@
 import {mount, ReactWrapper, shallow} from 'enzyme';
+import * as moment from 'moment';
 // tslint:disable-next-line:no-unused-variable
```
sroy3: This line can be removed

PR #637, opened 2018-05-11
Add callIfDefined util

gdostie:
Since we often do the following logic, I added a small util to make it a one liner:
```ts
if (someFunctionThatMayBeUndefined) {
    someFunctionThatMayBeUndefined(someArg, someOtherArg, ...);
}
```
Becomes this
```ts
callIfDefined(someFunctionThatMayBeUndefined, someArg, someOtherArg, ...);
```

### DIFF
--- a/src/utils/FalsyValuesUtils.ts
+++ b/src/utils/FalsyValuesUtils.ts
@@ -4,3 +4,9 @@ export const convertUndefinedAndNullToEmptyString = (value: any) =>
     _.isUndefined(value) || _.isNull(value)
         ? ''
         : value;
+
+export const callIfDefined = (callback: any, ...args: any[]) => {
+    if (typeof callback === typeof Function) {
+        callback(...args);
+    }
+};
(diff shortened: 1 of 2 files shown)


sroy3:
Are you using it in react-vapor? If not, I feel like this would be better suited for the common repo.

gdostie:
@sroy3 Not yet, but since this logic happens *all the time* with optional props, I figured it would quickly get usefull to have it here instead.

src/utils/FalsyValuesUtils.ts
```diff
@@ -4,3 +4,9 @@ export const convertUndefinedAndNullToEmptyString = (value: any) =>
     _.isUndefined(value) || _.isNull(value)
         ? ''
         : value;
+
+export const callIfDefined = (callback: any, ...args: any[]) => {
+    if (typeof callback === typeof Function) {
```
MichelML: can't you just do typeof callback === 'function' ?

gdostie: yes, but I prefer to avoid string literals

PR #641, opened 2018-05-14
Cleanup modal stacking + only modal examples

sroy3:
I have commented all other examples so that it is easier to look at the demo. I think it could be a good idea to comment out what we do not need when working locally and when we push for a demo so that we aren't overwhelmed by all the other examples making the page slow. That is until we have a better structure.

 https://coveo.github.io/react-vapor/modals-cleanup/


MichelML:
what happen with the coverage @sroy3 ? seems a bit weird

src/components/modal/ModalHeader.tsx
```diff
@@ -21,15 +21,19 @@ export interface IModalHeaderDispatchProps {
...
 
     componentDidMount() {
-        this.canClose = _.result(this.props, 'lastOpened', true);
+        this.canClose = this.props.lastOpened;
     }
 
     componentDidUpdate() {
         this.canClose = false;
-        setTimeout(() => this.canClose = _.result(this.props, 'lastOpened', true), 1);
+        _.defer(() => this.canClose = this.props.lastOpened);
```
MichelML: `setTimeout(() => this.canClose = this.props.lastOpened);` would act the same here unless i'm wrong, _.defer uses setTimeout without a time specified under the hood, and if not specified, setTimeout uses `4` by default in most browsers https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout see `Timeouts throttled to >=4ms` that being said both are fine

MichelML:
demo works super well with any modal closing strategy

gdostie:
when the nested modals aren't the same size, (eg: `mod-big` on the outer one), the backdrop of the smaller nested modal doesn't overlay the bigger outer modal. 

image

That's what the defined `layer-x` classes were for.

GermainBergeron:
I'm not sure it behave as expected, the style of the modal connected is weird and the click on the backdrop closes both modals. Also when I reopen them they are both shown

sroy3:
For style issues, it's in the Vapor PR. It'll need updating once it's merged.

sroy3:
@gdostie @GermainBergeron style is fixed

PR #644, opened 2018-05-17
add interface for a data attribute

pochretien:
add data attribute for Flippable
add data attribute for LogoCard


pochretien:
Lib used: https://github.com/kimamula/ts-transformer-keys

MichelML:
I was about to vote against the use of a lib for that, but thinking about it twice it could probably be useful in a lot of places, and it's small https://bundlephobia.com/result?p=ts-transformer-keys@0.3.1  
  
I would consider leaving the id on the container div since it can be useful in many cases

MichelML:
interesting to see cases where it's relevant to play with the compiler https://github.com/kimamula/ts-transformer-keys/blob/master/transformer.ts

pochretien:
use: https://github.com/cevek/ttypescript

pochretien:
demo: https://coveo.github.io/react-vapor/add-data-attributes/

PR #653, opened 2018-05-25
Added the possibility of just chosing a date

sroy3:


sroy3:
https://github.com/coveo/vapor/pull/519 is needed for the demo to look good

sroy3:
Demo: https://coveo.github.io/react-vapor/simple-datepicker/
Search for "Simple date picker dropdown with single date, no time, and Redux state"

src/components/calendar/Calendar.tsx
```diff
@@ -33,6 +35,7 @@ export interface ICalendarOwnProps extends React.ClassAttributes<Calendar> {
     startingDay?: number;
     selectionRules?: ICalendarSelectionRule[];
     isLinkedToDateRange?: boolean;
+    simple?: boolean;
```
MichelML: can you have a default prop `simple` set to false ? will it not break every existing usage of the component otherwise ?

sroy3: simple is already set to undefined by default so everywhere it isn't set, it won't be evaluated to simple.

MichelML: indeed, using simple: false everywhere in https://github.com/coveo/react-vapor/pull/653/files#diff-d31af32b280c5ab5caa1d211f11b91fa fooled me, sorry for that

sroy3: That is in the date picker state where it is set to false by default.

PR #654, opened 2018-05-25
+ Add disabled prop to SingleSelect component

gdostie:


src/components/select/SingleSelectConnected.tsx
```diff
@@ -65,6 +67,7 @@ export class SingleSelectConnected extends React.Component<ISingleSelectProps, {
                 onMouseUp={props.onMouseUp}
                 onKeyDown={props.onKeyDown}
                 onKeyUp={props.onKeyUp}
+                disabled={this.props.disabled}
```
pochretien: {...this.props.buttonProps}

src/components/select/SingleSelectConnected.tsx
```diff
@@ -13,6 +13,7 @@ export interface ISingleSelectOwnProps extends ISelectProps {
     placeholder?: string;
     toggleClasses?: string;
     onSelectOptionCallback?: (option: string) => void;
+    disabled?: boolean;
```
pochretien: buttonProps: React.HTMLProps<HTMLButtonElement>

GermainBergeron: I would even go further and remove it. We should go the same way as #644 by using the `React.Component<ISingleSelectProps & React.HTMLProps<HTMLButtonElement>` in combinaison with `ts-transformer-keys` and the `{..._.omit(this.props, keys<ISingleSelectProps>()}`

gdostie:
demo: https://coveo.github.io/react-vapor/add-singleselect-disabled-prop/

PR #656, opened 2018-05-28
 add prop to add a disabledTooltip on options in IDropdownOption interface

MichelML:
screen shot 2018-05-27 at 10 32 43 pm


src/components/dropdownSearch/DropdownSearch.tsx
```diff
@@ -139,6 +141,11 @@ export class DropdownSearch extends React.Component<IDropdownSearchProps, {}> {
                         {this.getDropdownOptionAppend(opt)}
                     </li>
                 );
+
+                return opt.disabled && opt.disabledTooltip
+                    ? <Tooltip {...opt.disabledTooltip}>{formattedOption}</Tooltip>
```
GermainBergeron: This will wrap the `li` element in a span, I'd prefer to put the span inside the li if possible

MichelML: this doesn't seem to affect a list per se:  

screen shot 2018-05-28 at 10 04 17 am  
  
I tried your suggestion but it breaks the style :S   
  
is it a must do for you ?

PR #660, opened 2018-05-30
Remove table rows selection action when there are no actions to display

gdostie:
Made each table row selectable only when the `getActions` function returns one or more item.


gdostie:
Demo: https://coveo.github.io/react-vapor/add-table-without-select-action/

PR #668, opened 2018-06-05
Infinite scroll dropdown

sroy3:


types/react-infinite-scroll-component/index.d.ts
```diff
@@ -0,0 +1,101 @@
+/// <reference types="react" />
```
sroy3: I'll make a PR to DefinitelyTyped for this file.

sroy3:
I have move some functions around in the `DropdownSearch` file. This is why there is a red square in the coverage, but as you can see from the summary below, the coverage did not decrease.

src/components/dropdownSearch/DropdownSearch.tsx
```diff
@@ -87,21 +88,51 @@ export class DropdownSearch extends React.Component<IDropdownSearchProps, {}> {
...
+        return (
+            <div className={this.getClasses()} style={this.getStyles()}>
+                {this.getMainInput()}
+                {this.getDropdownMenu()}
+            </div>
+        );
     }
 
     protected getDropdownOptions(): JSX.Element[] {
+        const ElementTag: string = this.props.infiniteScroll ? 'div' : 'li';
```
MichelML: til

src/components/dropdownSearch/DropdownSearch.tsx
```diff
@@ -139,11 +174,11 @@ export class DropdownSearch extends React.Component<IDropdownSearchProps, {}> {
                             {valueToShow}
                         </span>
                         {this.getDropdownOptionAppend(opt)}
-                    </li>
+                    </ElementTag>
                 );
 
                 return opt.disabled && opt.disabledTooltip
-                    ? <Tooltip noSpanWrapper {...opt.disabledTooltip}>{formattedOption}</Tooltip>
+                    ? <Tooltip key={opt.value} noSpanWrapper {...opt.disabledTooltip}>{formattedOption}</Tooltip>
```
MichelML: thanks I forgot to add this

src/components/dropdownSearch/DropdownSearch.tsx
```diff
@@ -152,8 +187,16 @@ export class DropdownSearch extends React.Component<IDropdownSearchProps, {}> {
...
+    }
+
+    protected getSelectedOptions(): IDropdownOption[] {
+        return _.where(this.props.options, {selected: true});
+    }
+
     protected getDropdownOptionAppend(option?: IDropdownOption): JSX.Element {
-        return !_.isUndefined(option && option.append)
+        return option && option.append
             ? <div className='dropdown-option-append'>{option && option.append}</div>
```
MichelML: you can write `<div className='dropdown-option-append'>{option.append}</div>` since we already know it's defined at this point

src/components/dropdownSearch/DropdownSearchInfiniteScrollOptions.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+export interface DropdownSearchInfiniteScrollOptionsProps {
+    onMouseEnter: () => void;
+    options: JSX.Element[];
+    ulElementRefFunction: (menu: HTMLElement) => void;
+    infiniteScroll: InfiniteScroll.InfiniteScrollProps;
+}
+
+export class DropdownSearchInfiniteScrollOptions extends React.Component<DropdownSearchInfiniteScrollOptionsProps, {}> {
+    private id: string;
```
MichelML: why can't you have a normal id in the props instead of setting it in the props ?

src/components/dropdownSearch/DropdownSearchInfiniteScrollOptions.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+    render() {
+        return (
+            <div
+                id={this.id}
+                className='dropdown-menu'
+                ref={(menu: HTMLElement) => this.props.ulElementRefFunction(menu)}
+                onMouseEnter={() => this.props.onMouseEnter()}
+            >
+                <InfiniteScroll {...this.props.infiniteScroll} scrollableTarget={this.id} style={{...this.props.infiniteScroll.style, overflow: 'initial'}}>
```
MichelML: could you put each prop on its own line?

src/components/dropdownSearch/DropdownSearchInfiniteScrollOptions.tsx
sroy3: I only need it inside the component. It is used to specify the scrollable target. It has no other use. It is easier to define it once here than having to pass it every time we want a `DropdownSearchInfiniteScrollOptions`.

src/components/dropdownSearch/DropdownSearchInfiniteScrollOptions.tsx
sroy3: I thought it was only if the line is too long. I feel like it's easier to read this way. Just like HTML.

src/components/dropdownSearch/examples/DropdownSearchExamples.tsx
```diff
@@ -1,20 +1,53 @@
...
 import {MultiSelectDropdownSearchConnected} from '../MultiSelectDropdownSearch/MultiSelectDropdownSearchConnected';
 
-export class DropdownSearchExamples extends React.Component<any, any> {
+interface DropdownSearchExamplesProps {
+    onOptionsChanged?: (id: string, options: IDropdownOption[]) => void;
+}
+
+const mapStateToProps = () => ({});
+
+const mapDispatchToProps = (dispatch: (action: IReduxAction<IOptionsDropdownSearchPayload>) => void): DropdownSearchExamplesProps => ({
```
MichelML: you can use IDispatch here as a type

src/components/dropdownSearch/examples/DropdownSearchExamples.tsx
```diff
@@ -1,20 +1,53 @@
...
+    onOptionsChanged: (id: string, options: IDropdownOption[]) => dispatch(updateOptionsDropdownSearch(id, options)),
+});
+
+@ReduxConnect(mapStateToProps, mapDispatchToProps)
```
MichelML: if you rewrite ReduxConnect as :  
  
```  
export class ReduxUtils {
    static mergeProps(...setOfProps: Array<KeyValue<any>>) {
        return extend({}, ...setOfProps);
    }
}  
```

I think you'll just be able to do this `@ReduxConnect(mapDispatchToProps)` without defining an empty mapStateToProps ,

MichelML: i think it would be a good idea for the future components we write as well, I know here it's just an example but overall it would be a good idea

sroy3:
Demo: https://coveo.github.io/react-vapor/infinitescroll-dropdown/ you can check the last dropdown.

MichelML:
screen shot 2018-06-06 at 9 59 58 am
  
is this expected ? options do not look the same as other dropdowns

src/components/dropdownSearch/DropdownSearchInfiniteScrollOptions.tsx
MichelML: ok I thought the line was pretty long but maybe not long enough 👍

sroy3:
@MichelML I've updated coveo-styleguide, it should be good on the next build. I think there is no npm update when building the PR.

MichelML:
there is this error `Warning: Unknown prop `noSpanWrapper` on <div> tag. Remove this prop from the element` popping a bit late when you play with the dropdowns , this is probably my fault though will try to find where to omit it

src/components/dropdownSearch/examples/DropdownSearchExamples.tsx
MichelML: comment to make the comment up to date

src/components/dropdownSearch/examples/DropdownSearchExamples.tsx
```diff
@@ -1,20 +1,53 @@
...
+    onOptionsChanged?: (id: string, options: IDropdownOption[]) => void;
+}
+
+const mapStateToProps = () => ({});
+
+const mapDispatchToProps = (dispatch: IDispatch): DropdownSearchExamplesProps => ({
+    onOptionsChanged: (id: string, options: IDropdownOption[]) => dispatch(updateOptionsDropdownSearch(id, options)),
+});
+
+@ReduxConnect(mapStateToProps, mapDispatchToProps)
```
MichelML: this has been outdated https://github.com/coveo/react-vapor/pull/668#discussion_r193418478

src/components/dropdownSearch/examples/DropdownSearchExamples.tsx
sroy3: It is like that to be like the connect function which is:
```
connect<TStateProps, no_dispatch, TOwnProps>(
    mapStateToProps: MapStateToPropsParam<TStateProps, TOwnProps>,
    mapDispatchToProps: null | undefined,
    mergeProps: null | undefined,
    options: Options<TStateProps, TOwnProps>
)
```
`mapStateToProps` cannot be omitted.

src/components/dropdownSearch/examples/DropdownSearchExamples.tsx
sroy3: I've seen it anyway. I make sure to read all comments, even if they are hidden.

MichelML:
@sroy3 it could be nice to test the inifinte scroll with 5000+ options and see if the search and behavior are still smooth, what do you think? Maybe replace the 5000 options example with infinite scroll?

src/components/dropdownSearch/examples/DropdownSearchExamples.tsx
MichelML: good to know 👍

sroy3:
@MichelML I've updated the demo to use 5000 options instead. I would be nice, but useless to test more than that. I'm pretty sure no one will ever scroll down through 25 000 options without using the search first.

MichelML:
Am I wrong or currently it searches only through the loaded results ? I think we'll need to add something to be able to implement a custom search if we have an infinite scroll. In the 5000 options example, if I type 4000 in the search for example, I expect it to pop even if it's not loaded. It's not the case as of now, and it does not seem something was implemented for it. Am I wrong?

sroy3:
We will need to add an external search function. I'll add this soon.

MichelML:
everything lgtm 👍 very nice, discussed with steph and the custom search feature should appear in this PR or another pretty soon

MichelML:
fine by me for the custom search 👍

PR #669, opened 2018-06-06
Fix collapsible table behavior

gdostie:
Improved those following use cases:

* When a table had some collapsible rows mixed with some standard rows, it looked like the screenshot bellow. I added an empty `<td></td>` to the standard rows only when there is one or more collapsible table rows that exists in the table.
image

* When a table row has no actions, it is not selectable and that is the expected behavior. The problem is that when its a collapsible row, the select action was not triggered, therefore the row wouldn't uncollapse. I separated the select action into 2: selectRow and toggleRow and added a reducer that goes with it.

### DIFF
--- a/src/components/tables/TableHeadingRow.tsx
+++ b/src/components/tables/TableHeadingRow.tsx
@@ -14,6 +14,7 @@ export interface ITableHeadingRowOwnProps extends React.ClassAttributes<TableHea
     className?: string;
     isMultiSelect?: boolean;
     selectionDisabled?: boolean;
+    isPartOfCollapsibleTable?: boolean;
 }
 
 export interface ITableHeadingRowStateProps {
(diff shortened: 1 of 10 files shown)


src/components/tables/TableHeadingRow.tsx
```diff
@@ -14,6 +14,7 @@ export interface ITableHeadingRowOwnProps extends React.ClassAttributes<TableHea
     className?: string;
     isMultiSelect?: boolean;
     selectionDisabled?: boolean;
+    isPartOfCollapsibleTable?: boolean;
```
MichelML: isPartOfCollapsibleTable --> tableHasCollapsibleRows ? 🤔

gdostie: Thanks, I was clueless as to how to name this :stuck_out_tongue:

PR #679, opened 2018-06-12
add inline-doc-link class on header title linksvg

MichelML:
main use case at hand as of now is if we want to add meta analytics on doc links in the header of each section of the admin

### DIFF
--- a/src/components/svg/tests/LinkSvg.spec.tsx
+++ b/src/components/svg/tests/LinkSvg.spec.tsx
@@ -55,15 +55,6 @@ describe('<LinkSvg>', () => {
             expect(linkSvgComponent.find(`a[href="${linkSvgProps.url}"]`).length).toBe(1);
         });
 
-        it('should set the onClick on the anchor element', () => {
-            const onClick = jasmine.createSpy('onClick');
-
-            renderLinkSvg({onClick});
-            linkSvgComponent.find('a').prop('onClick')(undefined);
-
-            expect(onClick).toHaveBeenCalledTimes(1);
-        });
-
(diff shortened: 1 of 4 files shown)


sroy3:
No need for meta analytics on links, the navigation will log the meta analytics needed

MichelML:
@sroy3 i'm talking about doc links , the one that you click on to go to the docs....wouldn't that be useful?

sroy3:
Might want to check with UA people. We do not have this anywhere else. But I was wondering the same thing as well. I guess it is already logged in some way though.

MichelML:
screen shot 2018-06-12 at 2 39 19 pm  
  
@GermainBergeron also says it would make sens @sroy3

MichelML:
@pochretien comment applied

src/components/svg/tests/LinkSvg.spec.tsx
```diff
@@ -55,6 +55,15 @@ describe('<LinkSvg>', () => {
             expect(linkSvgComponent.find(`a[href="${linkSvgProps.url}"]`).length).toBe(1);
         });
 
+        it('should set the onClick on the anchor element', () => {
+            const onClick = jasmine.createSpy('onClick');
+
+            renderLinkSvg({onClick});
+            linkSvgComponent.find('a').prop('onClick')(undefined);
+
+            expect(onClick).toHaveBeenCalledTimes(1);
```
sroy3: Can you simulate a click on the element to make sure it works? Since the page will change, some code might not run. I think you need to prevent default and once onClick is completed you have to manually navigate

PR #680, opened 2018-06-12
Set default initial position at 1

sroy3:


MichelML:
why?

sroy3:
@MichelML because by default our per page numbers are 25, 50 and 100 and we want to start with 50.  By setting it to 1 here by default, we don't have to change it anywhere else, but where there are exceptions. It makes it easier to add a new one.

MichelML:
@sroy3 all good, it would be wise to advise SiteCore and Dynamics of these changes I think, namely @francoislg and @sbelzile-Coveo

sroy3:
@francoislg it is for the per page so it starts with the middle option instead of the first one. I bumped a minor so that you can wait a little before applying the change. If you want your default per page number to be (stay) the first one, you'll have to pass `initialPosition={0}` as a prop to the navigation.

francoislg:
Yeah I get what it does, I just don't understand why it should be the
middle option by default now. Was it a UX decision?

The default perpages is 10 and 100, so now the default option is 100, which
I find very weird. It only picks the middle one if you have 3 options....

The first option is usually the default one, so I don't quite understand
the fundamental reason behind this change.

sroy3:
The second option is our default. Instead of adding `initialPosition={1}` every time we are using the per page navigation we can now omit it. I agree about the default options. They are not our default options. I'll set the initial position dynamically instead. It'll be the first one if there are only 2 options and the middle one if there are more. New PR coming soon.

sroy3:
Btw, the default perPages are 10, 20 and 30. Not 10 and 100.

PR #681, opened 2018-06-13
Default initial per page is now the middle option so that it can fit better with different implementations

sroy3:


sroy3:
I have no idea why the coverage has decreased.
screen shot 2018-06-13 at 11 03 14 am

src/components/navigation/perPage/NavigationPerPage.tsx
```diff
@@ -41,31 +44,23 @@ export class NavigationPerPage extends React.Component<INavigationPerPageProps,
...
 
-        const perPageSelects: JSX.Element[] = _.map(this.perPageNumbers, (num: number, index: number): JSX.Element => {
-            const shouldShowPerPageSelect: boolean =
-                topNumber > (this.perPageNumbers[index - 1] || 0);
+        const perPageSelects: JSX.Element[] = _.map(this.props.perPageNumbers, (num: number, index: number): JSX.Element => {
+            const shouldShowPerPageSelect: boolean = topNumber > (this.props.perPageNumbers[index - 1] || 0);
 
             if (shouldShowPerPageSelect) {
-                const selectId: string = 'perpage-' + (this.props.id || '') + num;
+                const selectId: string = `perpage-${this.props.id || ''}-${num}`;
```
francoislg: The ID now has an additional `-` before `num`, intended? 

It's only used as `key` so it should not be breaking, but just to be sure :)

sroy3: It's intended. I felt it would be cleaner and as you said, it's only used as a key

PR #682, opened 2018-06-13
Fix table collapsible row expand

gdostie:
Before:

collapsible-row-before

After:

collapsible-row-after


MichelML:
you'll need to fix this before merging https://travis-ci.org/coveo/react-vapor/builds/391818819#L630

PR #684, opened 2018-06-13
Add set collapsible row opened

gdostie:
We had a redux action to toggle a table collapsible row collapsed/expanded. I needed a way to set a collapsible row opened no matter what the current state is.

### DIFF
--- a/src/components/tables/tests/TableRowReducers.spec.ts
+++ b/src/components/tables/tests/TableRowReducers.spec.ts
@@ -1,5 +1,5 @@
 import {IReduxAction} from '../../../utils/ReduxUtils';
-import {ITableRowActionPayload, setRowOpened, TableRowActions, unselectAllRows} from '../TableRowActions';
+import {ITableRowActionPayload, TableRowActions, unselectAllRows} from '../TableRowActions';
 import {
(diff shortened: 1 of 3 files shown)


gdostie:
@MichelML good idea for the toggleAction, I'll do that thanks

src/components/tables/tests/TableRowReducers.spec.ts
```diff
@@ -131,7 +131,39 @@ describe('Tables', () => {
...
+                    type: TableRowActions.toggleOpen,
+                    payload: {
+                        id: 'row1',
+                        opened: false,
+                    },
+                };
+                let collapsibleRowsState: ITableRowState[] = tableRowsReducer(oldState, setToTrue);
+
+                expect(collapsibleRowsState.length).toBe(oldState.length);
+                expect(collapsibleRowsState.filter((row) => row.id === setToTrue.payload.id)[0].opened).toBe(true);
```
acollin1: Sa serais pas mieux un findWhere? Genre _.findWhere(collapsibleRowsState, {id: setToTrue.payload.id}).

Comme ca pas besoin d'aller chercher le premiere element de la liste

Issue #621, opened 2018-05-01
Errors are not shown if it happens on initial load


francoislg:
I am wondering why no errors are shown when no initial data is provided.

With this line:

```
this.isInitialLoad = props.initialTableData == DEFAULT_TABLE_DATA;
```

(https://github.com/coveo/react-vapor/blob/master/src/components/tables/Table.tsx#L123)

coupled with this condition:

```
const tableBodyNode: React.ReactNode = tableData.displayedIds.length || this.props.tableCompositeState.isLoading || this.isInitialLoad
            ? this.getTableBody()
            : <TableChildBlankSlate {...this.props} />;
```

(https://github.com/coveo/react-vapor/blob/master/src/components/tables/Table.tsx#L173)

The state of table can be `isLoading: false` and `isInError: true`, but the `this.isInitialLoad` condition prevents the error from being shown, thus the UI is shown in "loading" mode, even though there is an error.

I wonder if this "initialLoad" is required. Can we put the `isLoading` state to true by default instead?

Or we could explicitely handle the "isInError" state in the condition:

`!this.props.tableCompositeState.isInError && (tableData.displayedIds.length || this.props.tableCompositeState.isLoading || this.isInitialLoad)`

Thoughts?

MichelML:
- the loading behavior on first load is different than the following loads, we kept this behavior for legacy purposes as the tables in the Admin UI were loading this way prior to react-vapor.  So isInitialLoad is kind of required.
  
- There is surely a problem to solve here, and I'll consider your solution for the fix.

MichelML:
addressed in https://github.com/coveo/react-vapor/pull/686

PR #686, opened 2018-06-14
Fix table initial load

MichelML:
tableinitialload

As shown in the above gif, the initial load is currently not working properly. The blankslate pops up while the initial results are retrieved. The initial loading behavior should be displayed instead. This PR fixes the issue.    
  
also adressed the following issue:  https://github.com/coveo/react-vapor/issues/621

### DIFF
--- a/src/components/tables/Table.tsx
+++ b/src/components/tables/Table.tsx
@@ -141,7 +141,7 @@ export class Table extends React.Component<ITableProps> {
     }
 
     componentDidUpdate() {
-        if (this.isInitialLoad && !_.isUndefined(this.props.tableCompositeState.data)) {
+        if (this.isInitialLoad && JSON.stringify(this.props.tableCompositeState.data) !== JSON.stringify(DEFAULT_TABLE_DATA)) {
             this.isInitialLoad = false;
         }
     }
--- a/src/components/tables/Table.tsx
+++ b/src/components/tables/Table.tsx
@@ -175,8 +175,7 @@ export class Table extends React.Component<ITableProps> {
             this.props.tableClasses,
         );
 
-        const tableData = this.props.tableCompositeState.data || this.props.initialTableData;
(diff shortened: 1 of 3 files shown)


src/components/tables/Table.tsx
```diff
@@ -141,7 +141,7 @@ export class Table extends React.Component<ITableProps> {
     }
 
     componentDidUpdate() {
-        if (this.isInitialLoad && !_.isUndefined(this.props.tableCompositeState.data)) {
+        if (this.isInitialLoad && JSON.stringify(this.props.tableCompositeState.data) !== JSON.stringify(DEFAULT_TABLE_DATA)) {
```
gdostie: `DEFAULT_TABLE_DATA` --> `this.props.initialTableData`

src/components/tables/Table.tsx
```diff
@@ -264,4 +263,11 @@ export class Table extends React.Component<ITableProps> {
                 </tbody>
             );
     }
+
+    private shouldShowTableBody(): boolean {
+        const tableData = this.props.tableCompositeState.data || this.props.initialTableData;
+
+        return !this.props.tableCompositeState.isInError
+            && !!(tableData.displayedIds.length || this.props.tableCompositeState.isLoading || this.isInitialLoad);
```
gdostie: `!_.isEmpty(tableData.displayedIds)` is safer than `tableData.displayedIds.length`,but maybe `tableData.displayedIds` never is undefined in that case thought. Very minor, your call.

src/components/tables/Table.tsx
MichelML: we really want DEFAULT_TABLE_DATA here

src/components/tables/Table.tsx
MichelML: will never be undefined 👍

MichelML: but you're right 😄

Issue #97, opened 2017-03-23
MomentJs should be external dependency


sbelzile-Coveo:
Problem when referencing .d.ts: one of the class referenced an unknown type "Moment". The issue is caused when generating .d.ts: everything that originates from node_modules is stripped out, leaving an unexisting type in the definition.

MichelML:
this PR should adress the issue https://github.com/coveo/react-vapor/pull/687

PR #687, opened 2018-06-15
Reduce bundle size, and make libs that should be external actually external

MichelML:
unminified: 2.72 MiB —> ~1.06mb
minified: 1.32 mb—> ~500kb  
  
see edited readme https://github.com/coveo/react-vapor/pull/687/files#diff-04c6e90faac2675aa89e2176d2eec7d8R19 and whole PR for the reasoning behind the changes


sroy3:
Have you tried cherry picking dependencies or not including any of them? Are there any warnings or errors? I think that your change is justify and would sure help, but I just want to make sure it does work better than before.

MichelML:
@sroy3 if you omit certain dependencies it won't work yes, there are errors. You can try it if you want.  
  
However, I would bump a major for this since it implies that react-vapor's users will now have to load these dependencies if not already the case.

MichelML:
this PR also adresses this issue "par la bande" https://github.com/coveo/react-vapor/issues/97

README.md
```diff
@@ -17,11 +17,34 @@ npm install react-vapor
 > The project is at its early stages, some components can still have _lots_ of breaking changes between versions.
 
 ### Dependencies
-Those are required in order to use react-vapor in your own project:
-- [Tether](http://tether.io/)  (tested with 1.3.7)
-- [React](https://facebook.github.io/react/) (tested with 15.3.1)
-- [React DOM](https://facebook.github.io/react/) (tested with 15.3.1)
-- [Underscore](http://underscorejs.org/)  (tested with 1.8.3)
+React-Vapor is bundled with the popular bundler tool [Webpack](https://webpack.js.org/). However, we specify many dependencies as [externals](https://webpack.js.org/configuration/externals/#src/components/Sidebar/Sidebar.jsx) to maintain the library at a reasonable size, and to allow developers to cherry pick missing dependencies according to their own projects. The following packages are the external dependencies utilized by React-Vapor:
```
sroy3: This is very picky of me and I acknowledge that this is really more of a personal preference (or a pet peeve in this case). So, I won't hold it against you if you do not change it, but I would change `utilized` for `used`. Thishttps://english.stackexchange.com/a/143945 explains my reasoning (especially the last box). But in the end we write code, not books, so as long as documentation is clear, it's fine by me.

gulpfile.js
```diff
@@ -110,6 +113,45 @@ gulp.task('test:watch', 'Run all tests in PhantomJS and watch', (done) => {
...
+        'node_modules/moment/min/moment-with-locales.min.js',
+        'node_modules/coveo-styleguide/dist/js/VaporSVG.js',
+    ])
+        .pipe(gulp.dest('./dist/dependencies/'))
+        .pipe(concat('react-vapor.dependencies.js'))
+        .pipe(gulp.dest('./dist/'));
+});
+
+gulp.task('dependencies', ['dependencies:prod', 'dependencies:docs']);
+
```
sroy3: Can we do all of this with webpack instead?

README.md
```diff
@@ -17,11 +17,34 @@ npm install react-vapor
 > The project is at its early stages, some components can still have _lots_ of breaking changes between versions.
```
sroy3: While we're editing the README, can we remove this line? I feel like it's outdated.

README.md
MichelML: nice KT, will do

README.md
MichelML: done 👍

README.md
MichelML: 👍

gulpfile.js
GermainBergeron: We can probably add an entry point in the webpack config and automatically put everything that comes from node_modules in another chunk. It will handle the dev and prod version automatically

See https://stackoverflow.com/a/49429500 for inspiration

MichelML: gulp is really good for stream processing, since we're only taking a bunch of files, concatenate them and move them to another places, i think gulp is really well suited for the job.  Can you specify how you would see this implemented in webpack?  
  
On a related topic, one thing I had in mind though is to provide a file containing the proper externals setup in webpack. Meaning we could have a js file, say `react-vapor.webpackexternals.js` :  
  
```  
module.exports = {
        codemirror: 'CodeMirror',
        jquery: '$',
        react: 'React',
        'react-dom': 'ReactDOM',
        'react-bootstrap': 'ReactBootstrap',
        'react-redux': 'ReactRedux',
        Slider: {root: ['rc-slider', 'Slider']},
        createSliderWithTooltip: {root: ['rc-slider', 'createSliderWithTooltip']},
        'coveo-styleguide': 'VaporSVG',
        redux: 'Redux',
        underscore: '_',
        moment: 'moment',
}  
```  
  
This would then be really simple for react-vapor users having webpack as their bundler to setup react-vapor properly in their build. They would just have to load `dist/react-vapor.dependencies.js` in the browser and to destructure this file under `externals` in their webpack.config and they would be good to go after that.

MichelML: @GermainBergeron will look into it. Feel free to comment on my comment as well.

sroy3: It was more of a suggestion than anything. I think it would work the way @GermainBergeron said, but if it is too complicated, it can be made some other time as well.

gdostie:
Reselect has been removed from the peerDependencies by https://github.com/coveo/react-vapor/commit/d5322e073c38bd599f924784e9c85d9ea4a316db, can you re-add it while you're playing with the bundle please? :smile:

gulpfile.js
MichelML: not sure @GermainBergeron 's proposition is the best strategy after reading this https://webpack.js.org/guides/author-libraries/#src/components/Sidebar/Sidebar.jsx  , will continue tomorrow

PR #690, opened 2018-06-18
reduce build logs by changing karma reporter in travis

MichelML:
i was tired of scrolling all the way down the build logs to get the demo link, this should help reduce the logs of the build. It will only print the failed UTs if any to the console, and the warnings.


MichelML:
progress was still causing many logs, used dots in travis instead

PR #691, opened 2018-06-18
Fix keyboard nav on infinite scroll dropdown search

sroy3:
The origin of the problem was hard to find in this messy component so I first thought it would be a good idea to duplicate and simplify the component. I got to something really clean when I figured out where the problem originated. Since creating a new component is less desirable (also there were some behaviours that weren't copied and I would have had to had tests), I fixed what was wrong with the component.

### DIFF
--- a/src/components/dropdownSearch/DropdownSearch.tsx
+++ b/src/components/dropdownSearch/DropdownSearch.tsx
@@ -407,10 +407,7 @@ export class DropdownSearch extends React.Component<IDropdownSearchProps, {}> {
     }
 
     private isSearchOn(): boolean {
-        if (this.props.infiniteScroll) {
-            return true;
-        }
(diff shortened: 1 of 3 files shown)


src/components/dropdownSearch/DropdownSearch.tsx
```diff
@@ -407,6 +407,9 @@ export class DropdownSearch extends React.Component<IDropdownSearchProps, {}> {
     }
 
     private isSearchOn(): boolean {
+        if (this.props.infiniteScroll) {
+            return true;
+        }
         return this.props.options.length > this.props.searchThresold;
```
gdostie: `return !!this.props.infiniteScroll || this.props.options.length > this.props.searchThresold;`

PR #692, opened 2018-06-18
No render on facets without rows

sroy3:
This was previously done in the other repos, but the facets were not added to the state. By doing this here, we remove repetition and make sure everything is cleanly in the state.

### DIFF
--- a/src/components/facets/Facet.tsx
+++ b/src/components/facets/Facet.tsx
@@ -85,20 +85,25 @@ export class Facet extends React.Component<IFacetProps, any> {
     }
 
     render() {
+        if (!this.props.facetRows.length && !this.props.selectedFacetRows.length) {
+            return null;
+        }
         const removeSelectedClass: string = 'facet-header-eraser' + (this.props.selectedFacetRows.length ? '' : ' hidden');
         const selected: IFacet[] = this.sortFacetRows(this.props.selectedFacetRows);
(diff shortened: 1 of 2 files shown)


src/components/facets/tests/Facet.spec.tsx
```diff
@@ -54,6 +52,15 @@ describe('Facets', () => {
             facetComponent.detach();
         });
 
+        it('should not render anything when there are no rows to display', () => {
+            expect(facetComponent.html()).toBeNull();
+        });
+
+        it('should render som ehtml when there are rows to display', () => {
```
MichelML: fix typo here

src/components/facets/Facet.tsx
```diff
@@ -85,20 +85,25 @@ export class Facet extends React.Component<IFacetProps, any> {
...
         const rows: JSX.Element[] = _.map(facetRows, (facetRow: IFacet) => {
-            return (<FacetRow
-                key={facetRow.name}
-                facet={this.props.facet.name}
-                facetRow={facetRow}
-                onToggleFacet={this.buildFacet}
-                isChecked={_.contains(_.pluck(this.props.selectedFacetRows, 'name'), facetRow.name)}
-                maxTooltipLabelLength={this.props.maxTooltipLabelLength}
-            />);
+            return (
```
gdostie: remove return?

PR #695, opened 2018-06-21
Fix default to empty table data on table initial load

MichelML:
since we now make a comparison with `JSON.stringify` to toggle  the initial load, an empty table data could be identical to DEFAULT_TABLE_DATA, and we need something to differentiate the two for the initial load to work properly.

### DIFF
--- a/src/components/tables/TableReducers.ts
+++ b/src/components/tables/TableReducers.ts
@@ -27,6 +27,11 @@ export interface ITableData {
     totalEntries: number;
     totalPages: number;
     selectedIds?: string[];
+    /**
+     * Only used by DEFAULT_TABLE_DATA object in TableConstants.tsx.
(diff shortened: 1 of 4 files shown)


src/components/tables/TableReducers.ts
```diff
@@ -27,6 +27,11 @@ export interface ITableData {
     totalEntries: number;
     totalPages: number;
     selectedIds?: string[];
+    /**
+     * Only used by DEFAULT_TABLE_DATA object in TableConstants.tsx.
+     * Useful for differientiating it from empty table data during the table initial load logic
+     */
+    DEFAULT_TABLE_DATA?: boolean;
```
pochretien: its not better to have a "has" before the variable name because its a boolean ?

MichelML:
@pochretien added `IS_`

PR #701, opened 2018-06-27
add border by default for code editors

MichelML:
requested by the UX team  
  
screen shot 2018-06-27 at 5 53 59 pm

### DIFF
--- a/src/components/editor/tests/CodeEditor.spec.tsx
+++ b/src/components/editor/tests/CodeEditor.spec.tsx
@@ -114,5 +114,9 @@ describe('CodeEditor', () => {
             expect(newList).not.toEqual(currentKeywords);
             expect(newList).toEqual(currentKeywords.concat(expectedNewKeywords));
         });
+
+        it('should have a border by default', () => {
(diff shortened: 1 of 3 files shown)


src/components/editor/tests/CodeEditor.spec.tsx
```diff
@@ -114,5 +114,9 @@ describe('CodeEditor', () => {
             expect(newList).not.toEqual(currentKeywords);
             expect(newList).toEqual(currentKeywords.concat(expectedNewKeywords));
         });
+
+        it('should have a border by default', () => {
+            expect(CodeEditor.defaultProps).toContain('mod-border');
```
sroy3: Your test isn't working. You're not checking the `className` prop. Maybe make it a little more complex like checking the prop when not sending `className` to make sure there is a default. You're only testing the content of a static variable for now.

MichelML: knowing how react works, isn't enough to test that?

MichelML: will add a test to look if the default class is added to the rendered code editor

PR #709, opened 2018-07-05
Add a props to remove the on hover on table rows

pochretien:
add props withoutHOverOnRow

### DIFF
--- a/src/components/tables/Table.tsx
+++ b/src/components/tables/Table.tsx
@@ -117,6 +117,7 @@ export class Table extends React.Component<ITableProps> {
         } as Partial<ITableCompositeState>,
         initialTableData: DEFAULT_TABLE_DATA,
         rowsMultiSelect: false,
+        withoutHoverOnRow: false,
     } as Partial<ITableOwnProps>;
 
     constructor(props: ITableProps) {
@@ -252,6 +253,7 @@ export class Table extends React.Component<ITableProps> {
                     handleOnRowClick={this.props.handleOnRowClick}
                     additionalRowClasses={this.props.additionalRowClasses}
                     isMultiSelect={this.props.rowsMultiSelect}
+                    withoutHoverOnRow={this.props.withoutHoverOnRow}
                 />
             );
         });
--- a/src/components/tables/examples/TableExamples.tsx
+++ b/src/components/tables/examples/TableExamples.tsx
@@ -233,6 +233,29 @@ export class TableExamples extends React.Component<any, any> {
                     />
                 </div>
                 <div className='form-group'>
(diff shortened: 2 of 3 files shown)


src/components/tables/examples/TableExamples.tsx
```diff
@@ -232,6 +232,29 @@ export class TableExamples extends React.Component<any, any> {
                         blankSlateDefault={{title: 'No results!'}}
                     />
                 </div>
+                <div className='form-group'>
+                    <label className='form-control-label'>Table without hover on row</label>
+                    <TableConnected
+                        id={'react-vapor-table-without-hover'}
```
sroy3: strings do not need to be wrapped inside `{}`

src/components/tables/examples/TableExamples.tsx
```diff
@@ -232,6 +232,29 @@ export class TableExamples extends React.Component<any, any> {
...
+                                attributeName: 'attribute4',
+                                titleFormatter: _.identity,
+                            },
+                            {
+                                attributeName: 'attribute3',
+                                titleFormatter: _.identity,
+                            },
+                        ]}
+                        blankSlateDefault={{title: 'No results!'}}
+                        withoutHoverOnRow={true}
```
sroy3: No need for the `={true}` adding the property will set it to true.

src/components/tables/Table.tsx
```diff
@@ -117,6 +117,7 @@ export class Table extends React.Component<ITableProps> {
         } as Partial<ITableCompositeState>,
         initialTableData: DEFAULT_TABLE_DATA,
         rowsMultiSelect: false,
+        withoutHoverOnRow: false,
```
gdostie: Shouldn't it be added to `ITableOwnProps` aswell ?

src/components/tables/Table.tsx
```diff
@@ -252,6 +253,7 @@ export class Table extends React.Component<ITableProps> {
                     handleOnRowClick={this.props.handleOnRowClick}
                     additionalRowClasses={this.props.additionalRowClasses}
```
gdostie: I think the prop `withoutHoverOnRow` and all the logic that goes with it is not necessary as it could be passed down using the `additionnalRowClasses`.

src/components/tables/Table.tsx
pochretien: nope, this interface extends ITableBodyInheritedFromTableProps with this prop

src/components/tables/Table.tsx
pochretien: I think the additionalRowClasses is for a custom behaviour. But in this case, its a case more common and its more simple to have a boolean prop to handle that

PR #711, opened 2018-07-06
Center the tooltip on ellipse styled text

acollin1:


gdostie:
Demo link: https://coveo.github.io/react-vapor/fix-tooltip-position-too-long-label-facetRow/ doesn't seem to have built properly not sure why.

PR #713, opened 2018-07-09
- Remove fixed-actions logic

gdostie:
This logic is no longer needed, will release a minor for this


src/components/actions/ActionBar.tsx
```diff
@@ -28,15 +28,14 @@ export interface IActionBarOwnProps extends React.ClassAttributes<ActionBar> {
     withSmallActions?: boolean;
     prefixContent?: IContentProps;
     width?: number;
-    maxScroll?: number;
+    // maxScroll?: number;
```
GermainBergeron: remove?

gdostie: woops yes thx

PR #715, opened 2018-07-09
Add exclude behavior on facets

acollin1:


src/components/facets/FacetRow.tsx
```diff
@@ -9,48 +10,96 @@ export interface IFacetRowProps extends React.ClassAttributes<FacetRow> {
...
+
+    private getLabel(): JSX.Element {
+        const className = this.props.isChecked && this.isExclude ? ' text-exclude' : '';
+        const maxCalculatedNameLength = this.props.facetRow.count
+            ? this.props.maxTooltipLabelLength - this.props.facetRow.count.length
+            : this.props.maxTooltipLabelLength;
+
+        if (this.props.facetRow.formattedName.length > maxCalculatedNameLength) {
+            return (
+                <Tooltip title={this.props.facetRow.tooltipLabel || this.props.facetRow.formattedName} placement='top' className={`inline-block full-content-x${className}`}>
```
sroy3: This line is a bit long, you should break every prop on its own line.

src/components/facets/FacetRow.tsx
```diff
@@ -9,48 +10,96 @@ export interface IFacetRowProps extends React.ClassAttributes<FacetRow> {
...
+                    {this.props.facetRow.formattedName}
+                </Tooltip>
+            );
+        } else {
```
sroy3: You could use a ternary in the return instead.

src/components/facets/FacetRow.tsx
```diff
@@ -9,48 +10,96 @@ export interface IFacetRowProps extends React.ClassAttributes<FacetRow> {
...
+            return <span className={className}>{this.props.facetRow.formattedName}</span>;
+        }
+    }
+
+    private getCount(): JSX.Element {
+        if (!_.isUndefined(this.props.facetRow.count)) {
+            return <span className={`facet-value-count ${this.props.isChecked && this.isExclude ? 'text-exclude' : ''}`}>{this.props.facetRow.count}</span>;
```
sroy3: This line seems a bit long. You can probably break it down a little.

src/components/facets/Facet.tsx
```diff
@@ -141,8 +135,22 @@ export class Facet extends React.Component<IFacetProps, any> {
                     {rows}
                     {moreRowsToggle}
                 </ul>
-                {moreRows}
+                {this.getMoreRows(!!moreRowsToggle, rows.splice(rowsToShow))}
             </div>
         );
     }
+
+    private getMoreRows(isNeedMore: boolean, rows: JSX.Element[]): JSX.Element {
```
gdostie: I think renaming `isNeedMore` for `needMoreRows` would make it clearer.

src/components/facets/Facet.tsx
```diff
@@ -41,6 +42,7 @@ export interface IFacetDispatchProps {
 export interface IFacetChildrenProps {
     moreLabel?: string;
     filterPlaceholder?: string;
+    enabledExclude?: boolean;
```
gdostie: I think renaming `enabledExclude` for `excludeEnabled`, `exclusionsEnabled` or `enableExclusions` would make more sense

src/components/facets/Facet.tsx
```diff
@@ -14,6 +14,7 @@ export interface IFacet {
     formattedName: string;
     tooltipLabel?: string;
     count?: string;
+    isExclude?: boolean;
```
gdostie: I think renaming `isExclude` for `isExcluded` or just `exclude` would make it clearer.

src/components/facets/FacetConnected.tsx
```diff
@@ -9,7 +9,7 @@ import {IFacetState} from './FacetReducers';
 
 const mapStateToProps = (state: IReactVaporState, ownProps: IFacetOwnProps): IFacetStateProps => {
     const item: IFacetState = _.findWhere(state.facets, {facet: ownProps.facet.name});
-    const selectedFacetRows = item ? ownProps.facetRows.filter((row) => _.findWhere(item.selected, {name: row.name})) : [];
+    const selectedFacetRows: IFacet[] = item ? getSelectedRows(ownProps.facetRows, item.selected) : [];
```
gdostie: I don't understand that change... Why didn't the previous line work? It seems like your new way of filtering the selected rows does exactly the same, but with more lines of code. I may be missing something.

src/components/facets/FacetRow.tsx
```diff
@@ -9,48 +10,96 @@ export interface IFacetRowProps extends React.ClassAttributes<FacetRow> {
     onToggleFacet: (facetRow: IFacet) => void;
     isChecked: boolean;
     maxTooltipLabelLength?: number;
+    enabledExclude?: boolean;
```
gdostie: Same here, I think renaming `enabledExclude` for `excludeEnabled`, `exclusionsEnabled` or `enableExclusions` would make more sense

src/components/facets/FacetRow.tsx
```diff
@@ -9,48 +10,96 @@ export interface IFacetRowProps extends React.ClassAttributes<FacetRow> {
...
-            : <span>{this.props.facetRow.formattedName}</span>;
-        const count: JSX.Element = !_.isUndefined(this.props.facetRow.count)
-            ? <span className='facet-value-count'>{this.props.facetRow.count}</span>
-            : null;
+    get isExclude(): boolean {
+        return !!this.props.enabledExclude && !!this.props.facetRow.isExclude;
+    }
 
+    render() {
+        const className: string = `facet-value facet-selectable ${!!this.props.enabledExclude ? 'facet-exclude' : ''}`;
```
gdostie: You could use the `classNames` util here (`import * as classNames from 'classnames'`):

```
const className: string = classNames(
    'facet-value',
    'facet-selectable',
    {'facet-exclude': this.props.enabledExclude}
);
```

src/components/facets/FacetConnected.tsx
acollin1: Is about the format of the selected rows in state. It's not always similar to facetsRow in props especially when they are excluded. So I need to extend the facetRow with the selectedRow if match finds.

sroy3:
How do you exclude a row? 

You should run `npm run lintfix`. The EOL changed for the `package.json`.

acollin1:
To exclude a facet you need to click on hidden button to the full right of the row -> http://nimb.ws/ZaVGzu. Lint fix is already done. If the EOL changed in package.json it's because I merged master and lintfix seems to not affect package.json file.

PR #717, opened 2018-07-11
Added sortable option on the Multi-Select component

GermainBergeron:
Added keyboard navigation for the single and multiselect
Added unit tests
Fixed webpack config to be able to share the url

Démo: https://coveo.github.io/react-vapor/feature-sortable-multiselect/

sortable


src/components/select/MultiSelectConnected.tsx
```diff
@@ -56,22 +67,58 @@ export class MultiSelectConnected extends React.Component<IMultiSelectProps, {}>
         );
     }
 
-    private getSelectedOptionComponents(): JSX.Element | JSX.Element[] {
+    private getSelectedOptionComponents() {
```
gdostie: type: `React.ReactNode` ?

types/react-dnd/index.d.ts
```diff
@@ -0,0 +1,37 @@
...
+    type ReactProps = any;
+
+    type _DragSourceTypePrimitives = string | symbol;
+    type DragSourceType = _DragSourceTypePrimitives | ((props: ReactProps) => _DragSourceTypePrimitives);
+
+    type _DropTargetTypesPrimitives = string | symbol | Array<string | symbol>;
+    type DropTargetTypes = _DropTargetTypesPrimitives | ((props: ReactProps) => _DropTargetTypesPrimitives);
+
+    interface IDragSource {
+        beginDrag: (props: ReactProps, monitor: DragSourceMonitor, component: YourReactClass) => any;
```
gdostie: Where does `YourReactClass` come from?

types/react-dnd/index.d.ts
```diff
@@ -0,0 +1,37 @@
+declare module ReactDnD {
```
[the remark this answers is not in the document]
GermainBergeron: There's already types for react-dnd in their repo, but I couldn't use the latest version since it depends on React 16

webpack.config.js
```diff
@@ -102,5 +102,7 @@ module.exports = {
     },
     devServer: {
         contentBase: './docs',
+        host: '0.0.0.0',
+        disableHostCheck: true,
```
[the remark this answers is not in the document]
GermainBergeron: It allows someone on the same network to access your "demo" url. E.g., http://wks-000999:8080/

MichelML:
I won't go through the code, but I noticed two things:   
  
1. is the style just not the good one in the demo (looks cleaner in your gif)?
screen shot 2018-07-16 at 9 05 12 am  
  
2. Shouldn't we hide the selected answers in the dropdown like in DropdownSearch ?

types/react-dnd/index.d.ts
GermainBergeron: Good question, I took it from their website. I replaced with `React.Component`

gdostie:
Sometimes when droping items in a specific spot makes them disappear, otherwise this looks very clean, GJ!

weeajarbassethound-size_restricted

MichelML:
good job @GermainBergeron , i'll approve right after you answered @gdostie last comment

GermainBergeron:
I found a solution and fixed the conflicts.

If you check https://github.com/coveo/react-vapor/pull/717/files#diff-036ddb4b7fbdfd7633b2756a8378cddcR48 the parent handles the drop "outside" the children and simply do nothing (so the last position where you hovered will be used as the drop location)

PR #724, opened 2018-07-18
Fix exclusion behavior and style on facets

acollin1:
Facets' rows style should respect these criteria:
- Icon like this https://thenounproject.com/search/?q=dont&i=10550.
- Show the icon en grey on hover the facet's row. Became blue on hover the button hit zone.
- Show a tooltip “Exclude [row_name] logs“ on hover the button hit zone.

When you click on exclude facet already exclude it should not be add again in state.

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "classnames": "2.2.5",
     "codecov": "3.0.2",
     "codemirror": "5.38.0",
-    "coveo-styleguide": "3.2.0",
+    "coveo-styleguide": "3.2.3",
     "css-loader": "0.28.11",
     "del": "3.0.0",
     "dts-generator": "2.1.0",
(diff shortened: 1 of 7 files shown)


acollin1:
Demo: https://coveo.github.io/react-vapor/fix-exclusion-behavior-and-style-on-facets/

otaillon:
Semble y avoir un petit bug avec l'icon, elle est croppée à gauche et en haut
screen shot 2018-07-19 at 10 01 04

acollin1:
@otaillon It's like this for all facets with exclude icons or not. I don't know if it expected, I can fix it but it's also a vapor issue. Just be sure it isn't expected

otaillon:
@acollin1 Je parle de l'icon d'exclude (pour préciser)

PR #725, opened 2018-07-19
Change the tooltip's DOM level

acollin1:
In firefox ellipsis doesn't display on a too long row.

### DIFF
--- a/src/components/facets/FacetRow.tsx
+++ b/src/components/facets/FacetRow.tsx
@@ -30,6 +30,7 @@ export class FacetRow extends React.Component<IFacetRowProps, any> {
             'facet-selectable',
             {'facet-exclude': this.props.enableExclusions},
         );
+        const labelClassName = this.props.isChecked && this.isExclude ? ' text-exclude' : '';
         return (
             <li className={className}>
                 {this.getExcludeCheckbox()}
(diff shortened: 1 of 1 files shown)


src/components/facets/FacetRow.tsx
```diff
@@ -30,6 +30,7 @@ export class FacetRow extends React.Component<IFacetRowProps, any> {
             'facet-selectable',
             {'facet-exclude': this.props.enableExclusions},
         );
+        const labelClassName = this.props.isChecked && this.isExclude ? ' text-exclude' : '';
```
rousselm4: remove useless space before text-exclude

GermainBergeron: It would be more standard to do it like this:
```
const labelClassName = classNames({
  'text-exclude': this.props.isChecked && this.isExclude
})
```

src/components/facets/FacetRow.tsx
```diff
@@ -52,15 +53,15 @@ export class FacetRow extends React.Component<IFacetRowProps, any> {
                     <button type='button' className={this.props.isChecked && this.isExclude ? 'exclude-box' : ''}>
                         <Svg svgName='clear' className='icon hide exclude-icon' svgClass='fill-red' />
                     </button>
-                    <span className='label'>{this.getLabel()}</span>
+                    {this.getLabel(<span className={labelClassName}>{this.props.facetRow.formattedName}</span>)}
```
GermainBergeron: I find it weird to call a method getLabel but this method as a "label" parameter. Instead of using a parameter in the function you should create a const in the actual function:
```

private getLabel(): JSX.Element {
  const labelClassName = classNames({
    'text-exclude': this.props.isChecked && this.isExclude
  });
  const label = <span className={labelClassName}>{this.props.facetRow.formattedName}</span>;
  ...
```

PR #728, opened 2018-07-23
Be sure to not reference to state when selected rows are modified

acollin1:


acollin1:
Demo: https://coveo.github.io/react-vapor/fix-reference-error-in-facetconnected-state/

src/components/facets/FacetConnected.tsx
```diff
@@ -20,7 +20,7 @@ const mapStateToProps = (state: IReactVaporState, ownProps: IFacetOwnProps): IFa
 
 const getSelectedRows = (rows: IFacet[], selected: IFacet[]): IFacet[] => _.without(_.map(rows, (row: IFacet) => {
     const element = _.findWhere(selected, {name: row.name});
-    return element ? _.extend(element, row) : null;
+    return element ? _.extend({}, element, row) : null;
```
gdostie: I think `return element && {...element, ...row};` would do the same and is shorter.

src/components/facets/FacetReducers.ts
```diff
@@ -46,24 +46,32 @@ const changeFacet = (state: IFacetState, action: (IReduxAction<IReduxActionsPayl
         return state;
     }
 
-    let selected = state.selected;
+    let selected = _.clone(state.selected);
     if (_.some(state.selected, (facetRow: IFacet) => facetRow.name === action.payload.facetRow.name)) {
         const selectedIndex: number = _.findIndex(state.selected, {name: action.payload.facetRow.name});
         if (!selected[selectedIndex].exclude && action.payload.facetRow.exclude) {
-            selected[selectedIndex] = {...selected[selectedIndex], exclude: true};
```
GermainBergeron: This won't modify the original object

src/components/facets/FacetReducers.ts
```diff
@@ -46,24 +46,32 @@ const changeFacet = (state: IFacetState, action: (IReduxAction<IReduxActionsPayl
...
+                name: selected[selectedIndex].name,
+                formattedName: selected[selectedIndex].formattedName,
+                exclude: true,
+            };
         } else {
             selected = _.reject(state.selected, (facetRow: IFacet, index: number) => index === selectedIndex);
         }
     } else {
         selected = [
-            action.payload.facetRow,
```
GermainBergeron:
```
selected = [
 {...action.payload.facetRow},
...state.selected,
];
```

acollin1: It is to not pass the count attr. Maybe I can just do an ..._.omit(action.payload.facetRow, 'count')

GermainBergeron: Then I prefer your solution

PR #730, opened 2018-07-23
Change the value of the input when sending a new value prop

sroy3:
Will create other PR for the release-2.21


sroy3:
See https://github.com/coveo/react-vapor/pull/731 instead

PR #737, opened 2018-07-25
Feature collapsible information box

gdostie:
Implemented a new component: `CollapsibleInfoBox`.

Based on the following mockup:
image

Also did these:
- Extracted the collapsible logic from the existing `CollapsibleContainer` into a new more generic component: `Collapsible` and `CollapsibleConnected`
- Refactored the existing `CollapsibleContainer` and `CollapsibleContainerConnected` components to use that new generic component
- Tried to make no breaking changes by deprecating things, but still **this may cause unforeseen breaking changes**
- The new components use CSS modules, so all the style comming from vapor for collapsible is unused here

Demo: https://coveo.github.io/react-vapor/feature-collapsible-information-box/


gdostie:
I've seen the drop in code coverage, wokring on it.

src/components/collapsible/CollapsibleActions.ts
```diff
@@ -0,0 +1,51 @@
...
+    expanded?: boolean;
+}
+
+/**
+ * @deprecated use addCollapsible instead
+ */
+export const addCollapsibleContainer = (id: string, expanded: boolean): IReduxAction<ILoadingActionPayload> => ({
+    type: CollapsibleActions.add,
+    payload: {id, expanded},
+});
```
sroy3: You could call `addCollapsible(id, expanded)` instead of repeating the whole action. (same for other actions)

src/components/collapsible/CollapsibleConnected.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+    return {expanded: collapsibleState && collapsibleState.expanded};
+};
+
+const mapDispatchToProps = (
+    dispatch: IDispatch,
+    ownProps: CollapsibleOwnProps,
+): CollapsibleDispatchProps => ({
+    onMount: () => dispatch(addCollapsibleContainer(ownProps.id, !!ownProps.expandedOnMount)),
+    onUnmount: () => dispatch(removeCollapsibleContainer(ownProps.id)),
```
sroy3: You could use the other actions that are not deprecated instead

src/components/collapsible/CollapsibleReducers.ts
```diff
@@ -0,0 +1,79 @@
...
+export interface ICollapsibleContainerState extends CollapsibleState {}
+
+export const collapsibleInitialState: CollapsibleState = {
+    id: undefined,
+    expanded: undefined,
+};
+
+/**
+ * @deprecated use collapsibleInitialState instead
+ */
```
sroy3: Other than in tests, I'm not sure someone would actually need to import this.  I think this is safe to remove.

sroy3: Same with `collapsibleContainersInitialState`

src/components/collapsible/styles/collapsibleContainer.scss
```diff
@@ -0,0 +1,14 @@
+$header-height: 50px;
+$content-padding: 20px 40px;
```
sroy3: You could add these in vapor and import them here so they can be used elsewhere. I think content padding already exists too.

src/components/collapsible/styles/collapsibleContainer.scss
```diff
@@ -0,0 +1,14 @@
...
+    margin-top: -1px;
+}
+
+.header {
+    min-height: $header-height;
+}
+
+.content {
+    padding: $content-padding;
+}
```
sroy3: This file needs to end on a new line

src/components/collapsible/styles/collapsibleInfoBox.scss
```diff
@@ -0,0 +1,10 @@
+$info-icon-width: 20px;
+$collapsible-border-radius: 4px;
```
sroy3: There is already a border radius variable in vapor and it is not set to 4. You should import vapor's variables instead.

src/components/collapsible/styles/collapsibleInfoBox.scss
```diff
@@ -0,0 +1,10 @@
...
+
+.roundedBorders {
+    border-radius: $collapsible-border-radius;
+}
+
+.alignWithIcon {
+    margin-left: calc(1rem + #{$info-icon-width});
+}
```
sroy3: Should end on a new line

sroy3:
Is it normal in the "CollapsibleExamples" that the toggle makes the page scroll change but not in the "CollapsibleContainerExamples"?

docs/Index.tsx
```diff
@@ -121,6 +125,7 @@ class App extends React.Component<any, any> {
             <Provider store={ReactVaporStore}>
                 <div className='coveo-form'>
                     {[
+                        {component: CollapsibleInfoBoxExamples, componentName: 'CollapsibleInfoBox'},
```
MichelML: could you remove all the `Examples` suffix in `componentName` it will mirror better the name of our components

gdostie: thought so too

gdostie:
@sroy3 
> Is it normal in the "CollapsibleExamples" that the toggle makes the page scroll change but not in the "CollapsibleContainerExamples"?
Yes, its because the 2 examples have the same id, therefore the action is dispatch for both collapsibles, I'll fix that thx

src/components/collapsible/styles/collapsibleInfoBox.scss
gdostie: Took that value out of UX's mockup. How do we import vapor variables in css modules?

src/components/collapsible/CollapsibleReducers.ts
```diff
@@ -0,0 +1,79 @@
...
+                collapsibleReducer(undefined, action),
+            ];
+        case CollapsibleActions.remove:
+            return _.reject(state, (collapsible: CollapsibleState) => collapsible.id === action.payload.id);
+        case CollapsibleActions.setExpanded:
+            return state.map((collapsible: CollapsibleState) => collapsibleReducer(collapsible, action));
+        default:
+            return state;
+    }
+};
```
MichelML:
```+export const collapsiblesReducer = (
+    state: CollapsibleState[] = collapsiblesInitialState,
+    action: IReduxAction<IReduxActionsPayload>,
+): CollapsibleState[] => {
+    switch (action.type) {
+        case CollapsibleActions.add:
+            return [
+                ...state,
+                action.payload,
+            ];
+        case CollapsibleActions.remove:
+            return _.reject(state, (collapsible: CollapsibleState) => collapsible.id === action.payload.id);
+        case CollapsibleActions.setExpanded:
+            return state.map((collapsible: CollapsibleState) => collapsible.id === action.payload.id ? action.payload : collapsible);
+        default:
+            return state;
+    }
+};```

MichelML: if you rewrite `collapsiblesReducer` this way you can remove the rest of the file imho, most importantly `collapsibleReducer`

src/components/collapsible/styles/collapsibleContainer.scss
MichelML: agree with @sroy3 this is too generic to be at the component level. You have one of them already in vapor but I don't think content-padding exists in vapor, header-height is `$collapsible-height`. To use the variables in vapor you can do `@import '~coveo-styleguide/scss/variables.scss';` at the top of the file and they will be available to you afterward.

MichelML: you can look at BorderedLine.tsx for an example

src/components/collapsible/CollapsibleReducers.ts
gdostie: I could also refactor to use an object instead of an array in the state, but I don't think its really necessary + it could introduce bugs. I'll go with your solution thx

src/components/collapsible/CollapsibleConnected.tsx
gdostie: Thx forgot to do this

src/components/collapsible/styles/collapsibleInfoBox.scss
sroy3: You can look at the DiffViewer, I've imported vapor's variables

MichelML:
you have lint errors making the build fail I think

PR #740, opened 2018-07-26
add the new menu connected component

pochretien:
add actions for the menu connected
add reducers for the menu connected
add uts
Add examples for the Menu connected

image
image

### DIFF
--- a/src/components/menu/MenuConnected.tsx
+++ b/src/components/menu/MenuConnected.tsx
@@ -48,7 +48,7 @@ const mapDispatchToProps = (
     onRender: () => dispatch(addMenu(ownProps.id, ownProps.items)),
     onDestroy: () => dispatch(removeMenu(ownProps.id)),
     onToggleMenu: () => dispatch(toggleMenu(ownProps.id)),
-    onDocumentClick: () => dispatch(toggleMenu(ownProps.id)),
+    onDocumentClick: () => dispatch(toggleMenu(ownProps.id, false)),
 });
 
 @ReduxConnect(mapStateToProps, mapDispatchToProps)
@@ -124,10 +124,10 @@ export class MenuConnected extends React.Component<IMenuProps, {}> {
     private setListPosition() {
         const menuToggle = this.getMenuToggle();
         if (menuToggle) {
-            this.list.style.minWidth = menuToggle.clientWidth + 2 + 'px';
+            this.list.style.minWidth = `${menuToggle.clientWidth}2px`;
             this.props.positionRight
-                ? this.list.style.right = menuToggle.offsetLeft + 'px'
-                : this.list.style.left = menuToggle.offsetLeft + 'px';
+                ? this.list.style.right = `${menuToggle.offsetLeft}px`
+                : this.list.style.left = `${menuToggle.offsetLeft}px`;
         }
     }
 
--- a/src/components/menu/MenuConnected.tsx
+++ b/src/components/menu/MenuConnected.tsx
@@ -82,7 +82,6 @@ export class MenuConnected extends React.Component<IMenuProps, {}> {
         );
         const dropdownClasses = classNames('select-dropdown-container absolute bg-pure-white', {
             hidden: !this.props.isOpen,
-            right: this.props.positionRight,
         });
         return (
             <div className={pickerClasses} ref={(ref: HTMLDivElement) => this.icon = ref}>
--- a/src/components/menu/MenuConnected.tsx
+++ b/src/components/menu/MenuConnected.tsx
@@ -1,7 +1,6 @@
 import * as classNames from 'classnames';
 import * as React from 'react';
 import * as ReactDOM from 'react-dom';
-import * as _ from 'underscore';
 import {IReactVaporState} from '../../ReactVapor';
 import {IDispatch, ReduxConnect} from '../../utils/ReduxUtils';
 import {IItemBoxProps} from '../itemBox/ItemBox';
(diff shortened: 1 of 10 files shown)


src/components/menu/MenuConnected.tsx
```diff
@@ -0,0 +1,144 @@
...
+};
+
+const mapDispatchToProps = (
+    dispatch: (action: IReduxAction<IReduxActionsPayload>) => void,
+    ownProps: IMenuOwnProps,
+): IMenuDispatchProps => ({
+    onRender: () => dispatch(addMenu(ownProps.id, ownProps.items)),
+    onDestroy: () => dispatch(removeMenu(ownProps.id)),
+    onToggleMenu: () => dispatch(toggleMenu(ownProps.id)),
+    onDocumentClick: () => dispatch(toggleMenu(ownProps.id)),
```
sroy3: You could remove `onDocumentClick` and simply use `onToggleMenu` instead

src/components/menu/MenuConnected.tsx
```diff
@@ -0,0 +1,144 @@
...
+        this.props.onToggleMenu();
+    }
+
+    private setListPosition() {
+        const menuToggle = this.getMenuToggle();
+        if (menuToggle) {
+            this.list.style.minWidth = menuToggle.clientWidth + 2 + 'px';
+            this.props.positionRight
+                ? this.list.style.right = menuToggle.offsetLeft + 'px'
+                : this.list.style.left = menuToggle.offsetLeft + 'px';
```
sroy3: These should all be template strings

src/components/menu/MenuReducers.ts
```diff
@@ -0,0 +1,54 @@
...
+        case MenuActions.add:
+            return [
+                ...state,
+                menuReducer(undefined, action),
+            ];
+        case MenuActions.remove:
+            return _.reject(state, (listBox: IMenuState) => action.payload.id === listBox.id);
+        case MenuActions.toggle:
+        case MenuActions.updateList:
+            return state.map((menu: IMenuState) => menuReducer(menu, action));
```
sroy3: You call it `menu` here and `listBox` on line 47. It does not have any real impact, but for consistency it might be a good idea to keep the same name.

src/components/menu/MenuConnected.tsx
```diff
@@ -0,0 +1,144 @@
...
+import {IReduxAction, ReduxConnect} from '../../utils/ReduxUtils';
+import {IItemBoxProps} from '../itemBox/ItemBox';
+import {ListBox} from '../listBox/ListBox';
+import {Svg} from '../svg/Svg';
+import {addMenu, removeMenu, toggleMenu} from './MenuActions';
+import {IMenuState} from './MenuReducers';
+
+export interface IMenuOwnProps {
+    items: IItemBoxProps[];
+    classes?: string[];
```
GermainBergeron: Please use string instead (this way we can use classNames)

src/components/menu/MenuConnected.tsx
```diff
@@ -0,0 +1,144 @@
...
+    id: string;
+    svgName?: string;
```
GermainBergeron: Instead of having a svg, maybe we should use the children and use the tag like so:

```jsx
<MenuConnected><Svg svgName='svg-name' svgClass='fill-medium-blue icon mod-lg' /></MenuConnected>
<MenuConnected>Some String</MenuConnected>
<MenuConnected><OtherTag /></MenuConnected>
```

src/components/menu/MenuConnected.tsx
```diff
@@ -0,0 +1,143 @@
...
+            hidden: !this.props.isOpen,
+        });
+        return (
+            <div className={pickerClasses} ref={(ref: HTMLDivElement) => this.icon = ref}>
+                <button
+                    className={classNames('btn menu-toggle', {'bg-light-grey': this.props.isOpen})}
+                    type='button'
+                    onMouseUp={(e: React.MouseEvent<HTMLElement>) => this.onToggleMenu(e)}
+                >
+                    {this.getSvg()}
```
GermainBergeron: With my proposed change for the children you could still have a default

`{this.props.children ? this.props.children : this.getDefaultSvg()}`

src/components/menu/MenuConnected.tsx
```diff
@@ -0,0 +1,143 @@
...
+        this.setListPosition();
+
+        e.stopPropagation();
+        e.preventDefault();
+
+        this.props.onToggleMenu();
+    }
+
+    private setListPosition() {
+        const menuToggle = this.getMenuToggle();
```
GermainBergeron: I don't think you need a function for this, you could have a ref instead

src/components/menu/MenuConnected.tsx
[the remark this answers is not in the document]
pochretien: we use a string array for other component. I only to be uniform. Same for the array in the state

src/components/menu/MenuConnected.tsx
```diff
@@ -0,0 +1,138 @@
...
+});
+
+@ReduxConnect(mapStateToProps, mapDispatchToProps)
+export class MenuConnected extends React.Component<IMenuProps, {}> {
+    private icon: HTMLDivElement;
+    private list: HTMLDivElement;
+    private button: HTMLButtonElement;
+
+    static defaultProps: Partial<IMenuProps> = {
+        className: '',
```
gdostie: This is not required, classNames utils can handle undefined values

src/components/menu/MenuConnected.tsx
```diff
@@ -0,0 +1,138 @@
...
+            this.props.className,
+            {
+                open: this.props.isOpen,
+            },
+        );
+        const dropdownClasses = classNames('select-dropdown-container absolute bg-pure-white', {
+            hidden: !this.props.isOpen,
+        });
+        return (
+            <div className={pickerClasses} ref={(ref: HTMLDivElement) => this.icon = ref}>
```
gdostie: You can remove the ref here

pochretien:
image

PR #741, opened 2018-07-26
! Make callIfDefined util return what the callback returns

gdostie:


src/utils/FalsyValuesUtils.ts
```diff
@@ -5,8 +5,8 @@ export const convertUndefinedAndNullToEmptyString = (value: any) =>
         ? ''
         : value;
 
-export const callIfDefined = (callback: any, ...args: any[]) => {
+export const callIfDefined = (callback: any, ...args: any[]): any => {
```
acollin1: Use a template as return Type could be nice!

gdostie: Not sure how TBH, do you have a snippet?

acollin1: `export const callIfDefined = <T>(callback: () => T, ...args: any[]): T => {
    if (_.isFunction(callback)) {
        callback(...args);
        return callback(...args);
     }
};`

PR #746, opened 2018-07-30
fix-non-ES5-compatible-file

MichelML:
the root cause was that unidiff uses `let` and `const` in a file and these are not converted to `var` through the typescript compilation layer. I've added a babel-loader making this transformation for the file in question, which fixes the problem all the way down to the bundle-v2 js minification step.


package.json
```diff
@@ -144,12 +145,12 @@
     "sass-loader": "7.0.3",
     "shelljs": "0.8.2",
     "style-loader": "0.21.0",
-    "ts-loader": "4.3.1",
+    "ts-loader": "4.4.2",
```
pochretien: I think we can remove this dependencies ?

GermainBergeron: Can your remove ts-loader from package.json (and use awesome-typescript-loader in the tests too)

MichelML: it actually causes the test run to hang and never start...

MichelML: see https://github.com/coveo/react-vapor/pull/746#discussion_r206294859 as of now

MichelML:
done @sroy3 👍

PR #760, opened 2018-08-07
Add textarea autosize

pochretien:
add forder examples and tests for the textarea
add a new lib for a textarea with autosize
add examples for it


package.json
```diff
@@ -1,167 +1,169 @@
...
+  "pre-commit": [
+    "lintfix"
+  ],
+  "dependencies": {
+    "diff2html": "2.4.0",
+    "react-bootstrap": "0.31.3",
+    "react-codemirror2": "5.1.0",
+    "react-infinite-scroll-component": "4.1.0",
+    "react-tether": "0.5.7",
+    "react-textarea-autosize": "7.0.4",
```
pochretien: added this library with his d.ts

package.json
```diff
@@ -1,167 +1,169 @@
-{
```
sroy3: In Webstorm, go in "Preferences", then "Editor > Code Style" and change the line separator to use the Windows one. This should fix the EOL problem.

src/components/textarea/examples/TextareaAutosizeExamples.tsx
```diff
@@ -0,0 +1,33 @@
+import * as React from 'react';
+import TextareaAutosize from 'react-textarea-autosize';
```
sroy3: I think we should have a boolean in our `TextArea` to have it autosize or not instead of importing the component from another lib.

pochretien: this lib use autosize to handle the resize of the textarea. I think its useless to code the same component. this lib is very small and it add only this component.

pochretien:
Yep, I tried to change the config in webstorm, and save again the file but the diff remains the same :(

src/components/textarea/TextArea.tsx
```diff
@@ -69,8 +72,9 @@ export class TextArea extends React.Component<ITextAreaProps, {}> {
     }
 
     render() {
+        const TextareaEl: any = this.props.isAutosize ? TextareaAutosize : 'textarea';
```
gdostie: `TextareaTagName` or `TextareaComponentClass` would be clearer IMHO.

PR #761, opened 2018-08-08
Textarea available without redux

pochretien:


src/components/textarea/tests/TextArea.spec.tsx
```diff
@@ -77,6 +77,15 @@ describe('TextArea', () => {
...
 
+        it('should call prop onChangeCallBack on textarea change', () => {
+            const onChangeCallBack = jasmine.createSpy('onChangeCallBack');
+
+            wrapper.setProps({onChangeCallBack});
+            textArea.simulate('change');
+
+            expect(onChangeCallBack).toHaveBeenCalledTimes(1);
+        });
+
```
sroy3: You'll need to add tests calling `handleOnChange` without `onChange` and `onChangeCallBack` and check that it does not throw or else the coverage will go down.

src/components/textarea/TextArea.tsx
```diff
@@ -24,6 +24,8 @@ export interface ITextAreaOwnProps {
     disabledOnMount?: boolean;
 
     isAutosize?: boolean;
+
+    onChangeCallBack?: (event: React.KeyboardEvent<HTMLTextAreaElement>) => void;
```
sroy3: Why not just use `onChange` if you aren't using Redux?

pochretien: I added this function for the redux use case ( if you want to do something onChange without removing the redux behaviour.

src/components/textarea/tests/TextArea.spec.tsx
pochretien: I tried to do something for the onChange not defined but the not.throw want a function as argument. I check for an other solution to test.

src/components/textarea/TextArea.tsx
```diff
@@ -80,10 +82,20 @@ export class TextArea extends React.Component<ITextAreaProps, {}> {
...
-                onChange={(e: React.ChangeEvent<HTMLTextAreaElement>) => this.props.onChange(e)}
+                onChange={(e: React.ChangeEvent<HTMLTextAreaElement>) => this.handleOnChange(e)}
             />
         );
     }
+
+    private handleOnChange(e: React.ChangeEvent<HTMLTextAreaElement>) {
+        if (this.props.onChange) {
+            this.props.onChange(e);
+        }
```
gdostie: `callIfDefined(this.props.onChange, e);`

src/components/textarea/TextArea.tsx
```diff
@@ -80,10 +82,20 @@ export class TextArea extends React.Component<ITextAreaProps, {}> {
...
+
+        if (this.props.onChangeCallBack) {
+            this.props.onChangeCallBack(e);
+        }
```
gdostie: `callIfDefined(this.props.onChangeCallBack, e);`

src/components/textarea/TextArea.tsx
pochretien: oh nice!

PR #764, opened 2018-08-09
Add onClick and additional classes on cell for the table

pochretien:
add utils for uts
add uts for the additionalClasses on cell
add props for onClick on cell click with a condition

### DIFF
--- a/src/utils/EventUtils.ts
+++ b/src/utils/EventUtils.ts
@@ -1,6 +1,11 @@
 import * as React from 'react';
 import {isEmpty} from 'underscore';
 
+export interface IConditionalOnClickCallBack {
+    callBack: (event?: any, data?: any) => void;
+    condition?: (data: any) => boolean;
+}
+
 export class EventUtils {
     static isClickingInsideElementWithClassname = (e: React.MouseEvent<any>, classname: string): boolean => {
         let currentTarget: HTMLElement = e && (e.target as HTMLElement);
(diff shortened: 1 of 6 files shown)


pochretien:
Uts for the onClick coming soon

src/utils/TestUtils.ts
```diff
@@ -133,3 +133,7 @@ export const defaultTooltipProps: ITooltipProps = {
     placement: 'bottom',
     container: 'body',
 };
+
+export const triggerAlertFunction = () => {
+    alert(`Alert function triggered`);
+};
```
sroy3: I like that!

src/utils/EventUtils.ts
```diff
@@ -1,6 +1,11 @@
 import * as React from 'react';
 import {isEmpty} from 'underscore';
 
+export interface IConditionalOnClickCallBack {
+    callBack: (event?: any, data?: any) => void;
```
MichelML: if you have the data in the callback, why create a condition too with the same argument? What i'm trying to say is you could get away with only the callback and write a condition inside it if you need to, no need for the `condition` attribute/property

pochretien: Yep, its a good point but I want to stay with the same idea than the additionalClass. I think its more clear to have the condition in the table props instead of inside each callBack. You can handle only the callback behaviour without the condition eveytime

MichelML:
@gdostie @pochretien is right, talked to him on the phone and made total sense it's easier and less tricky to have an `onClickCell` prop

PR #765, opened 2018-08-09
Update list box each item on click

pochretien:


src/components/listBox/tests/ListBox.spec.tsx
```diff
@@ -75,6 +77,13 @@ describe('ListBox', () => {
             expect(onDestroySpy).toHaveBeenCalledTimes(1);
         });
 
+        it('should render items with events on onOptionClick', () => {
+            renderListBox();
+
+            (listBoxComponent.find(ItemBox) as any).node.handleOnOptionClick({target: 'target'});
```
GermainBergeron: Here you click on every boxes? Wouldn't it be better to simulate a click on the 2nd box?

src/components/listBox/ListBox.tsx
```diff
@@ -69,7 +70,7 @@ export class ListBox extends React.Component<IListBoxProps, {}> {
             .map((item: IItemBoxProps) => <ItemBox
                 key={item.value}
                 {...item}
-                onOptionClick={(option: IItemBoxProps) => this.onSelectItem(option)}
+                onOptionClick={(option: IItemBoxProps) => this.onSelectItem(option, item)}
```
GermainBergeron: Why are you sending both the option and the item?
IMO you should use: `onOptionClick={() => this.onSelectItem(item)}`

gdostie: It would be a breaking change, he needs option in the `onSelectItem` method

src/components/listBox/examples/ListBoxExamples.tsx
```diff
@@ -23,9 +27,11 @@ export class ListBoxExamples extends React.Component {
...
         ];
 
-        const triggerAlertFunction = (item: IItemBoxProps) => {
-            alert(`The item value triggered is ${item.value}`);
-        };
+        const clickableItems: IItemBoxProps[] = defaultItems.concat([{
+            value: 'test8',
+            onOptionClick: triggerAlertFunction,
+        },
+        ]);
```
MichelML: const clickableItems: IItemBoxProps[] = defaultItems.concat([
            {
                value: 'test8',
                onOptionClick: triggerAlertFunction,
            },
        ]);

src/components/listBox/ListBox.tsx
pochretien: its not the same object item and option. I think its can break something if we change that

src/components/listBox/tests/ListBox.spec.tsx
pochretien: Fixed, but when we do .node, it target the node from the list. Modified to target the node 2.

src/components/listBox/ListBox.tsx
GermainBergeron: It's not a breaking change since we're sending an element with the same type

PR #766, opened 2018-08-10
Banners

sroy3:
screen shot 2018-08-10 at 11 23 21 am

https://github.com/coveo/vapor/pull/568 needed


src/components/banner/Banner.tsx
```diff
@@ -0,0 +1,69 @@
...
+import * as _ from 'underscore';
+
+import {Svg} from '../svg/Svg';
+import * as styles from './styles/Banner.scss';
+
+export interface BannerProps {
+    name: string;
+    nameSubtitle?: string;
+    messageTitle?: string;
+    messageDescription?: React.ReactNode;
```
gdostie: use children instead

src/components/banner/Banner.tsx
```diff
@@ -0,0 +1,69 @@
...
+                {this.props.bottomRightInfos && <div className={classNames(styles.bannerRight, 'right-align')}>
+                    <div className={styles.bannerBottomRightInfos}>{this.props.bottomRightInfos}</div>
+                </div>}
+            </div>
+        );
+    }
+
+    private getTitleSvg(): JSX.Element {
+        if (_.contains([BannerMessageStates.Warning, BannerMessageStates.Error], this.props.messageState)) {
+            return <Svg svgName='info' svgClass='icon mx1' className={styles.bannerWarningIcon} />;
```
pochretien: It's possible to have an other svg here ? we can add a prop for the svg and add this one by default. what do you think ?

src/components/banner/styles/Banner.scss
```diff
@@ -0,0 +1,92 @@
...
+
+.banner {
+    background: linear-gradient(45deg, $mirage 0%, $science-blue 100%);
+    box-shadow: $banner-box-shadow;
+    padding: $banner-padding;
+    height: $banner-height;
+    box-sizing: border-box;
+    position: relative;
+
+    button {
```
pochretien: dont use the element directly, add a class on it

src/components/banner/styles/Banner.scss
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        &:global(.primary) {
+            background-color: $orange;
+            border: none;
+        }
+    }
+}
+
+.bannerName {
```
pochretien: use a - for classes : banner-name

pochretien: or you use this naming because its more simple to access your variable inside the component ? like styles.bannerWarningTitle

src/components/banner/Banner.tsx
sroy3: I feel like it's unclear what the children should be. I think it's easier to know what goes where if there is a prop for it.

src/components/banner/styles/Banner.scss
sroy3: I don't have access to it, it's the one sent with the content. It is scoped to the banner context.

src/components/banner/styles/Banner.scss
sroy3: You can only use camelCase in CSS Modules

src/components/banner/Banner.tsx
sroy3: I mean there are three props receiving a React.ReactNode.

src/components/banner/Banner.tsx
sroy3: I'll add this.

src/components/banner/styles/Banner.scss
gdostie: If its sent in the content, I say let the content style itself. Nothing more frustrating then syles overriding each other.

src/components/banner/Banner.tsx
gdostie: To me it is pretty clear that the children should be the main content in the middle of the banner. With `topRightInfos` and `bottomRightInfos` as props, it gets even clearer

src/components/banner/styles/Banner.scss
gdostie: You could front this style by putting it in a class and front it as static variable of the Banner component, then the user of the component could choose wether or not to use the class in its buttons

sroy3: You're right. At first it was because I found an old mockup with special buttons, but since they're "normal" now anyway, better not style them. I will remove.

src/components/banner/Banner.tsx
sroy3: Right now, there are no other svgs and it's for warnings and errors. I think it's better to keep this behaviour here and to expand later if we need to.

PR #777, opened 2018-08-17
Query Expression Editor - Logic Review

huppeje:
This is the first part of the Query Expression Editor project (the tests have not been added in this PR). It is not 100% completed and there might still be some little bugs!

Test it out : https://coveo.github.io/react-vapor/qeb-develop/

First, sorry about the fact that this is a big PR. I will reassure you, the Query Expression Editor only plugs together existing react-vapor components with the right data. 

The Query Expression Editor can be divided in three parts : 

1 - The BUILD FROM RESULT Editor & the PREVIEW. 
Concerned files : 
- **buildFromResult**
- **preview**
- **resultItem**
- **resultList**

2 -  The FORM MODE
Concerned files : 
- **formMode**
- **expressionEditor**
- **fieldSelect**
- **operatorSelect**
- **valueSelect**
- **valueSelectDate**
- **valueSelectNumber**
- **valueSelectString**
- **booleanOperatorSelect_**

3 - The ADVANCED MODE
Concerned files : 
- **advancedMode**

There is also two files for the communication with index : **queryTrigger** & **responseParser**.
And finally there is the **editors** file that regroups all the editor and the QueryExpressionEditor file that regroups the editors and the preview.

I made some simple graphs to better understand the data flow in the Query Expression Editor (see the images linked to the PR).
 
### Important notes : 

I temporary changed the index.tsx file so you can go right away and test the component.

The css is not done yet, don't bother checking it!

I left some TODOs for myself.
Some TODOs have a QUESTION tag (// TODO QUESTION ) those are questions^^
The questions for the react-vapor team are tagged with : // TODO QUESTION R-V

// TODO QUESTION
For now all the strings are hard coded, should these be regrouped in a string.json file?

// TODO QUESTION R-V
I had to change the  "target": "ES5" to "target": "es2015", to be able to use the promise constructor (Promise<T>) with the async and await keywords is it ok?

Finally, don’t hesitate to ask me questions!

### Graphs

basic_data_flow

expression_editor_terminology

update_expression_editor_redux_state


src/components/queryExpressionEditor/expressionEditor/ExpressionEditor.tsx
```diff
@@ -0,0 +1,205 @@
...
+        // TODO : Review if all types are handled
+        if (originalFieldType === OriginalFieldType.Date) {
+            return FieldType.Date;
+        }
+
+        if (originalFieldType === (OriginalFieldType.Long64 || OriginalFieldType.Long || OriginalFieldType.Double)) {
+            return FieldType.Number;
+        }
+
+        if (originalFieldType === OriginalFieldType.LargeString) {
```
olamothe: I'd change that to default to "string" at the end, instead of null

tsconfig.json
```diff
@@ -12,7 +12,7 @@
     "removeComments": true,
     "skipLibCheck": true,
     "sourceMap": true,
-    "target": "ES5",
+    "target": "es2015",
```
olamothe: Erhm, this might cause a lot of problem, should probably not touch this.

docs/index.html
```diff
@@ -11,7 +11,7 @@
         React Vapor
     </div>
 </div>
-<div class="container p2">
+<div class="container p2 bg-dark-blue">
```
gdostie: Pleaze revert, our demo page is meant to be white.

src/components/queryExpressionEditor/QueryExpressionEditor.scss
```diff
@@ -0,0 +1,24 @@
+.container {
+    overflow: auto;
+}
+
+.leftContainer {
+    float: left;
```
gdostie: Not sure about the use of float here, we tend to prefer the use of flex-boxes

src/components/queryExpressionEditor/QueryExpressionEditor.scss
```diff
@@ -0,0 +1,24 @@
...
+    width: 35%;
+}
+
+.topLeftContainer {
+    height: 114px;
+}
+
+.queryExpressionContainer {
+    height: 40px;
+    display: -webkit-flex;
```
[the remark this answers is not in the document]
gdostie: You're right, vendors prefix aren't necessary, we have an autoprefixer which adds them when building the main css file.

src/components/queryExpressionEditor/QueryExpressionEditor.tsx
```diff
@@ -0,0 +1,71 @@
...
+        return (
+            <div className={containerClasses}>
+                <div className={styles.leftContainer}>
+                    <div className={styles.topLeftContainer}>
+                        <div className='h1 text-medium-blue mt5 ml4 semibold'>
+                            Query Expression Builder
+                        </div>
+                        <div className={`h3 mt4 text-pure-white bg-light-blue ${styles.queryExpressionContainer}`}>
+                            <div className='ml4'>
+                                <span>Generated Query Expression : </span>
```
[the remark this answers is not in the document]
gdostie: The standard that we use is that every string should be configurable via props. Say the project that uses this component has to support other languages, then it couldn't translate those strings that are not accessible throught props.

src/components/queryExpressionEditor/QueryExpressionEditorUtils.tsx
```diff
@@ -0,0 +1,17 @@
+// TODO : Review isWrappedWithQuotes condition
+export const addQuotesIfContainsWhiteSpace = (value: string) => {
```
gdostie: Not sure this whole util is really necessary as you can do exactly this using
```ts
JSON.parse(JSON.stringify('your string here'));
```

src/components/queryExpressionEditor/advancedMode/AdvancedMode.tsx
```diff
@@ -0,0 +1,50 @@
...
+    private onChange(content: string) {
+        this.codeEditorContent = content;
+        this.props.updateQueryExpression(content);
+    }
+
+    render() {
+        return (
+            <div>
+                <div className='text-medium-blue mt4 ml4'>
+                    Write your query using <a href='https://onlinehelp.coveo.com/en/cloud/coveo_cloud_query_syntax_reference.htm'>Coveo query syntax </a>.
```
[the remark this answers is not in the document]
gdostie: I agree, and the string too should be configurable

src/components/queryExpressionEditor/booleanOperatorSelect/BooleanOperatorSelect.tsx
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export class BooleanOperatorSelect extends React.Component<IBooleanOperatorSelectProps> {
+
+    render() {
+        return (
+            <SingleSelectConnected
+                id={`${this.props.expressionEditorId}-${BOOLEAN_OPRATOR_SELECT_ID}`}
+                items={BOOLEAN_OPERATOR_ITEMS}
+                placeholder='Add rule'
```
gdostie: Add prop for placeholder

src/components/queryExpressionEditor/booleanOperatorSelect/BooleanOperatorSelect.tsx
```diff
@@ -0,0 +1,28 @@
+
+import * as React from 'react';
+import {IItemBoxProps} from '../../itemBox/ItemBox';
+import {SingleSelectConnected} from '../../select/SingleSelectConnected';
+
+export const BOOLEAN_OPRATOR_SELECT_ID: string = 'boolean-operator-select';
+export const BOOLEAN_OPERATOR_ITEMS: IItemBoxProps[] = [{value: 'AND', displayValue: 'And'}, {value: 'OR', displayValue: 'Or'}];
```
gdostie: Move those two const as static variables of `BooleanOperatorSelect`

src/components/queryExpressionEditor/buildFromResult/BuildFromResult.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+    private buildFieldExpression(fieldExpression: IFieldExpression): string {
+        if (fieldExpression.field && fieldExpression.fieldValue) {
+            return `${fieldExpression.field}=${fieldExpression.fieldValue}`;
+        } else {
+            return null;
+        }
+    }
+
+    private async onSearch(searchBarText: string) {
```
gdostie: IIRC `async` & `await` functions aren't supported by IE 11 which we need to support, use promises instead

gdostie:
Main commnets I have seen a lot of `setState` calls, but I think all those should be managed using redux with proper reducers and actions.

Also, many strings are written in plain english directly embedded directly in components. These should all be configurable using props in some way that its easy for the lib user to translate their app.

src/components/queryExpressionEditor/QueryExpressionEditor.tsx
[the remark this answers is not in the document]
MichelML: @gdostie is right, the currently provided strings could however be in the defaultProps

tsconfig.json
samisayegh: Yeah, changing the target would break the project in browsers that do not support es6/es2015.

Instead you could specify es2015 in the `lib` option if you want to use `async/await` or promise. By default, when the lib is not specified, if the target is es5, then the lib will be es5. However, you can set it to something else. Further info:

https://www.typescriptlang.org/docs/handbook/compiler-options.html

src/components/queryExpressionEditor/QueryExpressionEditor.scss
```diff
@@ -0,0 +1,24 @@
+.container {
```
MichelML: you can use `.overflow-auto` instead

MichelML: https://github.com/coveo/vapor/blob/0696ab8feb29ac04e5846cfb0b35ee2bc31b7fee/scss/utility/layout.scss

src/components/queryExpressionEditor/QueryExpressionEditor.scss
```diff
@@ -0,0 +1,24 @@
...
+}
+
+.topLeftContainer {
+    height: 114px;
+}
+
+.queryExpressionContainer {
+    height: 40px;
+    display: -webkit-flex;
+    display: flex;
```
MichelML: `.flex` and `.flex-center` in vapor can help you here https://github.com/coveo/react-vapor/pull/777/files

src/components/queryExpressionEditor/buildFromResult/BuildFromResult.tsx
```diff
@@ -0,0 +1,113 @@
...
+        }
+        return this.buildFieldExpression(fieldExpression);
+    }
+
+    private getFieldExpression(result: IResult): IFieldExpression {
+        if (!result) {
+            return null;
+        }
+
+        if (result.objectType) {
```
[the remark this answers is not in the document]
olamothe: Check for not null AND not undefined.

Also @huppeje can you double check that this works properly here ? It seems to me that you are using case sensitive check, so this always fails and never selects either objecttype and filetype.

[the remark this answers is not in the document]
samisayegh: When using `!= null` with one equal sign, the expression will be true only if the variable is not `null` and not `undefined`. However, I am still not clear on the benefit over using a simple truthy check. What is the case you are trying to screen for?

src/components/queryExpressionEditor/QueryExpressionEditor.scss
MichelML: `.left` class in vapor can be used if you really need to float

src/components/queryExpressionEditor/advancedMode/AdvancedMode.tsx
```diff
@@ -0,0 +1,50 @@
...
+    initialContent: string;
+}
+
+export class AdvancedMode extends React.Component<IAdvancedModeProps, IAdvancedModeState> {
+    private codeEditorContent: string;
+
+    constructor(props: IAdvancedModeProps) {
+        super(props);
+        this.state = {initialContent: ''};
+        this.codeEditorContent = '';
```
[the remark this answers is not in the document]
MichelML: I agree with what Oli and Sami are saying, codeEditorContent is basically the queryExpression. You should find a way not to duplicate this value in two entities. Why not debouncing input changes and updating the queryExpression after a certain timeout? I think it would work and typing in the code editor would feel smoother.

```
export class AdvancedMode extends React.Component<IAdvancedModeProps, IAdvancedModeState> {
    private codeEditorContent: string;

    constructor(props: IAdvancedModeProps) {
        super(props);
        this.state = {initialContent: ''};
        this.codeEditorContent = '';
    }

    private debouncedUpdateQueryExpression = _.debounce((content: string) => this.props.updateQueryExpression(content), 200);

    render() {
        return (
            <div>
                <div className='text-medium-blue mt4 ml4'>
                    Write your query using <a href='https://onlinehelp.coveo.com/en/cloud/coveo_cloud_query_syntax_reference.htm'>Coveo query syntax </a>.
                </div>
                <div className={`mt3 ml4 ${styles.codeEditor}`}>
                    <CodeEditor
                        value={this.props.queryExpression} mode={CodeMirrorModes.Python}
                        onChange={(content: string) => this.debouncedUpdateQueryExpression(content)}
                    />
                </div>
            </div>
        );
    }
}  

```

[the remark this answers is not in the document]
MichelML: Also, if there is an existing queryExpression , we should be able to see it on mount/first, currently it seems like it is always an empty.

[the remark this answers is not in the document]
MichelML: https://underscorejs.org/#debounce

src/components/queryExpressionEditor/booleanOperatorSelect/BooleanOperatorSelect.tsx
```diff
@@ -0,0 +1,28 @@
...
+export const BOOLEAN_OPRATOR_SELECT_ID: string = 'boolean-operator-select';
+export const BOOLEAN_OPERATOR_ITEMS: IItemBoxProps[] = [{value: 'AND', displayValue: 'And'}, {value: 'OR', displayValue: 'Or'}];
+
+export interface IBooleanOperatorSelectProps {
+    expressionEditorId: string;
+    isExpressionComplete: boolean;
+    onBooleanOperatorSelect: (booleanOperator: string) => void;
+}
+
+export class BooleanOperatorSelect extends React.Component<IBooleanOperatorSelectProps> {
```
MichelML: you could refactor it with the arrow function to make it shorter:   

```
export const BooleanOperatorSelect = (props: IBooleanOperatorSelectProps) =>
    <SingleSelectConnected
        id={`${this.props.expressionEditorId}-${BOOLEAN_OPRATOR_SELECT_ID}`}
        items={BOOLEAN_OPERATOR_ITEMS}
        placeholder='Add rule'
        disabled={!this.props.isExpressionComplete}
        onSelectOptionCallback={(selectedBooleanOperator) => this.props.onBooleanOperatorSelect(selectedBooleanOperator)}
    />;  
```

src/components/queryExpressionEditor/buildFromResult/BuildFromResult.tsx
MichelML: even promises will break unless `ES2018.Promise` is added in `lib` in the tsconfig. The quickest path to replace this would be to use jQuery which we try to move away from. I'm still not sure what is the best solution here, I think it will depend on whether we decide to keep this component in react-vapor or not also.

MichelML: in addition, the way we support asynchronous stuff for state management is usually through redux-thunk https://www.npmjs.com/package/redux-thunk , however that does not play well with react internal state , still continuing to think about possible options

src/components/queryExpressionEditor/buildFromResult/BuildFromResult.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+    private async onSearch(searchBarText: string) {
+        this.setState({searching: true});
+        await this.updateResults(searchBarText);
+        this.setState({searching: false});
+    }
+
+    private async updateResults(searchBarText: string) {
+        const results = await this.props.queryTrigger.getResultsWithBasicExpression(searchBarText);
+        this.setState({results: results});
```
MichelML: use short syntax when you can, this.setState({results});

src/components/queryExpressionEditor/buildFromResult/BuildFromResult.tsx
```diff
@@ -0,0 +1,113 @@
...
+        const parsedFieldExpression: string = this.getParsedFieldExpression(result);
+        this.props.updateQueryExpression(parsedFieldExpression);
+    }
+
+    private getParsedFieldExpression(result: IResult) {
+        const fieldExpression = this.getFieldExpression(result);
+        if (!fieldExpression) {
+            return null;
+        }
+        return this.buildFieldExpression(fieldExpression);
```
MichelML: use the ternary operator whenever you have a dual logic like this:  

```
return !fieldExpression 
    ? null
    : this.buildFieldExpression(fieldExpression);  
```

src/components/queryExpressionEditor/buildFromResult/BuildFromResult.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+        return null;
+    }
+
+    private buildFieldExpression(fieldExpression: IFieldExpression): string {
+        if (fieldExpression.field && fieldExpression.fieldValue) {
+            return `${fieldExpression.field}=${fieldExpression.fieldValue}`;
+        } else {
+            return null;
+        }
```
MichelML: ternary here

src/components/queryExpressionEditor/examples/QueryExpressionEditorExample.tsx
```diff
@@ -0,0 +1,16 @@
...
+import {QueryExpressionEditor} from '../QueryExpressionEditor';
+import * as styles from './TempLayout.scss';
+
+export class QueryExpressionEditorExample extends React.Component<any, any> {
+
+    render() {
+        return (
+            <div className='mt2'>
+                <div className={`form-group ${styles.containerTest}`}>
+                    <QueryExpressionEditor accessToken={'xx564559b1-0045-48e1-953c-3addd1ee4457'} organizationId={'searchuisamples'} />
```
GermainBergeron: This is a fake token I hope

src/components/queryExpressionEditor/expressionEditor/ExpressionEditorConnected.tsx
```diff
@@ -0,0 +1,45 @@
...
+// Note : Duplication of code with ValueSelectStringConnected.tsx
+function getSelectedOption(options: IDropdownOption[]): string {
+    const selectedOption: IDropdownOption = _.findWhere(options, {selected: true});
+    return selectedOption.value;
+}
+
+const mapStateToProps = (state: IReactVaporState, ownProps: IExpressionEditorOwnProps): IExpressionEditorStateProps => {
+    const expressionEditors: IExpressionEditorState[] = state.expressionEditors && state.expressionEditors;
+    const expressionEditor: IExpressionEditorState = _.findWhere(state.expressionEditors, {id: ownProps.id});
+    const fieldSelect: IDropdownSearchState = _.findWhere(state.dropdownSearch, {id: `${ownProps.id}-${FIELD_SELECT_ID}`});
```
MichelML: you should have a method that retrieve child ids for fieldSelect and operatorSelect, we a caught a lot of bugs in the past because we were writing ids composition by hand like this. This is error prone. You can check `TableUtils` for examples.

src/components/queryExpressionEditor/expressionEditor/ExpressionEditorConnected.tsx
```diff
@@ -0,0 +1,45 @@
...
+import {IListBoxState} from '../../listBox/ListBoxReducers';
+import {FIELD_SELECT_ID} from '../fieldSelect/FieldSelect';
+import {OPERATOR_SELECT_ID} from '../operatorSelect/OperatorSelect';
+import {ExpressionEditor, IExpressionEditorDispatchProps, IExpressionEditorOwnProps, IExpressionEditorProps, IExpressionEditorStateProps} from './ExpressionEditor';
+import {removeExpressionEditor, updateExpressionEditor} from './ExpressionEditorActions';
+import {IExpressionEditorState} from './ExpressionEditorReducers';
+
+// TODO QUESTION R-V :
+// Is this the only way to get the selected option from a DropdownSearchConnected selector?
+// Note : Duplication of code with ValueSelectStringConnected.tsx
```
MichelML: we have `getDropdownSearchSelectedOption` in DropdownSearchUtils but that does not do exactly this.   

you could make a generic util in DropdownSearchUtils with getSelectedOption that would be useful eslewhere for sure. I would make it so that if there is no selected value that would return undefined.  

```
const getSelectedOptions = (options: IDropdownOptions[] = []): string => options.filter((option) => option.selected)[0];  
```

src/components/queryExpressionEditor/formMode/FormMode.tsx
```diff
@@ -0,0 +1,142 @@
...
+        this.expressionEditorNewId++;
+    }
+
+    private deleteExpressionEditor(id: string) {
+        this.removeExpressionEditor(id);
+        this.ensureLastEditorCanAddRule();
+    }
+
+    private removeExpressionEditor(id: string) {
+        const updatedExpressionEditors = this.state.expressionEditors;
```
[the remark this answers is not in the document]
MichelML: underscore reject is what you want here, it returns a new function. the method should just be:

```
return _.reject(this.state.expressionEditors, (editor) => editor.id === id)  
```  
https://underscorejs.org/#reject

src/components/queryExpressionEditor/formMode/FormMode.tsx
```diff
@@ -0,0 +1,142 @@
...
+                key={`${EXPRESSION_EDITOR_ID}-${this.expressionEditorNewId}`}
+                id={`${EXPRESSION_EDITOR_ID}-${this.expressionEditorNewId}`}
+                fields={this.state.fields}
+                queryTrigger={this.props.queryTrigger}
+                addExpressionEditor={() => this.addExpressionEditor()}
+                deleteExpressionEditor={(id: string) => this.deleteExpressionEditor(id)}
+            />,
+        );
+        this.setState({expressionEditors: newSate});
+        this.expressionEditorNewId++;
```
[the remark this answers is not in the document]
MichelML: i would have a getNextExpressionEditor method, and since the next expressionEditor id is related to this.state.expressionEditors.length , i wouldn't bother managing this reference and simply remove it.   
  
```
    private addExpressionEditor() {
        this.setState({expressionEditors: [...this.state.expressionEditors, this.getNextExpressionEditor()]});
    }

    private getNextExpressionEditor() {
        return (
            <ExpressionEditorConnected
                key={`${EXPRESSION_EDITOR_ID}-${this.state.expressionEditors.length}`}
                id={`${EXPRESSION_EDITOR_ID}-${this.state.expressionEditors.length}`}
                fields={this.state.fields}
                queryTrigger={this.props.queryTrigger}
                addExpressionEditor={() => this.addExpressionEditor()}
                deleteExpressionEditor={(id: string) => this.deleteExpressionEditor(id)}
            />
        );
    }  
```

src/components/queryExpressionEditor/formMode/FormMode.tsx
```diff
@@ -0,0 +1,142 @@
...
+            return editor.props.id === id;
+        });
+        if (index !== -1) {
+            updatedExpressionEditors.splice(index, 1);
+        }
+        this.setState({expressionEditors: updatedExpressionEditors});
+    }
+
+    private ensureLastEditorCanAddRule() {
+        const lastEditorId = this.props.expressionEditorsState[this.props.expressionEditorsState.length - 2].id;
```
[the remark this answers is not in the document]
MichelML: why not length - 1 here ? not sure I can the intention

if you really want the last element of the list you can also do `_.last(this.props.expressionEditorsState)`

src/components/queryExpressionEditor/formMode/FormModeConnected.tsx
```diff
@@ -0,0 +1,17 @@
+import {connect} from 'react-redux';
+import {IReactVaporState, IReduxActionsPayload} from '../../../ReactVapor';
+import {IReduxAction, ReduxUtils} from '../../../utils/ReduxUtils';
+import {selectListBoxOption} from '../../listBox/ListBoxActions';
+import {FormMode, IFormModeDispatchProps, IFormModeProps, IFormModeStateProps} from './FormMode';
+
+const mapStateToProps = (state: IReactVaporState): IFormModeStateProps => {
+    return {
+        expressionEditorsState: state.expressionEditors && state.expressionEditors,
```
MichelML: i don't get it, this is twice the same expression no? why not just `state.expressionEditors` ?

src/components/queryExpressionEditor/fieldSelect/FieldSelect.tsx
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    render() {
+        return (
+            <span className={`${styles.container}`}>
+                {/*
+                // TODO QUESTION R-V : BUG?
+                // When loading more values in the infinte scroll,
+                // The selected value resets to 'Select an option'
+                // Why does the value gets reset?
```
MichelML: @sroy3 any idea on top of your head?

src/components/queryExpressionEditor/valueSelectString/ValueSelectStringConnected.tsx
```diff
@@ -0,0 +1,31 @@
...
+import * as _ from 'underscore';
+import {IReactVaporState, IReduxActionsPayload} from '../../../ReactVapor';
+import {IReduxAction, ReduxUtils} from '../../../utils/ReduxUtils';
+import {IDropdownOption} from '../../dropdownSearch/DropdownSearch';
+import {IDropdownSearchState} from '../../dropdownSearch/DropdownSearchReducers';
+import {clearListBoxOption} from '../../listBox/ListBoxActions';
+import {FIELD_SELECT_ID} from '../fieldSelect/FieldSelect';
+import {IValueSelectStringDispatchProps, IValueSelectStringOwnProps, IValueSelectStringProps, IValueSelectStringStateProps, ValueSelectString} from './ValueSelectString';
+
+// TODO QUESTION R-V :
```
MichelML: see https://github.com/coveo/react-vapor/pull/777/files/293da03659567567970674e809696bd4a4a28e73#r211723764

src/components/queryExpressionEditor/fieldSelect/FieldSelect.tsx
sroy3: Not quite sure why this would happen, but our `DropdownSearch` is very unstable and does not handle change too well. In theory this component is even deprecated. We added the infinite scroll on it because the org picker was already using the component, but you could use another dropdown with the infinite scroll that would be a lot more stable.

MichelML: the issue was disccues with Jé

PR #778, opened 2018-08-20
add single data entry actions for table

MichelML:
demo: https://coveo.github.io/react-vapor/single-table-data-entry-actions/ go in `Table` and try the Complex table (look into rows actions for create/delete/update)

### DIFF
--- a/src/components/tables/Table.tsx
+++ b/src/components/tables/Table.tsx
@@ -168,9 +168,17 @@ export class Table extends React.Component<ITableProps> {
         const {tableCompositeState} = this.props;
 
         if (this.hasTableCompositeStateChanged(tableCompositeState, nextProps.tableCompositeState)) {
-            // if the change occurs outside the navigation (per page, pagination), reset the pagination to 0
-            const shouldResetPage = tableCompositeState.page === nextProps.tableCompositeState.page
-                && tableCompositeState.perPage === nextProps.tableCompositeState.perPage;
+            const shouldResetPage =
+                // if the change occurs outside the navigation (per page, pagination), reset the pagination to 0
+                (
+                    tableCompositeState.page === nextProps.tableCompositeState.page
+                    && tableCompositeState.perPage === nextProps.tableCompositeState.perPage
+                )
+                /**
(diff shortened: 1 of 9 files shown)


src/components/tables/Table.tsx
```diff
@@ -168,9 +168,17 @@ export class Table extends React.Component<ITableProps> {
         const {tableCompositeState} = this.props;
 
         if (this.hasTableCompositeStateChanged(tableCompositeState, nextProps.tableCompositeState)) {
-            // if the change occurs outside the navigation (per page, pagination), reset the pagination to 0
-            const shouldResetPage = tableCompositeState.page === nextProps.tableCompositeState.page
-                && tableCompositeState.perPage === nextProps.tableCompositeState.perPage;
+            const shouldResetPage =
```
sroy3: What does this do exactly / how does it work? I feel like it would be annoying to reset paging for every case where we add an entry. Maybe have some prop to toggle it would be nice.

MichelML: I think you're right the logic needs to be reworked and more flexible, you can try the example and give me some feedback if you'd like

MichelML:
@sroy3 the behavior is now simpler and better, see `Complex Table in default mode`  

We check for deletion in `hasTableCompositeStateChanged` but not for addition (to avoid trigering sort and filter logic on newly added data), and it produces the good behavior in all cases.

src/components/tables/Table.tsx
```diff
@@ -247,6 +249,7 @@ export class Table extends React.Component<ITableProps> {
             )
             || currentTableCompositeState.from !== nextTableCompositeState.from
             || currentTableCompositeState.to !== nextTableCompositeState.to
+            || currentTableCompositeState.data.allIds.length > nextTableCompositeState.data.allIds.length
```
GermainBergeron: Should probably be `|| currentTableCompositeState.data.allIds.length !== nextTableCompositeState.data.allIds.length`

MichelML: the goal is to check for deletion only and let addition to avoid trigering sort and filter logic on newly added data

PR #779, opened 2018-08-20
Slidey duration

sroy3:


GermainBergeron:
I find it weird that the coverage decreases, any clue where it comes from?

(a library update maybe?)

sroy3:
I'll look into the coverage decrease. I find it weird also.

PR #790, opened 2018-08-28
Status card

sroy3:
https://github.com/coveo/vapor/pull/580 needed for a successful build.

### DIFF
--- a/src/components/statusCard/styles/StatusCard.scss
+++ b/src/components/statusCard/styles/StatusCard.scss
@@ -1,5 +1,4 @@
 @import '~coveo-styleguide/scss/common/palette.scss';
-@import '~coveo-styleguide/scss/common/palette-map.scss';
 @import '~coveo-styleguide/scss/variables.scss';
 
 .statusCard {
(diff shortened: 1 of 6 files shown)


src/components/statusCard/styles/StatusCard.scss
```diff
@@ -0,0 +1,54 @@
...
+    background: $pure-white;
+
+    &:global(.simple) {
+        background: transparent;
+        box-shadow: none;
+    }
+}
+
+@each $color, $value in $palette {
+    &:global(.border-color-#{$color}) {
```
GermainBergeron: This should probably go in https://github.com/coveo/vapor/blob/b952495364c87f5e057de405c7d8f5c7666cb8eb/scss/utility/border.scss

src/components/statusCard/styles/StatusCard.scss
```diff
@@ -0,0 +1,56 @@
+@import '~coveo-styleguide/scss/common/palette.scss';
```
pochretien: can we create an alias with this 3 imports ? if possible to do that. I think we will use these imports in many files

sroy3:
Demo: https://coveo.github.io/react-vapor/status-card/index.html

src/components/statusCard/styles/StatusCard.scss
sroy3: I don't think this is possible. I didn't need the map anymore and we don't always want to import both other files.

Issue #789, opened 2018-08-28
Codemirror dependency should also be in peerDependencies


huppeje:
```

"peerDependencies": {
    "codemirror": "^5.38.0",
     ...

```

MichelML:
You're right, can you make a PR with explanation for this issue? we also plan on making some changes regarding the way we bundle dependencies in the project, but it will be in a mid term future.

PR #795, opened 2018-08-30
Codemirror dependency should also be in peerDependencies

huppeje:


MichelML:
relates to https://github.com/coveo/react-vapor/issues/789

PR #804, opened 2018-09-05
Side-Navigation broken

sbelzile-Coveo:
New class `navigation-opened` was added to vapor, which means that the side-navigation in Dynamics is hidden by default.

- I added a property to control its state (`opened`)
- and used the `withReduxState` to control it and have it opened by default (not sure this is the best idea though)


MichelML:
i think it's fine and it won't affect anything on our side, fine by me and sorry for the breaking changes, my fault.

PR #806, opened 2018-09-06
Allow to specify table header classes

francoislg:
There was no way to add classes to specify attributes in the **header cells** like it is possible to do with row cells, so here it is.

Plus, added a way to specify a class directly on the **thead** too.

### DIFF
--- a/src/components/tables/table-children/TableChildHeader.tsx
+++ b/src/components/tables/table-children/TableChildHeader.tsx
@@ -19,7 +19,8 @@ export const TableChildHeader = (props: ITableProps): JSX.Element => {
         return {id, title, className, withFixedHeader: props.withFixedHeader, ...tableSortInformation} as ITableHeaderCellOwnProps;
     });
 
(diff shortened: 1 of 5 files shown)


src/components/tables/table-children/TableChildHeader.tsx
```diff
@@ -11,14 +11,15 @@ export const TableChildHeader = (props: ITableProps): JSX.Element => {
...
         const tableSortInformation = !!headingAttribute.sort
             ? {tableId: props.id, attributeToSort: headingAttribute.attributeName}
             : {};
 
-        return {id, title, withFixedHeader: props.withFixedHeader, ...tableSortInformation};
+        return {id, title, className, withFixedHeader: props.withFixedHeader, ...tableSortInformation} as ITableHeaderCellOwnProps;
     });
 
-    const headerClass = classNames(
+    const headerClass = classNames(props.tableHeaderClasses,
```
sroy3: Is this a string? It should also be on the next line.

PR #807, opened 2018-09-06
RadioSelectConnected

MichelML:
https://coveo.github.io/react-vapor/radioselect-connected/  

look for the `Radio` section last example 

the disabled option is still selectable, on it to fix it

### DIFF
--- a/docs/Index.tsx
+++ b/docs/Index.tsx
@@ -220,7 +220,9 @@ class App extends React.Component<any, any> {
                         {component: BannerExamples, componentName: 'Banner'},
                         {component: SlideYExamples, componentName: 'SlideY'},
                         {component: StatusCardExamples, componentName: 'StatusCard'},
-                    ].map((component) => <ExampleWrapper key={component.componentName} componentName={component.componentName} component={component.component} />)}
+                    ]
+                    .sort((componentA, componentB) => componentA.componentName.toLowerCase().localeCompare(componentB.componentName.toLowerCase()))
+                    .map((component) => <ExampleWrapper key={component.componentName} componentName={component.componentName} component={component.component} />)}
                 </div>
             </Provider>
         );
--- a/docs/Index.tsx
+++ b/docs/Index.tsx
@@ -106,7 +106,13 @@ import {ReactVaporStore} from './ReactVaporStore';
 interface ExampleWrapperState {
     shown: boolean;
 }
-class ExampleWrapper extends React.Component<{component: any, componentName: string}, ExampleWrapperState> {
+
+interface ExampleProps {
+    component: any;
+    componentName: string;
+}
+
+class ExampleWrapper extends React.Component<ExampleProps, ExampleWrapperState> {
     state: ExampleWrapperState = {shown: false};
 
     render() {
@@ -120,6 +126,9 @@ class ExampleWrapper extends React.Component<{component: any, componentName: str
         );
     }
 }
+
+const sortComponentsByName = (a: ExampleProps, b: ExampleProps) => a.componentName.toLowerCase().localeCompare(b.componentName.toLowerCase());
+const formatComponentsExamples = (example: ExampleProps) => <ExampleWrapper key={example.componentName} componentName={example.componentName} component={example.component} />;
 
 class App extends React.Component<any, any> {
     render() {
@@ -221,8 +230,8 @@ class App extends React.Component<any, any> {
                         {component: SlideYExamples, componentName: 'SlideY'},
                         {component: StatusCardExamples, componentName: 'StatusCard'},
                     ]
-                        .sort((componentA, componentB) => componentA.componentName.toLowerCase().localeCompare(componentB.componentName.toLowerCase()))
-                        .map((component) => <ExampleWrapper key={component.componentName} componentName={component.componentName} component={component.component} />)}
+                        .sort(sortComponentsByName)
+                        .map(formatComponentsExamples)}
                 </div>
             </Provider>
         );
--- a/docs/Index.tsx
+++ b/docs/Index.tsx
@@ -221,8 +221,8 @@ class App extends React.Component<any, any> {
                         {component: SlideYExamples, componentName: 'SlideY'},
                         {component: StatusCardExamples, componentName: 'StatusCard'},
                     ]
-                    .sort((componentA, componentB) => componentA.componentName.toLowerCase().localeCompare(componentB.componentName.toLowerCase()))
-                    .map((component) => <ExampleWrapper key={component.componentName} componentName={component.componentName} component={component.component} />)}
+                        .sort((componentA, componentB) => componentA.componentName.toLowerCase().localeCompare(componentB.componentName.toLowerCase()))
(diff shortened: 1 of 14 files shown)


MichelML:
the coverage is not up to date and is now back to being equal as before this PR

docs/Index.tsx
```diff
@@ -220,7 +220,9 @@ class App extends React.Component<any, any> {
                         {component: BannerExamples, componentName: 'Banner'},
                         {component: SlideYExamples, componentName: 'SlideY'},
                         {component: StatusCardExamples, componentName: 'StatusCard'},
-                    ].map((component) => <ExampleWrapper key={component.componentName} componentName={component.componentName} component={component.component} />)}
+                    ]
+                        .sort((componentA, componentB) => componentA.componentName.toLowerCase().localeCompare(componentB.componentName.toLowerCase()))
```
sroy3: I was thinking about the same thing yesterday. Good job!!

src/components/radio/RadioSelect.tsx
```diff
@@ -1,19 +1,49 @@
...
 
 export interface IRadioSelectProps {
+    id?: string;
     name?: string;
-    value?: string;
     disabled?: boolean;
-    onChange?: (value: string) => void;
     children?: Array<React.ReactElement<Radio>> | Array<React.ReactElement<ToggleForm>>;
+    /**
+     * used in RadioSelectConnected component only
```
sroy3: Same for `id` usually. Maybe a single line comment would be enough here.

docs/Index.tsx
francoislg: Wrapping this into a `const caseInsensitiveCompare = (a, b) => a.toLowerCase().localeCompare(b.toLowerCase())` would help a little bit on the readability:

`.sort((componentA, componentB) => caseInsensitiveCompare(componentA.componentName, componentB.componentName))`

src/components/radio/RadioSelect.tsx
```diff
@@ -1,19 +1,49 @@
...
+    value?: string;
+    disabledValues?: string[];
 }
 
-export class RadioSelect extends React.Component<IRadioSelectProps, any> {
-    private handleToggle(value: string) {
+export interface IRadioSelectAllProps extends IRadioSelectProps, IRadioSelectDispatchProps, IRadioSelectStateProps {}
+
+export class RadioSelect extends React.Component<IRadioSelectAllProps, any> {
+    private handleToggle(value: string, e: React.MouseEvent<HTMLElement>) {
```
gdostie: Put the private method after the public ones

src/components/radio/RadioSelect.tsx
```diff
@@ -1,19 +1,49 @@
...
         if (this.props.onChange) {
-            this.props.onChange(value);
+            this.props.onChange(value, this.props.id, e);
+        }
```
gdostie: Very facultative suggestion for the 3 `if` in this file:
`callIfDefined(this.props.onChange, value, this.props.id, e);`

src/components/radio/RadioSelect.tsx
francoislg: The comment could probably be removed, why not let the user decide if they need this attribute? :)

If you don't want to remove it, is this strictly the only use case where this attribute is necessary? If so, I would split the prop into a different interface so that standard props don't include it, like `IRadioSelectConnectedProps`.

src/components/radio/RadioSelectActions.ts
```diff
@@ -0,0 +1,27 @@
+import {IReduxAction} from '../../utils/ReduxUtils';
+
+export interface IRadioSelectActionPayload {
+    id: string;
+    value?: string;
+    disabledValues?: string[];
+}
+
+export const RadioSelectActions = {
```
gdostie: Suggestion: rename this const `RadioSelectActionTypes`

Then instead of exporting the actions individually, export the actions at the end of the file like the following:
```ts
export const RadioSelectActions = {
    set: setRadioSelect,
    remove: removeRadioSelect,
};
```
or even better rename the actions so you can do this `export const RadioSelectActions = {set, remove};`.

I know this is not how we typically export actions in this repo, and this is only a suggestion to avoid exporting too much things. I have a personnal preference for this, because it allows to reuse the same action names accross components.

src/components/radio/RadioSelectReducers.ts
```diff
@@ -0,0 +1,29 @@
...
+import {IRadioSelectActionPayload, RadioSelectActions} from './RadioSelectActions';
+
+export interface IRadioSelectState {
+    id: string;
+    value: string;
+    disabledValues: string[];
+}
+
+export const radioSelectInitialState: IRadioSelectState = {id: undefined, value: undefined, disabledValues: undefined};
+export const radioSelectsInitialState: IRadioSelectState[] = [];
```
gdostie: Any reason to use an array instead of an object ?

src/components/radio/RadioSelect.tsx
```diff
@@ -1,19 +1,49 @@
...
+}
+
+export interface IRadioSelectDispatchProps {
+    onMount?: (id: string, valueOnMount: string, disabledValues: string[]) => void;
+    onUnmount?: (id: string) => void;
+    onChange?: (value: string, id?: string, e?: React.MouseEvent<HTMLElement>) => void;
+}
+
+export interface IRadioSelectStateProps {
+    value?: string;
```
francoislg: Changing this here is kind of a breaking change.

Is there a way to keep `value` in the `IRadioSelectProps` and tag it as deprecated and comment on it like `use IRadioSelectStateProps or IRadioSelectAllProps instead.`?

People that upgrade would get a direction as to what to use instead of just being presented with a compilation error

src/components/radio/RadioSelectActions.ts
MichelML: you should suggest this in a jsadmin update and see if the group agrees with that, i'll leave it as it is for now

src/components/radio/RadioSelectReducers.ts
MichelML: this should also be a group decision, it can get messy if we start to go toward the Map strategy instead of array,

src/components/radio/RadioSelect.tsx
MichelML: made some changes that now makes it not-breaking

src/components/radio/tests/RadioSelectReducers.spec.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+            oldState = radioSelectsState;
+            action.payload.id = 'ialsodontexistinthestateyet';
+            radioSelectsState = radioSelectsReducer(oldState, action);
+
+            expect(radioSelectsState.length).toBe(oldState.length + 1);
+            expect(radioSelectsState.filter((radioSelect) => radioSelect.id === action.payload.id).length).toBe(1);
+        });
+
+        describe('"RadioSelectAction.set" when id is arleady in the state', () => {
```
GermainBergeron: already*

PR #817, opened 2018-09-11
Radio option disabled tooltip

MichelML:
need this to implement 
screen shot 2018-09-11 at 2 35 54 pm

### DIFF
--- a/src/components/input/Input.tsx
+++ b/src/components/input/Input.tsx
@@ -2,6 +2,8 @@ import * as classNames from 'classnames';
 import * as React from 'react';
 import {contains, isUndefined} from 'underscore';
 import {IClassName} from '../../utils/ClassNameUtils';
+import {TooltipPlacement} from '../../utils/TooltipUtils';
+import {Tooltip} from '../tooltip/Tooltip';
 import {ILabelProps, Label} from './Label';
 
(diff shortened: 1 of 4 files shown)


gdostie:
code lgtm!
Shouldn't there be a demo link automatically posted here?

src/components/input/Input.tsx
```diff
@@ -154,29 +157,37 @@ export class Input extends React.Component<IInputProps, IInputState> {
...
+                onChange={() => this.handleChange()}
+                onKeyUp={(event: React.KeyboardEvent<HTMLInputElement>) => this.handleKeyUp(event)}
+                placeholder={this.props.placeholder}
+                checked={!!this.props.checked}
+                disabled={!!this.props.disabled}
+                name={this.props.name}
+                required
+                readOnly={!!this.props.readOnly}
+                autoFocus={!!this.props.autoFocus}
+                step={this.props.type === 'number' ? 'any' : null}
```
sroy3: You are going to need a key on the input, the label and children.

MichelML: since i am destructuring the array i don't think i need that unless i'm wrong

MichelML:
@gdostie since i am not pointing to master i think this is ignored, so is the coverage

PR #821, opened 2018-09-13
popover connected

MichelML:
will need this for a use case  

screen shot 2018-09-13 at 2 26 08 pm


MichelML:
@gdostie this is just an example, you can popover any component or element you want

PR #825, opened 2018-09-21
type as any to be able to use jqueryxhr returned by thunk actions

MichelML:


src/utils/ReduxUtils.ts
```diff
@@ -2,8 +2,8 @@ import * as ReactRedux from 'react-redux';
 import * as Redux from 'redux';
 import {extend} from 'underscore';
 
-export type IDispatch = (action: IReduxAction<any> | IThunkAction) => void;
-export type IThunkAction = (dispatch: IDispatch, getState?: () => any) => void;
+export type IDispatch = (action: IReduxAction<any> | IThunkAction) => any;
+export type IThunkAction = (dispatch: IDispatch, getState?: () => any) => any;
```
GermainBergeron: I think it would be more flexible to do: 
```typescript
export type IDispatch<T = void> = (action: IReduxAction<any> | IThunkAction) => T;
export type IThunkAction<T = void> = (dispatch: IDispatch, getState?: () => any) => T;
````

MichelML:
@GermainBergeron I'll have to use the same call for ml midel crration in a "add" context and in a "add with json" context, depending on the context, I'll want to perform different things after the call failed or succeeded, so creating a generic action returning the jqueryxhr makes sense here I think. redux-thunk provides examples of that on npm if you want to look at it

PR #831, opened 2018-09-26
Fix INavigationChildrenProps's onPerPageClick

tommy-tran:
Changes:
- Change to onPerPageClick definition in Navigation and NavigationPerPage interfaces
- Added test for proper arguments being used in onPerPageClick
- Alert the value of onPerPageClick and onPageClick in component examples

### DIFF
--- a/src/components/navigation/Navigation.tsx
+++ b/src/components/navigation/Navigation.tsx
@@ -21,7 +21,7 @@ export interface INavigationChildrenProps {
     onPageClick?: (pageNb: number) => void;
     perPageLabel?: string;
     perPageNumbers?: number[];
-    onPerPageClick?: () => void;
+    onPerPageClick?: (newPerPage: number, currentPerPage: number, currentPage: number) => void;
     hidePages?: boolean;
     currentPerPage?: number;
     currentPage?: number;
--- a/src/components/navigation/Navigation.tsx
+++ b/src/components/navigation/Navigation.tsx
@@ -21,7 +21,7 @@ export interface INavigationChildrenProps {
     onPageClick?: (pageNb: number) => void;
     perPageLabel?: string;
     perPageNumbers?: number[];
-    onPerPageClick?: (newPerPage: number, currentPerPage: number, currentPage: number) => void;
+    onPerPageClick?: (newPerPage: number, currentPerPage?: number, currentPage?: number) => void;
     hidePages?: boolean;
     currentPerPage?: number;
     currentPage?: number;
--- a/src/components/navigation/tests/Navigation.spec.tsx
+++ b/src/components/navigation/tests/Navigation.spec.tsx
@@ -6,6 +6,7 @@ import {Loading} from '../../loading/Loading';
 import {INavigationProps, Navigation} from '../Navigation';
 import {NavigationPagination} from '../pagination/NavigationPagination';
 import {NavigationPerPage, PER_PAGE_NUMBERS} from '../perPage/NavigationPerPage';
+import {NavigationPerPageSelect} from '../perPage/NavigationPerPageSelect';
 
 describe(' navigation', () => {
     const basicNavigationProps: INavigationProps = {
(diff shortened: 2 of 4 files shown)


GermainBergeron:
It would be nice to add a test in `src/components/navigation/Navigation.spec.tsx` to validate that the parameters are correctly sent when the user change the "perPage"

src/components/navigation/Navigation.tsx
```diff
@@ -21,7 +21,7 @@ export interface INavigationChildrenProps {
     onPageClick?: (pageNb: number) => void;
     perPageLabel?: string;
     perPageNumbers?: number[];
-    onPerPageClick?: () => void;
+    onPerPageClick?: (newPerPage: number, currentPerPage: number, currentPage: number) => void;
```
sroy3: I think the parameters should be optional. The user might not need them and might send a function without these parameters.

GermainBergeron: I think you're confused, the onPerPageClick is optional, but here we define what we send to the function. If the "user" don't want to use them they can simply omit them in the receiver

[Example](http://www.typescriptlang.org/play/#src=interface%20SomeProps%20%7B%0D%0A%20%20%20%20f%3F%3A%20(a%3A%20string%2C%20b%3A%20number)%20%3D%3E%20any%3B%0D%0A%7D%0D%0A%0D%0Aconst%20One%3A%20SomeProps%20%3D%20%7B%7D%3B%0D%0A%0D%0Aconst%20Two%3A%20SomeProps%20%3D%20%7B%0D%0A%20%20%20%20f%3A%20()%20%3D%3E%20null%2C%0D%0A%7D%3B%0D%0A%0D%0Aconst%20Three%3A%20SomeProps%20%3D%20%7B%0D%0A%20%20%20%20f%3A%20(a%3A%20string)%20%3D%3E%20a%2C%0D%0A%7D%0D%0A%0D%0Aconst%20Four%3A%20SomeProps%20%3D%20%7B%0D%0A%20%20%20%20f%3A%20(a%3A%20string%2C%20b%3A%20number)%20%3D%3E%20b%2C%0D%0A%7D)

src/components/navigation/tests/Navigation.spec.tsx
```diff
@@ -82,5 +83,14 @@ describe(' navigation', () => {
...
             expect(navigation.find(NavigationPerPage).props().currentPerPage).toBe(expectedPerPage);
         });
+
+        it('should call onPerPageClick prop with the correct values when it is set', () => {
+            const onPerPageClick = jasmine.createSpy('mockOnPerPageClick');
+            const perPageNumbers: number[] = [2, 3, 4];
+            const currentPerPage: number = 3;
+            const newNavigationProps: INavigationProps = _.extend({}, basicNavigationProps, {currentPerPage, perPageNumbers, onPerPageClick});
+            navigation.setProps(newNavigationProps).update().find(NavigationPerPageSelect).last().simulate('click');
+            expect(onPerPageClick).toHaveBeenCalledWith(4, 3, undefined); // newPerPage, currentPerPage, currentPage (currentPerPage and currentPage are undefined unless passed as props)
```
GermainBergeron: Test looks god, I'd change the expect a bit:

```typescript
const currentPerPage = perPageNumbers[1];
const expectedPerPage = perPageNumbers[perPageNumber.length - 1];
...
expect(onPerPageClick).toHaveBeenCalledWith(expectedPerPage, currentPerPage, undefined);
```
This way if you add/modify the perPageNumbers const the test will stay green

src/components/navigation/Navigation.tsx
MichelML: I thought the same as @sroy3 but it seems you're right according to your example

tommy-tran: After looking at it again, I think that currentPerPage and currentPage are optional, and the only required one would be newPerPage. The other arguments are only set if Navigation receives currentPerPage and currentPage as props (in the current implementation).

PR #834, opened 2018-09-27
Updated jasmine and karma

sroy3:
There were a few tests with side effects and there might still be. Try to keep an eye out for this.


package.json
```diff
@@ -116,10 +116,10 @@
     "gulp-util": "3.0.8",
     "imports-loader": "0.8.0",
     "istanbul-instrumenter-loader": "3.0.1",
-    "jasmine-core": "2.8.0",
+    "jasmine-core": "3.2.1",
```
GermainBergeron: Can you also update the `@types/jasmine`?

sroy3: Currently doing this

PR #838, opened 2018-10-01
Update listbox list action

pochretien:
Add an action to update items in the store
Add items in the store for the listBox

### DIFF
--- a/src/components/listBox/ListBoxConnected.tsx
+++ b/src/components/listBox/ListBoxConnected.tsx
@@ -15,6 +15,7 @@ const mapStateToProps = (state: IReactVaporState, ownProps: IListBoxOwnProps): I
     return {
         selected: list ? list.selected : [],
         active: list ? list.active : undefined,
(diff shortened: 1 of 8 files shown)


src/components/listBox/ListBoxConnected.tsx
```diff
@@ -15,6 +15,7 @@ const mapStateToProps = (state: IReactVaporState, ownProps: IListBoxOwnProps): I
     return {
         selected: list ? list.selected : [],
         active: list ? list.active : undefined,
+        items: list ? list.items : ownProps.items,
```
pochretien: I can use the same value from the store and the ownProps.

With this change, the connected and not connected can work in the same way with the same interface.

If you have more question, you can DM me for more information about this change and the mergePropsWithStatePropsPriority.

sroy3: I really like this. It makes things more seamless when the same prop may come from one place or the other.

PR #841, opened 2018-10-01
Make collapsible container transition twice as fast

sroy3:


MichelML:
@gdostie is right changing the timeout doesn't change the animation speed, I remember having tried that in the past

gdostie:
Now it is fast! More than twice as fast because the previous duration value (200 ms) wasn't really set properly. I guess the previous duration value was 300ms so twice as fast would be 150ms, not 100

sroy3:
Other demo to compare with: https://coveo.github.io/react-vapor/update-listbox-list-action/

PR #842, opened 2018-10-02
Created a dependencies file with all externals bundle

sroy3:


webpack.config.prod.js
```diff
@@ -6,12 +6,14 @@ const isTravis = process.env.TRAVIS;
  * Config file for the packaged library
  */
 const config = {
-    entry: './Index.ts',
+    entry: {
+        'react-vapor': ['./Index.ts'],
+        'react-vapor.dependencies': ['jquery', 'underscore', 'react', 'react-dom', 'react-redux', 'redux', 'codemirror'],
```
samisayegh: Hmm, maybe there were further discussions that I am not aware of. When I created the jira, it was with the goal of avoiding the need to manually add certain dependencies using e.g. `webpack.ProvidePlugin`. Specifically I was questioning why underscore should be marked as an external? Why not package it OOTB?

On the other hand, with dependencies like react, react-dom, it makes more sense to have those be externals (i.e. if you want to use react-vapor, you need to first download react -> makes sense to me).

I am not sure if this change solves the initial issue. If we include the file `react-vapour.dependencies`, we will be including react twice in order to get access to underscore? I would prefer to not include this file and just expose underscore globally in this case as we are currently doing.

So in summary, my question is if we could remove underscore as an external? and if not, why not?

sroy3: We already install underscore and those other dependencies in our projects and react-vapor bundle is big enough as it is. We do not want to include dependencies twice or more. We even plan on having more external dependencies. I could maybe remove everything react related from the react-vapor.dependencies?

MichelML:
@sroy3 did you test if something breaks in our projects with that change?

sroy3:
Nothing should have changed, but I'll try it to make sure.

webpack.config.prod.js
samisayegh: Ok, if there is absolutely no more room to add to the react-vapor bundle, then I would say the best would be to keep it like it is now, including all the dependencies in one bundle. The way I see the bundle is as an alternative to installing the externals one-by-one, and should include all of them. It also has the benefit that the RV team can maintain it, so as new external dependencies are added, project consumers will not need to install anything new.

MichelML:
yeah no sorry just noticed you didn't make any changes to the externals actually, you just made them available as a bundled file

PR #843, opened 2018-10-02
Filter box match filter props

pochretien:
Review only the code!

Its part of a big change for the MultiSelect

### DIFF
--- a/src/components/filterBox/FilterBox.tsx
+++ b/src/components/filterBox/FilterBox.tsx
@@ -1,6 +1,7 @@
 import * as classNames from 'classnames';
 import * as React from 'react';
 import {Svg} from '../svg/Svg';
+import {MatchFilter} from './FilterBoxActions';
 
 export interface IFilterBoxOwnProps extends React.ClassAttributes<FilterBox> {
     id?: string;
@@ -15,6 +16,7 @@ export interface IFilterBoxOwnProps extends React.ClassAttributes<FilterBox> {
     truncate?: boolean;
     onFilterCallback?: (id: string, filterText: string) => void;
     className?: string;
+    matchFilter?: MatchFilter;
 }
 
 export interface IFilterBoxStateProps {
@@ -22,7 +24,7 @@ export interface IFilterBoxStateProps {
 }
 
 export interface IFilterBoxDispatchProps {
-    onRender?: (id: string) => void;
+    onRender?: (id: string, matchFilter: MatchFilter) => void;
     onDestroy?: (id: string) => void;
     onFilter?: (id: string, filterText: string) => void;
 }
@@ -87,7 +89,7 @@ export class FilterBox extends React.Component<IFilterBoxProps, any> {
 
     componentWillMount() {
         if (this.props.onRender) {
-            this.props.onRender(this.props.id);
+            this.props.onRender(this.props.id, this.props.matchFilter);
         }
     }
 
--- a/src/components/filterBox/FilterBoxActions.ts
+++ b/src/components/filterBox/FilterBoxActions.ts
@@ -1,3 +1,4 @@
+import * as _ from 'underscore';
 import {IReduxAction} from '../../utils/ReduxUtils';
 
 export const FilterActions = {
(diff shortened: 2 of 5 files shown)


MichelML:
@pochretien can you describe what the big change is in details so we can better review?

pochretien:
Add the MatchFilter in the State.

The MatchFilter is useful when you want a custom logic for your filter instant of the default behaviour.  You can add as arguments a list of properties in your object to test.

eg :
```
<MultiSelectWithPredicateAndFilter
        id={'multiSelectFilterPredicate'}
        items={this.state.hoc}
        options={defaultFlatSelectOptions}
        matchPredicate={(p: string, i: IItemBoxProps) => this.matchPredicate(p, i)}
/>

private matchPredicate(predicate: string, item: IItemBoxProps) {
        const value = parseInt(item.value, 10);
        if (predicate === defaultFlatSelectOptions[0].id) {
            return true;
        } else if (predicate === defaultFlatSelectOptions[1].id) {
            return value % 2 === 0;
        } else if (predicate === defaultFlatSelectOptions[2].id) {
            return value % 2 === 1;
        } else {
            return true;
        }
    }
```

src/components/filterBox/FilterBox.tsx
```diff
@@ -87,7 +89,7 @@ export class FilterBox extends React.Component<IFilterBoxProps, any> {
 
     componentWillMount() {
         if (this.props.onRender) {
-            this.props.onRender(this.props.id);
+            this.props.onRender(this.props.id, this.props.matchFilter);
         }
```
gdostie: Minor suggestion: 
```ts
callIfDefined(this.props.onRender, this.props.id, this.props.matchFilter);
```

src/components/filterBox/FilterBoxActions.ts
```diff
@@ -6,18 +7,35 @@ export const FilterActions = {
...
 };
 
+export const defaultMatchFilter = (filterValue: string, item: any, propertiesName: string[]) => {
+    if (filterValue === '') {
+        return true;
+    }
+
+    if (filterValue && item) {
+        const regex = new RegExp(filterValue, 'gi');
+        return _.some(propertiesName, (property: string) => item[property] ? regex.test(item[property]) : false);
```
gdostie:
```ts
return _.some(propertiesName, (property: string) => item[property] && regex.test(item[property]));
```

src/components/filterBox/FilterBoxActions.ts
```diff
@@ -6,18 +7,35 @@ export const FilterActions = {
...
     id: string;
+    matchFilter?: MatchFilter;
 }
 
 export interface IChangeFilterActionPayload extends IFilterActionPayload {
     filterText: string;
 }
 
-export const addFilter = (id: string): IReduxAction<IFilterActionPayload> => ({
+export const addFilter = (id: string, matchFilter: MatchFilter = defaultMatchFilter): IReduxAction<IFilterActionPayload> => ({
```
gdostie: If `defaultMatchFilter`'s type is defined properly, typescript can guess the type by the given default value, so no need to type directly `matchFilter`

src/components/filterBox/tests/FilterBox.spec.tsx
```diff
@@ -47,6 +47,18 @@ describe('FilterBox', () => {
             expect(renderSpy.calls.count()).toBe(1);
         });
 
+        it('should call prop onRender with the id and the matchFilter to set in the store', () => {
+            const renderSpy = jasmine.createSpy('onRender');
+            const matchFilter = jasmine.createSpy('matchFilter');
```
gdostie: `matchFilterSpy`

src/components/filterBox/tests/FilterBox.spec.tsx
```diff
@@ -47,6 +47,18 @@ describe('FilterBox', () => {
...
 
+        it('should call prop onRender with the id and the matchFilter to set in the store', () => {
+            const renderSpy = jasmine.createSpy('onRender');
+            const matchFilter = jasmine.createSpy('matchFilter');
+
+            expect(() => filterBoxInstance.componentWillMount()).not.toThrow();
+
+            filterBox.setProps({id: id, onRender: renderSpy, matchFilter: matchFilter});
+            filterBox.unmount();
+            filterBox.mount();
```
sroy3: You really need to unmount and mount after a setProps? If it doesn't work, maybe try an update after instead

PR #844, opened 2018-10-02
+ Make sure list box items get updated when changing SelectConnected items prop

gdostie:
I had many `SingleSelectConnected` components rendered by the `attributeFormatter` inside a `TableConnected`. Their `items` prop is changing to reflect the changes occuring in the table's state, but when the prop changes, it isn't actually updating the items of the `SingleSelectConnected`. 

The problem was that the `ListBoxConnected` used by the underlying `SelectConnected` component only updates its items if they change inside the redux store, but not when the `items` prop itself changes. 

To counter that, I dispatched an action to update the `ListBox` items in the store when the `SelectConnected`'s `items` prop changes. With that addition, my table now works flawlessly without any hassle.

### DIFF
--- a/src/components/itemBox/ItemBox.tsx
+++ b/src/components/itemBox/ItemBox.tsx
@@ -1,5 +1,6 @@
 import * as classNames from 'classnames';
 import * as React from 'react';
+
 import {Content, IContentProps} from '../content/Content';
 import {PartialStringMatch} from '../partial-string-match/PartialStringMatch';
 import {ITooltipProps, Tooltip} from '../tooltip/Tooltip';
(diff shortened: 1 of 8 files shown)


MichelML:
@gdostie if you could explain the context in which you had the bug that would be nice

MichelML:
the change is fine by me, but I keep wondering if that wouldn't be to the listbox itself to be aware of whether it should update its items or not. I know we ruled out this option, but I think if you do the check on `componentDidUpdate` of the listbox that could work and avoid the double dispatch we discussed

gdostie:
I agree with you @MichelML, I'll try it and update the PR if it works.

MichelML:
nvm that @gdostie i don't think it'll work since props.items after mount will always be what's in the state and not what you pass to the "own props"

PR #848, opened 2018-10-04
Update listbox reducer with filter

pochretien:


pochretien:
Simple change for the *Update* action with the listBox to keep the selected values.

Refactoring for Interface

src/ReactVapor.ts
```diff
@@ -81,6 +81,10 @@ export interface IReactVaporState {
     lastAction?: Redux.Action;
 }
 
+export interface IReduxActionsDefaultPayload {
```
gdostie: BasePayload already exists somewhere in the repo

src/components/listBox/ListBoxReducers.ts
```diff
@@ -79,21 +86,53 @@ export const listBoxReducer = (state: IListBoxState = listBoxInitialState, actio
...
+            const items: IItemBoxProps[] = _.map(state.items, (item: IItemBoxProps) => {
+                const hidden: boolean = _.some(state.selected, (selectedValue: string) => selectedValue === item.value);
+                return {
+                    ...item,
+                    hidden,
+                };
+            });
             return {
                 ...state,
+                items: items,
```
gdostie: `items,`

MichelML:
```
items: state.items((item) => ({...item, hidden: _.contains(state.selected, item.value)}))  
```  
you can remove the `const items` logic altogether if you just write the items property this way

PR #849, opened 2018-10-04
Connect SelectWithPredicate with the store

pochretien:
Update SelectWithPredicate to be connect with the store

I need to change items in the store with Dispatch to change the listbox related to the SelectComponent

### DIFF
--- a/src/components/select/SelectWithPredicate.tsx
+++ b/src/components/select/SelectWithPredicate.tsx
@@ -2,43 +2,76 @@ import * as React from 'react';
 import {keys} from 'ts-transformer-keys';
 import * as _ from 'underscore';
 import {IReactVaporState} from '../../ReactVapor';
-import {ReduxConnect} from '../../utils/ReduxUtils';
+import {callIfDefined} from '../../utils/FalsyValuesUtils';
+import {IDispatch, ReduxConnect} from '../../utils/ReduxUtils';
 import {FlatSelectConnected} from '../flatSelect/FlatSelectConnected';
 import {IFlatSelectOptionProps} from '../flatSelect/FlatSelectOption';
 import {IItemBoxProps} from '../itemBox/ItemBox';
-import {ISelectProps} from './SelectConnected';
+import {updateListBoxOption} from '../listBox/ListBoxActions';
+import {ISelectWithPredicateAndFilterProps} from './SelectComponents';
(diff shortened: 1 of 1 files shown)


src/components/select/SelectWithPredicate.tsx
```diff
@@ -2,43 +2,76 @@ import * as React from 'react';
...
-    class WrappedComponent extends React.Component<ISelectWithPredicateProps> {
+    const mapDispatchToProps = (dispatch: IDispatch, ownProps: ISelectWithPredicateOwnProps & ISelectOwnProps): ISelectWithPredicateDispatchProps => ({
+        updateItems: (predicate: string, items: IItemBoxProps[], matchPredicate: MatchPredicate) => {
+            const newItems = _.map(items, (item: IItemBoxProps) => {
+                const visible = matchPredicate(predicate, item);
+                return {...item, hidden: !visible};
+            });
+            dispatch(updateListBoxOption(
+                ownProps.id,
+                [...newItems],
```
MichelML: map is already returning a new array so you don't need to deconstruct here

src/components/select/SelectWithPredicate.tsx
```diff
@@ -2,43 +2,76 @@ import * as React from 'react';
...
+        const listbox = _.findWhere(state.listBoxes, {id: ownProps.id});
         return {
-            items,
+            items: listbox ? listbox.items : ownProps.items,
         };
     };
 
-    @ReduxConnect(mapStateToProps)
-    class WrappedComponent extends React.Component<ISelectWithPredicateProps> {
+    const mapDispatchToProps = (dispatch: IDispatch, ownProps: ISelectWithPredicateOwnProps & ISelectOwnProps): ISelectWithPredicateDispatchProps => ({
```
MichelML: i think this could be simplified and still be very readable:  

```
const mapDispatchToProps = (dispatch: IDispatch, ownProps: ISelectWithPredicateOwnProps & ISelectOwnProps): ISelectWithPredicateDispatchProps => ({
    updateItems: (predicate: string, items: IItemBoxProps[], matchPredicate: MatchPredicate) => dispatch(updateListBoxOption(
        ownProps.id,
        _.map(items, (item: IItemBoxProps) => ({...item, hidden: !matchPredicate(predicate, item)})),
    )),
});
```

PR #850, opened 2018-10-04
Connect the SelectWithFilter with the store

pochretien:
Update Interface

### DIFF
--- a/src/components/select/SelectConnected.tsx
+++ b/src/components/select/SelectConnected.tsx
@@ -4,6 +4,7 @@ import * as ReactDOM from 'react-dom';
 import * as _ from 'underscore';
 import {IReactVaporState, IReduxActionsPayload} from '../../ReactVapor';
 import {mod} from '../../utils/DataStructuresUtils';
+import {callIfDefined} from '../../utils/FalsyValuesUtils';
 import {keyCode} from '../../utils/InputUtils';
 import {IReduxAction, ReduxConnect} from '../../utils/ReduxUtils';
 import {Content} from '../content/Content';
@@ -17,16 +18,17 @@ import {ISelectState} from './SelectReducers';
 export interface ISelectSpecificProps {
     button: React.ReactNode;
     multi?: boolean;
+    customValues?: boolean;
 }
 
 export interface ISelectOwnProps {
     id: string;
     placeholder?: string;
-    customValues?: boolean;
 }
 
 export interface ISelectStateProps {
     items?: IItemBoxProps[];
+    selected?: string[];
     isOpen?: boolean;
     active?: number;
     selectedValues?: string[];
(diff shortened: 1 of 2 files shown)


src/components/select/SelectConnected.tsx
```diff
@@ -133,8 +135,14 @@ export class SelectConnected extends React.Component<ISelectProps & ISelectSpeci
             const newChildren = React.Children.map(this.props.children, (child: React.ReactElement<any>) => {
                 if (child) {
                     return React.cloneElement(child, {
-                        onKeyDown: (e: React.KeyboardEvent<HTMLElement>) => this.onKeyDown(e),
-                        onKeyUp: (e: React.KeyboardEvent<HTMLElement>) => this.onKeyUp(e),
+                        onKeyDown: (e: React.KeyboardEvent<HTMLElement>) => {
```
pochretien: I need to do this to avoid an overwrite from the parent component. I want to call the child props before the new parent keyUp function

src/components/select/SelectWithFilter.tsx
```diff
@@ -4,71 +4,82 @@ import {keys} from 'ts-transformer-keys';
...
-    };
+const SelectWithFilterPropsToOmit = keys<ISelectWithFilterOwnProps>();
 
+export const selectWithFilter = (Component: (React.ComponentClass<ISelectProps> | React.StatelessComponent<ISelectProps>)): React.ComponentClass<ISelectWithPredicateAndFilterProps> => {
     const mapStateToProps = (state: IReactVaporState, ownProps: ISelectWithFilterProps): ISelectWithFilterStateProps => {
+        const listbox = _.findWhere(state.listBoxes, {id: ownProps.id});
         const filter = _.findWhere(state.filters, {id: ownProps.id});
         const filterValue = filter && filter.filterText || '';
 
-        const items = _.map(ownProps.items, (item: IItemBoxProps) => {
```
pochretien: Moved in the reducer where items are stored

src/components/select/SelectWithFilter.tsx
```diff
@@ -91,13 +102,18 @@ export const selectWithFilter = (Component: (React.ComponentClass<ISelectWithFil
                     <FilterBoxConnected
                         id={this.props.id}
                         containerClasses={[filterBoxContainerClassNames]}
-                        onKeyDown={(this.props as any).onKeyDown}
-                        onKeyUp={(this.props as any).onKeyUp}
+                        onKeyDown={(e: React.KeyboardEvent<HTMLElement>) => callIfDefined((this.props as any).onKeyDown, e)}
```
gdostie: Why the as any here?

pochretien: this kind of structure doesnt know if the Parent component has or not the prop onKeyUp.
We need to set the props as any to call onKeyUp. Depending of what kind of parent you use, you can have it or not

sroy3:
I guess tests will be in the main PR?

src/components/select/SelectWithFilter.tsx
MichelML: can't you just add it as optional in `ISelectWithPredicateAndFilterProps` instead? same for the onKeyUp prop?

PR #851, opened 2018-10-04
Update inteface to avoid TS errors :)

pochretien:
Add items in the State props for the MultiSelectConnected to use the right items. Not only from the ownProps

### DIFF
--- a/src/components/select/tests/SingleSelectConnected.spec.tsx
+++ b/src/components/select/tests/SingleSelectConnected.spec.tsx
@@ -20,7 +20,7 @@ describe('Select', () => {
 
         const id: string = 'list-box-connected';
 
-        const mountSingleSelect = (items: IItemBoxProps[] = [], props: Partial<ISingleSelectProps> = {}) => {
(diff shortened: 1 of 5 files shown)


src/components/select/tests/SingleSelectConnected.spec.tsx
```diff
@@ -20,7 +20,7 @@ describe('Select', () => {
 
         const id: string = 'list-box-connected';
 
-        const mountSingleSelect = (items: IItemBoxProps[] = [], props: Partial<ISingleSelectProps> = {}) => {
+        const mountSingleSelect = (items: IItemBoxProps[] = [], props: ISingleSelectProps & any = {}) => {
```
MichelML: why do you need to add this here ?

pochretien: I have a problem with 2 interface, `selected` is defined with the button interface and with the ISingleSelectProps. I dont know how to handle this use case. For a simple uts, I think its legit to use any to avoid a ts error

PR #856, opened 2018-10-09
Default merge props redux utils

pochretien:


src/components/listBox/ListBoxConnected.tsx
```diff
@@ -26,4 +26,4 @@ const mapDispatchToProps = (dispatch: IDispatch, ownProps: IListBoxOwnProps): IL
 });
 
 export const ListBoxConnected: React.ComponentClass<IListBoxProps> =
-    connect(mapStateToProps, mapDispatchToProps, ReduxUtils.mergePropsWithStatePropsPriority)(ListBox);
+    connect(mapStateToProps, mapDispatchToProps, ReduxUtils.defaultMergeProps)(ListBox);
```
gdostie: I don't understand why you need to redefine the redux default merge props function. By default, if you don't specify any third argument, it will fallback to the default mergeprops function defined by redux

pochretien: connect always need 3 params

gdostie: Taken out of [Redux docs](https://github.com/reduxjs/react-redux/blob/master/docs/api.md):
>[mergeProps(stateProps, dispatchProps, ownProps): props] (Function): If specified, it is passed the result of mapStateToProps(), mapDispatchToProps(), and the parent props. The plain object you return from it will be passed as props to the wrapped component. You may specify this function to select a slice of the state based on props, or to bind action creators to a particular variable from props. If you omit it, Object.assign({}, ownProps, stateProps, dispatchProps) is used by default.

pochretien: merge props its not optional for connect but its not for `@ReduxConnect(mapStateToProps, mapDispatchToProps)`

pochretien: now, we always use the wrong mergeProps for components connected with connect

PR #857, opened 2018-10-09
ListBoxExample

pochretien:
Revert change on listBox to remove items in the store
Add example with a component to wrap the listBox with a custom list in the store.
Add a button to update the current list in the store to show a listbox with a dynamic item list

image

### DIFF
--- a/src/components/listBox/ListBoxConnected.tsx
+++ b/src/components/listBox/ListBoxConnected.tsx
@@ -15,12 +15,12 @@ const mapStateToProps = (state: IReactVaporState, ownProps: IListBoxOwnProps): I
     return {
         selected: list ? list.selected : [],
         active: list ? list.active : undefined,
-        items: list ? list.items : ownProps.items,
+        items: ownProps.items || [],
     };
 };
 
 const mapDispatchToProps = (dispatch: IDispatch, ownProps: IListBoxOwnProps): IListBoxDispatchProps => ({
-    onRender: () => dispatch(addListBox(ownProps.id, ownProps.items)),
+    onRender: () => dispatch(addListBox(ownProps.id)),
     onDestroy: () => dispatch(removeListBox(ownProps.id)),
     onOptionClick: (option: IItemBoxProps) => dispatch(selectListBoxOption(ownProps.id, ownProps.multi, option.value)),
 });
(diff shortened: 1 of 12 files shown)


src/components/listBox/ListBoxReducers.ts
```diff
@@ -22,15 +20,10 @@ export const listBoxReducer = (state: IListBoxState = listBoxInitialState, actio
 
     switch (action.type) {
         case ListBoxActions.add:
-            const selected = _.chain(action.payload.items)
-                .where({selected: true})
-                .pluck('value')
-                .value();
```
GermainBergeron: I wonder if this is still needed. Otherwise you can't have a default selected value?

src/components/listBox/examples/ListBoxExampleConnected.tsx
```diff
@@ -0,0 +1,77 @@
...
+    },
+    addListBoxExample: () => dispatch(addListBoxExample(ownProps.id, convertItemsBoxToOptions(ownProps.items))),
+    removeListBoxExample: () => dispatch(removeListBoxExample(ownProps.id)),
+});
+
+@ReduxConnect(mapStateToProps, mapDispatchToProps, ReduxUtils.defaultMergeProps)
+export class ListBoxExampleConnected extends React.Component<IListBoxExamplesProps> {
+
+    componentWillMount() {
+        callIfDefined(this.props.addListBoxExample);
```
GermainBergeron: In your example `this.props.addListBoxExample` and the other methods will always be defined so no need for callIfDefined

src/components/listBox/ListBoxConnected.tsx
```diff
@@ -15,12 +15,12 @@ const mapStateToProps = (state: IReactVaporState, ownProps: IListBoxOwnProps): I
     return {
         selected: list ? list.selected : [],
         active: list ? list.active : undefined,
-        items: list ? list.items : ownProps.items,
+        items: ownProps.items || [],
```
gdostie: You can remove this whole line. No need to redefine items' value in mapStateToProps if you don't take any value from the state.

src/components/listBox/ListBoxReducers.ts
```diff
@@ -113,7 +91,6 @@ export const listBoxesReducer = (
         case ListBoxActions.select:
         case ListBoxActions.reorder:
```
gdostie: reorder ... is it still relevant, because there are no items to reorder anymore?

pochretien: its a simple reorder for the selected values. you can sent a new array with the selected values. It's used by the drag and drop

PR #863, opened 2018-10-16
Add string list reusable

pochretien:
A simple string list we can use with other component connected

### DIFF
--- a/src/reusableState/customList/StringListActions.ts
+++ b/src/reusableState/customList/StringListActions.ts
@@ -1,0 +1,34 @@
+import {IReduxAction} from '../../utils/ReduxUtils';
+
+export const StringListActions = {
+    add: 'ADD_STRING_LIST',
+    remove: 'REMOVE_STRING_LIST',
+    addValue: 'ADD_VALUE_STRING_LIST',
+    removeValue: 'REMOVE_VALUE_STRING_LIST',
+};
+
+export interface IStringListPayload {
+    id: string;
+    list?: string[];
+    value?: string;
+}
+
+export const addStringList = (id: string, list?: string[]): IReduxAction<IStringListPayload> => ({
+    type: StringListActions.add,
+    payload: {id, list},
+});
+
+export const removeStringList = (id: string): IReduxAction<IStringListPayload> => ({
+    type: StringListActions.remove,
+    payload: {id},
+});
+
+export const addValueStringList = (id: string, value: string): IReduxAction<IStringListPayload> => ({
+    type: StringListActions.addValue,
+    payload: {id, value},
+});
+
+export const removeValueStringList = (id: string, value: string): IReduxAction<IStringListPayload> => ({
+    type: StringListActions.removeValue,
+    payload: {id, value},
+});
--- a/src/reusableState/customList/StringListActions.ts
+++ b/src/reusableState/customList/StringListActions.ts
@@ -14,7 +14,7 @@ export interface IStringListPayload {
     reset?: boolean;
 }
 
-export const addStringList = (id: string, list?: string[]): IReduxAction<IStringListPayload> => ({
+export const addStringList = (id: string, list: string[] = []): IReduxAction<IStringListPayload> => ({
     type: StringListActions.add,
     payload: {id, list},
 });
--- a/src/reusableState/customList/StringListActions.ts
+++ b/src/reusableState/customList/StringListActions.ts
@@ -11,6 +11,7 @@ export interface IStringListPayload {
     id: string;
     list?: string[];
     value?: string;
+    reset?: boolean;
 }
 
 export const addStringList = (id: string, list?: string[]): IReduxAction<IStringListPayload> => ({
@@ -23,9 +24,9 @@ export const removeStringList = (id: string): IReduxAction<IStringListPayload> =
     payload: {id},
 });
 
-export const addValueStringList = (id: string, value: string): IReduxAction<IStringListPayload> => ({
+export const addValueStringList = (id: string, value: string, reset: boolean = false): IReduxAction<IStringListPayload> => ({
     type: StringListActions.addValue,
-    payload: {id, value},
+    payload: {id, value, reset},
 });
 
 export const removeValueStringList = (id: string, value: string): IReduxAction<IStringListPayload> => ({
--- a/src/reusableState/customList/StringListReducers.ts
+++ b/src/reusableState/customList/StringListReducers.ts
@@ -27,7 +27,6 @@ export const stringListCompositeState = (
         return state;
     }
 
-    const stateList: IStringListState = {...state[action.payload.id]};
     switch (action.type) {
         case StringListActions.add:
             return {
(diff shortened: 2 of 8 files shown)


src/reusableState/customList/StringListActions.ts
```diff
@@ -0,0 +1,34 @@
...
+    removeValue: 'REMOVE_VALUE_STRING_LIST',
+};
+
+export interface IStringListPayload {
+    id: string;
+    list?: string[];
+    value?: string;
+}
+
+export const addStringList = (id: string, list?: string[]): IReduxAction<IStringListPayload> => ({
```
MichelML: `(id: string, list: string[] = [])`  will allow you to remove your default array here https://github.com/coveo/react-vapor/pull/863/files#diff-b1aec608a3062febfd6d1d9a3f6d70e3R34 I think it would be cleaner

src/reusableState/customList/StringListReducers.ts
```diff
@@ -0,0 +1,60 @@
...
+export const convertStringListToItemsBox = (list: string[], itemsBoxParams: Partial<IItemBoxProps> = {}): IItemBoxProps[] => _.map(list, (value: string) => ({...itemsBoxParams, value}));
+export const convertItemsBoxToStringList = (items: IItemBoxProps[]): string[] => _.pluck(items, 'value');
+
+export const stringListCompositeState = (
+    state: IStringListCompositeState = stringListInitialState,
+    action: IReduxAction<IStringListPayload>,
+): IStringListCompositeState => {
+
+    if (!action.payload || (!state[action.payload.id] && action.type !== StringListActions.add)) {
+        return {...state};
```
MichelML: you can just return the state you don't need to destructure it since you make no modifications to it.

src/reusableState/customList/StringListReducers.ts
```diff
@@ -0,0 +1,60 @@
...
+    const stateList: IStringListState = {...state[action.payload.id]};
+    switch (action.type) {
+        case StringListActions.add:
+            return {
+                ...state,
+                [action.payload.id]: {
+                    id: action.payload.id,
+                    list: action.payload.list || [],
+                },
+            };
```
MichelML:
```
            return {
                ...state,
                [action.payload.id]: action.payload,
            };
```

src/reusableState/customList/StringListReducers.ts
```diff
@@ -0,0 +1,60 @@
...
+        case StringListActions.remove:
+            return {
+                ..._.omit(state, [action.payload.id]),
+            };
+        case StringListActions.addValue:
+            return {
+                ...state,
+                [action.payload.id]: {
+                    ...stateList,
```
MichelML: i think you can avoid having a const `stateList` and destructure `{......state[action.payload.id]}`  

directly

src/reusableState/customList/StringListReducers.ts
```diff
@@ -0,0 +1,60 @@
...
+                    list: _.uniq([...stateList.list, action.payload.value]),
+                },
+            };
+        case StringListActions.removeValue:
+            return {
+                ...state,
+                [action.payload.id]: {
+                    ...stateList,
```
MichelML: i think you can avoid having a const `stateList` and destructure `...state[action.payload.id]`  

directly

src/reusableState/customList/tests/StringListReducers.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+
+        const genericAction: IReduxAction<IStringListPayload> = {
+            type: 'pokeball',
+        };
+
+        it('should return the default state if the action has no payload and the state is undefined', () => {
+            const newState: IStringListCompositeState = stringListCompositeState(undefined, genericAction);
+
+            expect(newState).toEqual(stringListInitialState);
+            expect(newState).not.toBe(stringListInitialState);
```
MichelML: `expect(newState).not.toBe(stringListInitialState);` doesn't have to be, it could actually be the same object if the state was not modified after an action.

src/reusableState/customList/StringListReducers.ts
```diff
@@ -0,0 +1,60 @@
...
+                    list: [..._.without(stateList.list, action.payload.value)],
+                },
+            };
+        default:
+            return {...state};
```
MichelML: `return state` , you don't have to modify the state if no modifications is made to it

src/reusableState/customList/tests/StringListReducers.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+        });
+
+        it(`should return the default state if the payload id is not in the state, the action type is not ${StringListActions.add} and the state is undefined`, () => {
```
MichelML: no variables in test description, it will make the test unsearchable

MichelML: same for ut below

src/reusableState/customList/tests/StringListReducers.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+            const newState: IStringListCompositeState = stringListCompositeState(oldState, {...genericAction, payload: {id: 'pokemon'}});
+
+            expect(newState).toEqual(oldState);
+        });
+
+        it(`should return the oldState state if the id is in the state but the action type do not exist in the StringListActions`, () => {
+            const newState: IStringListCompositeState = stringListCompositeState(oldState, {type: 'custom_action', payload: {id: stateId}});
+
+            expect(newState).toEqual(oldState);
+            expect(newState).not.toBe(oldState);
```
MichelML: this `expect(newState).not.toBe(oldState);` is unnecessary

src/reusableState/customList/StringListReducers.ts
pochretien: its too avoid any prob if we add other param in the futur. I create a new object to make sure to not use a reference

src/reusableState/customList/tests/StringListReducers.spec.ts
pochretien: Im not complete agree with that, if we rework the name of the variable, the test name will be wrong after that. We can search the test name with the uts structure easily.

src/reusableState/customList/StringListReducers.ts
MichelML: it doesn't use a reference since you are destructuring state[action.payload.id] anyways, not sure to get your point maybe i'm missing something

src/reusableState/customList/StringListReducers.ts
```diff
@@ -0,0 +1,60 @@
...
+        case StringListActions.add:
+            return {
+                ...state,
+                [action.payload.id]: {
+                    id: action.payload.id,
+                    list: action.payload.list || [],
+                },
+            };
+        case StringListActions.remove:
+            return {
```
gdostie: `_.omit` returns a new object so no need to deconstruct it in another object.

PR #867, opened 2018-10-16
Implement a fixed prepend for the SingleSelectedComponent's toggle button

gdostie:
I needed a way to prepend some text before the selected value of a SingleSelectedComponent, but I did not wanted this prepend to be shown on options in the dropdown.

### DIFF
--- a/src/components/select/SingleSelectConnected.tsx
+++ b/src/components/select/SingleSelectConnected.tsx
@@ -15,6 +15,7 @@ export interface ISingleSelectOwnProps extends ISelectProps {
     toggleClasses?: string;
     onSelectOptionCallback?: (option: string) => void;
     items?: IItemBoxProps[];
+    togglePrepend?: React.ReactNode;
 }
 
 export interface ISingleSelectStateProps {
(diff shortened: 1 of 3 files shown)


src/components/select/SingleSelectConnected.tsx
```diff
@@ -15,6 +15,7 @@ export interface ISingleSelectOwnProps extends ISelectProps {
     toggleClasses?: string;
     onSelectOptionCallback?: (option: string) => void;
     items?: IItemBoxProps[];
+    togglePrepend?: React.ReactNode;
```
MichelML: why "toggle" prepend?

pochretien:
you can add an example

pochretien:
The actual prepend props can do the job to add a text I think

src/components/select/SingleSelectConnected.tsx
GermainBergeron: The toggle is the button I guess, maybe buttonPrepend would be clearer

gdostie:
@pochretien the prepend prop of each list item prepends the text in the dropdown values aswell which is not what I want, I want it to be displayed only in the button

PR #874, opened 2018-10-18
Add automatic infinite scroll for DropdownSearch

ThibodeauJF:
I wanted to give the DropdownSearch component a bunch of options and tell it to manage the infinite scrolling itself by setting a number of options per pages
I wanted the filter to search through all the options given instead of just those displayed at a given moment by the infinite scroll which was the default behaviour.
All this without having to refactor the DropdownSearch too much!

Comments and tips are greatly appreciated! 😄


MichelML:
If I would give some comments they would be very minor so I'll refrain myself from giving any. I think you're demo works super well and the code lgtm. very nice 👍

PR #881, opened 2018-10-23
Feature Color Component

GermainBergeron:
The component will use the color passed as a prop and automatically detect if it exists in the Coveo palette. If it's the case it will use the associated bg class. If it doesn't exists (e.g., for rgba, hex, hsl, CSS color) it will take the prop and put it in the background color

image


src/components/color/Color.tsx
```diff
@@ -0,0 +1,33 @@
+import * as classNames from 'classnames';
+import * as React from 'react';
+import * as _ from 'underscore';
+
+export interface ColorProps {
+    color: string;
+}
+
+// tslint:disable-next-line
+const VaporColors: {[key: string]: string} = require('!sass-variable-loader?preserveVariableNames!coveo-styleguide/scss/common/palette.scss');
```
pochretien: I think we can export this variable, we can use this one to select the color in other project after that

MichelML:
What about creating a main section `Colors` and a section `Components` and rendering the colors directly expanded and on top of the page, could serve as a reference ? thinking outloud

PR #882, opened 2018-10-23
Fix modal header truncate

gdostie:
Before
image

image

After
image

image

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -83,7 +83,7 @@ module.exports = {
             },
             {
                 test: /\.scss$/,
-                exclude: [/src\/components/],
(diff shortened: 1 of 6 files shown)


webpack.config.js
```diff
@@ -83,7 +83,7 @@ module.exports = {
             },
             {
                 test: /\.scss$/,
-                exclude: [/src\/components/],
+                exclude: path.join(__dirname, 'src/components'),
```
gdostie: Without this, `npm start` is broken on Windows OS

gdostie:
Just FYI: coverage is dropping because of this file, but I did not make any change in it.

image

PR #883, opened 2018-10-23
Better Demo Page

GermainBergeron:
Auto-reload will stay on the current component
The active component will be in the url to share more easily

image

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -5,14 +5,14 @@
 
     <meta charset="UTF-8">
 </head>
-<body class="coveo-styleguide">
+<body class="coveo-styleguide" style="height: 100%; display: flex">
 <div class="header flex" style="min-height: 65px;">
     <div class="header-section h1 p2 flex flex-center">
         React Vapor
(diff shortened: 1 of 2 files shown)


docs/index.html
```diff
@@ -5,14 +5,14 @@
 
     <meta charset="UTF-8">
 </head>
-<body class="coveo-styleguide">
+<body class="coveo-styleguide" style="height: 100%; display: flex">
```
gdostie: Could you use vapor classes here instead?

docs/index.html
```diff
@@ -5,14 +5,14 @@
...
 <div class="header flex" style="min-height: 65px;">
     <div class="header-section h1 p2 flex flex-center">
         React Vapor
     </div>
 </div>
-<div class="container p2">
-    <div id="App"></div>
+<div class="container" style="height: 0; flex-grow: 1;">
```
gdostie: Same here, I'm pretty sure there is a class in vapor that does this.

PR #888, opened 2018-10-24
Fix the getListState select selector

gdostie:
`getListState` selector was missing a check for `selectWithFilter` prop. Since its optional and can be undefined in the ReactVaporState the selector shouldn't throw IMHO. 

@pochretien does not agree with this. Feel free to argue with this here if you want.

### DIFF
--- a/src/components/select/SelectSelector.ts
+++ b/src/components/select/SelectSelector.ts
@@ -15,7 +15,8 @@ export const getFilterText = (state: IReactVaporState, ownProps: ISelectWithFilt
     return (filter && filter.filterText) || '';
 };
 
-export const getListState = (state: IReactVaporState, ownProps: ISelectWithFilterProps): string[] => state.selectWithFilter[ownProps.id] ? state.selectWithFilter[ownProps.id].list : [];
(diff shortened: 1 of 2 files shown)


src/components/select/SelectSelector.ts
```diff
@@ -15,7 +15,8 @@ export const getFilterText = (state: IReactVaporState, ownProps: ISelectWithFilt
     return (filter && filter.filterText) || '';
 };
 
-export const getListState = (state: IReactVaporState, ownProps: ISelectWithFilterProps): string[] => state.selectWithFilter[ownProps.id] ? state.selectWithFilter[ownProps.id].list : [];
+export const getListState = (state: IReactVaporState, ownProps: ISelectWithFilterProps): string[] =>
```
pochretien: IMO, I not agree with this kind of check because it's not a real use case to have an empty object for the React-Vapor State. In a real app, the redux store is alway initialized with all reducers and the * selectWithFilter* is always defined as an empty object (worst case).  If we need to add an check to make sure a UT passe. I think its the job of this UT to have the right State like the real app.

GermainBergeron:
I think the fix is fine but we'll still need to fix the state in the tests of our other repositories

PR #897, opened 2018-10-29
Add example with list of only selectable

pochretien:
Add example with list of only selectable without customValue
Add option to set disabled or not the button on noselected
Add items when no items in the list

### DIFF
--- a/src/components/select/MultiSelectConnected.tsx
+++ b/src/components/select/MultiSelectConnected.tsx
@@ -138,7 +138,7 @@ export class MultiSelectConnected extends React.Component<IMultiSelectProps, {}>
 
     private getButton(props: ISelectButtonProps): JSX.Element {
         const classes = classNames('multiselect-input', {'mod-sortable': this.props.sortable});
-        const buttonAttrs = {disabled: !this.props.noDisabled && this.props.selected && this.props.selected.length === this.props.items.length};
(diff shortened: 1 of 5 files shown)


src/components/select/MultiSelectConnected.tsx
```diff
@@ -136,7 +138,7 @@ export class MultiSelectConnected extends React.Component<IMultiSelectProps, {}>
 
     private getButton(props: ISelectButtonProps): JSX.Element {
         const classes = classNames('multiselect-input', {'mod-sortable': this.props.sortable});
-        const buttonAttrs = this.props.selected && this.props.selected.length === this.props.items.length ? {disabled: true} : {};
+        const buttonAttrs = {disabled: !this.props.noDisabled && this.props.selected && this.props.selected.length === this.props.items.length};
```
GermainBergeron: Use the new logic with the old way to add the attribute to avoid `disabled: false` on the element

src/components/select/tests/MultiSelectConnected.spec.tsx
```diff
@@ -125,7 +125,7 @@ describe('Select', () => {
                 {value: 'b', selected: true},
             ]);
 
-            expect(multiSelect.find('.multiselect-add').prop('disabled')).toBeUndefined();
+            expect(multiSelect.find('.multiselect-add').prop('disabled')).toBe(false);
```
GermainBergeron: With my suggested fix, it will go back to undefined

PR #899, opened 2018-10-30
Bases for the new table

GermainBergeron:
Added TableHOC component
Added TableWithFilter HOC
Added basics for a connected table row
Added basic example

### DIFF
--- a/src/components/table-hoc/TableHOC.tsx
+++ b/src/components/table-hoc/TableHOC.tsx
@@ -1,0 +1,49 @@
+import * as classNames from 'classnames';
+import * as React from 'react';
+import {ReduxConnect} from '../../utils/ReduxUtils';
+import {ActionBarConnected} from '../actions/ActionBarConnected';
+
+export interface ITableHOCOwnProps {
+    id: string;
+    hasActionButtons?: boolean;
+    data: any[];
+    renderData: (data: any) => React.ReactNode;
+    actions?: React.ReactNode[];
+    tableHeader?: React.ReactNode;
+    onUpdate?: () => void;
+}
+
+export interface ITableHOCProps extends ITableHOCOwnProps {}
+
+@ReduxConnect()
+export class TableHOC extends React.Component<ITableHOCProps & React.HTMLAttributes<HTMLTableElement>> {
+    static defaultProps: Partial<ITableHOCOwnProps> = {
+        hasActionButtons: false,
+        actions: [],
+        onUpdate: () => undefined,
+    };
+
+    render() {
+        return (
+            <div className='table-container'>
+                {this.renderActions()}
+                <table className={classNames(this.props.className)}>
+                    {this.props.tableHeader}
+                    <tbody>{this.props.renderData(this.props.data || [])}</tbody>
+                </table>
+                {this.props.children}
+            </div>
+        );
+    }
+
+    private renderActions() {
+        if (this.props.hasActionButtons || this.props.actions.length) {
+            return (
+                <ActionBarConnected>
+                    {this.props.actions.length && this.props.actions}
+                </ActionBarConnected>
+            );
+        }
+        return null;
+    }
+}
--- a/src/components/table-hoc/TableWithFilter.tsx
+++ b/src/components/table-hoc/TableWithFilter.tsx
@@ -1,0 +1,75 @@
+import * as React from 'react';
+import {keys} from 'ts-transformer-keys/index';
+import * as _ from 'underscore';
+import {IReactVaporState} from '../../ReactVapor';
+import {callIfDefined} from '../../utils/FalsyValuesUtils';
+import {ReduxConnect} from '../../utils/ReduxUtils';
+import {FilterBoxConnected} from '../filterBox/FilterBoxConnected';
+import {IFilterState} from '../filterBox/FilterBoxReducers';
+import {ITableHOCOwnProps} from './TableHOC';
+
+export interface ITableWithFilterConfig {
+    matchFilter?: (filterValue: string, datum: any) => boolean;
+}
+
+export interface ITableWithFilterStateProps {
+    filter: string;
+}
+
+export interface ITableWithFilterDispatchProps {
+    onRender: () => void;
+}
+
+export interface ITableWithFilterProps extends Partial<ITableWithFilterStateProps>,
+    Partial<ITableWithFilterDispatchProps>,
+    ITableHOCOwnProps {}
+
+const TableWithFilterPropsToOmit = keys<ITableWithFilterStateProps>();
+
+const defaultMatchFilter = (filter: string, datum: any) => JSON.stringify(_.values(datum).map((v: string) => v.toLowerCase())).indexOf(filter.toLowerCase()) !== -1;
+
+export const tableWithFilter = (config: ITableWithFilterConfig = {matchFilter: defaultMatchFilter}) => (Component: (React.ComponentClass<ITableWithFilterProps> | React.StatelessComponent<ITableWithFilterProps>)): React.ComponentClass<ITableWithFilterProps & React.HTMLAttributes<HTMLTableElement>> => {
+    const mapStateToProps = (state: IReactVaporState, ownProps: ITableWithFilterProps): ITableWithFilterStateProps | ITableHOCOwnProps => {
+        const filter: IFilterState = _.findWhere(state.filters, {id: ownProps.id});
+        const matchingData = (data: any[]) => {
+            if (filter && filter.filterText) {
+                return _.filter(ownProps.data, (datum: any) => config.matchFilter(filter.filterText, datum));
+            }
+            return data;
+        };
+        return {
+            filter: filter && filter.filterText,
+            data: ownProps.data && matchingData(ownProps.data),
+        };
+    };
+
+    @ReduxConnect(mapStateToProps)
+    class TableWithFilter extends React.Component<ITableWithFilterProps> {
+
+        componentDidUpdate(prevProps: ITableWithFilterProps) {
+            if (prevProps.filter !== this.props.filter) {
+                callIfDefined(this.props.onUpdate);
+            }
+        }
+
+        render() {
+            const filterAction = <FilterBoxConnected
+                key='FilterBox'
+                id={this.props.id}
+                className='coveo-table-actions'
+                isAutoFocus
+            />;
+            const newActions = [...[], ...this.props.actions, filterAction];
+            const newProps = {
+                ..._.omit(this.props, [...TableWithFilterPropsToOmit]),
+            };
+            return (
+                <Component {...newProps} actions={newActions}>
+                    {this.props.children}
+                </Component>
+            );
+        }
+    }
+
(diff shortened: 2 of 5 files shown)


src/components/table-hoc/TableHOC.tsx
```diff
@@ -0,0 +1,49 @@
+import * as classNames from 'classnames';
+import * as React from 'react';
+import {ReduxConnect} from '../../utils/ReduxUtils';
+import {ActionBarConnected} from '../actions/ActionBarConnected';
+
+export interface ITableHOCOwnProps {
+    id: string;
+    hasActionButtons?: boolean;
```
MichelML: why do you need to have this prop? can't you use `_.isEmpty(this.props.actions)` ?

MichelML: I think i know, I'd suggest you leave actions as undefined and not `[]` and do a falsy check with `_.isArray` instead of having another prop `hasActionButtons` , if we pass an empty array the action bar will still show

src/components/table-hoc/TableWithFilter.tsx
```diff
@@ -0,0 +1,75 @@
...
+    ITableHOCOwnProps {}
+
+const TableWithFilterPropsToOmit = keys<ITableWithFilterStateProps>();
+
+const defaultMatchFilter = (filter: string, datum: any) => JSON.stringify(_.values(datum).map((v: string) => v.toLowerCase())).indexOf(filter.toLowerCase()) !== -1;
+
+export const tableWithFilter = (config: ITableWithFilterConfig = {matchFilter: defaultMatchFilter}) => (Component: (React.ComponentClass<ITableWithFilterProps> | React.StatelessComponent<ITableWithFilterProps>)): React.ComponentClass<ITableWithFilterProps & React.HTMLAttributes<HTMLTableElement>> => {
+    const mapStateToProps = (state: IReactVaporState, ownProps: ITableWithFilterProps): ITableWithFilterStateProps | ITableHOCOwnProps => {
+        const filter: IFilterState = _.findWhere(state.filters, {id: ownProps.id});
+        const matchingData = (data: any[]) => {
```
MichelML: I'd suggest using a ternary here it'll be lighter to read

src/components/table-hoc/TableHOC.tsx
```diff
@@ -0,0 +1,49 @@
+import * as classNames from 'classnames';
+import * as React from 'react';
+import {ReduxConnect} from '../../utils/ReduxUtils';
+import {ActionBarConnected} from '../actions/ActionBarConnected';
+
+export interface ITableHOCOwnProps {
+    id: string;
+    hasActionButtons?: boolean;
+    data: any[];
+    renderData: (data: any) => React.ReactNode;
```
gdostie: Since the `data` prop has `any[]` type, I feel like the `data` parameter in `renderData` should have the same type.

src/components/table-hoc/TableWithFilter.tsx
```diff
@@ -0,0 +1,75 @@
...
+    class TableWithFilter extends React.Component<ITableWithFilterProps> {
+
+        componentDidUpdate(prevProps: ITableWithFilterProps) {
+            if (prevProps.filter !== this.props.filter) {
+                callIfDefined(this.props.onUpdate);
+            }
+        }
+
+        render() {
+            const filterAction = <FilterBoxConnected
```
MichelML: hate to be that guy, standards say to wrap components around parenthesis

src/components/table-hoc/TableWithFilter.tsx
```diff
@@ -0,0 +1,75 @@
...
+                key='FilterBox'
+                id={this.props.id}
+                className='coveo-table-actions'
+                isAutoFocus
+            />;
+            const newActions = [...[], ...this.props.actions, filterAction];
```
MichelML: i don't think the first `...[]` is necessary , array spread compiles to Array.concat which returns a new array

src/components/table-hoc/TableWithFilter.tsx
```diff
@@ -0,0 +1,75 @@
...
+            const newProps = {
+                ..._.omit(this.props, [...TableWithFilterPropsToOmit]),
```
MichelML: `..._.omit(this.props, TableWithFilterPropsToOmit)`

src/components/table-hoc/TableWithFilter.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+export interface ITableWithFilterProps extends Partial<ITableWithFilterStateProps>,
+    Partial<ITableWithFilterDispatchProps>,
+    ITableHOCOwnProps {}
+
+const TableWithFilterPropsToOmit = keys<ITableWithFilterStateProps>();
+
+const defaultMatchFilter = (filter: string, datum: any) => JSON.stringify(_.values(datum).map((v: string) => v.toLowerCase())).indexOf(filter.toLowerCase()) !== -1;
+
+export const tableWithFilter = (config: ITableWithFilterConfig = {matchFilter: defaultMatchFilter}) => (Component: (React.ComponentClass<ITableWithFilterProps> | React.StatelessComponent<ITableWithFilterProps>)): React.ComponentClass<ITableWithFilterProps & React.HTMLAttributes<HTMLTableElement>> => {
```
gdostie: Line too long

src/components/table-hoc/TableWithFilter.tsx
```diff
@@ -0,0 +1,75 @@
...
+    onRender: () => void;
+}
+
+export interface ITableWithFilterProps extends Partial<ITableWithFilterStateProps>,
+    Partial<ITableWithFilterDispatchProps>,
+    ITableHOCOwnProps {}
+
+const TableWithFilterPropsToOmit = keys<ITableWithFilterStateProps>();
+
+const defaultMatchFilter = (filter: string, datum: any) => JSON.stringify(_.values(datum).map((v: string) => v.toLowerCase())).indexOf(filter.toLowerCase()) !== -1;
```
gdostie: TIL datum is a word haha

src/components/table-hoc/TableWithFilter.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+export const tableWithFilter = (config: ITableWithFilterConfig = {matchFilter: defaultMatchFilter}) => (Component: (React.ComponentClass<ITableWithFilterProps> | React.StatelessComponent<ITableWithFilterProps>)): React.ComponentClass<ITableWithFilterProps & React.HTMLAttributes<HTMLTableElement>> => {
+    const mapStateToProps = (state: IReactVaporState, ownProps: ITableWithFilterProps): ITableWithFilterStateProps | ITableHOCOwnProps => {
```
gdostie: This logic should eventually go in a cached selector since filtering can be expensive.

src/components/table-hoc/TableWithFilter.tsx
```diff
@@ -0,0 +1,75 @@
...
+        const filter: IFilterState = _.findWhere(state.filters, {id: ownProps.id});
+        const matchingData = (data: any[]) => {
+            if (filter && filter.filterText) {
+                return _.filter(ownProps.data, (datum: any) => config.matchFilter(filter.filterText, datum));
+            }
+            return data;
+        };
```
gdostie:
```tsx
const matchingData = (data: any[]) =>  filter && filter.filterText
    ? _.filter(ownProps.data, (datum: any) => config.matchFilter(filter.filterText, datum))
    : data;
```

src/components/table-hoc/TableWithFilter.tsx
```diff
@@ -0,0 +1,75 @@
...
+import * as _ from 'underscore';
+import {IReactVaporState} from '../../ReactVapor';
+import {callIfDefined} from '../../utils/FalsyValuesUtils';
+import {ReduxConnect} from '../../utils/ReduxUtils';
+import {FilterBoxConnected} from '../filterBox/FilterBoxConnected';
+import {IFilterState} from '../filterBox/FilterBoxReducers';
+import {ITableHOCOwnProps} from './TableHOC';
+
+export interface ITableWithFilterConfig {
+    matchFilter?: (filterValue: string, datum: any) => boolean;
```
francoislg: Could `matchFilter`  be generic to allow library users to inject their own type? `matchFilter<T = any>(filterValue: string, datum: T)`

src/components/table-hoc/TableHOC.tsx
GermainBergeron: For the moment I don't need it, but I was thinking about when I'll connect the rows with "selectable" how I'll be able to add the rows actions. I can probably delete it in this PR and add it when needed

src/components/table-hoc/TableWithFilter.tsx
GermainBergeron: But this.props.actions can be undefined and in this case the code output doesn't work

image
[playground](http://www.typescriptlang.org/play/#src=interface%20IProps%20%7B%0D%0A%20%20%20%20maybeArr%3F%3A%20any%5B%5D%3B%0D%0A%7D%0D%0Aconst%20props%3A%20IProps%20%3D%20%7B%7D%3B%0D%0A%0D%0A%5B...props.maybeArr%2C%20%22a%22%5D%3B%0D%0A%0D%0A%0D%0A)

src/components/table-hoc/TableHOC.tsx
MichelML: in that sense, maybe we should have a `getActions` method instead of `actions`, not every row will have the same actions

src/components/table-hoc/TableWithFilter.tsx
MichelML: You're right on that, but in the other case if it's undefined you'll insert an undefined value in the array:

screen shot 2018-10-30 at 2 16 19 pm

i guess something like `[...(this.props.actions || []), filterAction]` should be used

src/components/table-hoc/TableHOC.tsx
GermainBergeron: Removed for now

src/components/table-hoc/TableWithFilter.tsx
francoislg: Cool stuff, did not notice this, I'm fine with it :)

src/components/table-hoc/TableWithFilter.tsx
```diff
@@ -0,0 +1,76 @@
...
+            data: ownProps.data && filterData(),
+        };
+    };
+
+    @ReduxConnect(mapStateToProps)
+    class TableWithFilter extends React.Component<ITableWithFilterProps> {
+
+        componentDidUpdate(prevProps: ITableWithFilterProps) {
+            if (prevProps.filter !== this.props.filter) {
+                callIfDefined(this.props.onUpdate);
```
pochretien: just to be sure, we add this logic with the right props to compare on each component we want to add a behaviour to update on change ? eg. with a predicate component, we valide de props predicate value and call onUpdate if it change ?

PR #900, opened 2018-10-30
Added blankslate HOC and examples

GermainBergeron:


src/components/table-hoc/TableWithBlankslate.tsx
```diff
@@ -0,0 +1,43 @@
+import * as React from 'react';
+import {keys} from 'ts-transformer-keys/index';
```
gdostie: @MichelML @pochretien Thinking out loud here. Do you guys think we should get away from `ttypescript` afterall? I know @pochretien had some issues with it lacking support for ES7/ES8 support of newer features.

MichelML: i don't have a real opinion on that, never used it. Wasn't aware that we had problems with that recently

pochretien: I talked about it with Germain. The idea would be to create our own function to create an array with the interface before we compile typescript. After that, we can use typescript as compiler

PR #901, opened 2018-10-30
Added WithSort for the new table

GermainBergeron:
image


src/ReactVapor.ts
```diff
@@ -70,6 +71,7 @@ export interface IReactVaporState {
     toastContainers?: IToastsState[];
     tableHeaderCells?: ITableHeaderCellsState;
     tables?: ITablesState;
+    tableHOCHeader?: ITableWithSortState[];
```
gdostie: As our state is getting bigger and bigger, what do you think of grouping related states under the same key. For example we could have a `tableHOC` key that contains all `tableHOC*` states?

src/components/table-hoc/TableWithSort.tsx
```diff
@@ -0,0 +1,56 @@
...
+export interface ITableWithSortStateProps {
+    isAsc?: boolean;
+    sortKey?: string;
+}
+
+export interface ITableWithSortProps extends Partial<ITableWithSortStateProps> {}
+
+const TableWithSortPropsToOmit = keys<ITableWithSortStateProps>();
+
+export const tableWithSort = (config: ITableWithSortConfig) => (Component: (React.ComponentClass<ITableHOCOwnProps> | React.StatelessComponent<ITableHOCOwnProps>)): React.ComponentClass<ITableWithSortProps & React.HTMLAttributes<HTMLTableElement>> => {
```
gdostie: Line too long

src/components/table-hoc/actions/TableHeaderActions.ts
```diff
@@ -0,0 +1,42 @@
...
+const addTableHeader = (id: string, tableId: string, isDefault: boolean): IReduxAction<ITableHeaderAddPayload> => ({
+    type: TableHeaderActionsType.add,
+    payload: {id, tableId, isDefault},
+});
+
+export interface ITableHeaderRemovePayload {
+    id: string;
+}
+
+const removeTableHeader = (id: string): IReduxAction<ITableHeaderRemovePayload> => ({
```
gdostie: You could use directly the `BasePayload` and remove `ITableHeaderRemovePayload`

src/components/table-hoc/actions/TableHeaderActions.ts
```diff
@@ -0,0 +1,42 @@
...
+    type: TableHeaderActionsType.remove,
+    payload: {id},
+});
+
+export interface ITableHeaderSortPayload {
+    id: string;
+}
+
+const sortTable = (id: string): IReduxAction<ITableHeaderSortPayload> => ({
```
gdostie: Same here, you could use directly the BasePayload and remove ITableHeaderSortPayload

src/components/table-hoc/actions/TableHeaderActions.ts
```diff
@@ -0,0 +1,42 @@
+import {IReduxAction} from '../../../utils/ReduxUtils';
+
+export const TableHeaderActionsType = {
```
gdostie: `TableHeaderActionTypes`

src/components/table-hoc/actions/TableHeaderActions.ts
```diff
@@ -0,0 +1,42 @@
+import {IReduxAction} from '../../../utils/ReduxUtils';
+
+export const TableHeaderActionsType = {
+    add: 'ADD_TABLE_HEADER',
+    remove: 'REMOVE_TABLE_HEADER',
+    sort: 'SORT_TABLE_HEADER',
+};
+
+export interface ITableHeaderAddPayload {
```
gdostie: I say we quit using `I` as interface prefix. https://github.com/coveo/tsjs#interface-no-i

src/ReactVapor.ts
GermainBergeron: I think it's better to keep things separated, the problem if we use the same key is that all reducers for this key will run which add some overhead

src/components/table-hoc/TableHeaderWithSort.tsx
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    componentWillUnmount() {
+        this.props.onUnmount();
+    }
+
+    render() {
+        const headerCellClasses = classNames(this.props.className, 'admin-sort', {
+            'admin-sort-ascending': this.props.sorted === true,
+            'admin-sort-descending': this.props.sorted === false,
```
MichelML:
```
true --> sort ascending
false --> sort descending  
undefined --> unsorted  
```  
?

src/components/table-hoc/TableWithSort.tsx
```diff
@@ -0,0 +1,58 @@
...
+const TableWithSortPropsToOmit = keys<ITableWithSortStateProps>();
+
+type SortableTableComponent = React.ComponentClass<ITableHOCOwnProps>;
+
+export const tableWithSort = (config: ITableWithSortConfig) => (Component: SortableTableComponent) => {
+    const mapStateToProps = (state: IReactVaporState, ownProps: ITableHOCOwnProps): ITableWithSortStateProps | ITableHOCOwnProps => {
+        const tableSort: ITableWithSortState = _.find(state.tableHOCHeader, (v: ITableWithSortState) => v.tableId === ownProps.id && _.isBoolean(v.isAsc));
+        if (tableSort && ownProps.data) {
+            return {
+                data: [...ownProps.data].sort((a, b) => config.sort(tableSort.id, tableSort.isAsc, a, b)),
```
MichelML: could tableSort.id be sortKey instead ?

src/components/table-hoc/TableHeaderWithSort.tsx
pochretien: If we want to continue with the same logic has we have now. Its the right think to do. IMO, if we have a sortable table, It better to sort asc or desc on a specific row on the first render. In this case, the unsorted was removed and we can set the right sort svg on sorted column.  It's look weird to have an initial state only available on the first render and not available after that.  @Charles-ErickBG I don't know the right answer but I think the UX can help to figure out what kind of behaviour we want in the new table.

GermainBergeron: @MichelML Yes, undefined is unsorted (for this column)

@pochretien there's a prop on the header that you need to set (isDefault) to make this column the default sort. This way you control which column is the default one

pochretien: Good, but its possible to have the default column sorted with undefined in the first render ? or we need to set asc or desc

src/components/table-hoc/reducers/TableWithSortReducers.ts
```diff
@@ -0,0 +1,54 @@
...
+                    isAsc: !header.isAsc,
+                };
+            }
+            return header.tableId === current.tableId ? {...header, isAsc: undefined} : header;
+        });
+    }
+    return state;
+};
+
+const TableWithSortActionReducers: {[key: string]: (...args: any[]) => any} = {
```
pochretien: we use 2 kind of structure for the reducer, I dont have preference but I dont think its a good idea to use both.  When we use the switch case and when this one ?

src/components/table-hoc/TableHeaderWithSort.tsx
GermainBergeron: Talked live, everything is fine like it is

src/components/table-hoc/reducers/TableWithSortReducers.ts
GermainBergeron: Personally I really prefer this one. The problem with the switch case is that you can't reuse let and const name between cases and that the action is not correctly typed

PR #902, opened 2018-10-30
+ Refactor the toast so that it is more generic

gdostie:
Now we can use `children` in toasts and the `className` prop.

```tsx
<Toast title='The toas title' className='bg-coveo-blue'>
    <div>The toast content.<div>
</Toast>
```

### DIFF
--- a/src/components/toast/Toast.tsx
+++ b/src/components/toast/Toast.tsx
@@ -1,6 +1,7 @@
 import * as classNames from 'classnames';
 import * as React from 'react';
 import * as _ from 'underscore';
+
 import {Svg} from '../svg/Svg';
 
 export interface IToastProps {
@@ -11,6 +12,7 @@ export interface IToastProps {
     dismiss?: number;
     dismissible?: boolean;
     animate?: boolean;
+    className?: string;
     content?: React.ReactNode;
     onRender?: () => void;
     onClose?: () => void;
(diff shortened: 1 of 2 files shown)


src/components/toast/Toast.tsx
```diff
@@ -65,27 +67,40 @@ export class Toast extends React.Component<IToastProps, {}> {
...
+                'mod-success': this.props.type === ToastType.Success,
+                'mod-warning': this.props.type === ToastType.Warning,
+                'mod-error': this.props.type === ToastType.Error,
+                'mod-animated': _.isUndefined(this.props.animate) || this.props.animate === true,
+            },
+            this.props.className,
+        );
+
+        const closeButton = this.props.dismissible ?
+            <span className='toast-close' onClick={() => this.close()}>
```
pochretien: can we use a button instead of a div with a onClick, now the Button can have a Svg as children

gdostie: Probably, I'll try refactoring this also

src/components/toast/tests/Toast.spec.tsx
```diff
@@ -118,7 +122,7 @@ describe('Toasts', () => {
         it('should allow JSX in the content', () => {
             const descriptionContainer = '.toast-description';
             const expectedDescription = 'description';
-            const newToastAttributes = _.extend({}, toastBasicAttributes, {content: () => <a href='#'>{expectedDescription}</a>});
+            const newToastAttributes = _.extend({}, toastBasicAttributes, {content: <a href='#'>{expectedDescription}</a>});
```
GermainBergeron: Is it still working with the old way?

gdostie: No

src/components/toast/Toast.tsx
```diff
@@ -65,27 +67,40 @@ export class Toast extends React.Component<IToastProps, {}> {
...
-        const classes = classNames('toast', {
-            'mod-success': this.props.type !== ToastType.Warning && this.props.type !== ToastType.Error,
-            'mod-warning': this.props.type === ToastType.Warning,
-            'mod-error': this.props.type === ToastType.Error,
-            'mod-animated': _.isUndefined(this.props.animate) || this.props.animate === true,
-        });
+        const classes = classNames(
+            'toast',
+            {
+                'mod-success': this.props.type === ToastType.Success,
```
MichelML: what if we don't pass anything now?

MichelML: will it still be green ?

src/components/toast/Toast.tsx
MichelML:
```
? (
 // your component
)
: null;
```

src/components/toast/Toast.tsx
```diff
@@ -65,27 +67,40 @@ export class Toast extends React.Component<IToastProps, {}> {
...
+        );
+
+        const closeButton = this.props.dismissible ?
+            <span className='toast-close' onClick={() => this.close()}>
+                <Svg svgName='close' className='icon mod-lg fill-pure-white' />
+            </span>
+            : null;
+
+        const toastContent = !!this.props.content || !!this.props.children
+            ? <div className='toast-description'>
```
MichelML: use parenthesis

src/components/toast/tests/Toast.spec.tsx
MichelML: have the same question 🤔

src/components/toast/Toast.tsx
gdostie: It won't

PR #904, opened 2018-10-30
Added Server Table HOC Example

GermainBergeron:
Démo: https://coveo.github.io/react-vapor/feature-hoc-table-server/#Table%20(server%20+%20hoc)

### DIFF
--- a/src/components/navigation/pagination/NavigationPaginationReducers.ts
+++ b/src/components/navigation/pagination/NavigationPaginationReducers.ts
@@ -2,6 +2,7 @@ import * as _ from 'underscore';
 import {contains} from 'underscore.string';
 import {IReduxActionsPayload} from '../../../ReactVapor';
 import {IReduxAction} from '../../../utils/ReduxUtils';
+import {FilterActions} from '../../filterBox/FilterBoxActions';
 import {TableActions} from '../../tables/TableActions';
 import {PaginationActions} from './NavigationPaginationActions';
 
@@ -37,6 +38,10 @@ export const paginationReducer = (state: IPaginationState = paginationInitialSta
             if (contains(state.id, action.payload.id) && action.payload.shouldResetPage) {
                 return {...state, pageNb: 0};
             }
+        case FilterActions.filterThrough:
+            if (contains(state.id, action.payload.id)) {
+                return {...state, pageNb: 0};
+            }
         default:
             return state;
     }
@@ -55,6 +60,7 @@ export const paginationCompositeReducer = (state: IPaginationState[] = paginatio
             });
         case PaginationActions.changePage:
         case PaginationActions.reset:
+        case FilterActions.filterThrough:
         case TableActions.modifyState:
             return state.map((pagination: IPaginationState) => paginationReducer(pagination, action));
         default:
--- a/src/components/navigation/pagination/NavigationPaginationReducers.ts
+++ b/src/components/navigation/pagination/NavigationPaginationReducers.ts
@@ -38,10 +38,12 @@ export const paginationReducer = (state: IPaginationState = paginationInitialSta
             if (contains(state.id, action.payload.id) && action.payload.shouldResetPage) {
                 return {...state, pageNb: 0};
             }
+            return state;
         case FilterActions.filterThrough:
             if (contains(state.id, action.payload.id)) {
                 return {...state, pageNb: 0};
             }
+            return state;
         default:
             return state;
     }
--- a/src/components/navigation/pagination/NavigationPaginationReducers.ts
+++ b/src/components/navigation/pagination/NavigationPaginationReducers.ts
@@ -1,5 +1,5 @@
 import * as _ from 'underscore';
-import {contains} from 'underscore.string';
+import * as s from 'underscore.string';
 import {IReduxActionsPayload} from '../../../ReactVapor';
 import {IReduxAction} from '../../../utils/ReduxUtils';
 import {FilterActions} from '../../filterBox/FilterBoxActions';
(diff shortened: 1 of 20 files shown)


src/components/table-hoc/examples/TableHOCServerExamples.tsx
```diff
@@ -0,0 +1,88 @@
...
+import {tableWithPagination} from '../TableWithPagination';
+import {tableWithSort} from '../TableWithSort';
+import {IExampleRowData, TableHOCServerActions} from './TableHOCServerExampleReducer';
+
+const ServerTable = _.compose(
+    tableWithBlankslate({title: 'No data caused the table to be empty'}),
+    tableWithFilter({isServer: true}),
+    tableWithBlankslate({title: 'Filter caused the table to be empty'}),
+    tableWithSort({isServer: true}),
+    tableWithPagination({isServer: true, perPageNumbers: [3, 5, 10]}),
```
francoislg: Looks a bit weird at first that `isServer` has to be passed in all of the composite components.

Is there really a use case to have `isServer: true/false` in each of those composite state?

Say, is it possible to filter server-side but then apply a sort client-side?

src/components/table-hoc/examples/TableHOCServerExamples.tsx
```diff
@@ -0,0 +1,88 @@
...
+)(TableHOC);
+
+interface TableHOCServerDispatchProps {
+    fetch: (compositeState: any) => void;
```
francoislg: Can `any` be typed here? At least put a generic.

src/components/table-hoc/examples/TableHOCServerExamples.tsx
```diff
@@ -0,0 +1,88 @@
...
+                    <ServerTable
+                        id={TableHOCServerExamples.TABLE_ID}
+                        className='table'
+                        data={this.props.serverData}
+                        renderData={generateRow}
+                        onUpdate={this.onUpdate}
+                        totalEntries={this.props.totalEntries}
+                        totalPages={this.props.totalPages}
+                        tableHeader={
+                            <thead>
```
francoislg: I'd suggest putting this closer to `generateRow` into a `tableHeader` variable. They most probably will change together, it makes sense to put them close.

src/components/table-hoc/examples/TableHOCServerExampleReducer.ts
```diff
@@ -0,0 +1,63 @@
...
+    setData: 'TABLE_HOC_SET_DATA',
+    fetch: 'TABLE_HOC_FETCH_DATA',
+};
+
+const setData = (data: any[]): IReduxAction<ISetExampleDataPayload> => ({
+    type: TableHOCServerActionsType.setData,
+    payload: {data},
+});
+
+const fetchData = (compositeState: any): IThunkAction => (dispatch: IDispatch) => {
```
francoislg: I'd rather have a typed `compositeState` here

src/components/table-hoc/TableWithSort.tsx
```diff
@@ -8,7 +8,8 @@ import {ITableWithSortState} from './reducers/TableWithSortReducers';
 import {ITableHOCOwnProps} from './TableHOC';
 
 export interface ITableWithSortConfig {
-    sort: (sortKey: string, isAsc: boolean, a: any, b: any) => number;
+    sort?: (sortKey: string, isAsc: boolean, a: any, b: any) => number;
```
francoislg: Can it be generic? 

`sort<T = any>(sortKey: string, isAsc: boolean, a: T, b: T)`

src/components/table-hoc/examples/TableHOCServerExamples.tsx
```diff
@@ -0,0 +1,88 @@
...
+                            </thead>
+                        }
+                    />
+                </div>
+            </div>
+        );
+    }
+
+    private onUpdate = () => {
+        const compositeState = TableHOCUtils.getCompositeState(TableHOCServerExamples.TABLE_ID, ReactVaporStore.getState());
```
gdostie: I know this is an example but just to avoid people from making `.getState()` calls everywhere `compositeState` could be a state props and `TableHOCUtils.getCompositeState` would be used in `mapStateToProps`.

src/components/table-hoc/TableHOCUtils.ts
```diff
@@ -0,0 +1,28 @@
+import * as _ from 'underscore';
+import {IReactVaporState} from '../../ReactVapor';
+import {IFilterState} from '../filterBox/FilterBoxReducers';
+import {ITableWithSortState} from './reducers/TableWithSortReducers';
+
+const getCompositeState = (id: string, state: IReactVaporState) => {
```
gdostie: Should be a reselect selector IMHO.

src/components/table-hoc/examples/TableHOCServerExamples.tsx
gdostie: @francoislg yes we have use cases of that in the Coveo's Administration Console (i.e. the Sources Table) We're not using a React component for this table yet, but someday we will.

src/components/table-hoc/examples/TableHOCServerExamples.tsx
GermainBergeron: Yes, I'll create an interface for this

src/components/table-hoc/TableWithSort.tsx
GermainBergeron: Same thing as the other method for filtering, since it's any you can use your own type (see examples)

src/components/table-hoc/examples/TableHOCServerExamples.tsx
GermainBergeron: I put the getState directly in the IThunk since the parameter after dispatch is getState

src/components/navigation/pagination/NavigationPaginationReducers.ts
```diff
@@ -38,10 +38,12 @@ export const paginationReducer = (state: IPaginationState = paginationInitialSta
             if (s.contains(state.id, action.payload.id) && action.payload.shouldResetPage) {
                 return {...state, pageNb: 0};
             }
+            return state;
```
gdostie: Ternaries ?

PR #908, opened 2018-11-01
Feature Table HOC with Predicates & Loading

GermainBergeron:
image

PR #909, opened 2018-11-01
Feature Table HOC with Actions

GermainBergeron:
See https://github.com/coveo/react-vapor/pull/908/files#diff-465068f1d1e6ba675d6cf11134870c53R73 for example

image

### DIFF
--- a/src/components/table-hoc/TableRowConnected.tsx
+++ b/src/components/table-hoc/TableRowConnected.tsx
@@ -1,9 +1,81 @@
+import * as classNames from 'classnames';
 import * as React from 'react';
+import * as _ from 'underscore';
+import {IReactVaporState} from '../../ReactVapor';
+import {IDispatch, ReduxConnect} from '../../utils/ReduxUtils';
+import {IActionOptions} from '../actions/Action';
+import {addActionsToActionBar} from '../actions/ActionBarActions';
+import {TableRowActions} from './actions/TableRowActions';
+import {ITableRowState} from './reducers/TableRowReducers';
 
-export class TableRowConnected extends React.Component<React.HTMLAttributes<HTMLTableRowElement>> {
+export interface ITableRowOwnProps {
+    id: string;
+    tableId: string;
+    actions: IActionOptions[];
+    isMultiselect?: boolean;
+}
+
+export interface ITableRowStateProps {
+    selected: boolean;
+}
+
+export interface ITableRowDispatchProps {
+    onMount: () => void;
+    onUnmount: () => void;
+    onClick: (isMulti: boolean) => void;
+}
+
+export interface ITableHeaderWithSortProps extends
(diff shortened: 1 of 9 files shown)


src/components/table-hoc/reducers/TableRowReducers.ts
```diff
@@ -0,0 +1,69 @@
+import * as _ from 'underscore';
+import {IReduxAction} from '../../../utils/ReduxUtils';
+import {PaginationActions} from '../../navigation/pagination/NavigationPaginationActions';
+import {PerPageActions} from '../../navigation/perPage/NavigationPerPageActions';
+import {ITableRowAddPayload, ITableRowRemovePayload, ITableRowSelectPayload, TableRowActionsType} from '../actions/TableRowActions';
+import {TableHOCComponentType, TableHOCUtils} from '../TableHOCUtils';
+
```
MichelML: this is an interesting way of composing reducers, but i'm wondering if we shouldn't stick to the switch case style since we have this style everywhere?

GermainBergeron: The problem I have with the switch case is that you can't reuse variable name (with const) in multiple cases. Also, the interface of the payload is the same for every reducer which can be confusing

src/components/table-hoc/TableRowConnected.tsx
```diff
@@ -1,9 +1,81 @@
...
 
-export class TableRowConnected extends React.Component<React.HTMLAttributes<HTMLTableRowElement>> {
     render() {
         return (
-            <tr {...this.props}>{this.props.children}</tr>
+            <tr
+                {..._.omit(this.props, ['actions', 'selected', 'tableId', 'isMultiselect', 'onUnmount', 'onMount', 'onClick'])}
+                className={classNames(this.props.className, {selected: this.props.selected})}
+                onClick={this.handleClick}
+            >{this.props.children}</tr>
```
MichelML: html tag on its own line

src/components/table-hoc/TableRowConnected.tsx
```diff
@@ -1,9 +1,81 @@
...
+    componentWillUnmount() {
+        this.props.onUnmount();
+    }
 
-export class TableRowConnected extends React.Component<React.HTMLAttributes<HTMLTableRowElement>> {
     render() {
         return (
-            <tr {...this.props}>{this.props.children}</tr>
+            <tr
+                {..._.omit(this.props, ['actions', 'selected', 'tableId', 'isMultiselect', 'onUnmount', 'onMount', 'onClick'])}
```
MichelML: can't you just use ttransform and get the props keys instead or we decided to stop using it ?

GermainBergeron: I can use ttransform

PR #917, opened 2018-11-06
Fix SelectConnected component's focus behavior

gdostie:
When rendering `SingleSelectConnected` in table rows, the underlying `SelectConnected` was loosing its focus somewhere in the life cycle making it impossible to use the keyboard to select items.

The solution was to fix the boolean check for `SelectConnected.keepFocus` by passing down a new prop called `hasFocusableChild` from the withFilter HOC. This way, the `keepFocus` variable stays to `true` for `SelectConnected` that don't have focusable children and is set to `false` for the ones that render focusable children.

Also did some minor refactoring while at it.

### DIFF
--- a/src/components/select/SelectConnected.tsx
+++ b/src/components/select/SelectConnected.tsx
@@ -25,10 +25,11 @@ export interface ISelectOwnProps {
     placeholder?: string;
     noResultItem?: IItemBoxProps;
     selectClasses?: string;
+    items?: IItemBoxProps[];
+    hasFocusableChild?: boolean;
 }
 
 export interface ISelectStateProps {
-    items?: IItemBoxProps[];
     isOpen?: boolean;
     active?: number;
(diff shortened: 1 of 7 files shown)


gdostie:
Had a look at the coverage report, the decrease in coverage is unrelated to the modified files.

src/components/select/SelectConnected.tsx
```diff
@@ -95,7 +96,7 @@ export class SelectConnected extends React.Component<ISelectProps & ISelectSpeci
 
     componentWillUpdate(nextProps: ISelectProps): any {
```
ThibodeauJF: won't `componentWillUpdate` be deprecated? It's a good opportunity here to use something else

pochretien: `componentDidUpdate ` can do the job

PR #922, opened 2018-11-12
handle maximum and minimum values (enable innate Input tag attributes…

TonyGreenwood:
… on React Wrapper class)

DEMO:
https://gfycat.com/FlamboyantUntidyCollie

### DIFF
--- a/src/components/input/Input.tsx
+++ b/src/components/input/Input.tsx
@@ -28,6 +28,8 @@ export interface IInputOwnProps {
     validateOnMount?: boolean;
     autoFocus?: boolean;
     disabledTooltip?: string;
+    minimum?: string;
+    maximum?: string;
 }
 
 export interface IInputStateProps {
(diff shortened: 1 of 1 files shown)


src/components/input/Input.tsx
```diff
@@ -28,6 +28,8 @@ export interface IInputOwnProps {
     validateOnMount?: boolean;
     autoFocus?: boolean;
     disabledTooltip?: string;
+    minimum?: string;
+    maximum?: string;
```
GermainBergeron: Shouldn't it be numbers?

TonyGreenwood: Some sites aren't too explicit about it but it's mostly used as a string
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input

GermainBergeron: React accept `number | string` but number seem a better fit IMO

TonyGreenwood: Done.

PR #928, opened 2018-11-13
Table with numbered rows

MichelML:
screen shot 2018-11-13 at 4 15 27 pm
  
i don't think it was HOC worthy, especially since renderData is custom to each table, but definitely open for changes here.  

you need to do three things to add numbered rows for now:

- Add <TableRowNumberHeader /> in the header
- Add <TableRowNumberColumn /> as a column in renderData
- add the .table-numbered class on the table  

maybe there would be something simpler, thoughts are welcome  

@otaillon


otaillon:
### Coooooooooooooooooooool stuff Michel !

- Is the box wider than in the mockup (60px vs 40px) to allow for bigger number ?
- Text color of the row number should be #5D7289
- There seems to be a weird display bug where the background color of the position column doesn't reach the row's border. There's like 1px of white
- Also the bottom border of the second column is weirdly cut. (See attached screenshot)
screen shot 2018-11-14 at 08 38 52

PR #930, opened 2018-11-13
implement table with numbered rows

MichelML:
screen shot 2018-11-13 at 4 15 27 pm  

https://coveo.github.io/react-vapor/table-with-numbered-rows/#Table%20(server%20+%20hoc)
  
i don't think it was HOC worthy, especially since renderData is custom to each table, but definitely open for changes here.  

you need to do three things to add numbered rows for now:

- Add <TableRowNumberHeader /> in the header
- Add <TableRowNumberColumn /> as a column in renderData
- add the .table-numbered class on the table  

maybe there would be something simpler, thoughts are welcome  

@otaillon

### DIFF
--- a/src/components/table-hoc/TableRowNumberColumn.tsx
+++ b/src/components/table-hoc/TableRowNumberColumn.tsx
@@ -1,0 +1,11 @@
+import * as classNames from 'classnames';
+import * as React from 'react';
+import * as styles from './styles/TableRowNumber.scss';
+
+export const TableRowNumberColumn = (props: {number: string | number}) => (
+    <td key='table-row-number' className={styles.tableNumberedRow}>
+        <div className={classNames('bg-grey-1 mod-border-right text-medium-blue flex flex-column flex-center center', styles.tableNumberedRowContainer)}>
+            {props.number}
+        </div>
+    </td>
+);
--- a/src/components/table-hoc/TableRowNumberColumn.tsx
+++ b/src/components/table-hoc/TableRowNumberColumn.tsx
@@ -4,7 +4,7 @@ import * as styles from './styles/TableRowNumber.scss';
 
 export const TableRowNumberColumn = (props: {number: string | number}) => (
     <td key='table-row-number' className={styles.tableNumberedRow}>
-        <div className={classNames('bg-grey-1 mod-border-right text-medium-blue flex flex-column flex-center center', styles.tableNumberedRowContainer)}>
+        <div className={classNames('bg-grey-1 mod-border-right text-lynch flex flex-column flex-center center', styles.tableNumberedRowContainer)}>
             {props.number}
         </div>
     </td>
--- a/src/components/table-hoc/TableRowNumberColumn.tsx
+++ b/src/components/table-hoc/TableRowNumberColumn.tsx
@@ -2,7 +2,7 @@ import * as classNames from 'classnames';
 import * as React from 'react';
 import * as styles from './styles/TableRowNumber.scss';
 
-export const TableRowNumberColumn = (props: {number: string | number}) => (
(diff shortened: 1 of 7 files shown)


otaillon:
### Coooooooooooooooooooool stuff Michel !

- Is the box wider than in the mockup (60px vs 40px) to allow for bigger number ?
- Text color of the row number should be #5D7289
- There seems to be a weird display bug where the background color of the position column doesn't reach the row's border. There's like 1px of white
- Also the bottom border of the second column is weirdly cut. (See attached screenshot)
screen shot 2018-11-14 at 08 38 52

src/components/table-hoc/TableRowNumberColumn.tsx
```diff
@@ -0,0 +1,11 @@
+import * as classNames from 'classnames';
+import * as React from 'react';
+import * as styles from './styles/TableRowNumber.scss';
+
+export const TableRowNumberColumn = (props: {number: string | number}) => (
```
pochretien: use ReactNode instead of a string or number

GermainBergeron: Or simply use props.children?

src/components/table-hoc/styles/TableRowNumber.scss
```diff
@@ -0,0 +1,27 @@
...
+     tr:global(.selected) td:nth-child(2) {
+        border-left-color: $orange;
+    }
+}
+
+:global(.coveo-styleguide .table) .tableNumberedRow:first-child {
+    padding:0;
+    border-left: none;
+    width: 60px;
+    height: 40px;
```
GermainBergeron: 60px and 40px should probably be variables

src/components/table-hoc/tests/TableRowNumberColumn.spec.tsx
```diff
@@ -0,0 +1,10 @@
+import {shallow} from 'enzyme';
+import * as React from 'react';
+import {TableRowNumberColumn} from '../TableRowNumberColumn';
+
+describe('TableRowNumberColumn', () => {
+    it('should render without error', () => {
+        expect(() => shallow(<TableRowNumberColumn number={1} />)).not.toThrow();
+        expect(() => shallow(<TableRowNumberColumn number={'10000'} />)).not.toThrow();
```
GermainBergeron: no need for the `{}` since you're using a string (but I think a children would be more flexible)

src/components/table-hoc/tests/TableRowNumberHeader.spec.tsx
```diff
@@ -0,0 +1,9 @@
+import {shallow} from 'enzyme';
+import * as React from 'react';
+import {TableRowNumberHeader} from '../TableRowNumberHeader';
+
+describe('TableRowNumberHeader', () => {
+    it('should render without error', () => {
+        expect(() => shallow(<TableRowNumberHeader />)).not.toThrow();
+    });
+});
```
GermainBergeron: it should contains a th?

src/components/table-hoc/TableRowNumberColumn.tsx
MichelML: thanks, on it

src/components/table-hoc/tests/TableRowNumberColumn.spec.tsx
MichelML: it's now a react node, but I added the children prop

src/components/table-hoc/tests/TableRowNumberHeader.spec.tsx
MichelML: 👍

PR #936, opened 2018-11-16
Added selectors for Table

GermainBergeron:
Moved filters selector in their own files, used them in the select selectors and in the table
Added Table selectors
Fixed file casing

### DIFF
--- a/docs/Index.tsx
+++ b/docs/Index.tsx
@@ -178,7 +178,7 @@ class App extends React.Component<{}, AppState> {
         {component: TableRowConnectedExamples, componentName: 'TableRowConnected'},
         {component: TableEmptyRowExamples, componentName: 'TableEmptyRow'},
         {component: TableHeaderExamples, componentName: 'TableHeader'},
-        {component: TableExamples, componentName: 'Table'},
+        {component: TableExamples, componentName: 'Table (Deprecated)'},
         {component: TableWithDisabledRowsExamples, componentName: 'TableWithDisabledRows'},
         {component: OptionsCycleExamples, componentName: 'OptionsCycle'},
         {component: OptionsCycleConnectedExamples, componentName: 'OptionsCycleConnected'},
--- a/src/components/filterBox/FilterBoxSelectors.ts
+++ b/src/components/filterBox/FilterBoxSelectors.ts
@@ -1,0 +1,27 @@
+import * as _ from 'underscore';
+import {IReactVaporState} from '../../ReactVapor';
+import {defaultMatchFilter, MatchFilter} from '../../utils/FilterUtils';
+import {IFilterState} from './FilterBoxReducers';
+
+export interface GetFilterTextProps {
+    id: string;
+}
+
+const getFilterText = (state: IReactVaporState, props: GetFilterTextProps): string => {
+    const filter: IFilterState = _.findWhere(state.filters, {id: props.id});
+    return (filter && filter.filterText) || '';
+};
+
+export interface GetMatchFilterTextProps {
+    matchFilter?: MatchFilter;
+}
+const getMatchFilter = (state: IReactVaporState, props: GetMatchFilterTextProps): MatchFilter => {
+    return _.isUndefined(props.matchFilter)
+        ? defaultMatchFilter
+        : props.matchFilter;
+};
+
+export const FilterBoxSelectors = {
+    getFilterText,
+    getMatchFilter,
+};
--- a/src/components/navigation/NavigationSelectors.ts
+++ b/src/components/navigation/NavigationSelectors.ts
@@ -1,0 +1,22 @@
+import * as _ from 'underscore';
+import {IReactVaporState} from '../../ReactVapor';
+import {PER_PAGE_NUMBERS} from './perPage/NavigationPerPage';
+
+export interface GetPaginationProps {
+    id: string;
+}
+
+const getPaginationPage = (state: IReactVaporState, props: GetPaginationProps): number => {
+    const pagination = _.findWhere(state.paginationComposite, {id: props.id});
+    return pagination && pagination.pageNb || 0;
+};
+
+const getPerPage = (state: IReactVaporState, props: GetPaginationProps): number => {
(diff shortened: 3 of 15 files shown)


docs/Index.tsx
```diff
@@ -178,7 +178,7 @@ class App extends React.Component<{}, AppState> {
         {component: TableRowConnectedExamples, componentName: 'TableRowConnected'},
         {component: TableEmptyRowExamples, componentName: 'TableEmptyRow'},
         {component: TableHeaderExamples, componentName: 'TableHeader'},
-        {component: TableExamples, componentName: 'Table'},
+        {component: TableExamples, componentName: 'Table (Deprecated)'},
```
MichelML: sadness

gdostie: She served us good, let her rest in peace. :dove:

src/components/filterBox/FilterBoxSelectors.ts
```diff
@@ -0,0 +1,27 @@
+import * as _ from 'underscore';
+import {IReactVaporState} from '../../ReactVapor';
+import {defaultMatchFilter, MatchFilter} from '../../utils/FilterUtils';
+import {IFilterState} from './FilterBoxReducers';
+
+export interface GetFilterTextProps {
+    id: string;
+}
+
+const getFilterText = (state: IReactVaporState, props: GetFilterTextProps): string => {
```
MichelML: are we suppose to follow this standard function signature for selectors?

MichelML: if so, could be nice to add this rule in tsjs about selectors

src/components/navigation/NavigationSelectors.ts
```diff
@@ -0,0 +1,22 @@
...
+}
+
+const getPaginationPage = (state: IReactVaporState, props: GetPaginationProps): number => {
+    const pagination = _.findWhere(state.paginationComposite, {id: props.id});
+    return pagination && pagination.pageNb || 0;
+};
+
+const getPerPage = (state: IReactVaporState, props: GetPaginationProps): number => {
+    const pagination = _.findWhere(state.perPageComposite, {id: props.id});
+    return pagination && pagination.perPage || PER_PAGE_NUMBERS[1];
```
MichelML: why not PER_PAGE_NUMBERS[0] ?

GermainBergeron: The default is `const defaultInitialPosition: number = Math.ceil(perPageNumber.length / 2) - 1;` from the NavigationPerPageConnected. I don't think it makes that much sense but I don't know how to do otherwise.

This will only apply if the pagination is not in the state, I could put 1 instead and it would still work

src/components/select/SelectSelector.ts
```diff
@@ -31,18 +24,18 @@ const getSelect = (state: IReactVaporState, ownProps: ISelectProps): ISelectStat
...
     matchFilter: MatchFilter,
 ): IItemBoxProps[] => _.map(items, (item: IItemBoxProps) => ({...item, hidden: !matchFilter(filterText, item) || !!item.hidden}));
 
 const getItemsWithFilter: (state: IReactVaporState, ownProps: ISelectWithFilterProps) => IItemBoxProps[] = createSelector(
-    [getItems, getFilterText, getMatchFilter],
+    [
+        getItems,
+        (state: IReactVaporState, ownProps: ISelectWithFilterProps) => FilterBoxSelectors.getFilterText(state, {id: ownProps.id}),
+        (state: IReactVaporState, ownProps: ISelectWithFilterProps) => FilterBoxSelectors.getMatchFilter(state, {matchFilter: ownProps.matchFilter}),
+    ],
```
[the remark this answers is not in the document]
GermainBergeron: Because I want to decouple the selector from the ISelectWithFilterProps, Typescript throws an error if the types are different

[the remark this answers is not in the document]
GermainBergeron: doesn't work with Typescript and I wanted to avoid creating a new selector which only call the FilterBoxSelector

src/components/filterBox/FilterBoxSelectors.ts
```diff
@@ -0,0 +1,27 @@
...
+};
+
+export interface GetMatchFilterTextProps {
+    matchFilter?: MatchFilter;
+}
+const getMatchFilter = (state: IReactVaporState, props: GetMatchFilterTextProps): MatchFilter => {
+    return _.isUndefined(props.matchFilter)
+        ? defaultMatchFilter
+        : props.matchFilter;
+};
```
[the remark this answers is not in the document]
GermainBergeron: I only moved it.

The method would exactly be the same , I wonder if we should differentiate utils form selectors

src/components/select/SelectSelector.ts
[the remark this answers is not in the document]
gdostie: then you can remove the array brackets

src/components/filterBox/FilterBoxSelectors.ts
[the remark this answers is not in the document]
pochretien: we can create a utils with that too but a little problem come with that, we cant use it like a selector for a combiner like 
```
const getItemsWithFilter: (state: IReactVaporState, ownProps: ISelectWithFilterProps) => IItemBoxProps[] = createSelector(
    [getItems, getFilterText, getMatchFilter],
    itemsWithFilterCombiner,
);
```
I think its legit to have a selector if its from the props objet and an other selector can use it with a combine. What do you think ?

src/components/filterBox/FilterBoxSelectors.ts
pochretien: I like it, only use the minimal object from the props instead of the full ownProps. avoid a problem with the interface late if we need this selector with an other component

src/components/navigation/NavigationSelectors.ts
pochretien: If my memory is good, the default per page is 50, Its a good idea to set with 1.

PR #937, opened 2018-11-16
Update many dependencies

gdostie:


package.json
```diff
@@ -85,21 +85,21 @@
...
     "axios": "0.18.0",
     "bithound": "1.7.0",
     "chosen-js": "1.8.2",
     "classnames": "2.2.6",
     "codecov": "3.1.0",
-    "codemirror": "5.40.2",
-    "coveo-styleguide": "3.6.x",
-    "css-loader": "1.0.0",
+    "codemirror": "5.41.0",
+    "coveo-styleguide": "3.6.28",
```
GermainBergeron: we should revert this

gdostie: Why should we?

PR #954, opened 2018-12-03
add better type for withEditing hoc

MichelML:
screen shot 2018-12-03 at 1 56 03 pm
  

good catch from @tommy-tran , the hoc was returning the wrong prop signature. You can now pass the props interface and state interface to the hoc if needed.


MichelML:
line 28 and 29 starting from this https://github.com/coveo/react-vapor/pull/954/files#diff-89afe580193fe1cb373d11e342971923R28 is the actual change

PR #965, opened 2018-12-07
Add input value selector

pochretien:
Add uts
Remove useless duplicate for IInputState


src/components/input/Input.tsx
```diff
@@ -49,10 +50,6 @@ export interface IInputDispatchProps {
 
 export interface IInputProps extends IInputOwnProps, IInputStateProps, IInputDispatchProps {}
 
-export interface IInputState {
-    valid: boolean;
-}
```
GermainBergeron: I think it make sense to have a State interface here since we don't have all of the attributes from the Redux state. Maybe we should name it `IInputComponentState` to avoid confusion

PR #971, opened 2018-12-10
Feature InfoBox Component

GermainBergeron:
image

### DIFF
--- a/src/components/infoBox/tests/InfoBoxLink.spec.tsx
+++ b/src/components/infoBox/tests/InfoBoxLink.spec.tsx
@@ -1,0 +1,25 @@
+import {shallow} from 'enzyme';
+import * as React from 'react';
+import {InfoBoxLink} from '../InfoBoxLink';
+
+describe('InfoBoxLink', () => {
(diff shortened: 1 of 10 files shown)


src/components/infoBox/tests/InfoBoxLink.spec.tsx
```diff
@@ -0,0 +1,25 @@
...
+        const wrapper = shallow(<InfoBoxLink className={expectedClass} />);
+
+        expect(wrapper.hasClass(expectedClass)).toBe(true);
+    });
+
+    it('should render children', () => {
+        const expectedChildren = <div className='to-find' />;
+        const wrapper = shallow(<InfoBoxLink>{expectedChildren}</InfoBoxLink>);
+
+        expect(wrapper.find('.to-find').exists()).toBe(true);
```
gdostie: Very minor:
You can expect wrappers to contain jsx elements directly instead of relying on classes. I find it cleaner in general.
```ts
expect(wrapper.contains(expectedChildren)).toBe(true);
```

PR #975, opened 2018-12-13
change section on hashchange in docs

MichelML:
navigate via the url and the section should change accordingly

### DIFF
--- a/docs/Index.tsx
+++ b/docs/Index.tsx
@@ -287,6 +287,14 @@ class App extends React.Component<{}, AppState> {
         document.addEventListener(SideNavigation.toggleEvent, () => {
             this.setState({sideNavOpened: !this.state.sideNavOpened});
         });
+
+        window.onhashchange = () => {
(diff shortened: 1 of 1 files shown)


docs/Index.tsx
```diff
@@ -287,6 +287,14 @@ class App extends React.Component<{}, AppState> {
         document.addEventListener(SideNavigation.toggleEvent, () => {
             this.setState({sideNavOpened: !this.state.sideNavOpened});
         });
+
+        window.onhashchange = () => {
+            const componentIdFromHash = this.getHash();
+            const firstComponentId = this.components.sort(this.sortComponentsByName);
+            this.setState({
+                activeComponentId: this.getSelectedComponent(componentIdFromHash) && componentIdFromHash || firstComponentId[0].componentName,
+            });
```
GermainBergeron: You're pretty much duplicating the constructor, you should probably extract a method to get the componentId from the url

MichelML: changed

PR #976, opened 2018-12-13
Fix Tooltip Throws an Error When the Element is Removed from the DOM

GermainBergeron:


src/components/tooltip/Tooltip.tsx
```diff
@@ -76,18 +77,35 @@ const OVERLAY_PROPS_TO_OMIT: string[] = [
     'container',
 ];
 
-export class Tooltip extends React.Component<ITooltipProps, {}> {
+export class Tooltip extends React.Component<ITooltipProps> {
     static defaultProps: Partial<ITooltipProps> = {
         className: '',
     };
 
+    private readonly overlay: React.RefObject<BootstrapTooltip>;
```
gdostie: I like the idea of putting refs in private readonly, good job

src/components/tooltip/Tooltip.tsx
```diff
@@ -76,18 +77,35 @@ const OVERLAY_PROPS_TO_OMIT: string[] = [
...
 
+    private readonly overlay: React.RefObject<BootstrapTooltip>;
+
+    constructor(props: ITooltipProps) {
+        super(props);
+        this.overlay = React.createRef();
+    }
+
+    componentWillUnmount() {
+        if (this.overlay && this.overlay.current) {
```
gdostie: No need to check for `this.overlay` it will always exists as it is instanciated in the constructor and is a readonly prop.

MichelML:
very nice, thanks @GermainBergeron

PR #978, opened 2018-12-18
single select fixed width to prevent dropdown width to change when changing value

MichelML:
https://coveo.github.io/react-vapor/singleselect-fixed-width/#SingleSelect

screen shot 2018-12-18 at 3 20 54 pm

### DIFF
--- a/src/components/select/SingleSelectConnected.tsx
+++ b/src/components/select/SingleSelectConnected.tsx
@@ -9,6 +9,7 @@ import {Content} from '../content/Content';
 import {IItemBoxProps} from '../itemBox/ItemBox';
 import {ISelectButtonProps, ISelectProps, SelectConnected} from './SelectConnected';
 import {SelectSelector} from './SelectSelector';
+import * as styles from './styles/SingleSelect.scss';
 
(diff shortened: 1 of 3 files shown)


src/components/select/SingleSelectConnected.tsx
```diff
@@ -9,13 +9,15 @@ import {Content} from '../content/Content';
...
 import {SelectSelector} from './SelectSelector';
+import * as styles from './styles/SingleSelect.scss';
 
 export interface ISingleSelectOwnProps extends ISelectProps {
     placeholder?: string;
     toggleClasses?: string;
     onSelectOptionCallback?: (option: string) => void;
     items?: IItemBoxProps[];
     buttonPrepend?: React.ReactNode;
+    fixedWidth?: boolean;
```
gdostie: I think it would be clearer if we did the inverse logic: the optional prop would be called `unfixedWidth` or a synonym and would have to be set to true if we want the dropdown to have an unset width. This would remove the need of setting a defaultProp

MichelML: done !

PR #984, opened 2018-12-21
Single select connected selected display value

gdostie:
I needed a way to display a different value for a specific item when it is selected in the `SingleSelectConnected`.

Example:
image

image


GermainBergeron:
Maybe add a UT?

PR #985, opened 2018-12-21
Feature Deep PartialStringMatch

GermainBergeron:
Replace wholeString with a React.ReactNode
Iterate on every children to match string by using a generator
Added example

image

### DIFF
--- a/src/components/partial-string-match/PartialStringMatch.tsx
+++ b/src/components/partial-string-match/PartialStringMatch.tsx
@@ -3,7 +3,7 @@ import * as React from 'react';
 import {escape} from 'underscore';
 
 export interface PartialStringMatchProps {
-    wholeString: string;
+    wholeString: React.ReactNode;
     partialMatch?: string;
     caseInsensitive?: boolean;
 }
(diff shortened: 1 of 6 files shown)


src/components/partial-string-match/PartialStringMatch.tsx
```diff
@@ -15,16 +15,50 @@ export class PartialStringMatch extends React.Component<PartialStringMatchProps>
...
+        const els = [];
+        while (!res.done) {
+            els.push(res.value);
+            res = it.next();
+        }
+
+        return els;
+    }
+
+    private *deepReplaceStrings(component: React.ReactNode | React.ReactNode[], regExp: RegExp): IterableIterator<React.ReactNode> {
```
gdostie: Do you really need the `| React.ReactNode[]` ?

Taken out of definitely typed code: 
image

src/components/partial-string-match/PartialStringMatch.tsx
```diff
@@ -3,7 +3,7 @@ import * as React from 'react';
 import {escape} from 'underscore';
 
 export interface PartialStringMatchProps {
-    wholeString: string;
+    wholeString: React.ReactNode;
```
gdostie: Now that `PartialStringMatch` can parse react nodes, I think it would be cleaner to parse its children. That being said, we can keep the wholeString prop for retro compatibility, but I think we should parse the children also. What do you think?

src/components/partial-string-match/PartialStringMatch.tsx
GermainBergeron: Removed, it was a fragment of the past

src/components/partial-string-match/PartialStringMatch.tsx
GermainBergeron: Yeah, it would be better I think. I'll try to make something

src/components/partial-string-match/PartialStringMatch.tsx
gdostie: a ReactFragment from the past

PR #988, opened 2018-12-29
Update single select listbox visible active element

pochretien:
image


src/components/listBox/ListBox.tsx
```diff
@@ -12,6 +12,7 @@ export interface IListBoxOwnProps {
     multi?: boolean;
     highlight?: string;
     items?: IItemBoxProps[];
+    ulElementRefFunction?: (menu: HTMLElement) => void;
```
gdostie: You probably can use the new react 16 feature of ref forwarding instead of using a ref function (see https://reactjs.org/docs/forwarding-refs.html)

PR #996, opened 2019-01-15
adjust popover to be shown properly in modals

MichelML:
screen shot 2019-01-15 at 3 35 24 pm


MichelML:
the latter @GermainBergeron 👍

PR #1000, opened 2019-01-18
fix: adding class "flippable-ie" to flippables when the browser is IE

gdostie:
- Added `BrowserUtils.isIE()`. It detects if the browser is IE (FYI: returns false in MsEdge);
- Flippables will now have the class `flippable-ie` when the browser is IE;


gdostie:
Related to https://github.com/coveo/vapor/pull/622

PR #1006, opened 2019-01-25
Fixed an issue where the element currently selected is not highlighte…

dinulade101:
…d in the dropdown menu when assigning access level to a resource.

Removed code which identified the active element incorrectly.
Copied logic from 'selected'

### DIFF
--- a/src/components/select/tests/SingleSelectWithFilter.spec.tsx
+++ b/src/components/select/tests/SingleSelectWithFilter.spec.tsx
@@ -179,7 +179,7 @@ describe('Select', () => {
                     .simulate('keydown', {keyCode: keyCode.enter})
                     .simulate('keyup', {keyCode: keyCode.enter});
 
-                expect(dispatchSpy).toHaveBeenCalledWith(selectListBoxOption(id, undefined, items[0].value));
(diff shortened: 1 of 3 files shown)


GermainBergeron:
It seems like it breaks the keyboard navigation on both the SingleSelect and the MultiSelect

dinulade101:
@GermainBergeron yes sure!

src/components/select/tests/SingleSelectWithFilter.spec.tsx
```diff
@@ -179,7 +179,7 @@ describe('Select', () => {
                     .simulate('keydown', {keyCode: keyCode.enter})
                     .simulate('keyup', {keyCode: keyCode.enter});
 
-                expect(dispatchSpy).toHaveBeenCalledWith(selectListBoxOption(id, undefined, items[0].value));
+                expect(dispatchSpy).toHaveBeenCalledWith(selectListBoxOption(id, undefined, items[1].value));
```
MichelML:
```
const expectedActiveItem = _.findWhere(items, {selected: true});
expect(dispatchSpy).toHaveBeenCalledWith(selectListBoxOption(id, undefined, expectedActiveElement);
```

would be more resilient, if someone refactor items for any reason, there is less chance that this test will break with this approach ⬆️

PR #1010, opened 2019-01-28
add possibility to disable the multiselect dropdown

MichelML:


src/components/select/MultiSelectConnected.tsx
```diff
@@ -66,7 +66,7 @@ const parentDropTarget = {
 @DropTarget(DraggableSelectedOptionType, parentDropTarget, (connect: any) => ({
     connectDropTarget: connect.dropTarget(),
 }))
-export class MultiSelectConnected extends React.Component<IMultiSelectProps, {}> {
+export class MultiSelectConnected extends React.Component<IMultiSelectProps & React.ButtonHTMLAttributes<HTMLButtonElement>> {
```
pochretien: Nice call. to help with other project who want to use the IMultiSelectProps interface, we need to do this
`interface IMultiSelectProps extends React.ButtonHTMLAttributes<HTMLButtonElement>`

MichelML: changed, thanks

src/components/select/MultiSelectConnected.tsx
```diff
@@ -19,7 +19,10 @@ import {Tooltip} from '../tooltip/Tooltip';
 import {ISelectButtonProps, ISelectProps, SelectConnected} from './SelectConnected';
 import {SelectSelector} from './SelectSelector';
 
-export interface IMultiSelectOwnProps extends ISelectProps, IDropTargetProps {
+export interface IMultiSelectOwnProps extends
```
pochretien: GG

src/components/select/SelectConnected.tsx
```diff
@@ -21,7 +21,7 @@ export interface ISelectSpecificProps {
 }
 
 export interface ISelectOwnProps {
-    id: string;
+    id?: string;
```
GermainBergeron: I'm pretty sure the id is required

MichelML: changed

PR #1012, opened 2019-01-29
Always show label with input

pochretien:
image
Input without label but the validation message is showed

### DIFF
--- a/src/components/input/Input.tsx
+++ b/src/components/input/Input.tsx
@@ -61,6 +61,7 @@ export class Input extends React.Component<IInputProps, IInputComponentState> {
         id: uniqueId('input'),
         type: 'text',
         valid: true,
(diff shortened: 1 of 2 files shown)


src/components/input/tests/Input.spec.tsx
```diff
@@ -1,205 +1,260 @@
...
-            const id = 'yo';
-            expect(input.find('input').first().prop('id')).not.toBe(id);
-
-            input.setProps({id});
-            expect(input.find('input').first().prop('id')).toBe(id);
+            const id = 'plume';
+            shallowInput({
+                id,
+            });
+            expect(input.find('input').first().props().id).toBe(id);
```
GermainBergeron: Personally I prefer the syntax `.prop('id')` instead of `.props().id`

PR #1013, opened 2019-01-29
Implement doubleclick behavior on table rows and disabled style

gdostie:
- `TableRowConnected` with prop `disabled` set to true will now output a `tr` with the `row-disabled` class. It won't prevent any action from being triggered, if you want such behavior you should handle it in your callback;
- `TableRowConnected` will now trigger all of its actions that have the prop `callOnDoubleClick` set to true;

### DIFF
--- a/src/components/table-hoc/TableRowConnected.tsx
+++ b/src/components/table-hoc/TableRowConnected.tsx
@@ -29,6 +29,7 @@ export interface ITableRowOwnProps {
     actions?: IActionOptions[];
     isMultiselect?: boolean;
     collapsible?: CollapsibleRowProps;
+    disabled?: boolean;
 }
 
 export interface ITableRowStateProps {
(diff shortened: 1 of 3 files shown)


src/components/table-hoc/TableRowConnected.tsx
```diff
@@ -171,6 +174,12 @@ class TableRowConnected extends React.PureComponent<ITableRowConnectedProps & Re
             this.props.onClick(isMulti);
         }
     }
+
+    private handleDoubleClick = () => {
+        _(this.props.actions)
+            .filter((action: IActionOptions) => action.callOnDoubleClick)
+            .forEach((action: IActionOptions) => action.trigger());
```
MichelML: is there really a case where we want more than one action to be triggered ?

MichelML: clean style by the way, you could do it without underscore too

gdostie: I don't know if there really is a use case for it, but that's the same logic we implemented on the previous table component.

PR #1015, opened 2019-01-29
Bind the code editor to its value prop

gdostie:
The code editor now uses the `Controlled` version of reac-codemirror-2 which allows to have more control over how it handles its state. I fixed it so that its state changes both when the user inputs something, but also when its value changes programatically.


MichelML:
i guess you meant `programatically` @gdostie

gdostie:
yes haha I used chrome spell checker without paying attention

PR #1018, opened 2019-01-31
add onItemClick prop for actionable items

MichelML:


src/components/actionable-item/ActionableItem.tsx
```diff
@@ -1,25 +1,36 @@
...
 import {IItemBoxProps} from '../itemBox/ItemBox';
 import {ListBox} from '../listBox/ListBox';
 import {PopoverConnected} from '../popover/PopoverConnected';
 import {Svg} from '../svg/Svg';
 import {actionableItemContainer, actionableItemContent, actionableItemDots, actionableItemDotsWidthWithPadding} from './styles/ActionableItem.scss';
 
-export interface IActionableItemProps extends React.HTMLAttributes<HTMLDivElement> {
+export interface IActionableItemProps {
     id: string;
+    onItemClick?: (e?: React.MouseEvent<HTMLDivElement>) => any;
```
pochretien: you can return void

MichelML: thanks, you're right, both are equivalent in this case i'd say

PR #1025, opened 2019-02-05
Add MultiineBoxContainer hoc

pochretien:
add interface for new tag
add examples

image

### DIFF
--- a/src/components/multilineBox/MultilineBox.tsx
+++ b/src/components/multilineBox/MultilineBox.tsx
@@ -9,7 +9,7 @@ import {UUID} from '../../utils/UUID';
 import {IMultiSelectOwnProps} from '../select/MultiSelectConnected';
 import {MultilineBoxSelectors} from './MultilineBoxSelector';
 
-export interface IMultilineSingleBoxProps<T> {
+export interface IMultilineSingleBoxProps<T = any> {
     id: string;
     isLast: boolean;
     props: T;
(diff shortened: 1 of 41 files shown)


src/components/multilineBox/MultilineBox.tsx
```diff
@@ -100,21 +104,23 @@ export class MultilineBox<T> extends React.PureComponent<IMultilineBoxProps<T>>
         };
     }
 
+    componentDidUpdate(prevProps: Readonly<IMultilineBoxProps<T>>, prevState: Readonly<IMultilineBoxState<T>>) {
+        if (prevState && JSON.stringify(this.state.initialData) !== JSON.stringify(prevState.initialData)) {
```
GermainBergeron: `if (prevState && !_.isEqual(this.state.initialData, prevState.initialData)) {`

pochretien:
image

src/reusableState/customList/StringListReducers.ts
```diff
@@ -54,6 +54,14 @@ export const stringListCompositeReducer = (
                     list: [..._.without(state[action.payload.id].list, action.payload.value)],
                 },
             };
+        case StringListActions.updateValues:
```
gdostie: `StringListActions.add` is almost identical, I'm pretty sure you can combine them so that they use the same logic.

pochretien: Good point. I added this action to avoid confusion when you use the action in your dispatch. I can be confusion to use an add when your action its supposed to only update an existing object.

pochretien: Fixed

PR #1026, opened 2019-02-05
Add multiline box remove button hoc

pochretien:
Add example for multiple use case
add the new multilinebox remove button hoc

image


src/components/multilineBox/examples/MultilineBoxExamples.tsx
```diff
@@ -23,6 +24,43 @@ const MultilineBoxWithContainer = _.compose(
...
+const MultilineBoxWithContainerAndTwoRemoveButton = _.compose(
+    multilineBoxWithRemoveButton(),
+    multilineBoxContainer({
+        container: {
+            tag: 'div',
+            props: {
+                className: 'mod-border p1',
+            },
+        },
+    }),
```
GermainBergeron: I'm looking at this and wouldn't it be easier to directly set the element like this:

```typescript
const MultilineBoxWithContainerAndTwoRemoveButton = _.compose(
    multilineBoxWithRemoveButton(),
    multilineBoxContainer((props: theCorrectType) => (<div className='mod-border p1'>{props.children}</div>),
...
```

src/components/multilineBox/examples/MultilineBoxExamples.tsx
```diff
@@ -23,6 +24,43 @@ const MultilineBoxWithContainer = _.compose(
...
+    multilineBoxWithRemoveButton(),
```
GermainBergeron: You could do the same thing for the remove button (but with a default value that render your actual svg)

src/components/multilineBox/examples/MultilineBoxExamples.tsx
pochretien: Updated with a global variable for each example

pochretien:
Examples Updated
image

PR #1030, opened 2019-02-06
Disable dropdowns when one available value and already selected

dinulade101:
Singleselect and Multiselect will automatically be disabled when there are no items to select in dropdown. For example, if the only items in the dropdown are disabled, dropdown will be disabled.


MichelML:
I don't understand why we would want this?

PR #1045, opened 2019-02-18
Use the react-modal lib for our modals

gdostie:
I refactored our ModalComposite and friends components to leverage [the react-modal lib](https://www.npmjs.com/package/react-modal). My main need was that our modals get rendered through portals, because I had display issues in the admin and this lib takes care of that challenge easily. It is also a very popular lib that handles modals in a dumbproof way and it is full accessibility compliant n stuff. 

This change is non-breaking and backward compatible so all existing modals should still work perfectly.

Associated style changes: https://github.com/coveo/vapor/pull/638


gdostie:
Note that it is important to set globally the modal portal root where the modals will be rendered by default.
example:
```ts
ReactModal.setAppElement('#App');
```

PR #1047, opened 2019-02-19
Implement a way to handle global defaults

gdostie:
Needed a way to set global variables outside the scope of the lib. I implemented the `defaults` class. Let me know what you think about it.

### DIFF
--- a/src/Defaults.ts
+++ b/src/Defaults.ts
@@ -1,25 +1,8 @@
 import * as ReactModal from 'react-modal';
 
-let _MODAL_ROOT: string = 'body';
-let _MODAL_TIMEOUT: number = 300;
-
-/* istanbul ignore next */
 export abstract class Defaults {
-    static get MODAL_ROOT(): string {
-        return _MODAL_ROOT;
-    }
-
-    static set MODAL_ROOT(root: string) {
-        _MODAL_ROOT = root;
-    }
-
-    static get MODAL_TIMEOUT(): number {
-        return _MODAL_TIMEOUT;
-    }
-
-    static set MODAL_TIMEOUT(timeout: number) {
-        _MODAL_TIMEOUT = timeout;
-    }
+    static MODAL_ROOT: string = 'body';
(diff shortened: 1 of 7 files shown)


src/Defaults.ts
```diff
@@ -0,0 +1,28 @@
+import * as ReactModal from 'react-modal';
+
+let _MODAL_ROOT: string = 'body';
+let _MODAL_TIMEOUT: number = 300;
+
+// tslint:disable-next-line:class-name
+/* istanbul ignore next */
+export abstract class defaults {
```
GermainBergeron: Why not `Defaults`?

src/Defaults.ts
```diff
@@ -0,0 +1,28 @@
...
+// tslint:disable-next-line:class-name
+/* istanbul ignore next */
+export abstract class defaults {
+    static get MODAL_ROOT(): string {
+        return _MODAL_ROOT;
+    }
+
+    static set MODAL_ROOT(root: string) {
+        _MODAL_ROOT = root;
+    }
```
GermainBergeron: What about 
```typescript
export abstract class Defaults {
  static MODAL_ROOT = 'body';

  static MODAL_TIMEOUT = 300;

  static set APP_ELEMENT(appElement: string | HTMLElement) {
    ReactModal.setAppElement(appElement);
   }
}
```

GermainBergeron: I think you'll be able to remove the tslint and istanbul disable

gdostie: The goal is that a user of the react-vapor lib will be able to set the modal root and modal timeout for all modals to whatever he wants. That's why I used getters and setters.

The modal root and the app element are 2 different things:
>The app element should not be a parent of the modal, to prevent modal content from being hidden to screenreaders while it is open.

https://reactcommunity.org/react-modal/#custom-parent

src/Defaults.ts
gdostie: No specific reason

PR #1050, opened 2019-02-20
Fix ReduxUtils generic types

gdostie:
`tslint` was complaining with our custom `IDispatch` and `IThunkAction` types. I refactored our types so that they use the proper types from redux-thunk.


gulpfile.js
```diff
@@ -49,7 +49,6 @@ gulp.task('ts:definitions', 'Generate the project definition file', (done) => {
 
 gulp.task('internalDefs', false, () =>
     dtsGenerator.default({
-        name: 'ReactVapor',
```
GermainBergeron: Did you try to load the dts in another repo?

PR #1051, opened 2019-02-21
Fix Select with Filter Highlight

GermainBergeron:
image


gdostie:
All SingleSelects are broken in the demo

GermainBergeron:
Fixed by #1058

PR #1057, opened 2019-02-25
Multiline box with remove button uts

pochretien:


src/components/multilineBox/tests/MultilineBoxWithRemoveButton.spec.tsx
```diff
@@ -0,0 +1,165 @@
...
+        )(MultilineBox);
+
+        it('should mount without errors', () => {
+            expect(() => shallowWithState(
+                <DefaultMultilineBoxWithRemoveButton
+                    id={id}
+                    data={[]}
+                    renderBody={_.identity}
+                    defaultProps={{}}
+                />, {}).dive().dive().dive(),
```
GermainBergeron: It would be nice to add a function to shallow with the props to avoid the triple dive everywhere

PR #1058, opened 2019-02-26
Feature Single Select Clear Button

GermainBergeron:
Added props on the SingleSelect to add a clear button and a tooltip
Added unit tests
Fixed demo for single select and Menu

image

### DIFF
--- a/src/components/select/SingleSelectConnected.tsx
+++ b/src/components/select/SingleSelectConnected.tsx
@@ -1,4 +1,5 @@
 import * as classNames from 'classnames';
+import * as VaporSVG from 'coveo-styleguide';
 import * as React from 'react';
 import * as _ from 'underscore';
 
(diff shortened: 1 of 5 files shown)


src/components/select/SingleSelectConnected.tsx
```diff
@@ -103,8 +117,17 @@ export class SingleSelectConnected extends React.Component<ISingleSelectProps &
 
         return <span className='dropdown-no-value'>{this.props.placeholder}</span>;
     }
+
+    private getDeselectOptionButton(): React.ReactNode {
+        return (
+            <Tooltip title={this.props.deselectTooltipText} placement='top' noSpanWrapper onClick={this.props.deselect}>
+                <Svg svgName='clear' svgClass='icon mod-12' className='btn-append center-align' />
```
pochretien: VaporSVG.svg.clear.name

src/components/select/SingleSelectConnected.tsx
```diff
@@ -18,15 +22,19 @@ export interface ISingleSelectOwnProps extends ISelectProps {
     items?: IItemBoxProps[];
     buttonPrepend?: React.ReactNode;
     noFixedWidth?: true;
+    canClear?: true;
```
gdostie: With the new typing file working properly, tslint cries over using `true` as type in common repo. I think we should revert to using `boolean` and for `noFixedWidth` also.

PR #1060, opened 2019-02-27
Multiline box with dnd uts

pochretien:


src/utils/tests/RTestUtils.tsx
```diff
@@ -0,0 +1,56 @@
...
+                    <ComponentClass {...props} >
+                        {child}
+                    </ComponentClass>
+                </Provider>
+            );
+        }
+    }
+
+    const Draggable = DD(Tester);
+    const ret = mount(
```
GermainBergeron: Maybe use a better name:

```suggestion
    const wrapper = mount(
```

src/utils/tests/RTestUtils.tsx
```diff
@@ -0,0 +1,56 @@
...
+        <Draggable {...props} />,
+    );
+
+    return {
+        ret,
```
GermainBergeron:
```suggestion
        wrapper,
```

PR #1062, opened 2019-02-28
Feature More Banner Flexibility

GermainBergeron:
Added BannerContainer for when you need something simple
Added unit tests
Added example
Fixed style issue

### DIFF
--- a/src/components/banner/BannerContainer.tsx
+++ b/src/components/banner/BannerContainer.tsx
@@ -4,7 +4,7 @@ import * as _ from 'underscore';
 
 import * as styles from './styles/Banner.scss';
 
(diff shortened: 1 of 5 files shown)


src/components/banner/BannerContainer.tsx
```diff
@@ -0,0 +1,14 @@
+import * as classNames from 'classnames';
+import * as React from 'react';
+import * as _ from 'underscore';
+
+import * as styles from './styles/Banner.scss';
+
+export const BannerContainer: React.FunctionComponent<React.HTMLProps<HTMLDivElement>> = (props: React.HTMLProps<HTMLDivElement>) => (
```
gdostie: IIRC you don't need to re-type props

GermainBergeron: You're right

PR #1065, opened 2019-03-01
Fix input component interface

pochretien:


src/components/input/tests/Input.spec.tsx
```diff
@@ -230,12 +230,12 @@ describe('<Input />', () => {
             });
         });
 
-        it('should set the step prop to any if the input is of type number', () => {
+        it('should set the step prop to undefined if the input is of type number', () => {
```
GermainBergeron: I think I would delete this test

src/utils/react/PropsToOmitUtils.ts
```diff
@@ -0,0 +1,8 @@
+const internalComponentPropsToOmit: string[] = ['children', 'dangerouslySetInnerHTML'];
+
+const inputTagPropsToOmit: string[] = [...internalComponentPropsToOmit, 'defaultValue', 'onClick', 'onChange', 'onBlur', 'value', 'valid'];
```
GermainBergeron: I think I would add a small comment about why those props are to omit something in the line of `We omit those props because when the input was created we didn't use the correct types.`

samisayegh:
Thanks for this. When will you release a new version containing the changes?

PR #1075, opened 2019-03-12
Fixed Snyk vulnerabilities

GermainBergeron:
Test to use a sh script to deploy
Replaced gulp with version 4
Removed bithound


package.json
```diff
@@ -114,10 +113,8 @@
     "eslint-plugin-react": "7.12.0",
     "faker": "4.1.0",
     "file-loader": "3.0.1",
-    "gulp": "3.9.1",
+    "gulp": "4.0.0",
     "gulp-footer": "2.0.2",
-    "gulp-help": "1.6.1",
-    "gulp-sequence": "1.0.0",
     "gulp-util": "3.0.8",
```
gdostie: I think it would be best to remove gulp-util while we're at it

package.json
```diff
@@ -114,10 +113,8 @@
     "eslint-plugin-react": "7.12.0",
     "faker": "4.1.0",
     "file-loader": "3.0.1",
-    "gulp": "3.9.1",
+    "gulp": "4.0.0",
     "gulp-footer": "2.0.2",
```
gdostie: IICR gulp-footer is not used in the gulpfile anymore so we should remove it.

PR #1076, opened 2019-03-13
Prototype: Tooltip HOC + dropPod Utils

pochretien:
Add DropPod for the portal element
Add tooltipHOC
Add tool to calcula position for the tooltips with a portal
Add examples

image
image
image

image

image

### DIFF
--- a/src/components/drop/Drop.tsx
+++ b/src/components/drop/Drop.tsx
@@ -2,8 +2,8 @@ import * as React from 'react';
 import * as ReactDOM from 'react-dom';
 import {DropPod} from './DropPod';
 
-export interface ITooltipHOCProps {
-    buttonContainer?: React.HTMLAttributes<HTMLDivElement>;
+export interface IDropProps {
+    buttonContainer?: React.HTMLProps<HTMLDivElement>;
     listContainer?: React.HTMLAttributes<HTMLDivElement>;
     renderOpenButton: (onClick: () => void) => React.ReactNode;
     closeOnClick?: boolean;
@@ -11,18 +11,18 @@ export interface ITooltipHOCProps {
     selector?: string;
 }
 
-export interface ITooltipHOCState {
+export interface IDropState {
     isOpen: boolean;
 }
 
-export class TooltipHOC extends React.PureComponent<ITooltipHOCProps, ITooltipHOCState> {
+export class Drop extends React.PureComponent<IDropProps, IDropState> {
     readonly button: React.RefObject<HTMLDivElement>;
 
-    static defaultProps: Partial<ITooltipHOCProps> = {
+    static defaultProps: Partial<IDropProps> = {
         closeOnClick: true,
     };
 
-    constructor(props: ITooltipHOCProps, state: ITooltipHOCState) {
+    constructor(props: IDropProps, state: IDropState) {
         super(props, state);
 
         this.button = React.createRef();
@@ -46,7 +46,7 @@ export class TooltipHOC extends React.PureComponent<ITooltipHOCProps, ITooltipHO
                     ref={this.button}
                     {...this.props.buttonContainer}
                 >
-                    {this.props.renderOpenButton(() => this.onClick())}
+                    {this.props.renderOpenButton(this.onClick)}
                 </div>
                 {this.createPortalMenu()}
             </>
@@ -58,9 +58,9 @@ export class TooltipHOC extends React.PureComponent<ITooltipHOCProps, ITooltipHO
             <DropPod
                 positions={this.props.positions}
                 isOpen={this.state.isOpen}
-                buttonRef={this.button}
-                renderTooltip={(style: React.CSSProperties, containerRef: React.RefObject<HTMLDivElement>) => (
-                    <div style={style} ref={containerRef} {...this.props.listContainer}>
+                ref={this.button}
+                renderTooltip={(style: React.CSSProperties, dropRef: React.RefObject<HTMLDivElement>): React.ReactNode => (
+                    <div style={style} ref={dropRef} {...this.props.listContainer}>
                         {this.props.children}
                     </div>
                 )}
@@ -68,8 +68,8 @@ export class TooltipHOC extends React.PureComponent<ITooltipHOCProps, ITooltipHO
         );
     }
 
-    private onClick() {
-        this.setState((state) => ({
+    private onClick = () => {
+        this.setState((state: IDropState) => ({
             isOpen: !state.isOpen,
         }));
     }
--- a/src/components/drop/Drop.tsx
+++ b/src/components/drop/Drop.tsx
@@ -3,9 +3,9 @@ import * as ReactDOM from 'react-dom';
 import {DropPod} from './DropPod';
 
 export interface IDropProps {
-    buttonContainer?: React.HTMLProps<HTMLDivElement>;
-    listContainer?: React.HTMLAttributes<HTMLDivElement>;
     renderOpenButton: (onClick: () => void) => React.ReactNode;
+    buttonContainerProps?: React.HTMLProps<HTMLDivElement>;
+    listContainerProps?: React.HTMLAttributes<HTMLDivElement>;
     closeOnClick?: boolean;
     positions?: string[];
     selector?: string;
@@ -44,12 +44,12 @@ export class Drop extends React.PureComponent<IDropProps, IDropState> {
             <>
                 <div
                     ref={this.button}
-                    {...this.props.buttonContainer}
+                    {...this.props.buttonContainerProps}
                 >
                     {this.props.renderOpenButton(this.onClick)}
                 </div>
                 {this.createPortalMenu()}
-            </>
+            </React.Fragment>
         );
     }
 
@@ -60,7 +60,7 @@ export class Drop extends React.PureComponent<IDropProps, IDropState> {
                 isOpen={this.state.isOpen}
                 ref={this.button}
                 renderTooltip={(style: React.CSSProperties, dropRef: React.RefObject<HTMLDivElement>): React.ReactNode => (
-                    <div style={style} ref={dropRef} {...this.props.listContainer}>
+                    <div style={style} ref={dropRef} {...this.props.listContainerProps}>
                         {this.props.children}
                     </div>
                 )}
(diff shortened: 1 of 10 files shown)


src/components/tooltips-hoc/DropPod.tsx
```diff
@@ -0,0 +1,90 @@
...
+        super(props, state);
+
+        this.tooltip = React.createRef();
+        this.state = {
+            offset: undefined,
+        };
+    }
+
+    componentDidMount() {
+        window.addEventListener('resize', this.updateOffset as any);
```
gdostie: Can you remove `as any` ?

src/components/tooltips-hoc/DropPod.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+export interface IDropPodState {
+    offset: ClientRect | DOMRect;
+}
+
+export class DropPod extends React.PureComponent<IDropPodProps, IDropPodState> {
+    readonly tooltip: React.RefObject<HTMLElement>;
+
+    static defaultProps: Partial<IDropPodProps> = {
+        selector: '#App',
```
gdostie: You should use the same kind of logic we did for the modals to specify the modal root. Have a look at `src/Defaults.ts`.

src/components/tooltips-hoc/DropPod.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+        const visibility = !_.isEmpty(style) && this.props.isOpen ? 'visible' : 'hidden';
+        return {
+            ...style,
+            visibility,
+        };
+    }
+
+    render() {
+        const tooltip: React.ReactNode = this.props.renderTooltip({
```
gdostie: This looks a lot like a ref forwarding use case. 
Have a look at https://reactjs.org/docs/forwarding-refs.html#forwarding-refs-to-dom-components.

I think it looks pretty good like this:
```tsx
const TheTooltip = React.forwardRef(this.props.renderTooltip);

return ReactDOM.createPortal(
    <TheTooltip ref={this.tooltip} />, 
    document.querySelector(this.props.selector),
);
```

src/components/tooltips-hoc/TooltipHOC.tsx
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    render() {
+        return (
+            <>
+                <div
+                    ref={this.button}
+                    {...this.props.buttonContainer}
+                >
+                    {this.props.renderOpenButton(() => this.onClick())}
```
gdostie: By using `() => this.onClick()`, you will create a new handler on each render.
To avoid that, you can simply use `this.props.renderOpenButton(this.onClick)` here and add `this.onClick.bind(this);` in the constructor.

src/components/tooltips-hoc/DropPod.tsx
pochretien: I have a ts error about my return interface, I dont know why because its the same everywhere... If I can, I will but for now, I dont know how hahah

src/components/tooltips-hoc/positionValidation.ts
```diff
@@ -0,0 +1,72 @@
+import {CSSProperties} from 'react';
+
+export const positionValidator: {[position: string]: (boxOffset: ClientRect | DOMRect, tooltipOffset: ClientRect | DOMRect, window: Window) => Partial<CSSProperties>} = {
```
gdostie: FIle should be named `postitionValidator` or `PositionValidator` (same name as exported const).

src/components/tooltips-hoc/DropPod.tsx
pochretien: I want to let the user choose where he want to put the DropPod, Iike the bootstrap tooltip. I will try to set something logic instead of App by default

src/components/tooltips-hoc/DropPod.tsx
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    private getStyle(): React.CSSProperties {
+        let style: Partial<React.CSSProperties> = {};
+        if (this.tooltip.current && this.state.offset) {
+            const tooltipOffset: ClientRect | DOMRect = this.tooltip.current.getBoundingClientRect();
+
+            let index = 0;
+            while (_.isEmpty(style) && index < this.props.positions.length) {
+                style = positionValidator[this.props.positions[index]](this.state.offset, tooltipOffset, window) || {};
```
GermainBergeron: I think you should validate that `positionValidator[this.props.positions[index]]` exists

```suggestion
                const validator = positionValidator[this.props.positions[index]];
                style = validator && validator(this.state.offset, tooltipOffset, window) || {};
```

src/components/tooltips-hoc/DropPod.tsx
```diff
@@ -0,0 +1,90 @@
...
+        const tooltip: React.ReactNode = this.props.renderTooltip({
+            position: 'absolute',
+            display: 'inline-block',
+            ...this.getStyle(),
+        }, this.tooltip);
+
+        return ReactDOM.createPortal((
+            <>
+                {tooltip}
+            </>
```
GermainBergeron: Do you really need to return a fragment or you can do

`return ReactDOM.createPortal(tooltip, document.querySelector(this.props.selector));`

src/components/tooltips-hoc/TooltipHOC.tsx
```diff
@@ -0,0 +1,94 @@
+import * as React from 'react';
+import * as ReactDOM from 'react-dom';
+import {DropPod} from './DropPod';
+
+export interface ITooltipHOCProps {
+    buttonContainer?: React.HTMLAttributes<HTMLDivElement>;
+    listContainer?: React.HTMLAttributes<HTMLDivElement>;
```
GermainBergeron:
```suggestion
    listContainerProps?: React.HTMLAttributes<HTMLDivElement>;
```

src/components/tooltips-hoc/TooltipHOC.tsx
```diff
@@ -0,0 +1,94 @@
+import * as React from 'react';
+import * as ReactDOM from 'react-dom';
+import {DropPod} from './DropPod';
+
+export interface ITooltipHOCProps {
+    buttonContainer?: React.HTMLAttributes<HTMLDivElement>;
```
GermainBergeron:
```suggestion
    buttonContainerProps?: React.HTMLAttributes<HTMLDivElement>;
```

src/components/tooltips-hoc/TooltipHOC.tsx
```diff
@@ -0,0 +1,94 @@
+import * as React from 'react';
+import * as ReactDOM from 'react-dom';
+import {DropPod} from './DropPod';
+
+export interface ITooltipHOCProps {
+    buttonContainer?: React.HTMLAttributes<HTMLDivElement>;
+    listContainer?: React.HTMLAttributes<HTMLDivElement>;
+    renderOpenButton: (onClick: () => void) => React.ReactNode;
```
GermainBergeron: The required props should appear at the top of the interface

src/components/tooltips-hoc/DropPod.tsx
```diff
@@ -0,0 +1,90 @@
...
+    left: 'left',
+    right: 'right',
+};
+
+export interface IDropPodProps {
+    isOpen: boolean;
+    positions?: string[];
+    buttonRef: React.RefObject<HTMLElement>;
+    renderTooltip: (style: React.CSSProperties, containerRef: React.RefObject<HTMLElement>) => React.ReactNode;
+    selector: string;
```
GermainBergeron: This should be optional since you have a default

```suggestion
    selector?: string;
```

src/components/tooltips-hoc/DropPod.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+export const DropPodPosition = {
+    bottom: 'bottom',
+    top: 'top',
+    left: 'left',
+    right: 'right',
+};
+
+export interface IDropPodProps {
+    isOpen: boolean;
```
GermainBergeron: This should be optional since you have a default

```suggestion
    isOpen?: boolean;
```

src/components/tooltips-hoc/TooltipHOC.tsx
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    componentWillUnmount() {
+        document.removeEventListener('mousedown', this.handleDocumentClick);
+    }
+
+    render() {
+        return (
+            <>
+                <div
```
GermainBergeron: Do you really need a div or you could simply put 
```typescript
render() {
  return (
    <>
      {this.props.renderOpenButton(() => this.onClick())}
      {this.createPortalMenu()}
    </>
  );
```

GermainBergeron:
Demo looks good 🎉

src/components/tooltips-hoc/DropPod.tsx
gdostie: Same for modals, the user of the lib will want to set one root for all tooltips I guess. they can do `Defaults.tooltipRoot = '#Whatever'`

src/components/drop/Drop.tsx
```diff
@@ -58,18 +58,18 @@ export class TooltipHOC extends React.PureComponent<ITooltipHOCProps, ITooltipHO
             <DropPod
                 positions={this.props.positions}
                 isOpen={this.state.isOpen}
-                buttonRef={this.button}
-                renderTooltip={(style: React.CSSProperties, containerRef: React.RefObject<HTMLDivElement>) => (
-                    <div style={style} ref={containerRef} {...this.props.listContainer}>
+                ref={this.button}
+                renderTooltip={(style: React.CSSProperties, dropRef: React.RefObject<HTMLDivElement>): React.ReactNode => (
```
gdostie: `renderDrop` ?

src/components/tooltips-hoc/DropPod.tsx
pochretien: Its a state, we can let him always because its not specified on the component attributes

src/components/tooltips-hoc/TooltipHOC.tsx
pochretien: I need my ref. Its necessary to calcul the offset

PR #1077, opened 2019-03-14
DropFromDropdown + example + drop connected

pochretien:
image


src/components/drop/hoc/DropWithDropdown.tsx
```diff
@@ -0,0 +1,41 @@
...
+import {DropPodPosition} from '../DropPod';
+import {DropSelectors} from '../redux/DropReducers';
+
+type TooltipWithDropdownComponent<T = any> = React.ComponentClass<IDropProps>;
+
+const mapStateToProps = (state: IReactVaporState, {id}: IDropOwnProps): IDropStateProps => ({
+    isOpen: DropSelectors.isOpen(state, {id}),
+});
+
+export const dropFormDropdown = (supplier: ConfigSupplier = {}) => (Component: TooltipWithDropdownComponent): TooltipWithDropdownComponent => {
```
gdostie: The name of the file and the name of the exported class should be the same.

src/components/drop/hoc/DropWithDropdown.tsx
```diff
@@ -0,0 +1,41 @@
...
+import {SlideY} from '../../../animations/SlideY';
+import {IReactVaporState} from '../../../ReactVapor';
+import {ConfigSupplier} from '../../../utils/HocUtils';
+import {PropsToOmitUtils} from '../../../utils/react/PropsToOmitUtils';
+import {ReduxConnect} from '../../../utils/ReduxUtils';
+import {IDropOwnProps, IDropProps, IDropStateProps} from '../Drop';
+import {DropPodPosition} from '../DropPod';
+import {DropSelectors} from '../redux/DropReducers';
+
+type TooltipWithDropdownComponent<T = any> = React.ComponentClass<IDropProps>;
```
gdostie: Should probably renamed to include `drop`

src/components/drop/hoc/DropWithDropdown.tsx
gdostie: Not sure I like the name drop from dropdown, could we simply use `Dropdown` ?

src/components/drop/redux/DropReducers.ts
```diff
@@ -0,0 +1,33 @@
+import {IReactVaporState} from '../../../ReactVapor';
+import {IReduxAction} from '../../../utils/ReduxUtils';
+import {DropReducerActions, IDropPayload} from './DropActions';
+
+export interface IDropCompositeState extends IDropPayload {}
+
+export const dropInitialState: IDropCompositeState = {id: undefined, isOpen: false};
+
+export const dropCompositeReducer = (
```
gdostie: I think `composite` doesn't add any value to our understanding, it should simply be named `dropReducer`

src/components/drop/Drop.tsx
```diff
@@ -69,21 +91,17 @@ export class Drop extends React.PureComponent<IDropProps, IDropState> {
...
-        }));
+        this.props.toggle(true);
     }
 
     private onClickMenu() {
         if (this.props.closeOnClick) {
-            this.setState({
-                isOpen: false,
-            });
+            this.props.toggle(true);
```
GermainBergeron: false?

src/components/drop/hoc/DropWithDropdown.tsx
GermainBergeron: Or dropWithDropdown

GermainBergeron:
> I don't understand why we need yet another dropdown like component. IMO the `Drop` HOC should be used under the hood of the components where we need to have the Drop behavior, like the SingleSelect. The new actions and reducer created here feel a lot like a duplicate of what we already have somewhere. Or maybe I'm missing something

I think that PO wants to drop the other action/reducers in things like the SingleSelect to reuse this one everywhere

src/components/drop/Drop.tsx
```diff
@@ -1,34 +1,57 @@
...
+const mapDispatchToProps = (
+    dispatch: IDispatch,
+    ownProps: IDropOwnProps,
+): IDropDispatchProps => ({
+    toggle: (isOpen?: boolean) => dispatch(DropActions.toggle(ownProps.id, ownProps.group, isOpen)),
+});
+
+@ReduxConnect(mapStateToProps, mapDispatchToProps)
+export class Drop extends React.PureComponent<IDropProps> {
     readonly button: React.RefObject<HTMLDivElement>;
```
GermainBergeron:
```suggestion
    private readonly button: React.RefObject<HTMLDivElement>;
```

src/components/drop/Drop.tsx
```diff
@@ -1,34 +1,57 @@
...
+    private dropRef: React.RefObject<HTMLDivElement>;
```
GermainBergeron:
```suggestion
    private readonly dropRef: React.RefObject<HTMLDivElement>;
```

src/components/drop/Drop.tsx
```diff
@@ -1,34 +1,57 @@
...
-        closeOnClick: true,
-    };
+    static defaultProps: Partial<IDropProps>;
 
-    constructor(props: IDropProps, state: IDropState) {
-        super(props, state);
+    constructor(props: IDropProps) {
+        super(props);
 
         this.button = React.createRef();
```
GermainBergeron:
```suggestion
        this.button = React.createRef();
        this.dropRef = React.createRef();
```

src/components/drop/Drop.tsx
```diff
@@ -56,11 +79,12 @@ export class Drop extends React.PureComponent<IDropProps, IDropState> {
...
             <DropPod
-                positions={this.props.positions}
-                isOpen={this.state.isOpen}
                 ref={this.button}
+                isOpen={this.props.isOpen}
+                positions={this.props.positions}
+                selector={this.props.selector}
                 renderDrop={(style: React.CSSProperties, dropRef: React.RefObject<HTMLDivElement>): React.ReactNode => (
-                    <div style={style} ref={dropRef} {...this.props.listContainerProps}>
+                    <div style={style} ref={this.dropRef = dropRef} className={classNames('show-on-top', this.props.listContainerProps.className)} {...this.props.listContainerProps} >
```
GermainBergeron:
```suggestion
                    <div style={style} ref={this.dropRef} className={classNames('show-on-top', this.props.listContainerProps.className)} {...this.props.listContainerProps} >
```

src/components/drop/redux/DropReducers.ts
```diff
@@ -0,0 +1,57 @@
...
+                    isOpen: action.payload.isOpen,
+                },
+            };
+        default:
+            return state;
+    }
+};
+
+const isDropOpen = (state: IReactVaporState, props: {id: string, group: string}): boolean =>
+    props.group && state.drop[props.group] && state.drop[props.group].id === props.id ? state.drop[props.group].isOpen : false;
```
gdostie: You could extract a `getDrop` selector from this, I think it could be useful somewhere and it makes the code more readable:
```ts
const getDrop = (state: IReactVaporState: {groupId}: {groupId: string = ''}): IDropState => 
    state.drop ? state.drop[groupId] : undefined;

const isDropOpen = (state: IReactVaporState, {id, groupId}: {id: string = '', groupId: string = ''}): boolean => {
    const drop: IDropState = getDrop(state, groupId); 
    return drop ? drop.id === id && drop.isOpen : false;
}

export const DropSelectors = {
    isOpen: isDropOpen,
    getDrop,
};
```

src/components/drop/Drop.tsx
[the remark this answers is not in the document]
pochretien: I dont want to create a ref here but only a variable to have a reference on it. The initialization for the variable is already done in the dropPod constructor

Comment added in createPortalMenu to explain why I need dropRef as a reference

src/components/drop/Drop.tsx
```diff
@@ -1,34 +1,57 @@
...
+    id: string;
     renderOpenButton: (onClick: () => void) => React.ReactNode;
     buttonContainerProps?: React.HTMLProps<HTMLDivElement>;
     listContainerProps?: React.HTMLAttributes<HTMLDivElement>;
-    closeOnClick?: boolean;
-    positions?: string[];
-    selector?: string;
+    closeOnClickOutside?: boolean;
+    closeOnClickDrop?: boolean;
+    group?: string;
```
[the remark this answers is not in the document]
pochretien: de we need to add id or group is good to understand the principe of drop by group.

If not, I can add id to avoid confusion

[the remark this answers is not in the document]
gdostie: Its just by convention, everywhere else we use `id` so I think its best if we continue in that way to avoid confusion.
Also in the admin we already have a notion of groups and default groups so it would add in to the confusion.

src/components/drop/Drop.tsx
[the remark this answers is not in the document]
gdostie: That's my point... by doing it here, you can remove the constructor and it does exactly the same thing.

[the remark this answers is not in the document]
pochretien: I tried and its not, I lost my ref on it if I did that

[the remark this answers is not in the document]
gdostie: Sorry you need to use this, my bad
```ts
private dropRef = React.createRef<HTMLDivElement>();
```

PR #1080, opened 2019-03-19
Feature HOC Table Date Picker

GermainBergeron:
Added examples in normal & connected tables
Added prop to add custom tooltip on filter in the table
Added selectors for the datepicker

image


francoislg:
Stumbled upon this in your live demo: 

image

src/components/table-hoc/TableWithDatePicker.tsx
```diff
@@ -0,0 +1,87 @@
...
+                <DatePickerDropdownConnected
+                    {...config as any}
+                    id={this.props.id}
+                    key={this.props.id}
+                    className={classNames('coveo-table-actions', config.className)}
+                    onRight
+                />
+            );
+            const newActions = [...(this.props.actions || []), datePickerAction];
+            const newProps = _.omit(this.props, [...TableWithFilterPropsToOmit]);
```
gdostie:
```suggestion
            const newProps = _.omit(this.props, ...TableWithFilterPropsToOmit);
```

src/components/table-hoc/TableWithDatePicker.tsx
```diff
@@ -0,0 +1,87 @@
...
+            const datePickerAction = (
+                <DatePickerDropdownConnected
+                    {...config as any}
+                    id={this.props.id}
+                    key={this.props.id}
+                    className={classNames('coveo-table-actions', config.className)}
+                    onRight
+                />
+            );
+            const newActions = [...(this.props.actions || []), datePickerAction];
```
gdostie:
```suggestion
            const newActions = [...this.props.actions, datePickerAction];
```
Use defaultProps ?

src/components/datePicker/DatePickerSelectors.ts
```diff
@@ -0,0 +1,22 @@
...
+        if (datePicker.id.indexOf(props.id) === 0) {
+            return datePicker;
+        }
+    }).filter(Boolean);
+    return datePickers && datePickers[0] || null;
+};
+
+const getDatePickerLimits = (state: IReactVaporState, props: {id: string}): [Date, Date?] => {
+    const picker = DatePickerSelectors.getDatePicker(state, props);
+    return picker ? [picker.appliedLowerLimit, picker.appliedUpperLimit] : [null, null];
```
pochretien: just to be sure, you have a interface with a required Date as first param but you return null with your selector
Do you think its will be better to return somthing like [0, null] juste to avoid a Date at null ?

src/components/table-hoc/TableWithDatePicker.tsx
```diff
@@ -0,0 +1,90 @@
...
+const defaultMatchDates = () => true;
+
+export const tableWithDatePicker = (supplier: ConfigSupplier<ITableWithDatePickerConfig> = {}) => (Component: FilterableTableComponent): FilterableTableComponent => {
+
+    const mapStateToProps = (state: IReactVaporState, ownProps: ITableWithDatePickerProps): ITableWithDatePickerStateProps | ITableHOCOwnProps => {
+        const config = HocUtils.supplyConfig(supplier);
+        const [lowerLimit, upperLimit] = DatePickerSelectors.getDatePickerLimits(state, {id: ownProps.id});
+        const matchDates = config.matchDates || defaultMatchDates;
+        const filterData = () => lowerLimit
+            ? _.filter(ownProps.data, (datum: any) => matchDates(datum, lowerLimit, upperLimit))
```
pochretien: We need a validation for lowerLimit because you can have a null returned by the selector. I think we can avoid to call the filter if the lowerLimit is null

PR #1081, opened 2019-03-19
Allow to use StickyFooter outside of the project

francoislg:
`.stickyFooterOpened` was compiled, so there was no direct way to use this class.

Furthermore, the component was *closed* by default, meaning that the only way we could use the component was by replicating the `stickyFooterOpened` into our project and applying it at all times.

I instead fronted a simple property to add this class when required.

### DIFF
--- a/src/components/stickyFooter/StickyFooter.tsx
+++ b/src/components/stickyFooter/StickyFooter.tsx
@@ -1,17 +1,25 @@
 import * as classNames from 'classnames';
 import * as React from 'react';
 import * as styles from './StickyFooter.scss';
+import { IClassName } from 'src/utils/ClassNameUtils';
 
 export interface IStickyFooterProps {
-    className?: string;
+    id?: string;
+    className?: IClassName;
+    isOpened: boolean;
 }
 
 export class StickyFooter extends React.Component<IStickyFooterProps> {
     static ID = 'StickyFooter';
 
     render() {
+        const { id, className, isOpened } = this.props;
         return (
-            <div id={StickyFooter.ID} className={classNames(styles.stickyFooter, this.props.className)}>
+            <div id={id || StickyFooter.ID} className={
+                classNames({
+                    [styles.stickyFooterOpened]: isOpened,
+                    [styles.stickyFooter]: true
+                }, className)}>
                 {this.props.children}
             </div>
         );
--- a/src/components/stickyFooter/StickyFooter.tsx
+++ b/src/components/stickyFooter/StickyFooter.tsx
@@ -10,11 +10,14 @@ export interface IStickyFooterProps {
 
 export class StickyFooter extends React.Component<IStickyFooterProps> {
     static ID = 'StickyFooter';
+    static defaultProps: Partial<IStickyFooterProps> = {
+        id: StickyFooter.ID
+    }
 
     render() {
         const {id, className, isOpened} = this.props;
         return (
-            <div id={id || StickyFooter.ID} className={classNames(styles.stickyFooter, {[styles.stickyFooterOpened]: isOpened}, className)}>
+            <div id={id} className={classNames(styles.stickyFooter, {[styles.stickyFooterOpened]: isOpened}, className)}>
                 {this.props.children}
             </div>
         );
--- a/src/components/stickyFooter/StickyFooter.tsx
+++ b/src/components/stickyFooter/StickyFooter.tsx
@@ -1,25 +1,20 @@
 import * as classNames from 'classnames';
 import * as React from 'react';
 import * as styles from './StickyFooter.scss';
-import { IClassName } from 'src/utils/ClassNameUtils';
 
 export interface IStickyFooterProps {
-    id?: string;
-    className?: IClassName;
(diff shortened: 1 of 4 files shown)


francoislg:
Forgot the UTs 🤦‍♂️, incoming

src/components/stickyFooter/StickyFooter.tsx
```diff
@@ -1,17 +1,25 @@
 import * as classNames from 'classnames';
 import * as React from 'react';
 import * as styles from './StickyFooter.scss';
+import { IClassName } from 'src/utils/ClassNameUtils';
 
 export interface IStickyFooterProps {
-    className?: string;
+    id?: string;
+    className?: IClassName;
+    isOpened: boolean;
```
GermainBergeron: isOpened should be the first prop of the interface since it's required

src/components/stickyFooter/StickyFooter.tsx
```diff
@@ -1,17 +1,25 @@
 import * as classNames from 'classnames';
 import * as React from 'react';
 import * as styles from './StickyFooter.scss';
+import { IClassName } from 'src/utils/ClassNameUtils';
 
 export interface IStickyFooterProps {
-    className?: string;
+    id?: string;
+    className?: IClassName;
```
GermainBergeron: To stay coherent with React className should be a string

src/components/stickyFooter/StickyFooter.tsx
```diff
@@ -1,17 +1,25 @@
...
 
 export class StickyFooter extends React.Component<IStickyFooterProps> {
     static ID = 'StickyFooter';
 
     render() {
+        const { id, className, isOpened } = this.props;
         return (
-            <div id={StickyFooter.ID} className={classNames(styles.stickyFooter, this.props.className)}>
+            <div id={id || StickyFooter.ID} className={
+                classNames({
```
GermainBergeron: `classNames(styles.stickyFooter, {[styles.stickyFooterOpened]: isOpened}, className)`

src/components/stickyFooter/examples/StickyFooterExamples.tsx
```diff
@@ -21,7 +20,7 @@ export class StickyFooterExamples extends React.Component<{}, {isOpened: boolean
             <div className='mt2'>
                 <Button name='toggle footer' onClick={() => this.setState({isOpened: !this.state.isOpened})} />
                 <div className='mt2'>{lorem}</div>
-                <StickyFooter className={classNames('sticky-footer-mod-header', {[styles.stickyFooterOpened]: this.state.isOpened})}>
+                <StickyFooter className={'sticky-footer-mod-header'} isOpened={this.state.isOpened} >
```
GermainBergeron:
```suggestion
                <StickyFooter className='sticky-footer-mod-header' isOpened={this.state.isOpened} >
```

src/components/stickyFooter/tests/StickyFooter.spec.tsx
```diff
@@ -1,16 +1,34 @@
...
+import * as styles from '../StickyFooter.scss';
 
 describe('StickyFooter', () => {
     it('should render without error', () => {
-        expect(() => shallow(<StickyFooter />)).not.toThrow();
-        expect(() => shallow(<StickyFooter className='someclass' />)).not.toThrow();
+        expect(() => shallow(<StickyFooter isOpened={true} />)).not.toThrow();
+        expect(() => shallow(<StickyFooter className='someclass' isOpened={true} />)).not.toThrow();
+        expect(() => shallow(<StickyFooter id='oyeah' isOpened={true} />)).not.toThrow();
     });
```
GermainBergeron: You should add tests for isOpened={false} or simply `expect(() => shallow(<StickyFooter />)).not.toThrow();`

src/components/stickyFooter/StickyFooter.tsx
```diff
@@ -3,15 +3,18 @@ import * as React from 'react';
...
 }
 
 export class StickyFooter extends React.Component<IStickyFooterProps> {
     static ID = 'StickyFooter';
 
     render() {
+        const {id, className, isOpened} = this.props;
         return (
-            <div id={StickyFooter.ID} className={classNames(styles.stickyFooter, this.props.className)}>
+            <div id={id || StickyFooter.ID} className={classNames(styles.stickyFooter, {[styles.stickyFooterOpened]: isOpened}, className)}>
```
gdostie:
```suggestion
            <div id={id} className={classNames(styles.stickyFooter, {[styles.stickyFooterOpened]: isOpened}, className)}>
```
I think it'd be more future-proof to use defaultProps instead. Imagine a case where you would need to reuse the `id` prop somewhere else, you would need to repeat this logic. 

Something like the following would solve this issue:
```ts
static defaultProps: Partial<IStickyFooterProps> = {
    id: StickyFooter.ID,
}
```

francoislg:
☝️ you guys OK with merging this? :)

GermainBergeron:
Yes! 🌮

PR #1087, opened 2019-03-20
Workaround issue react-modal/issues/745

gdostie:
Workaround for https://github.com/reactjs/react-modal/issues/745

### DIFF
--- a/src/components/modal/ModalComposite.tsx
+++ b/src/components/modal/ModalComposite.tsx
@@ -38,7 +38,10 @@ const modalPropsToOmit = keys<IModalCompositeProps>();
 export class ModalComposite extends React.PureComponent<IModalCompositeProps & Partial<ReactModal.Props>> {
     static defaultProps: Partial<IModalCompositeProps> = {
         id: _.uniqueId('modal'),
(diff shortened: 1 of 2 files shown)


src/components/modal/ModalComposite.tsx
```diff
@@ -112,4 +115,13 @@ export class ModalComposite extends React.PureComponent<IModalCompositeProps & P
     )
 
     private getParent = (): HTMLElement => document.querySelector(Defaults.MODAL_ROOT);
+
+    private handleRequestClose = () => {
+        // Workaround for https://github.com/reactjs/react-modal/issues/745
+        // Once issue is fixed, use onAfterClose={this.props.closeCallback} instead
+        callIfDefined(this.props.onClose);
```
pochretien: do you think Its a good idea to use the onClose to return a boolean if we need or not to do the closeCallback ?

gdostie: I don't really see the added value. If you have a use case where you don't want the `closeCallback` prop to be called afterClose, just don't provide the `closeCallback` prop to the `ModalComposite`.

PR #1090, opened 2019-03-26
Drop improvement

pochretien:
Improvement on event handling, no more event when the drop is not open to avoid useless process on the other on the page
Add new props to set the same width for the drop than the button
add little task to create a js file not minified to test in local with link
fix a little miss variable where the number was sent instead of a boolean ( work too because a number is true but fixed now )
Add uts
Fix uts with the old event system

### DIFF
--- a/src/components/drop/Drop.tsx
+++ b/src/components/drop/Drop.tsx
@@ -59,13 +59,15 @@ export class Drop extends React.PureComponent<IDropProps> {
         this.removeEventOnClickOnDocument();
     }
 
-    render() {
-        if (this.props.isOpen) {
+    componentDidUpdate(prevProps: Readonly<IDropProps>) {
+        if (!prevProps.isOpen && this.props.isOpen) {
             this.setEventOnClickOnDocument();
-        } else {
+        } else if (prevProps.isOpen && !this.props.isOpen) {
             this.removeEventOnClickOnDocument();
         }
+    }
 
+    render() {
         return (
             <>
                 <div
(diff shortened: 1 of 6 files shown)


pochretien:
Change the event `mousedown` with `click` because we by pass the click event on child element if we catch the mousedown before ( event ordering )

GermainBergeron:
There is a new miss, I don't know if it's easy to fix?

src/components/drop/Drop.tsx
```diff
@@ -52,17 +52,20 @@ export class Drop extends React.PureComponent<IDropProps> {
...
-        document.removeEventListener('mousedown', this.handleDocumentClick);
+        this.removeEventOnClickOnDocument();
     }
 
     render() {
+        if (this.props.isOpen) {
+            this.setEventOnClickOnDocument();
+        } else {
+            this.removeEventOnClickOnDocument();
+        }
```
gdostie: This if block should be done inside `componentDidUpdate`, because you will be setting a new event handler on each render if `this.props.isOpen` remains true.

src/components/drop/DropPod.tsx
```diff
@@ -134,6 +153,12 @@ class RDropPod extends React.PureComponent<IRDropPodProps, IDropPodState> {
     }
 
     render() {
+        if (this.props.isOpen) {
+            this.setEventsOnDocument();
+        } else {
+            this.removeEventsOnDocument();
+        }
```
gdostie: Same here. This if block should be done inside `componentDidUpdate`, because you will be setting a new event handler on each render if `this.props.isOpen` remains true.

pochretien:
coverage fixed

PR #1098, opened 2019-04-03
Hide SingleSelect clear button when disabled

gdostie:
The clear button was shown and was also effective even when the SingleSelect was disabled.

What I did:
- Prevented the clear button from showing when the SingleSelect is disabled
- Prevented the clear single select value action to be dispatched onclick when the SingleSelect is disabled

### DIFF
--- a/src/components/listBox/ListBox.tsx
+++ b/src/components/listBox/ListBox.tsx
@@ -1,6 +1,7 @@
 import * as classNames from 'classnames';
 import * as React from 'react';
 import * as _ from 'underscore';
+
 import {mod} from '../../utils/DataStructuresUtils';
 import {callIfDefined} from '../../utils/FalsyValuesUtils';
 import {IItemBoxProps, ItemBox} from '../itemBox/ItemBox';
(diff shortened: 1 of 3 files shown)


src/components/listBox/ListBox.tsx
```diff
@@ -83,9 +84,14 @@ export class ListBox extends React.Component<IListBoxProps, {}> {
             />)
             .value();
 
-        return items.length
-            ? items
-            : <ItemBox {...{...this.props.noResultItem, classes: ['multi-line']}} />;
+        const emptyItem = (
```
pochretien: We can allowed a ReactNode as props for the empty box I think, Its more flexible if we want a specific design not related to a ItemBox

src/components/select/SingleSelectConnected.tsx
```diff
@@ -121,11 +121,17 @@ export class SingleSelectConnected extends React.Component<ISingleSelectProps &
...
     private getDeselectOptionButton(): React.ReactNode {
         return (
-            <Tooltip title={this.props.deselectTooltipText} placement='top' noSpanWrapper onClick={this.props.deselect}>
+            <Tooltip title={this.props.deselectTooltipText} placement='top' noSpanWrapper onClick={this.handleDeselect}>
                 <Svg svgName={VaporSVG.svg.clear.name} svgClass='icon mod-12' className='btn-append center-align' />
             </Tooltip>
         );
     }
+
+    private handleDeselect = () => {
```
pochretien: return your event to deselect if we do this

gdostie: The `deselect` prop isn't typed to accept an event as argument, and the event isn't used in the `mapDispatchToProps` anyway.

src/components/listBox/ListBox.tsx
gdostie: I like the idea, not sure if it is a breaking change though, I will try it.

gdostie: It involves too much changes and its probably breaking, so I think its out of the scope of this fix. Would be nice to do though.

PR #1099, opened 2019-04-03
Improve footer component

pochretien:


src/hoc/withEditing/withEditing.tsx
```diff
@@ -23,41 +23,40 @@ export interface IWithEditDispatchProps {
...
 
-        return ComponentWithEditing;
-    };
+        componentWillUnmount() {
+            this.props.toggleDirtyComponent(false);
+        }
+
+        render() {
+            return (
+                <>
```
pochretien: only change <div/> for a Fragment

GermainBergeron: Can/did you run lintfix?

PR #1102, opened 2019-04-10
Update drop case out of the box

pochretien:
Update how the bounding limit with the drop projection is calculated.
The projection is always calculated if a new one is not available to be able to show the drop sticky with the bounding limit.


src/components/drop/DomPositionCalculator.ts
```diff
@@ -0,0 +1,162 @@
+import {CSSProperties} from 'react';
```
pochretien: add `dropUIPosition: IDropUIPosition` for each position to calculate the next position if the button position do not have space to render drop inside the bounding limit

src/components/drop/DomPositionCalculator.ts
```diff
@@ -0,0 +1,183 @@
...
+    left: 'left',
+    right: 'right',
+};
+
+export const OrientationByPosition: {[Position: string]: string[]} = {
+    bottom: [DropPodPosition.left, DropPodPosition.right],
+    top: [DropPodPosition.left, DropPodPosition.right],
+    left: [DropPodPosition.top, DropPodPosition.bottom],
+    right: [DropPodPosition.top, DropPodPosition.bottom],
+};
```
gdostie: By typing explicitly the const, you loose auto-completion, you could simply let typescript do the work here. Also, you could reuse the DropPodPosition to make sure there is no typo.
```ts
export const OrientationByPosition = {
    [DropPodPosition.bottom]: [DropPodPosition.left, DropPodPosition.right],
    [DropPodPosition.top]: [DropPodPosition.left, DropPodPosition.right],
    [DropPodPosition.left]: [DropPodPosition.top, DropPodPosition.bottom],
    [DropPodPosition.right]: [DropPodPosition.top, DropPodPosition.bottom],
};
```

PR #1104, opened 2019-04-11
Expose toast props in add toast action

gdostie:
I needed a way to pass down a class to the toast created with addToast action.

### DIFF
--- a/src/components/toast/ToastReducers.ts
+++ b/src/components/toast/ToastReducers.ts
@@ -10,12 +10,8 @@ export interface IToastsState {
 
 export interface IToastState {
     id: string;
-    title: string;
-    type?: string;
-    dismiss?: number;
-    dismissible?: boolean;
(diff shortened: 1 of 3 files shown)


src/components/toast/ToastReducers.ts
```diff
@@ -10,12 +10,8 @@ export interface IToastsState {
...
 export interface IToastState {
     id: string;
-    title: string;
-    type?: string;
-    dismiss?: number;
-    dismissible?: boolean;
-    animate?: boolean;
-    content?: React.ReactNode;
+    title: React.ReactNode;
+    [toastProp: string]: any;
```
GermainBergeron: Can you do something like `IToastState extends IToastProps` instead ? Otherwise I think we will lose a lot of type completion with this

If it doesn't work maybe something like `[toastProp: KeyOf<IToastProps>]: any;`

PR #1111, opened 2019-04-18
Feature Charts

GermainBergeron:
I added basic classes to create a responsive `<svg />` and different elements

In a next PR I'll add unit tests and a tooltip

image


package.json
```diff
@@ -55,6 +55,7 @@
   },
   "peerDependencies": {
     "codemirror": "^5.39.2",
+    "d3": "^3.5.17",
```
pochretien: remove ^ ?

src/components/chart/ChartContainer.tsx
```diff
@@ -0,0 +1,47 @@
...
+    renderChart: (width: number, height: number) => React.ReactNode;
+}
+
+export interface ChartBaseProps {
+    width: number;
+    height: number;
+}
+
+export const ChartContainer: React.FunctionComponent<ChartContainerProps> = (props) => {
+    const container = React.useRef<HTMLDivElement>(null);
```
pochretien: type: 
`container: React.RefObject<HTMLElement>`

src/components/chart/ChartContainer.tsx
```diff
@@ -0,0 +1,47 @@
...
+    useDebouncedWindowResize();
+
+    const toggleSvgs = (show: boolean) => {
+        if (container.current) {
+            const svgs = container.current.querySelectorAll('svg');
+            _.each(svgs, (svg: SVGElement) => svg.style.display = show ? 'block' : 'none');
+        }
+    };
+
+    toggleSvgs(false);
```
pochretien: why we need to toggle hidden and show after ? I think if we need to calculate something on render, its better to use a component instead of a function

src/components/chart/ChartContainer.tsx
```diff
@@ -0,0 +1,47 @@
...
+};
+
+const useForceUpdate = () => React.useState(undefined)[1];
+const useDebouncedWindowResize = () => {
+    const forceUpdate = useForceUpdate();
+    React.useEffect(() => {
+        const debouncedResize = () => window.requestAnimationFrame(forceUpdate);
+
+        window.addEventListener('resize', debouncedResize);
+        forceUpdate(1);
```
pochretien: What is the use of the 1 in forceUpdate

src/components/chart/ChartContainer.tsx
```diff
@@ -0,0 +1,47 @@
...
+    toggleSvgs(true);
+
+    return (
+        <div className='full-content' {..._.omit(props, 'renderChart')} ref={container}>
+            {width + height > 0 ? props.renderChart(width, height) : null}
+        </div>
+    );
+};
+
+const useForceUpdate = () => React.useState(undefined)[1];
```
pochretien: its safe to do that, it looks weird without validation

src/components/chart/XYAxis.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+const withDefaultConfig = (props: Partial<AxisProps> = {}): AxisProps => _.defaults(props, {
+    tickSize: props.show !== false ? 7 : 0,
+    size: props.show !== false ? 40 : 0,
+    innerPadding: 0,
+    show: true,
+    format: _.identity,
+});
+
+export const XYAxis: React.FunctionComponent<XYAxisProps> = ({x, y, children}) => {
```
pochretien: Nice!

src/components/chart/YGrid.tsx
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react';
+
+import {XYChartContext} from './XYChart';
+
+export interface YGridProps {
+    padding?: number;
+    color?: string;
+}
+
+export const YGrid: React.FunctionComponent<YGridProps> = ({padding = 0, color = 'rgba(0,0,0,0.2)'}) => {
```
pochretien: im not sure if its a good idea to set all default directly in the function arguments. I think it should be better to have an object with all default value. More ez to manage after

package.json
GermainBergeron: For the peer dependencies I think it make sense to allow more versions

src/components/chart/ChartContainer.tsx
GermainBergeron: I need to hide the svg to make sure It doesn't create scrollbar on the content (making the calculation wrong)

src/components/chart/ChartContainer.tsx
GermainBergeron: It's to change the value from undefined to something else. Maybe it would be clearer if I used Math.random?

src/components/chart/ChartContainer.tsx
GermainBergeron: useState always return an array `[currentState, setNewState]`. Here I just return setNewState because I don't care about the currentState

src/components/chart/ChartContainer.tsx
pochretien: oh ok I understand, Its fine for me to use 1

PR #1115, opened 2019-04-23
Feature Chart Tooltip

GermainBergeron:
image

### DIFF
--- a/src/components/chart/ChartTooltip.tsx
+++ b/src/components/chart/ChartTooltip.tsx
@@ -1,0 +1,84 @@
+import * as classNames from 'classnames';
+import * as React from 'react';
+
+import {VaporColors} from '../color/Color';
+import {DropPodPosition, IDropUIPosition} from '../drop/DomPositionCalculator';
+import {DropPod} from '../drop/DropPod';
+import {ChartTooltipContent} from './ChartTooltipContent';
+import {XYChartContext, XYPoint} from './XYChart';
+
+export interface ChartTooltipProps {
+    sort?: boolean;
+}
+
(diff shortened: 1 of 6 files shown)


src/components/chart/ChartTooltip.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+    return (
+        <g className='chart-tooltip-zones'>
+            {!!position.position && <rect width={2} x={xScale(position.index) - 1} y={yScale(yDomain[1])} height={yScale(yDomain[0])} fill={VaporColors.orange} />}
+            <circle cx={position.x} cy={position.y} r='1' fill='transparent' ref={dropRoot} />
+            <DropPod
+                ref={dropRoot}
+                isOpen={!!position.position}
+                positions={[position.position, DropPodPosition.left, DropPodPosition.right]}
+                parentSelector='body'
```
gdostie: `Defaults.DROP_ROOT`

src/components/chart/styles/ChartTooltipContent.scss
```diff
@@ -0,0 +1,20 @@
+@import '~coveo-styleguide/scss/common/palette';
+@import '~coveo-styleguide/scss/variables';
+
+$chart-tooltip-content-size: 3px;
+
+.chartTooltipContent {
+  font-size: 0.75rem;
```
pochretien: should be in the vapor as variable IMO

GermainBergeron: I'll change for $small-font-size which is 1px bigger

PR #1116, opened 2019-04-24
Add validation before DOMRect to avoid an error in Edge

pochretien:
Update interface
update logic with hasSameWidth to avoid an error with Edge

### DIFF
--- a/src/components/select/tests/SingleSelectConnected.spec.tsx
+++ b/src/components/select/tests/SingleSelectConnected.spec.tsx
@@ -252,7 +252,7 @@ describe('Select', () => {
             });
         });
 
-        xdescribe('click handler', () => {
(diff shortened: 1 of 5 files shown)


src/components/select/tests/SingleSelectConnected.spec.tsx
```diff
@@ -252,7 +252,7 @@ describe('Select', () => {
             });
         });
 
-        describe('click handler', () => {
+        xdescribe('click handler', () => {
```
GermainBergeron: Can you try changing the id of the div instead?

pochretien: tried and nothing good happen 😿

PR #1117, opened 2019-04-24
Feature Chart Unit Tests

GermainBergeron:


src/components/chart/tests/ChartTooltip.spec.tsx
```diff
@@ -0,0 +1,75 @@
...
+    it('should render a line when the user hover a serie point', () => {
+        const getAttributeSpy = jasmine.createSpy('getAttribute').and.returnValues(DropPodPosition.right, '10', '0');
+        const component = shallow(<ChartTooltip />);
+        const rect = component.find('rect').first();
+
+        rect.simulate('mouseMove', {currentTarget: {getAttribute: getAttributeSpy}});
+        // force update
+        component.setProps({sort: false});
+
+        // I can't seem to get this to work, the useState is never re-called
```
gdostie: Have you tried mounting the component instead of using shallow?

src/components/chart/tests/LineSeries.spec.tsx
```diff
@@ -0,0 +1,28 @@
...
+    it('should render a <g>', () => {
+        const component = shallow(<LineSeries />);
+        expect(component.find('g').exists()).toBe(true);
+    });
+
+    it('should render a path for every serie', () => {
+        spyOn(React, 'useContext').and.returnValue(XYChartContextMock);
+
+        const {series} = XYChartContextMock;
+        const component = shallow(<LineSeries />);
```
gdostie: Instead of spying on useContext, it is probably cleaner to pass down the context directly to shallow:
```tsx
const component = shallow(<LineSeries />, {context: XYChartContextMock});
```
Same comment goes for the other places where you spy on `useContext`

GermainBergeron: I haven't been able to make it work. It might be working with a mount but I would wait for better support from enzyme

src/components/chart/tests/ChartTooltip.spec.tsx
GermainBergeron: Drop throws some errors because it can't detect the boundingBox of the element

PR #1118, opened 2019-04-24
Implement disabled tabs and tab tooltips

gdostie:


src/components/tab/Tab.tsx
```diff
@@ -1,53 +1,65 @@
...
-    componentWillMount() {
-        if (this.props.onRender) {
-            this.props.onRender();
-        }
+        const tabTitle = _.isEmpty(this.props.tooltip)
+            ? this.props.title
+            : (
+                <Tooltip title={this.props.tooltip} placement={TooltipPlacement.Top}>
+                    {this.props.title}
+                </Tooltip>
```
GermainBergeron: I think the tooltip already handles this so you could always display:

```tsx
<Tooltip title={this.props.tooltip} placement={TooltipPlacement.Top}>
    {this.props.title}
</Tooltip>
```

PR #1122, opened 2019-05-03
Feature Select with Drop Component

GermainBergeron:
Dropdown can now overflow the modal :trophy:
Better positioning for the Select
Fix unit tests

image


src/components/drop/DropPod.tsx
```diff
@@ -88,11 +90,21 @@ class RDropPod extends React.PureComponent<IRDropPodProps, IDropPodState> {
     private setEventsOnDocument() {
         window.addEventListener('resize', this.updateOffset, true);
         window.addEventListener('scroll', this.updateOffset, true);
+
+        if (this.dropRef && this.dropRef.current) {
+            const config = {attributes: false, childList: true, subtree: true};
+            this.parentMutationObserver = new MutationObserver(this.updateOffset);
```
gdostie: TIL `MutationObserver`

gdostie:
The SingleSelect doesn't close on option click anymore

GermainBergeron:
I Fixed the first two issues in:
> * The SingleSelect doesn't close on option click anymore
> * The disabled SingleSelect works like its not disabled
> * Also this (probably unrelated) (gif)

But for the scrolling problem I don't see an easy solution still allowing the dropdown to overflow the modal 
image

I suggest that we leave it like that for now and if the problem happens in the admin we'll do a fix

PR #1125, opened 2019-05-07
Add possibility to set class on SelectConnected drop container

GermainBergeron:
Fix issue with button ignoring certain props
Always use the same import for ts-transformer-keys


src/components/button/Button.tsx
```diff
@@ -31,6 +35,7 @@ export class Button extends React.Component<IButtonProps & React.ButtonHTMLAttri
         let buttonElement: JSX.Element;
 
         let buttonAttrs = {
+            ..._.omit(this.props, ButtonPropsToOmit),
```
pochretien: GG

PR #1136, opened 2019-05-21
pressing arrow also open modals / pressing enter select item and clos…

nanani-coveo:
Arrows were already working, I added small changes to make the selectSingleModal more user friendly.

- if modal is closed and focused, pressing up or down arrow will open it

- if modal is open, pressing enter/tab will select the current element and close the modal

### DIFF
--- a/src/components/select/SelectConnected.tsx
+++ b/src/components/select/SelectConnected.tsx
@@ -193,7 +193,7 @@ export class SelectConnected extends React.PureComponent<ISelectProps & ISelectS
                 .filter((item: IItemBoxProps) => !item.hidden
                     && !this.props.multi
                     && !item.disabled
-                    || !_.contains(this.props.selectedValues, item.value)
+                    || !_.contains(this.props.selectedValues, item.value),
                 )
                 .value();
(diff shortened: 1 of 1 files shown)


src/components/select/SelectConnected.tsx
```diff
@@ -190,13 +190,15 @@ export class SelectConnected extends React.PureComponent<ISelectProps & ISelectS
 
         if (_.contains([keyCode.enter, keyCode.tab], e.keyCode) && this.props.isOpened) {
             const actives = _.chain(this.props.items)
-                .filter((item: IItemBoxProps) => !item.hidden && (!this.props.multi || !_.contains(this.props.selectedValues, item.value)) && !item.disabled)
-                .value();
+                .filter((item: IItemBoxProps) => !item.hidden && (!this.props.multi
+                    || !_.contains(this.props.selectedValues, item.value)) && !item.disabled).value();
```
GermainBergeron:
```
.filter((item: IItemBoxProps) => !item.hidden
    && !item.disabled
    && (!this.props.multi || !_.contains(this.props.selectedValues, item.value))
)
.value();
```

PR #1139, opened 2019-05-23
Update readme badges

gdostie:


gdostie:
Demo: https://github.com/coveo/react-vapor/blob/update-readme-badges/README.md

PR #1149, opened 2019-06-07
Fix unit test typings in vscode

gdostie:
### Description

vscode resolves a given file's typings by finding the nearest parent `tsconfig.json` file. When opening any `.spec` file, typings don't get resolved correctly, because all spec files are excluded in the `tsconfig.json` file. 

### Proposed Changes

- Removed all includes and excludes from `tsconfig.json`
- Added a `tsconfig.build.json` file which is now targeted by webpack and gulp for compilation and d.ts generation.

None

### Acceptance Criteria

- [x] [README.md](https://github.com/coveo/react-vapor/blob/master/README.md) is adjusted to reflect the proposed changes (if relevant)

### DIFF
--- a/packages/react-vapor/tsconfig.build.json
+++ b/packages/react-vapor/tsconfig.build.json
@@ -1,0 +1,13 @@
+{
+  "extends": "./tsconfig",
+  "compilerOptions": {
+    "outDir": "./dist/split/",
+    "sourceMap": true,
+    "rootDir": ".",
+  },
+  "exclude": [
+    "src/**/*.spec.*",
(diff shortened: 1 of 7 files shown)


packages/react-vapor/tsconfig.build.json
```diff
@@ -0,0 +1,13 @@
+{
+  "extends": "./tsconfig",
+  "compilerOptions": {
+    "outDir": "./dist/split/",
+    "sourceMap": true,
+    "rootDir": ".",
```
gdostie: I need to specify the rootDir or else dts-generator throws.

PR #1174, opened 2019-06-18
Feature - Add FlatSelect Selector in React Vapor [ADUI-4495]

dmgauthier:


packages/react-vapor/src/components/flatSelect/FlatSelectConnected.tsx
```diff
@@ -18,10 +18,7 @@ const mapStateToProps = (state: IReactVaporState, ownProps: IFlatSelectOwnProps)
     return {selectedOptionId: ownProps.options.length && ownProps.options[0].id};
```
GermainBergeron: Here you should use the new selector instead of duplicating the logic

packages/react-vapor/src/components/flatSelect/tests/FlatSelectSelectors.spec.tsx
```diff
@@ -0,0 +1,39 @@
+import { flatSelectInitialState, IFlatSelectState } from '../FlatSelectReducers';
+import { FlatSelectSelectors } from '../FlatSelectSelectors';
+
+describe('FlatSelectSelectors', () => {
+    describe('getInput', () => {
+        it('Should return the default flatSelect state', () => {
```
GermainBergeron: I'd change the name to be more specific

```suggestion
        it('should return the default flatSelect state when the flatSelect is not in the store', () => {
```

packages/react-vapor/src/components/flatSelect/FlatSelectSelectors.ts
```diff
@@ -0,0 +1,16 @@
+import {createSelector} from 'reselect';
+import * as _ from 'underscore';
+
+import {IReactVaporState} from '../../ReactVapor';
+import {flatSelectInitialState, IFlatSelectState } from './FlatSelectReducers';
+
+const getInput = (state: IReactVaporState, props: {id: string}): IFlatSelectState => {
+    return _.findWhere(state.flatSelect, {id: props.id}) || flatSelectInitialState;
```
gdostie: I think it would be best to let it return `undefined` instead of the flatSelectInitialState. From a conceptual standpoint, the flatSelect is really undefined in the state, so why should we lie about it :wink:

packages/react-vapor/src/components/flatSelect/tests/FlatSelectSelectors.spec.tsx
```diff
@@ -0,0 +1,39 @@
...
+            };
+            const selectedFlatSelect = FlatSelectSelectors.getInput({ flatSelect: [expectedSelectedFlatSelect] }, { id: expectedSelectedFlatSelect.id });
+            expect(selectedFlatSelect).toEqual(expectedSelectedFlatSelect);
+        });
+    });
+});
+
+describe('getSelectedOptionId', () => {
+    it('should return the value of the flatSelect initial state if no other value is chosen', () => {
+        flatSelectInitialState.id = 'the-first-id';
```
GermainBergeron: Here you modify the object which can be quite mind boggling when we try to debug a test (in react-vapor the unit tests are run in a random order)

Do you really need to change the default id and selectedOptionId?

packages/react-vapor/src/components/flatSelect/FlatSelectSelectors.ts
```diff
@@ -0,0 +1,16 @@
...
+import * as _ from 'underscore';
+
+import {IReactVaporState} from '../../ReactVapor';
+import {flatSelectInitialState, IFlatSelectState } from './FlatSelectReducers';
+
+const getInput = (state: IReactVaporState, props: {id: string}): IFlatSelectState => {
+    return _.findWhere(state.flatSelect, {id: props.id}) || flatSelectInitialState;
+};
+
+const getSelectedOptionId = createSelector(getInput, (flatSelect: IFlatSelectState) => flatSelect && flatSelect.selectedOptionId);
```
gdostie: You could add a typing for the return value so that we know what kind of value is returned
```suggestion
const getSelectedOptionId = createSelector(getInput, (flatSelect: IFlatSelectState): string => flatSelect && flatSelect.selectedOptionId);
```

packages/react-vapor/src/components/flatSelect/FlatSelectSelectors.ts
dmgauthier: `        it('should return the default flatSelect state when the flatSelect is not in the store', () => {
            const FlatSelect = FlatSelectSelectors.getInput({ flatSelect: [] }, { id: 'nothing is selected by the user' });
            expect(FlatSelect).toEqual(flatSelectInitialState);
        });`

gdostie: didn't you write the test?

dmgauthier: YES but my point is to expect that if nothing is selected by the user, the initial value is chosen.
but... I guess it is useless right ?

packages/react-vapor/src/components/flatSelect/FlatSelectConnected.tsx
```diff
@@ -1,27 +1,23 @@
...
+import {IReactVaporState} from '../../ReactVapor';
+import {IDispatch, ReduxUtils} from '../../utils/ReduxUtils';
 import {FlatSelect, IFlatSelectDispatchProps, IFlatSelectOwnProps, IFlatSelectProps, IFlatSelectStateProps} from './FlatSelect';
 import {addFlatSelect, removeFlatSelect, selectFlatSelect} from './FlatSelectActions';
 import {IFlatSelectOptionProps} from './FlatSelectOption';
 import {IFlatSelectState} from './FlatSelectReducers';
 
 const mapStateToProps = (state: IReactVaporState, ownProps: IFlatSelectOwnProps): IFlatSelectStateProps => {
-    const flatSelectState: IFlatSelectState = _.findWhere(state.flatSelect, {id: ownProps.id});
+    const flatSelect: IFlatSelectState = _.findWhere(state.flatSelect, {id: ownProps.id});
```
francoislg: Shouldn't this use the new selector?

packages/react-vapor/src/components/radio/tests/RadioSelectSelectors.spec.ts
```diff
@@ -1,10 +1,10 @@
-import {IRadioSelectState, radioSelectInitialState} from '../RadioSelectReducers';
-import {RadioSelectSelectors} from '../RadioSelectSelectors';
+import { IRadioSelectState, radioSelectInitialState } from '../RadioSelectReducers';
```
francoislg: This should probably be reverted 🤔

packages/react-vapor/src/components/flatSelect/FlatSelectConnected.tsx
pochretien: you can remove the return if you want
```
const mapStateToProps = (state: IReactVaporState, ownProps: IFlatSelectOwnProps): IFlatSelectStateProps => ({
    selectedOptionId: FlatSelectSelectors.getSelectedOptionId(state, {id: ownProps.id}),
)};
```

packages/react-vapor/src/components/flatSelect/FlatSelectSelectors.ts
```diff
@@ -0,0 +1,16 @@
+import {createSelector} from 'reselect';
+import * as _ from 'underscore';
+
+import {IReactVaporState} from '../../ReactVapor';
+import {IFlatSelectState} from './FlatSelectReducers';
+
+const getInput = (state: IReactVaporState, props: {id: string}): IFlatSelectState => {
+    return _.findWhere(state.flatSelect, {id: props.id});
```
pochretien: add an initial value to avoid an undefined as return

`const initialFlatSelectState = {id: undefiend, selectedOptionId: undefiend};`

pochretien:
```suggestion
    return _.findWhere(state.flatSelect, {id: props.id}) || initialFlatSelectState;
```

packages/react-vapor/src/components/flatSelect/FlatSelectSelectors.ts
```diff
@@ -0,0 +1,16 @@
...
+import * as _ from 'underscore';
+
+import {IReactVaporState} from '../../ReactVapor';
+import {IFlatSelectState} from './FlatSelectReducers';
+
+const getInput = (state: IReactVaporState, props: {id: string}): IFlatSelectState => {
+    return _.findWhere(state.flatSelect, {id: props.id});
+};
+
+const getSelectedOptionId = createSelector(getInput, (flatSelect: IFlatSelectState): string => flatSelect && flatSelect.selectedOptionId);
```
pochretien: you can remove your check for the flatSelect after that and use that
```suggestion
const getSelectedOptionId = createSelector(getInput, (flatSelect: IFlatSelectState): string => flatSelect.selectedOptionId);
```

PR #1180, opened 2019-06-19
Change ItemBox displayValue to ReactNode

gdostie:
### Proposed Changes

ItemBox displayValue was only allowed as string, now we can pass down a react-node

Custom match filters that assume displayValue is a string may break.

### Acceptance Criteria


francoislg:
Why though? I can't quite see the use can if the text content is extracted from the ReactNode anyway

gdostie:
I had a use case where I needed to have some part of text in bold as displayValue. 
image

With `displayValue: string` I simply cannot  achieve this

packages/vapor/scss/sprites.scss
```diff
@@ -1,78 +1 @@
...
...eight:12px;      text-indent:12px;    }i.coveo-sprites-misc-spinner_arrow_down, .coveo-sprites-misc-spinner_arrow_down {      background-position:-32px 0px;      width:12px;      height:12px;      text-indent:12px;    }i.coveo-sprites-misc-filter, .coveo-sprites-misc-filter {      background-position:-46px 0px;      width:12px;      height:12px;      text-indent:12px;    }i.coveo-sprites-misc-alert, .coveo-sprites-misc-alert {      background-position:-60px 0px;      width:11px;      height:11px;      text-indent:11px;    }i.coveo-sprites-misc-panel-expended, .coveo-sprites-misc-panel-expended {      background-position:-73px 0px;      width:10px;      height:10px;      text-indent:10px;    }i.coveo-sprites-misc-panel-collapsed, .coveo-sprites-misc-panel-collapsed {      background-position:-85px 0px;      width:10px;      height:10px;      text-indent:10px;    }i.coveo-sprites-tables-sort_desc, .coveo-sprites-tables-sort_desc {      background-position:-97px 0px;      width:6px;      height:6px;      text-indent:6px;    }i.coveo-sprites-tables-sort_asc, .coveo-sprites-tables-sort_asc {      background-position:-97px -8px;      width:6px;      height:6px;      text-indent:6px;    }
```
francoislg: Revert this..?

gdostie: No, this file is generated at compile time and prettier was then re-formatting it, causing `packages/vapor/scss/sprites.scss` to always appear in the diff. I added its path to `.prettierignore` so it should be the last time we see this diff in theory :slightly_smiling_face:

packages/react-vapor/src/utils/FilterUtils.spec.ts
```diff
@@ -45,7 +45,9 @@ describe('Utils', () => {
             });
 
             it('should return true if the item displayValue contain a exact match with the filterValue', () => {
-                expect(defaultMatchFilter(itemBoxWithDisplayValue.displayValue, itemBoxWithDisplayValue)).toBe(true);
+                expect(
+                    defaultMatchFilter(itemBoxWithDisplayValue.displayValue as string, itemBoxWithDisplayValue)
```
GermainBergeron: Can you remove the cast here and change the `defaultMatchFilter` to accept a ReactNode?

gdostie: The first parameter of the defaultMatchFilter is the filter value, I don't see any use case where it could be a ReactNode

PR #1183, opened 2019-06-20
Feature adui 4451 better selector logs table log browser

FelixBlaisThon:
### Proposed Changes

Added the table id for logs table to help qa feature test and differentiate it with the table.calendar-grid

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/table-hoc/TableHOC.tsx
+++ b/packages/react-vapor/src/components/table-hoc/TableHOC.tsx
@@ -35,7 +35,7 @@ export class TableHOC extends React.PureComponent<ITableHOCProps & React.HTMLAtt
                 })}
             >
                 {this.renderActions()}
-                <table className={classNames(this.props.className)}>
+                <table className={classNames(this.props.className)} id={this.props.id}>
                     {this.props.tableHeader}
                     <tbody>{this.props.renderBody(this.props.data || [])}</tbody>
                 </table>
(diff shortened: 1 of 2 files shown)


packages/react-vapor/src/components/table-hoc/TableHOC.tsx
```diff
@@ -35,7 +35,7 @@ export class TableHOC extends React.PureComponent<ITableHOCProps & React.HTMLAtt
                 })}
             >
                 {this.renderActions()}
-                <table className={classNames(this.props.className)}>
+                <table className={classNames(this.props.className)} id={this.props.id}>
```
francoislg: Since `ActionBarConnected` down here also has the same `this.props.id`, there is a slight possibility that the page would contain the same `id` element twice, for instance, if `ActionBar` is updated to actually output this ID.

I think either or both of those IDs should be prefixed, just to be sure :)

FelixBlaisThon: Yes, there is indeed two id="complex-example" (in the local demo); one for the table and for the actionBar, but i don't understand what you mean by prefixing the IDs :S

GermainBergeron:
I'm wondering why you didn't add a new class in the admin instead?

FelixBlaisThon:
I might just add a class in LogsTable.tsx in jsadmin-log-browser and if it's good, i will close this PR

packages/react-vapor/src/components/table-hoc/TableHOC.tsx
francoislg: `prefixing` => I mean adding a prefix, like `table-` :)

`id={`table-${this.props.id}`}`

GermainBergeron:
We won't merge the pull request since the fix was made directly in the admin console

PR #1186, opened 2019-06-21
fix tableRowConnected double click should trigger the link if defined

dmgauthier:


packages/react-vapor/src/components/table-hoc/TableRowConnected.tsx
```diff
@@ -190,9 +190,9 @@ class TableRowConnected extends React.PureComponent<
     };
 
     private handleDoubleClick = () => {
-        _(this.props.actions)
-            .filter((action: IActionOptions) => action.callOnDoubleClick)
-            .forEach((action: IActionOptions) => action.trigger());
+        _(this.props.actions).each((action: IActionOptions) =>
+            action.callOnDoubleClick && action.link ? window.location.replace(action.link) : action.trigger()
```
francoislg: Since `location.replace` [replaces the current history](https://developer.mozilla.org/en-US/docs/Web/API/Location/replace), I think that you won't be able to navigate back.

I think it would be more useful with `window.location.href = action.link` ?

dmgauthier: `window.location.href = action.link ` brings me to `localhost/action.link` I don't know why.

And I would seem that `location.replace` is superior exactly because the page is not kept in the history. I think I was able to go back though... but I'll test again to be sure.

good point.

packages/react-vapor/src/components/table-hoc/TableRowConnected.tsx
```diff
@@ -190,9 +190,13 @@ class TableRowConnected extends React.PureComponent<
     };
 
     private handleDoubleClick = () => {
-        _(this.props.actions)
-            .filter((action: IActionOptions) => action.callOnDoubleClick)
-            .forEach((action: IActionOptions) => action.trigger());
+        _(this.props.actions).each((action: IActionOptions) => {
+            if (action.callOnDoubleClick && action.link) {
```
francoislg: 🤔 

Why not 

```
 .filter((action: IActionOptions) => action.callOnDoubleClick))
 .forEach((action: IActionOptions) => {
    if (action.link) { 
       window.location.href = action.link;
     } else if (action.trigger) {
       action.trigger();
     }
 });
```

gdostie: The solution proposed by @dmgauthier is better performance wise as it iterates only one time over all actions whereas your solution iterates one time over all actions + one time over double click actions. Practically thought, it makes no real difference as we won't have enough action to make a significant difference. I would go with readability here.

PR #1188, opened 2019-07-02
move dependencies in peer and dev dependencies to remove vunerability

pochretien:
### Acceptance Criteria

### DIFF
--- a/packages/vapor/package.json
+++ b/packages/vapor/package.json
@@ -23,15 +23,17 @@
         "type": "git",
         "url": "https://github.com/coveo/react-vapor.git"
     },
-    "dependencies": {
-        "chosen-npm": "1.4.2",
-        "coveo-slider": "1.0.1",
-        "materialize-css": "0.98.2"
+    "peerDependencies": {
+        "chosen-npm": "^1.4.2",
+        "coveo-slider": "^1.0.1",
+        "materialize-css": "^0.98.2"
     },
     "devDependencies": {
         "ansi-colors": "3.2.4",
         "autosize": "3.0.21",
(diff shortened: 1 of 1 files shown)


packages/vapor/package.json
```diff
@@ -23,15 +23,17 @@
...
         "url": "https://github.com/coveo/react-vapor.git"
     },
-    "dependencies": {
-        "chosen-npm": "1.4.2",
-        "coveo-slider": "1.0.1",
-        "materialize-css": "0.98.2"
+    "peerDependencies": {
+        "chosen-npm": "^1.4.2",
+        "coveo-slider": "^1.0.1",
+        "materialize-css": "^0.98.2"
```
samisayegh: Does materialize-css need to be a peer dependency if we generate a file containing the necessary CSS using a gulp task?

GermainBergeron: We bundle the js files inside the gulp task to make the project easier to use. Someone could chose to use a different version of everyone of those project and it would still work if he bundle them himself. Vapor is mostly a style project and could also be used without the js

If you have any recommendation for the setup we're takers but I don't think it should only target `materialize-css` and it's a bit out of scope for this PR

samisayegh: Hmm, I see, however in the case of this specific dependency as well as `chosen-npm`, they appear to only be used by the gulp task. They are not used in the vapor source code.

Therefore, mentioning that these packages should be peer-dependencies seems unnecessary?

pochretien: For the first move, I will stay with the peer-dependencies but we can valid in the futur if we can do better with that.

PR #1195, opened 2019-07-17
Support withServerSideProcessing in SelectWithPredicate

gdostie:
### Proposed Changes

Added support for server side processing of predicate filtering in select components. Followed the same pattern we use in TableHOC withServerSideProcessing.

Basically, the idea is that when processing is done by the server, no filtering is applied on the UI side, so we pass down `ownProps.items` directly to the wrapped component. Also, when the predicate value changes, we trigger the `onUpdate` callback so that a call to the server can be sent.

None.

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/select/SelectWithPredicate.tsx
+++ b/packages/react-vapor/src/components/select/SelectWithPredicate.tsx
@@ -1,10 +1,14 @@
 import * as React from 'react';
+import {connect} from 'react-redux';
 import {keys} from 'ts-transformer-keys';
 import * as _ from 'underscore';
+
+import {WithServerSideProcessingProps} from '../../hoc/withServerSideProcessing/withServerSideProcessing';
 import {IReactVaporState} from '../../ReactVapor';
-import {ReduxConnect} from '../../utils/ReduxUtils';
+import {callIfDefined} from '../../utils/FalsyValuesUtils';
 import {FlatSelectConnected} from '../flatSelect/FlatSelectConnected';
 import {IFlatSelectOptionProps} from '../flatSelect/FlatSelectOption';
+import {FlatSelectSelectors} from '../flatSelect/FlatSelectSelectors';
 import {IItemBoxProps} from '../itemBox/ItemBox';
 import {ISelectProps} from './SelectConnected';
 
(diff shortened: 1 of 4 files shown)


packages/react-vapor/src/components/select/tests/MultiSelectWithPredicate.spec.tsx
```diff
@@ -150,5 +150,28 @@ describe('Select', () => {
                 expect(spy).toHaveBeenCalledWith(unselectListBoxOption(id, items[1].value));
             });
         });
+
+        describe('when predicates are processed on the server side', () => {
```
gdostie: Forgot to add a test for the onUpdate prop being called when `predicate` changes...

packages/react-vapor/src/components/select/tests/MultiSelectWithPredicate.spec.tsx
```diff
@@ -150,5 +152,44 @@ describe('Select', () => {
...
+                    store
+                )
+                    .dive()
+                    .dive();
+
+                expect(onUpdateSpy).not.toHaveBeenCalled();
+
+                component.setProps({predicate: 'some-other-predicate'});
+
+                expect(onUpdateSpy).not.toHaveBeenCalledTimes(1);
```
GermainBergeron: remove `.not` ?

gdostie: woops indeed and the test is probably failing then :disappointed:

gdostie: fixed

packages/react-vapor/src/components/select/SelectWithPredicate.tsx
```diff
@@ -13,44 +17,60 @@ export interface ISelectWithPredicateOwnProps {
...
 
 export const selectWithPredicate = (
-    Component: React.ComponentClass<ISelectProps> | React.StatelessComponent<ISelectProps>
-): React.ComponentClass<ISelectWithPredicateProps> => {
-    const mapStateToProps = (state: IReactVaporState, ownProps: ISelectWithPredicateProps): Partial<ISelectProps> => {
-        const flatSelect = _.findWhere(state.flatSelect, {id: ownProps.id});
-        const predicate = (flatSelect && flatSelect.selectedOptionId) || ownProps.options[0].id;
+    Component: React.ComponentType<ISelectProps>
+): React.ComponentType<ISelectWithPredicateProps> => {
+    const mapStateToProps = (
```
francoislg: Ça change pas grand chose, mais tu pourrais sortir cette méthode de `selectWithPredicate`. 

Présentement, tu peux techniquement accéder à `Component` et ça serait plus safe d'enlever cette possibilité.

gdostie: Done

PR #1200, opened 2019-07-18
Add props filterPlaceholder to FilterBoxConnected

rosalie-liu:
### Proposed Changes

Add props filterPlaceholder to FilterBoxConnected in order to changing the default text of filter box in drop down menu

Before:
before

After:
after


packages/react-vapor/src/components/select/SelectWithFilter.tsx
```diff
@@ -28,6 +28,7 @@ export interface ISelectWithFilterOwnProps {
     noItemsText?: string;
     filterButton?: IButtonProps;
     filter?: IFilterBoxOwnProps;
+    filterPlaceholder?: string;
```
GermainBergeron: I think you can send the filterPlaceholder inside the `filter` props, so you won't need the new props

PR #1204, opened 2019-07-26
Add Component Code in Demo Components

GermainBergeron:
### Proposed Changes

Add libraries
Prepare structure to split demo in multiple sections
Replace custom router with react-router

Links and urls in demo

### Acceptance Criteria


GermainBergeron:
image

packages/react-vapor/docs/src/components/ComponentsApp.tsx
```diff
@@ -0,0 +1,57 @@
...
+import * as _ from 'underscore';
+import {ReactVaporStore} from '../../ReactVaporStore';
+import {ComponentCode} from './ComponentCode';
+import {IComponent} from './ComponentsInterface';
+import {ComponentsMenu} from './ComponentsMenu';
+
+export const ComponentsApp: React.FunctionComponent = () => {
+    const req = require.context('../../../src/components/', true, /Examples?\.tsx?$/i);
+    const components: IComponent[] = [];
+    req.keys().forEach((key) => {
```
francoislg: This is basically a `.map` :)

```
const components: IComponent[] = req.keys().map((key) => {
    const component = req(key);
    const code = require('!raw-loader!../../../src/components/' + key.replace('./', '')).default;
    const name = _.keys(component)[0].replace(/Examples?/i, '');
    return {name, code, path: key, component: _.values(component)[0]}
});
```

packages/react-vapor/docs/src/components/ComponentsApp.tsx
```diff
@@ -0,0 +1,57 @@
...
+        const component = req(key);
+        const code = require('!raw-loader!../../../src/components/' + key.replace('./', '')).default;
+        const name = _.keys(component)[0].replace(/Examples?/i, '');
+        components.push({name, code, path: key, component: _.values(component)[0]});
+    });
+
+    const routes = components.map((c: IComponent) => (
+        <Route
+            key={c.path}
+            path={`/${c.name}`}
```
francoislg: You are building a couple of times this `/${c.name}`, maybe you should directly embed it in your `Component` interface as a `link` property ...?

GermainBergeron:
> This looks very nice and I am all for React-Router and such, but you are losing a quite important feature in the demo: Deprecation notices.
> 
> Could you keep a hand-crafted array of deprecated component and tag them? Or simply completely filter them out ?

@francoislg I re-added the notices for the deprecated components. In a second phase we'll want to move components around and create a 2 (or 3) level menu, deprecated components will be grouped together in a section

PR #1207, opened 2019-07-29
Select with infinite scroll HOC.

gdostie:
### Proposed Changes

- Added a `wrapItems` prop to the `ListBox` component
- Exposed all `ListBoxOwnProps` on the `SelectConnected`
- Implemented a `selectWithInfiniteScroll` HOC component that can be used with any SelectComponent (single or multi). While best used in combination with the `withServerSideProcessing` HOC, it still can be used in a UI side fashion. See example file for how to use it.

None.

### Acceptance Criteria


packages/react-vapor/src/components/select/examples/ServerSideSelectExamples.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+import {IReactVaporExampleState} from '../../../../docs/Reducers';
+import {withServerSideProcessing} from '../../../hoc/withServerSideProcessing/withServerSideProcessing';
+import {FilterBoxSelectors} from '../../filterBox/FilterBoxSelectors';
+import {IItemBoxProps} from '../../itemBox/ItemBox';
+import {selectWithFilter} from '../hoc/SelectWithFilter';
+import {selectWithInfiniteScroll, SelectWithInfiniteScrollProps} from '../hoc/SelectWithInfiniteScroll';
+import {ISingleSelectOwnProps, SingleSelectConnected} from '../SingleSelectConnected';
+
+const TOTAL_ENTRIES = 5000;
```
GermainBergeron: Do we really need to set this value?

packages/react-vapor/src/components/select/examples/ServerSideSelectExamples.tsx
```diff
@@ -0,0 +1,127 @@
...
+            <span>{title}</span>
+        </div>
+    );
+};
+
+function mapStateToProps(state: IReactVaporExampleState, props: {id: string}) {
+    return {
+        filterValue: FilterBoxSelectors.getFilterText(state, props),
+    };
+}
```
GermainBergeron:
```typescript
const mapStateToProps = (state: IReactVaporExampleState, props: {id: string}) =>({
    filterValue: FilterBoxSelectors.getFilterText(state, props),
})
```

packages/react-vapor/src/components/select/examples/ServerSideSelectExamples.tsx
```diff
@@ -0,0 +1,127 @@
...
+    albumId: string;
+    id: string;
+    title: string;
+    url: string;
+    thumbnailUrl: string;
+}
+
+const clean = <T extends object>(o: T) => _(o).pick(_.identity);
+
+function usePhotosAPIMock(initialPhotos: any[] = []): [any[], (params?: any, overwrite?: boolean) => void] {
```
GermainBergeron: you can set the default value of `overwrite` to `false`. I think it makes the code a bit clearer

GermainBergeron:
Code looks good but somehow the demo is not working. 🤔

packages/react-vapor/src/components/select/examples/ServerSideSelectExamples.tsx
gdostie: yes, its used to determine if the `hasMore` prop of the InfiniteScroll

gdostie:
@GermainBergeron demo should be working now

packages/react-vapor/src/components/select/examples/ServerSideSelectExamples.tsx
gdostie: I discovered that the json-server exposes the total count in the request headers at `x-total-count` so I used this instead of the hardcoded value.

packages/react-vapor/src/components/select/examples/ServerSideSelectExamples.tsx
gdostie: overwrite already has a default value of `true`

pochretien:
image
The loading stay after the fetch

packages/react-vapor/src/components/listBox/ListBox.tsx
```diff
@@ -32,6 +33,7 @@ export class ListBox extends React.Component<IListBoxProps, {}> {
         noResultItem: {
             value: 'No Items',
         },
+        wrapItems: _.identity,
```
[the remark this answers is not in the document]
gdostie: If you look at the `wrapItems` type what we want is to return a `ReactNode`. `_.noop` returns `void` if I'm not mistaken. `_.identity` however returns directly what it receives as arguments which is what we want here.

PR #1221, opened 2019-08-19
Add Fields table checkbox width class

rosalie-liu:
### Proposed Changes

- Add Fields table checkbox width class
- Fix webpack configuration for Windows
image

### Acceptance Criteria

### DIFF
--- a/packages/vapor/scss/tables/table.scss
+++ b/packages/vapor/scss/tables/table.scss
@@ -157,6 +157,10 @@
             position: relative;
             top: -1px; // Small hack to make it work in all browsers...
         }
+
+        &.coveo-checkbox-width {
+            width: 142px;
+        }
     }
 
     &.row-selected td,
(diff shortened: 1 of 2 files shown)


packages/vapor/scss/tables/table.scss
```diff
@@ -157,6 +157,10 @@
             position: relative;
             top: -1px; // Small hack to make it work in all browsers...
         }
+
+        &.coveo-checkbox-width {
+            width: 142px;
```
GermainBergeron: I'm kind of confused about this 142px, it seems to be the width of the label in the table and it has nothing to do with checkboxes

image

IMHO this style won't be shared with any other tables, thus it should go in the admin

pochretien: if we want to add this classes, we need to remove the coveo- before, nothing related to coveo should be added in react-vapor

PR #1223, opened 2019-08-20
Adui 4690/refactor side navigation

gdostie:
#### Refactored the `SideNavigationMenuSection` component

Now the `SideNavigationMenuSection` supports passing down the header props directly on the SideNavigationMenuSection (backwards compatibility has been preserved):

```tsx
// Before
<SideNavigationMenuSection
    header={{
        title: 'the title',
        svgName: 'theIcon',
        svgClass: 'theClasses',
        onClick: doSomething,
    }}
    {...otherProps}
/>

// After (you can still use the header prop although it its deprecated)
<SideNavigationMenuSection 
    title='the title' 
    svgName='theIcon' 
    svgClass='theClasses' 
    onClick={doSomething}
    {...otherProps}
/>
```

#### Refactored the `SideNavigationItem` component
Made the `SideNavigationItem` more generic while preserving backwards compatibility.

```tsx
// Before
<SideNavigationItem href="http://coveo.com" title="Link to Coveo" />

// After (you can still use the old way)
<SideNavigationItem>
    <a
        href="http://coveo.com"
        title="Link to Coveo"
        className="navigation-menu-section-item-link"
    >
        Link to Coveo
    </a>
</SideNavigationItem>
```

#### Deprecate `SideNavigationHeader` component. 

Now it is embedded directly inside `SideNavigationMenuSection`'s implementation because it doesn't make sense to use such header outside a side navigation section.

### Potention breaking changes

None.

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/sideNavigation/SideNavigationHeader.tsx
+++ b/packages/react-vapor/src/components/sideNavigation/SideNavigationHeader.tsx
@@ -1,37 +1,40 @@
 import * as classNames from 'classnames';
 import * as React from 'react';
+
 import {Svg} from '../svg/Svg';
 
+/**
+ * @deprecated Use SideNavigationHeaderProps instead
+ */
 export interface ISideNavigationHeaderProps {
-    title: string;
+    title?: string;
     svgName?: string;
     svgClass?: string;
-    onClick?: () => void;
+    onClick?: (event: React.MouseEvent<HTMLDivElement, MouseEvent>) => void;
 }
 
-export class SideNavigationHeader extends React.Component<ISideNavigationHeaderProps> {
-    private handleClick() {
-        if (this.props.onClick) {
-            this.props.onClick();
-        }
-    }
+/**
+ * @deprecated Will be removed in version 5
+ */
+export const SideNavigationHeader: React.FunctionComponent<ISideNavigationHeaderProps> = ({
+    onClick,
+    svgClass,
+    svgName,
+    children,
+}) => {
+    const icon = svgName ? (
+        <Svg
+            svgName={svgName}
+            svgClass={classNames('navigation-menu-section-header-icon icon mod-lg transparency-3', svgClass)}
+        />
+    ) : (
+        <span className="navigation-menu-section-header-icon" />
+    );
 
-    render() {
-        const svgClass = classNames(
-            'navigation-menu-section-header-icon icon mod-lg transparency-3',
-            this.props.svgClass
-        );
-        const icon = this.props.svgName ? (
-            <Svg svgName={this.props.svgName} svgClass={svgClass} />
-        ) : (
-            <span className="navigation-menu-section-header-icon" />
-        );
-        return (
-            <div className="navigation-menu-section-header text-white" onClick={() => this.handleClick()}>
-                {icon}
-                {this.props.title}
-                {this.props.children}
-            </div>
-        );
-    }
-}
+    return (
+        <div className="navigation-menu-section-header text-white" onClick={onClick}>
+            {icon}
+            {children}
+        </div>
+    );
+};
--- a/packages/react-vapor/src/components/sideNavigation/SideNavigationHeader.tsx
+++ b/packages/react-vapor/src/components/sideNavigation/SideNavigationHeader.tsx
@@ -7,34 +7,38 @@ import {Svg} from '../svg/Svg';
  * @deprecated Use SideNavigationHeaderProps instead
  */
 export interface ISideNavigationHeaderProps {
-    title?: string;
+    title: string;
     svgName?: string;
     svgClass?: string;
-    onClick?: (event: React.MouseEvent<HTMLDivElement, MouseEvent>) => void;
+    onClick?: () => void;
 }
 
 /**
  * @deprecated Will be removed in version 5
  */
-export const SideNavigationHeader: React.FunctionComponent<ISideNavigationHeaderProps> = ({
-    onClick,
-    svgClass,
-    svgName,
-    children,
-}) => {
-    const icon = svgName ? (
-        <Svg
-            svgName={svgName}
-            svgClass={classNames('navigation-menu-section-header-icon icon mod-lg transparency-3', svgClass)}
-        />
-    ) : (
-        <span className="navigation-menu-section-header-icon" />
-    );
+export class SideNavigationHeader extends React.Component<ISideNavigationHeaderProps> {
+    private handleClick() {
+        if (this.props.onClick) {
+            this.props.onClick();
+        }
+    }
 
-    return (
-        <div className="navigation-menu-section-header text-white" onClick={onClick}>
-            {icon}
-            {children}
-        </div>
-    );
-};
+    render() {
+        const svgClass = classNames(
+            'navigation-menu-section-header-icon icon mod-lg transparency-3',
+            this.props.svgClass
+        );
+        const icon = this.props.svgName ? (
+            <Svg svgName={this.props.svgName} svgClass={svgClass} />
+        ) : (
+            <span className="navigation-menu-section-header-icon" />
+        );
+        return (
+            <div className="navigation-menu-section-header text-white" onClick={() => this.handleClick()}>
+                {icon}
+                {this.props.title}
+                {this.props.children}
+            </div>
+        );
+    }
+}
--- a/packages/react-vapor/src/components/sideNavigation/SideNavigationItem.tsx
+++ b/packages/react-vapor/src/components/sideNavigation/SideNavigationItem.tsx
@@ -29,6 +29,7 @@ export const SideNavigationItem: React.FunctionComponent<SideNavigationItemProps
 
     const itemClasses = classNames('navigation-menu-section-item', {'state-active': isActive});
 
+    // Rendering an anchor tag from href and title support for retrocompatibility
     return href && title ? (
         <a className={classNames('block', itemClasses)} href={href} target={target} ref={ref}>
             <span className="navigation-menu-section-item-link">{title}</span>
--- a/packages/react-vapor/src/components/sideNavigation/SideNavigationItem.tsx
+++ b/packages/react-vapor/src/components/sideNavigation/SideNavigationItem.tsx
@@ -1,13 +1,22 @@
+import * as classNames from 'classnames';
 import * as React from 'react';
 
-export interface ISideNavigationItemProps {
-    href: string;
-    title: string;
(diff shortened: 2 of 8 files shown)


packages/react-vapor/src/components/sideNavigation/SideNavigationItem.tsx
```diff
@@ -6,8 +7,36 @@ export interface ISideNavigationItemProps {
...
+    const ref = React.useRef(null);
+
+    React.useEffect(() => {
+        if (isActive && ref.current) {
+            ref.current.scrollIntoView({behavior: 'instant', block: 'nearest'});
+        }
+    }, [isActive]);
+
+    return href && title ? (
+        <a
```
francoislg: I think I'd put a comment here that states this is for retrocompatibility.

packages/react-vapor/src/components/sideNavigation/SideNavigationMenuSection.tsx
```diff
@@ -1,63 +1,86 @@
...
-        return (
-            <div className="block navigation-menu-section">
-                {this.buildHeader()}
-                <SlideY in={!(this.props.expandable && !this.props.expanded)} timeout={350}>
-                    <div className="navigation-menu-section-items">{this.props.children}</div>
-                </SlideY>
-            </div>
+const SideNavigationHeader: React.FunctionComponent<SideNavigationHeaderProps> = (props) => {
+    let icon: React.ReactNode = null;
+    if (props.customIcon) {
```
francoislg: 🤔 what do you think about changing it a little to have a `const icon` instead? 

```
const icon = props.customIcon || (props.svgName ? <this.SvgIcon /> : <span className="navigation-menu-section-header-icon" />
```

packages/react-vapor/src/components/sideNavigation/SideNavigationMenuSection.tsx
```diff
@@ -1,63 +1,86 @@
...
-            return (
-                <SideNavigationHeader {...this.props.header} onClick={() => this.handleClick()}>
-                    {arrow}
-                </SideNavigationHeader>
-            );
-        }
-        return <SideNavigationLoadingHeader />;
-    }
-}
+    return !_.isEmpty(React.Children.map(props.children, _.identity)) ? (
```
francoislg: what is this sorcery? :P

Looks like a `hasChildren`? Can it be named as such?

On a higher level, why is this check required at all?

packages/react-vapor/src/components/sideNavigation/SideNavigationMenuSection.tsx
```diff
@@ -1,63 +1,86 @@
...
+    const items = <div className="navigation-menu-section-items">{children}</div>;
+
+    return expandable ? (
+        <Collapsible
+            className="navigation-menu-section"
+            id={_.uniqueId('nav-section')}
+            headerContent={sectionHeader}
+            toggleIconClassName="fill-white transparency-3"
+            onToggleExpandedState={onClick}
+            expanded={!!expanded}
```
francoislg: `expanded` is already a `boolean`, do you really need this `!!`?

packages/react-vapor/src/components/sideNavigation/examples/SideNavigationExample.tsx
```diff
@@ -25,23 +26,40 @@ export class SideNavigationExample extends React.Component<any, any> {
...
-                            header={{title: 'Section 3', svgName: 'menu-content'}}
+                            title="Collapsible Section"
+                            svgName="menu-content"
                             expandable
                             expanded={this.state.expanded}
                             onClick={() => this.click()}
                         >
-                            <SideNavigationItem href="http://coveo.com" target="_blank" title="Link to Coveo" />
-                            <SideNavigationItem href="http://coveo.com" title="Another link to Coveo" />
+                            <SideNavigationItem>
```
francoislg: To be quite honest, I think I prefer the old style.

No concern about "how", just fill in the props and the component will be rendered with the right style.

With the new way, I have to know that a `<a href>` goes inside and that I must put a `className="navigation-menu-section-item-link"`, which will be duplicated for all of my links.

If you really want to go the "super flexible" way, I think it would be better to also offer a `SideNavigationItemLink` that provides this full easy functionality

packages/react-vapor/src/components/sideNavigation/SideNavigationMenuSection.tsx
gdostie: This high Griffindor sorcery is indeed `hasChildren`.

The use case is when the side navigation section has no title. In that situation, you don't want to display a header with an empty content, you simply don't want to display the header at all. 

It visually looks like this:
image

I can simply verify if the title is empty a little earlier and if so not render a `SideNavigationHeader` at all. Would that be simpler?

packages/react-vapor/src/components/sideNavigation/SideNavigationMenuSection.tsx
gdostie: This is me being paranoid, expanded can be `undefined` instead of false. It doesn't do much of a difference except in the unit tests.

packages/react-vapor/src/components/sideNavigation/examples/SideNavigationExample.tsx
gdostie: I preserved the old way of doing things, because I agree it looks better, but the new way is required if you want to bind the navigation item to `react-router-dom` properly like I did in the demo:

```tsx
const NavLink: React.FunctionComponent<{href: string; name: string; baseUrl?: string}> = ({
    href,
    name,
    baseUrl,
    ...rest
}) => {
    const path = `${baseUrl}${href}`;
    return (
        <Route
            path={path}
            children={(routeProps: RouteChildrenProps) => (
                <SideNavigationItem isActive={!!routeProps.match}>
                    <Link to={path} className="block navigation-menu-section-item-link" {...rest}>
                        {name}
                    </Link>
                </SideNavigationItem>
            )}
        />
    );
};
```

`Link` already outputs an anchor tag, so if the SideNavigationItem does too by default you'll end-up having 2 embedded `<a>`;

packages/react-vapor/src/components/sideNavigation/SideNavigationMenuSection.tsx
gdostie: So you would convert `SideNavigationHeader` in a class component to define `<this.SvgIcon />` ?

francoislg: Oh hum, definitely don't want to convert :)

What about a local functional component? 🤔

```
const SvgIcon = () => <Svg ... />;
const icon = props.customIcon || (props.svgName ? <SvgIcon /> : <span className="navigation-menu-section-header-icon" />
```

I mean, I find it slightly easier to follow, but your original code is not that bad either ;)

gdostie: Aight good idea I'll convert it to something like this

```tsx
const HeaderIcon: React.FunctionComponent<SideNavigationHeaderProps> = ({svgName, svgClass}) =>
    svgName ? (
        <Svg
            svgName={svgName}
            svgClass={classNames('navigation-menu-section-header-icon icon mod-lg transparency-3 fill-white', svgClass)}
        />
    ) : (
        <span className="navigation-menu-section-header-icon" />
    );
```

packages/react-vapor/src/components/sideNavigation/examples/SideNavigationExample.tsx
GermainBergeron: Maybe we could add the `className="navigation-menu-section-item-link"` automatically on the children?

gdostie: Done

packages/react-vapor/src/components/sideNavigation/SideNavigationHeader.tsx
```diff
@@ -1,37 +1,40 @@
...
-                {icon}
-                {this.props.title}
-                {this.props.children}
-            </div>
-        );
-    }
-}
+    return (
+        <div className="navigation-menu-section-header text-white" onClick={onClick}>
+            {icon}
```
dmgauthier: is `{this props title}` replaced by something else?

packages/react-vapor/src/components/sideNavigation/SideNavigationItem.tsx
```diff
@@ -1,13 +1,42 @@
...
+            ref.current.scrollIntoView({behavior: 'instant', block: 'nearest'});
+        }
+    }, [isActive]);
+
+    const itemClasses = classNames('navigation-menu-section-item', {'state-active': isActive});
+
+    // Rendering an anchor tag from href and title support for retrocompatibility
+    return href && title ? (
+        <a className={classNames('block', itemClasses)} href={href} target={target} ref={ref}>
+            <span className="navigation-menu-section-item-link">{title}</span>
```
dmgauthier: I guess the title is right here right ?

packages/react-vapor/src/components/sideNavigation/SideNavigationHeader.tsx
gdostie: Good catch I removed it by error.

packages/react-vapor/src/components/sideNavigation/SideNavigationItem.tsx
gdostie: Here its the title of the SideNavigationItem, the other file is the title of the navigation section

packages/react-vapor/src/components/sideNavigation/SideNavigationHeader.tsx
gdostie: I reverted the changes to this whole file except the deprecation warning because I'm not using it anyway.

packages/vapor/scss/components/navigation.scss
```diff
@@ -11,7 +11,8 @@
 }
 
 .navigation {
-    width: $navigation-width;
+    min-width: $navigation-width;
+    max-width: $navigation-width;
```
dmgauthier: What does it change to do this ?

gdostie: Somehow the width of the side nav was changing when clicking on smaller items. I don't really understand why, but by setting a min and a max width instead the width of the side nav always stay the same :taco:

PR #1227, opened 2019-08-20
Fix active state in SingleSelect when clicking or using key navigation

acote-coveo:
### Proposed Changes

When playing with the value selection in SingleSelect, we were not previously updating the activate state. It was only updated when using the navigation key.

I now keep track of the item index and change the active cells when selecting an item through a click or key navigation.

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/listBox/ListBox.tsx
+++ b/packages/react-vapor/src/components/listBox/ListBox.tsx
@@ -6,10 +6,7 @@ import {mod} from '../../utils/DataStructuresUtils';
 import {callIfDefined} from '../../utils/FalsyValuesUtils';
 import {IItemBoxProps, ItemBox} from '../itemBox/ItemBox';
 
-export interface IITemBoxPropsWithIndex {
-    index?: number;
-    item: IItemBoxProps;
-}
+export type IItemBoxPropsWithIndex = {index?: number} & IItemBoxProps;
 
 export interface IListBoxOwnProps {
     noResultItem?: IItemBoxProps;
@@ -65,7 +62,7 @@ export class ListBox extends React.Component<IListBoxProps, {}> {
             .filter(shouldShow)
             .map((item: IItemBoxProps) => {
                 let active = false;
-                const itemWithIndex: IITemBoxPropsWithIndex = {item: {...item}};
+                const itemWithIndex: IItemBoxPropsWithIndex = {...item};
                 if (!item.disabled) {
                     if (this.props.active === null) {
                         active = _.contains(this.props.selected, item.value);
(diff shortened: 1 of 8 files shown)


packages/react-vapor/src/components/listBox/ListBox.tsx
```diff
@@ -54,39 +59,45 @@ export class ListBox extends React.Component<IListBoxProps, {}> {
...
             .length;
 
-        let index = 0;
+        let realIndex = 0;
         let activeSet = false;
         const items = _.chain(this.props.items)
             .filter(shouldShow)
             .map((item: IItemBoxProps) => {
                 let active = false;
+                const itemWithIndex: IITemBoxPropsWithIndex = {item: {...item}};
```
GermainBergeron: Can you do this instead?

```suggestion
                const itemWithIndex: IITemBoxPropsWithIndex = {item};
```

packages/react-vapor/src/components/listBox/ListBox.tsx
```diff
@@ -54,39 +59,45 @@ export class ListBox extends React.Component<IListBoxProps, {}> {
...
-            .map((item: IItemBoxProps) => {
+            .map(({item, index}: IITemBoxPropsWithIndex) => {
                 let active = item.active;
                 if (!item.disabled && activeSet === false) {
                     active = true;
                     activeSet = true;
                 }
-                return {...item, active};
+                return {item: {...item, active}, index};
+            })
```
GermainBergeron:
```typescript
.map((itemWithIndex: IITemBoxPropsWithIndex) => {
    let active = itemWithIndex.active;
    if (!itemWithIndex.disabled && activeSet === false) {
        active = true;
        activeSet = true;
    }
    return {...itemWithIndex, active};
})
```

acote-coveo: So *itemWithIndex* has the following props: **item**, **index** 

In your suggestion, active would be a new prop to *itemWithIndex* instead of *item*.

acote-coveo: I could probably use something like that:
`return _.merge(itemWithIndex, {item: {active}}}`

packages/react-vapor/src/components/listBox/ListBox.tsx
acote-coveo: If I mutate the same item, it seems to have some side effects on some states holder. If I clone the object with a spread operator it seems to work fine.

packages/react-vapor/src/components/listBox/ListBox.tsx
```diff
@@ -6,6 +6,11 @@ import {mod} from '../../utils/DataStructuresUtils';
 import {callIfDefined} from '../../utils/FalsyValuesUtils';
 import {IItemBoxProps, ItemBox} from '../itemBox/ItemBox';
 
+export interface IITemBoxPropsWithIndex {
```
gdostie:
```suggestion
export interface IItemBoxPropsWithIndex {
```

packages/react-vapor/src/components/listBox/ListBox.tsx
```diff
@@ -6,6 +6,11 @@ import {mod} from '../../utils/DataStructuresUtils';
 import {callIfDefined} from '../../utils/FalsyValuesUtils';
 import {IItemBoxProps, ItemBox} from '../itemBox/ItemBox';
 
+export interface IITemBoxPropsWithIndex {
+    index?: number;
+    item: IItemBoxProps;
+}
```
gdostie: Wouldn't it be simpler to use a simple type intersection?
```ts
{index?: number} & IItemBoxProps
```

gdostie: That way you wouldn't have to use `itemWithIndex.item.something`, and could simply use `itemWithIndex.active` instead.

PR #1229, opened 2019-08-21
Refactor unMount method

rosalie-liu:
### Proposed Changes

- At first, I've tried replace deselectAll with dispatching deselect a single row, but it didn't work. Because the ActionBarReducer has no knowledge about the current selection. Also it may cause unexpected effect. 
- In a nutshell, IMO the safest solution till now is dispatching addActionsToActionBar without action on unMount if a row is selected, in another word, remove existing actions.

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/table-hoc/TableRowConnected.tsx
+++ b/packages/react-vapor/src/components/table-hoc/TableRowConnected.tsx
@@ -39,7 +39,7 @@ export interface ITableRowStateProps {
 
 export interface ITableRowDispatchProps {
     onMount: () => void;
-    onUnmount: () => void;
+    onUnmount: (isSelected: boolean) => void;
     handleClick: (isMulti: boolean, isOpened: boolean) => void;
     onUpdateToCollapsibleRow: () => void;
     onActionBarActionsChanged: () => void;
@@ -78,9 +78,11 @@ const mapDispatchToProps = (dispatch: IDispatch, ownProps: ITableRowOwnProps): I
                 dispatch(TableHOCRowActions.toggleCollapsible(ownProps.id, true));
             }
         },
-        onUnmount: () => {
+        onUnmount: (isSelected?: boolean) => {
             dispatch(TableHOCRowActions.remove(ownProps.id));
-            dispatch(TableHOCRowActions.deselectAll(ownProps.tableId));
+            if (isSelected) {
+                dispatch(addActionsToActionBar(ownProps.tableId, []));
+            }
         },
         handleClick: (isMulti: boolean, isOpened: boolean) => {
             refreshActionBarActions(isMulti);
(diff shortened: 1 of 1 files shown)


packages/react-vapor/src/components/table-hoc/TableRowConnected.tsx
```diff
@@ -78,9 +78,11 @@ const mapDispatchToProps = (dispatch: IDispatch, ownProps: ITableRowOwnProps): I
                 dispatch(TableHOCRowActions.toggleCollapsible(ownProps.id, true));
             }
         },
-        onUnmount: () => {
+        onUnmount: (isSelected?: boolean) => {
             dispatch(TableHOCRowActions.remove(ownProps.id));
-            dispatch(TableHOCRowActions.deselectAll(ownProps.tableId));
+            if (isSelected) {
+                dispatch(addActionsToActionBar(ownProps.tableId, []));
```
gdostie:
```suggestion
                dispatch(TableHOCRowActions.deselectAll(ownProps.tableId));
```

packages/react-vapor/src/components/table-hoc/tests/TableRowConnected.spec.tsx
```diff
@@ -94,8 +94,9 @@ describe('Table HOC', () => {
             expect(store.getActions()).toContain(expectedAction);
         });
 
-        it('should dispatch an TableHOCRowActions.deselectAll on componentWillUnmount', () => {
-            const expectedAction = TableHOCRowActions.deselectAll(defaultProps.tableId);
+        it('should dispatch an addActionsToActionBar without action on componentWillUnmount when a row is selected', () => {
+            const expectedAction = addActionsToActionBar(defaultProps.tableId, []);
```
gdostie:
```suggestion
            const expectedAction = TableHOCRowActions.deselectAll(defaultProps.tableId);
```

packages/react-vapor/src/components/table-hoc/TableRowConnected.tsx
rosalie-liu: I've exactly used TableHOCRowActionsType.deselectAll in my previous code. But it caused some bugs, like in the description. https://coveord.atlassian.net/secure/RapidBoard.jspa?rapidView=569&modal=detail&selectedIssue=ADUI-4695 Then Germain discussed with me, maybe we should try dispatching deselect a single one row instead of deselectAll in the table. But the problem is, after removing a row while unmount, that doesn't make sense of deselect it, because it's already gone, the effect of button is simply for the actions. So I chose to dispatch(addActionsToActionBar(ownProps.tableId, [])); for empty actions.

gdostie: oh ok yeah makes sense

PR #1231, opened 2019-08-21
Add class name to children util

gdostie:
### Proposed Changes

Implemented a small util that adds className to children (or any react node).

```tsx
// node
<span className="old-class">Hello react-vapor!</span>;

const newNode = addClassNameToChildren(node, 'new-class');

// newNode
<span className="old-class new-class">Hello react-vapor!</span>
```

None

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/utils/JSXUtils.spec.tsx
+++ b/packages/react-vapor/src/utils/JSXUtils.spec.tsx
@@ -1,7 +1,9 @@
+import {shallow} from 'enzyme';
 import * as React from 'react';
-import {getReactNodeTextContent} from './JSXUtils';
 
-describe('JSXUtils', () => {
+import {addClassNameToChildren, getReactNodeTextContent} from './JSXUtils';
+
+fdescribe('JSXUtils', () => {
     describe('getReactNodeTextContent', () => {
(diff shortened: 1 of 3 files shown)


packages/react-vapor/src/utils/JSXUtils.spec.tsx
```diff
@@ -25,4 +27,21 @@ describe('JSXUtils', () => {
             ).toBe("Hello there! Can't you see me? I can.");
         });
     });
+
+    describe('addClassNameToChildren', () => {
+        it('should not add any class if the children is not a react element', () => {
+            const myElement = 'a string is not a react element';
+            expect(addClassNameToChildren(myElement, 'new-class')).toContain(myElement);
```
GermainBergeron: Would it work with toBe instead of toContain?
```suggestion
            expect(addClassNameToChildren(myElement, 'new-class')).toBe(myElement);
```

gdostie: `addClassNameToChildren` returns an array

gdostie:
I was thinking about this util and I think it would better if we wrapped the child with a span that has the className when the child is a string.

PR #1235, opened 2019-08-22
Reset page for predicates

rosalie-liu:
### Proposed Changes

After a few tries, I adopted this solution for the issue with predicates pagination. I've added ListBoxActions.select to paginationReducer and paginationCompositeReducer, if it's the same tableId, we will go to the first page after making change in predicates. Basically, it's the same logic as FilterActions.filterThrough. 

### Acceptance Criteria


packages/react-vapor/src/components/navigation/pagination/NavigationPaginationReducers.ts
```diff
@@ -37,6 +38,10 @@ export const paginationReducer = (
                 : state;
         case FilterActions.filterThrough:
             return s.contains(state.id, action.payload.id) ? {...state, pageNb: 0} : state;
+        case ListBoxActions.select: {
+            const [tableId] = s.words(action.payload.id, '--');
```
GermainBergeron: I understand what you're doing here but I think it would be better to add an util inside `TableHOCUtils` to get the TableId from the predicateid, this way here you wouldn't need to know that the separator is `--`

Also, I think it's easier to understand with a split or a regex than with the words function

packages/react-vapor/src/components/table-hoc/TableHOCUtils.ts
```diff
@@ -69,4 +76,5 @@ export const TableHOCUtils = {
     getCompositeState,
     getPredicateId,
     getPaginationId,
+    getTableIdFromPredicateId,
```
gdostie: I know there is no test file for the TableHOCUtils but I think those are worth testing. If you don't feel like writing unit tests for all the other existing and untested utils you can just write tests for `getTableIdFromPredicateId` for now, although it would be nice if they were all tested :smile:.

rosalie-liu: No problem!

packages/react-vapor/src/components/navigation/pagination/NavigationPaginationReducers.ts
```diff
@@ -37,6 +39,10 @@ export const paginationReducer = (
                 : state;
         case FilterActions.filterThrough:
             return s.contains(state.id, action.payload.id) ? {...state, pageNb: 0} : state;
+        case ListBoxActions.select: {
+            const tableId = TableHOCUtils.getTableIdFromPredicateId(action.payload.id);
+            return tableId && s.contains(state.id, tableId) ? {...state, pageNb: 0} : state;
```
dmgauthier: Since the real method in underscore string library is `.include` from which `.contains` is only an aliase. I would use the real method name IMO. it's only a detail though.

rosalie-liu: ah you have a good point! but in this file I think I would keep .contains, because I've followed the filterThrough. I'm not sure if I must change all to .include or it's simply kind of convention to use .contains hmm

PR #1242, opened 2019-08-29
changed the readonly behavior for the datePickerDropdownConnected

dmgauthier:
### Proposed Changes

The readOnly mode changed the dropdown to an input component.
Now the component stays the same but the `onclick` functionality is disabled and the text is grayed out.

none as far as I know

### DIFF
--- a/packages/react-vapor/src/components/datePicker/DatePickerDropdown.tsx
+++ b/packages/react-vapor/src/components/datePicker/DatePickerDropdown.tsx
@@ -150,7 +150,7 @@ export class DatePickerDropdown extends React.Component<IDatePickerDropdownProps
         return (
             <div className={classNames('date-picker-dropdown', this.props.className)}>
                 <div className={dropdownClasses} ref={(dropdown: HTMLDivElement) => (this.dropdown = dropdown)}>
-                    <span className={toggleClasses} onClick={() => !this.props.readonly && this.handleClick()}>
+                    <span className={toggleClasses} onClick={() => !disabled && this.handleClick()}>
                         <span className={!disabled ? 'dropdown-selected-value' : 'text-medium-grey'}>
                             <label>
                                 {label}
--- a/packages/react-vapor/src/components/datePicker/DatePickerDropdown.tsx
+++ b/packages/react-vapor/src/components/datePicker/DatePickerDropdown.tsx
@@ -124,8 +124,7 @@ export class DatePickerDropdown extends React.Component<IDatePickerDropdownProps
                 if (formattedUpper !== label) {
                     toLabel = (
                         <span className={this.props.readonly ? 'text-medium-grey' : 'text-dark-grey'}>
-                            {' '}
-                            {this.props.toLabel}{' '}
+                            {this.props.toLabel}
                         </span>
                     );
                     labelSecondPart = formattedUpper;
--- a/packages/react-vapor/src/components/datePicker/DatePickerDropdown.tsx
+++ b/packages/react-vapor/src/components/datePicker/DatePickerDropdown.tsx
@@ -123,7 +123,9 @@ export class DatePickerDropdown extends React.Component<IDatePickerDropdownProps
                 const formattedUpper = this.formatDate(this.props.datePicker.appliedUpperLimit);
                 if (formattedUpper !== label) {
                     toLabel = (
-                        <span className={this.props.readonly ? 'text-medium-grey' : 'text-dark-grey'}>
+                        <span
+                            className={classNames('mx1', this.props.readonly ? 'text-medium-grey' : 'text-dark-grey')}
+                        >
                             {this.props.toLabel}
                         </span>
                     );
@@ -150,7 +152,7 @@ export class DatePickerDropdown extends React.Component<IDatePickerDropdownProps
         return (
             <div className={classNames('date-picker-dropdown', this.props.className)}>
                 <div className={dropdownClasses} ref={(dropdown: HTMLDivElement) => (this.dropdown = dropdown)}>
-                    <button className={toggleClasses} onClick={() => this.handleClick()} disabled={this.props.readonly}>
+                    <button className={toggleClasses} onClick={this.handleClick} disabled={this.props.readonly}>
                         <span className="dropdown-selected-value">
                             <label>
                                 {label}
--- a/packages/react-vapor/src/components/datePicker/DatePickerDropdown.tsx
+++ b/packages/react-vapor/src/components/datePicker/DatePickerDropdown.tsx
@@ -124,7 +124,7 @@ export class DatePickerDropdown extends React.Component<IDatePickerDropdownProps
                 const formattedUpper = this.formatDate(this.props.datePicker.appliedUpperLimit);
                 if (formattedUpper !== label) {
                     toLabel = (
-                        <span className={disabled ? 'text-medium-grey' : 'to-label'}> {this.props.toLabel} </span>
+                        <span className={disabled ? 'text-medium-grey' : 'text-dark-grey'}> {this.props.toLabel} </span>
                     );
                     labelSecondPart = formattedUpper;
                 }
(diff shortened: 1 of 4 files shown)


dmgauthier:
Must fix failing tests

packages/react-vapor/src/components/datePicker/DatePickerDropdown.tsx
```diff
@@ -160,15 +150,15 @@ export class DatePickerDropdown extends React.Component<IDatePickerDropdownProps
         return (
             <div className={classNames('date-picker-dropdown', this.props.className)}>
                 <div className={dropdownClasses} ref={(dropdown: HTMLDivElement) => (this.dropdown = dropdown)}>
-                    <span className={toggleClasses} onClick={() => this.handleClick()}>
-                        <span className="dropdown-selected-value">
+                    <span className={toggleClasses} onClick={() => !disabled && this.handleClick()}>
```
gdostie: Check for the `disabled` condition inside the `handleClick` method.

packages/react-vapor/src/components/datePicker/examples/DatePickerDropdownConnectedExamples.tsx
```diff
@@ -31,6 +31,16 @@ export class DatePickerDropdownConnectedExamples extends React.Component<any, an
...
+                        readonly={false}
+                    />
+                </div>
+                <div className="form-group">
+                    <label className="form-control-label">Date picker dropdown disabled</label>
+                    <DatePickerDropdownConnected
+                        id="date-picker-dropdown"
+                        datesSelectionBoxes={SELECTION_BOXES}
+                        selectionRules={CALENDAR_SELECTION_RULES}
+                        readonly={true}
```
gdostie:
```suggestion
                        readonly
```

packages/react-vapor/src/components/datePicker/examples/DatePickerDropdownConnectedExamples.tsx
```diff
@@ -31,6 +31,16 @@ export class DatePickerDropdownConnectedExamples extends React.Component<any, an
                         datesSelectionBoxes={SELECTION_BOXES}
                         selectionRules={CALENDAR_SELECTION_RULES}
                         initialDateRange={DATE_RANGE_EXAMPLE}
+                        readonly={false}
```
gdostie: False is the default
```suggestion

```

packages/react-vapor/src/components/datePicker/DatePickerDropdown.tsx
gdostie: You could add the `disabled` class in the `toggleClasses` I think it already does some magic for us, like graying everything. 

Also, have a look at [`pointer-events` rule in css](https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events), setting this to none could help prevent the onClick prop to be triggered in the first place. Maybe we could add this rule to the `disabled` class in vapor if it isn't already there.

GermainBergeron:
I think I would keep the arrow but make it grey, otherwise it look strange IMO 

image

packages/react-vapor/src/components/datePicker/DatePickerDropdown.tsx
```diff
@@ -169,15 +158,15 @@ export class DatePickerDropdown extends React.Component<IDatePickerDropdownProps
...
+                    </button>
                     <div className={menuClasses}>{this.getDatePickerBox()}</div>
                 </div>
             </div>
         );
     }
 
-    private handleClick = () => {
-        if (this.props.onClick) {
+    private handleClick = (disabled: boolean) => {
```
gdostie: You can access `this.props.readonly` from here, so no need to pass `disabled` as parameter.

packages/react-vapor/src/components/datePicker/DatePickerDropdown.tsx
```diff
@@ -156,11 +146,10 @@ export class DatePickerDropdown extends React.Component<IDatePickerDropdownProps
                 'dropdown-toggle-placeholder': !this.props.datePicker || !this.props.datePicker.appliedLowerLimit,
             }
         );
-
         return (
             <div className={classNames('date-picker-dropdown', this.props.className)}>
                 <div className={dropdownClasses} ref={(dropdown: HTMLDivElement) => (this.dropdown = dropdown)}>
-                    <span className={toggleClasses} onClick={() => this.handleClick()}>
+                    <button className={toggleClasses} onClick={() => this.handleClick(disabled)} disabled={disabled}>
```
gdostie:
```suggestion
                    <button className={toggleClasses} onClick={this.handleClick} disabled={disabled}>
```

packages/react-vapor/src/components/datePicker/tests/DatePickerDropdown.spec.tsx
```diff
@@ -104,14 +104,27 @@ describe('Date picker', () => {
...
+            );
 
-            expect(datePickerDropdown.find('.date-picker-dropdown').length).toBe(0);
-            expect(datePickerDropdown.find('Input').length).toBe(1);
+            expect(datePickerDropdownInstance.props.readonly).toBeTruthy();
+            expect(
+                datePickerDropdown
+                    .children()
+                    .find('button.dropdown-toggle')
+                    .matchesElement(expectedElement)
```
gdostie: Instead of having to define the expectedElement and aserting `.matchesElement`, we could simply check that it has the `disabled` prop set to true.

packages/react-vapor/src/components/datePicker/DatePickerDropdown.tsx
```diff
@@ -117,30 +116,21 @@ export class DatePickerDropdown extends React.Component<IDatePickerDropdownProps
         let label: string = this.props.label;
         let toLabel: JSX.Element = null;
         let labelSecondPart: string = '';
+        const disabled: boolean = this.props.readonly;
```
gdostie: I don't see the added value of doing this so I would tend to use the `readonly` prop directly everywhere.

packages/react-vapor/src/components/datePicker/DatePickerDropdown.tsx
```diff
@@ -117,30 +116,21 @@ export class DatePickerDropdown extends React.Component<IDatePickerDropdownProps
...
+        const disabled = this.props.readonly;
+
         if (this.props.datePicker && this.props.datePicker.appliedLowerLimit) {
             label = this.formatDate(this.props.datePicker.appliedLowerLimit);
             if (this.props.datePicker.isRange) {
                 const formattedUpper = this.formatDate(this.props.datePicker.appliedUpperLimit);
                 if (formattedUpper !== label) {
-                    toLabel = <span className="to-label"> {this.props.toLabel} </span>;
+                    toLabel = (
+                        <span className={disabled ? 'text-medium-grey' : 'to-label'}> {this.props.toLabel} </span>
```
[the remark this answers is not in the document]
GermainBergeron: Nvm, I didn't think it was used to change the text color. In this case what do you think of:
```suggestion
                        <span className={disabled ? 'text-medium-grey' : 'text-dark-grey'}> {this.props.toLabel} </span>
```

dmgauthier:
a test is failing since the last commit.

packages/react-vapor/src/components/datePicker/tests/DatePickerDropdown.spec.tsx
```diff
@@ -347,14 +338,13 @@ describe('Date picker', () => {
             expect(onClickSpy).toHaveBeenCalled();
         });
 
-        it('should not set a click listener to handleDocumentClick if it has the readonly prop', () => {
+        xit('should not set a click listener to handleDocumentClick if it has the readonly prop', () => {
```
GermainBergeron: It this a wanted change?

PR #1245, opened 2019-09-03
Better reflect actual callIfDefined signature

francoislg:
### Proposed Changes

We tried using `callIfDefined` with `strictNullCheck: true` and the method is failing because it does not actually accept `null` and `undefined` values!

Since its purpose is to validate whether it is defined, I think it makes sense to allow passing those values :P

What I have done here is activate `strictNullCheck` temporarily and tested that the unit tests did not return an error. Since `someDeclaredButNotAssignedCallback` was also invalid (assigning `undefined` to a callback type), I instead put `any`.

None, it is more permissive now.

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/utils/FalsyValuesUtils.spec.tsx
+++ b/packages/react-vapor/src/utils/FalsyValuesUtils.spec.tsx
@@ -39,7 +39,7 @@ describe('callIfDefined', () => {
     });
 
     it('should not throw errors when calling with undefined values', () => {
-        const someDeclaredButNotAssignedCallback: () => any = undefined;
+        const someDeclaredButNotAssignedCallback: any = undefined;
 
         expect(() => callIfDefined(undefined)).not.toThrow();
         expect(() => callIfDefined(null)).not.toThrow();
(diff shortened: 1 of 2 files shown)


francoislg:
image

packages/react-vapor/src/utils/FalsyValuesUtils.spec.tsx
```diff
@@ -39,7 +39,7 @@ describe('callIfDefined', () => {
     });
 
     it('should not throw errors when calling with undefined values', () => {
-        const someDeclaredButNotAssignedCallback: () => any = undefined;
+        const someDeclaredButNotAssignedCallback: any = undefined;
```
GermainBergeron: I think `const someDeclaredButNotAssignedCallback: () => any;` would be better to reflect what we actually want to validate

francoislg: If you activate `strictNullCheck` in this repo, this line breaks a compilation error when you keep the type  :P

I mean... This line gets compiled to `var someDeclaredButNotAssignedCallback = undefined` either way, so I'm not sure what it is actually testing

GermainBergeron: You're right, it's only testing that typescript work as expected 😛

GermainBergeron:
The build is failing because a unit test throws an error

francoislg:
☝️ last failure was a random failure, I think it can be merged now

PR #1250, opened 2019-09-10
+id in tableHeaderWithSort

FelixBlaisThon:
### Proposed Changes

There was an id attribute in the TableHeaderWithSort component, but it was never used. I've added it to help QA FT

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/table-hoc/TableHeaderWithSort.tsx
+++ b/packages/react-vapor/src/components/table-hoc/TableHeaderWithSort.tsx
@@ -64,7 +64,7 @@ export class TableHeaderWithSort extends React.Component<
         });
 
         return (
-            <th className={headerCellClasses} onClick={() => this.props.onSort()}>
+            <th id={this.props.id} className={headerCellClasses} onClick={() => this.props.onSort()}>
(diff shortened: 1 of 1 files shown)


packages/react-vapor/src/components/table-hoc/TableHeaderWithSort.tsx
```diff
@@ -64,7 +64,7 @@ export class TableHeaderWithSort extends React.Component<
         });
 
         return (
-            <th className={headerCellClasses} onClick={() => this.props.onSort()}>
+            <th id={this.props.id} className={headerCellClasses} onClick={() => this.props.onSort()}>
```
francoislg: I think I would also add the `tableId`, just in case multiple tables in the same page use the same `id` for their rows.

FelixBlaisThon: I could add the attribute tableId and Id but seperate them, rather than mixing them:

`tableId={this.props.tableId}` id={this.props.id}` 
Since i'm using `this.props.id` to the attribute id, i prefer not to mix them in the same attribute

PR #1253, opened 2019-09-10
Fix add dangerously set description in the blank slate component

dmgauthier:
### Proposed Changes

Added a dangerouslySetDescription in the BlankSlate props

none

### Acceptance Criteria

-   [ x ] The proposed changes are covered by unit tests
-   [ x ] The potential breaking changes are clearly identified
-   [ - ] [README.md](https://github.com/coveo/react-vapor/blob/master/README.md) is adjusted to reflect the proposed changes (if relevant)

### DIFF
--- a/packages/react-vapor/src/components/blankSlate/BlankSlate.tsx
+++ b/packages/react-vapor/src/components/blankSlate/BlankSlate.tsx
@@ -7,7 +7,7 @@ import {Svg} from '../svg/Svg';
 
 export interface IBlankSlateProps extends React.ClassAttributes<BlankSlate> {
     title?: string;
-    description?: string;
+    description?: React.ReactNode;
     buttons?: IBaseActionOptions[];
     withModal?: boolean;
     classes?: string[];
--- a/packages/react-vapor/src/components/blankSlate/examples/BlankSlateExample.tsx
+++ b/packages/react-vapor/src/components/blankSlate/examples/BlankSlateExample.tsx
@@ -1,5 +1,4 @@
 import * as React from 'react';
-
 import {BlankSlate} from '../BlankSlate';
 
 export class BlankSlateExample extends React.Component<any, any> {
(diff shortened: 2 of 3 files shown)


packages/react-vapor/src/components/blankSlate/BlankSlate.tsx
```diff
@@ -8,6 +8,7 @@ import {Svg} from '../svg/Svg';
 export interface IBlankSlateProps extends React.ClassAttributes<BlankSlate> {
     title?: string;
     description?: string;
+    dangerouslySetDescription?: string;
```
francoislg: Why though? 

Could `description` just be of type `React.ReactNode` instead?

dmgauthier: changed the `description` type to `React.ReactNode` but got this : 

Screenshot from 2019-09-11 15-52-29

Is there something else I don't know of I could do to make that work ?

GermainBergeron: When you use the prop you'll need to send 
```typescript
<BlankSlate
    title="title test"
    description={<p dangerouslySetInnerHTML={{__html: "description test <a href='https://www.google.com'>the link</a>"}} />
/>
```

dmgauthier: Oh I see. So we still have to use the `DangerouslySetInnerHTML` in both case. If both of you think that it is better this way than adding another props lets do that. Thank you guys for the `React.ReactNode` type, I will have to read a bit more documentation about it !

dmgauthier: is it ok if I just skipped the `dangerouslySetThing` ?
`<div className="form-group">
                    <label className="form-control-label">BlankSlate with title and description with a link</label>
                    <BlankSlate title="title test" description={<p>this is a description with a link to <a href='https//www.google.com'>this website</a></p>} />
                </div>`
Screenshot from 2019-09-11 20-03-54
:1st_place_medal:

PR #1255, opened 2019-09-11
Add a header to every components demo section

gdostie:
### Proposed Changes

- I added a basic header to every component page in the demo.
- Implemented some `// start-print` & `// stop-print` tags to scope down the examples source code.
- Added a way to set a custom name for the first tab.
- Added a way to set a custom title for a section.
- Added the `npm start` command to the root folder, it will run the start command in all packages in parallel.

#### Usage

```tsx

import {ExampleComponent} from '../../../../docs/src/components/ComponentsInterface';

// start-print
const Fish = () => <div>Bloup bloup</div>;
// stop-print

export const FishExamples: ExampleComponent = Fish;

FishExamples.firstTabLabel = 'ocean';
FishExamples.description = 'Lovers of the fishes';
```
Will render this:

image

None, it only impacts the demo.

### Acceptance Criteria


packages/react-vapor/docs/src/components/ComponentPage.tsx
```diff
@@ -1,47 +1,64 @@
...
+    const {name, tabs, component} = props;
+    const hasMarkdownTabs = tabs.length > 0;
+    const getTabId = buildTabIdTemplate(name);
+    const mapTabConfigToProps = ({tabName}: TabConfig): ITabProps => ({
+        id: getTabId(tabName),
+        title: tabName,
+    });
+
+    const tabProps: ITabProps[] = hasMarkdownTabs && [
+        {id: getTabId('development'), title: component.firstTabLabel || 'Develop'},
```
GermainBergeron: Label instead of firstTabLabel?

package.json
```diff
@@ -44,5 +44,8 @@
         "last 1 versions",
         "IE 11",
         "not dead"
-    ]
+    ],
+    "scripts": {
+        "start": "lerna run start --parallel"
```
dmgauthier: > Stream output from child processes immediately, prefixed with the originating package name. This allows output from different packages to be interleaved. [...] but completely disregards concurrency and topological sorting, running a given command or script immediately in all matching packages with prefixed streaming output. This is the preferred flag for long-running processes [...]

You are using this on `run start`, why ?

gdostie: This allows to run `npm start` in the root of react-vapor. It'll run in parallel `npm start` in all the sub packages (react-vapor and vapor) in parallel which is pretty convenient.

PR #1247, opened 2019-09-09
Add Form and FormGroup components

francoislg:
### Proposed Changes

The idea is to have simple wrappers for vapor styles and, ideally, if you always use the `Form` > `FormGroup` > `Input` structure, you should get consistent styling everywhere.

`Form` wraps `coveo-form` and offers simple props to configure it (like `asCard` which I find really helpful for discoverability) and `FormGroup` wraps `form-group`.

### Potential Breaking Changes

None, the only style change requires a new mod class

### Acceptance Criteria

-   [ ] The proposed changes are covered by unit tests
-   [ ] The potential breaking changes are clearly identified
-   [ ] [README.md](https://github.com/coveo/react-vapor/blob/master/README.md) is adjusted to reflect the proposed changes (if relevant)

PR #1262, opened 2019-09-19
Add Form and FormGroup components

francoislg:
### Proposed Changes

The idea is to have simple wrappers for vapor styles and, ideally, if you always use the `Form` > `FormGroup` > `Input` structure, you should get consistent styling everywhere.

`Form` wraps `coveo-form` and offers simple props to configure it (like `asCard` which I find really helpful for discoverability) and `FormGroup` wraps `form-group`.

Reopens #1247 

None, the only style change requires a new mod class

### Acceptance Criteria


francoislg:
☝️ I changed some stuff in the last commit:

Mainly ensuring that by default, anything will have *some* padding in both `coveo-form` and `form-group`. This is done non-breakingly by adding a mandatory `mod-padding-children` on both elements.

packages/react-vapor/src/components/form/examples/FormExamples.tsx
```diff
@@ -0,0 +1,69 @@
...
+import {Label} from '../../input/Label';
+import {Radio} from '../../radio/Radio';
+import {RadioSelect} from '../../radio/RadioSelect';
+import {Form} from '../Form';
+import {FormGroup} from '../FormGroup';
+
+export class FormExamples extends React.Component<any, any> {
+    render() {
+        return (
+            <div className="mt2">
```
gdostie: add `// start-print` on a new line before line 13

packages/react-vapor/src/components/form/examples/FormExamples.tsx
```diff
@@ -0,0 +1,69 @@
...
+                                <Label>Value 1</Label>
+                            </Radio>
+                            <Radio value="2">
+                                <Label>Value 2</Label>
+                            </Radio>
+                        </RadioSelect>
+                    </FormGroup>
+                </Form>
+            </div>
+        );
```
gdostie: add `// stop-print` on a new line after line 66

packages/react-vapor/src/components/form/Form.tsx
```diff
@@ -0,0 +1,29 @@
+import * as classNames from 'classnames';
+import * as React from 'react';
+
+type FormMods = 'mod-header-padding' | 'mod-form-top-bottom-padding';
+
+export interface IFormProps {
+    title?: string;
+    className?: string;
+    asCard?: boolean;
```
GermainBergeron: Will you apply the suggestion of removing asCard and using a mod instead?

francoislg: I could but never got a response here: https://github.com/coveo/react-vapor/pull/1247#discussion_r322359734
Thoughts? :P

GermainBergeron: The way I see it it modifies the way the form is displayed so it should be a mod

Eventually we could rename the class but it would be breaking for now

PR #1271, opened 2019-09-24
Better Performance in the SlideY and Svg Components

GermainBergeron:
### Proposed Changes

Don't re-calculate the SlideY on unmount
Use a string replace in Svg instead of creating and updating a DOM node

Load all exemples and not just those in the components folder

All the tests are still green and the examples are working so I expect everything to be good

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/docs/src/components/ComponentPage.tsx
+++ b/packages/react-vapor/docs/src/components/ComponentPage.tsx
@@ -31,7 +31,7 @@ const ComponentPage: React.FunctionComponent<ComponentPageProps> = (props) => {
         };
         const loadAll = () => {
             const mdFiles = require.context(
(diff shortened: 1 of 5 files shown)


gdostie:
The `Collapsible`, `CollapsibleContainer` and `CollapsibleInfoBox` doesn't seem to work anymore. Whats even weirder is that they don't work on the master's demo either.

GermainBergeron:
> The `Collapsible`, `CollapsibleContainer` and `CollapsibleInfoBox` doesn't seem to work anymore. Whats even weirder is that they don't work on the master's demo either.

If you refresh the page in those, the collapsible will work. I didn't investigate more then that since the behaviour is the same on master

PR #1277, opened 2019-09-25
Add ColorPicker component

olamothe:
### Proposed Changes

Adds a new component which display a color picker, using `react-color`: https://casesandberg.github.io/react-color/

It's synched with an hidden "InputConnected" so that it's possible to retrieve the currently selected color from the state

- None that I know of ;) 

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/package.json
+++ b/packages/react-vapor/package.json
@@ -43,7 +43,7 @@
         "diff2html": "2.5.0",
         "react-bootstrap": "0.32.4",
         "react-codemirror2": "5.1.0",
-        "react-color": "^2.17.3",
+        "react-color": "2.17.3",
         "react-infinite-scroll-component": "4.2.0",
         "react-modal": "3.10.1",
         "react-tether": "1.0.4",
@@ -63,7 +63,7 @@
     },
     "devDependencies": {
         "@loadable/component": "5.10.2",
-        "@types/react-color": "^3.0.1",
+        "@types/react-color": "3.0.1",
         "@types/chosen-js": "1.6.2",
         "@types/classnames": "2.2.6",
         "@types/codemirror": "0.0.62",
--- a/packages/react-vapor/package.json
+++ b/packages/react-vapor/package.json
@@ -43,6 +43,7 @@
         "diff2html": "2.5.0",
         "react-bootstrap": "0.32.4",
         "react-codemirror2": "5.1.0",
+        "react-color": "^2.17.3",
         "react-infinite-scroll-component": "4.2.0",
         "react-modal": "3.10.1",
         "react-tether": "1.0.4",
@@ -62,6 +63,7 @@
     },
     "devDependencies": {
         "@loadable/component": "5.10.2",
+        "@types/react-color": "^3.0.1",
         "@types/chosen-js": "1.6.2",
         "@types/classnames": "2.2.6",
         "@types/codemirror": "0.0.62",
--- a/packages/react-vapor/src/components/color-picker/ColorPicker.spec.tsx
+++ b/packages/react-vapor/src/components/color-picker/ColorPicker.spec.tsx
@@ -2,9 +2,10 @@ import {mount, shallow} from 'enzyme';
 import * as React from 'react';
 import {ChromePicker} from 'react-color';
 import {Provider} from 'react-redux';
-import {getStoreMock} from '../../utils/tests/TestUtils';
+import {getStoreMock, TestUtils} from '../../utils/tests/TestUtils';
 import {InputConnected} from '../input/InputConnected';
-import {ColorPicker} from './ColorPicker';
+import {InputSelectors} from '../input/InputSelectors';
+import {ColorPicker, ColorPickerConnected} from './ColorPicker';
 
 describe('ColorPicker', () => {
     it('should mount and unmount without error', () => {
@@ -37,5 +38,15 @@ describe('ColorPicker', () => {
         const picker = shallow(<ColorPicker color="#fff" />);
         expect(picker.find(ChromePicker).prop('color')).toBe('#fff');
         expect(picker.find(InputConnected).prop('defaultValue')).toBe('#fff');
+    });
+
+    it('should make the color available from state', () => {
+        const store = TestUtils.buildStore();
+        mount(
+            <Provider store={store}>
+                <ColorPickerConnected defaultColor="#47FF21" id="color-picker-test" />
+            </Provider>
+        );
+        expect(InputSelectors.getValue(store.getState(), {id: 'color-picker-test'})).toBe('#47FF21');
     });
 });
--- a/packages/react-vapor/src/components/color-picker/ColorPicker.spec.tsx
+++ b/packages/react-vapor/src/components/color-picker/ColorPicker.spec.tsx
@@ -1,4 +1,4 @@
-import {shallowWithStore} from 'enzyme-redux';
+import {mountWithStore, shallowWithStore} from 'enzyme-redux';
 import * as React from 'react';
 import {ChromePicker} from 'react-color';
 import {getStoreMock, ReactVaporMockStore, TestUtils} from '../../utils/tests/TestUtils';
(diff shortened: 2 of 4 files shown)


francoislg:
🤔 

image

olamothe:
> 🤔
> 
> image

Hum, you are right, there seems to be a problem with the example if you load another page first, then switch to the ColorPicker example, it returns undefined.

If you only load the example page first, it's fine. I'll investigate.

packages/react-vapor/src/components/color-picker/ColorPicker.tsx
```diff
@@ -0,0 +1,84 @@
...
+import {addInput, changeInputValue, removeInput} from '../input/InputActions';
+import {InputConnected} from '../input/InputConnected';
+import {InputSelectors} from '../input/InputSelectors';
+
+export interface IColorPickerProps extends ChromePickerProps {
+    id?: string;
+    defaultColor?: string;
+}
+
+export interface IColorPickerDispatchProps {
```
GermainBergeron: Remove this type, instead we use `ReturnType<typeof mapDispatchToProps>`. E.g,

`export class ColorPicker extends React.Component<IColorPickerProps & ReturnType<typeof mapStateToProps> & ReturnType<typeof mapDispatchToProps>> {`

packages/react-vapor/src/components/color-picker/ColorPicker.tsx
```diff
@@ -0,0 +1,84 @@
...
+const mapDispatchToProps = (dispatch: IDispatch, ownProps: IColorPickerProps): IColorPickerDispatchProps => ({
+    componentDidMount: (value: string = '', valid = true, disabled = false) => {
+        dispatch(addInput(ownProps.id, value, valid, disabled));
+        dispatch(changeInputValue(ownProps.id, value, true));
+    },
+    onDestroy: () => dispatch(removeInput(ownProps.id)),
+    onChangeComplete: (colorPicked: ColorResult) => dispatch(changeInputValue(ownProps.id, colorPicked.hex, true)),
+});
+
+export class ColorPicker extends React.Component<IColorPickerProps & IColorPickerDispatchProps> {
```
GermainBergeron: We stopped creating both version (connected & not connected) of components since we only used the connected ones in the admin. The non-connected version should not be exported

```typescript
class ColorPickerDisconnected extends React.PureComponent<...> {
...
export const ColorPicker = connect(
```

(also your component uses the InputConnected so it wouldn't work in a non-connected environment)

packages/react-vapor/src/components/color-picker/ColorPicker.tsx
```diff
@@ -0,0 +1,84 @@
...
+import {connect} from 'react-redux';
+import {isString, noop, uniqueId} from 'underscore';
+import {IReactVaporState} from '../../ReactVapor';
+import {IDispatch} from '../../utils/ReduxUtils';
+import {addInput, changeInputValue, removeInput} from '../input/InputActions';
+import {InputConnected} from '../input/InputConnected';
+import {InputSelectors} from '../input/InputSelectors';
+
+export interface IColorPickerProps extends ChromePickerProps {
+    id?: string;
```
GermainBergeron: id should be required otherwise it will be impossible to get the input value via Redux

packages/react-vapor/src/components/color-picker/ColorPickerExamples.tsx
```diff
@@ -0,0 +1,56 @@
...
+import {ColorResult} from 'react-color';
+import {Provider} from 'react-redux';
+import {debounce} from 'underscore';
+import {ReactVaporStore} from '../../../docs/ReactVaporStore';
+import {Button} from '../button/Button';
+import {InputSelectors} from '../input/InputSelectors';
+import {ColorPicker, ColorPickerConnected} from './ColorPicker';
+
+const logColorPicked = debounce((colorPicked: ColorResult) => {
+    /* tslint:disable */
```
GermainBergeron: `// tslint:disable-next-line`

packages/react-vapor/package.json
```diff
@@ -43,6 +43,7 @@
         "diff2html": "2.5.0",
         "react-bootstrap": "0.32.4",
         "react-codemirror2": "5.1.0",
+        "react-color": "^2.17.3",
```
GermainBergeron: We use fixed dependencies version to avoid issue

```suggestion
        "react-color": "2.17.3",
```

packages/react-vapor/package.json
```diff
@@ -62,6 +63,7 @@
     },
     "devDependencies": {
         "@loadable/component": "5.10.2",
+        "@types/react-color": "^3.0.1",
```
GermainBergeron:
```suggestion
        "@types/react-color": "3.0.1",
```

packages/react-vapor/src/components/color-picker/ColorPicker.spec.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+    it('should add state input on mount', () => {
+        const store = getStoreMock();
+        spyOn(store, 'dispatch');
+        mount(
+            <Provider store={store}>
+                <ColorPicker id="foo" />
+            </Provider>
+        );
+        expect(store.dispatch).toHaveBeenCalledWith(addInput('foo'));
```
GermainBergeron: We use enzyme-redux's `shallowWithStore` and `mountWithStore` to get the actions from the state, it's a bit easier to read

```
it('should add state input on mount', () => {
    shallowWithStore(<ColorPicker id="foo" />, getStoreMock());
    expect(store.getActions()).toContain(addInput('foo'));
});
```

packages/react-vapor/src/components/color-picker/ColorPicker.spec.tsx
```diff
@@ -0,0 +1,61 @@
...
+describe('ColorPicker', () => {
+    let store: ReactVaporMockStore;
+
+    beforeEach(() => {
+        store = getStoreMock();
+    });
+
+    it('should mount and unmount without error', () => {
+        expect(() => {
+            const picker = mountWithStore(<ColorPicker />, store);
```
gdostie: use shallow instead, no need to mount here

packages/react-vapor/src/components/color-picker/ColorPicker.spec.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+    it('should sync the InputConnected and ChromePicker with color props', () => {
+        const picker = shallowWithStore(<ColorPicker defaultColor="#fff" />, store).dive();
+        expect(picker.find(ChromePicker).prop('color')).toBe('#fff');
+        expect(picker.find(InputConnected).prop('defaultValue')).toBe('#fff');
+    });
+
+    it('should make the color available from state', () => {
+        const nonMockStore = TestUtils.buildStore();
+        mountWithStore(<ColorPicker defaultColor="#47FF21" id="color-picker-test" />, nonMockStore);
```
gdostie: is it really necessary to mount?

packages/react-vapor/src/components/color-picker/ColorPicker.spec.tsx
```diff
@@ -0,0 +1,61 @@
...
+        expect(InputSelectors.getValue(nonMockStore.getState(), {id: 'color-picker-test'})).toBe('#47FF21');
+    });
+
+    it('should add state input on mount', () => {
+        mountWithStore(<ColorPicker id="foo" />, store);
```
gdostie: is it really necessary to mount?

packages/react-vapor/src/components/color-picker/ColorPicker.spec.tsx
```diff
@@ -0,0 +1,61 @@
...
+        expect(store.getActions()).toContain(addInput('foo'));
+    });
+
+    it('should remove state input on destroy', () => {
+        const picker = mountWithStore(<ColorPicker id="foo" />, store);
```
gdostie: is it really necessary to mount?

packages/react-vapor/src/components/color-picker/ColorPicker.tsx
```diff
@@ -0,0 +1,81 @@
...
+const mapStateToProps = (state: IReactVaporState, ownProps: IColorPickerProps) => {
+    const color = InputSelectors.getValue(state, {id: ownProps.id}) || ownProps.defaultColor;
+    return {
+        color,
+    };
+};
+
+const mapDispatchToProps = (dispatch: IDispatch, ownProps: IColorPickerProps) => ({
+    componentDidMount: (value: string = '', valid = true, disabled = false) => {
+        dispatch(addInput(ownProps.id, value, valid, disabled));
```
gdostie: I don't understand why we add an input here. We already render an `InputConnected` which is supposed to add itself in the store IIRC.

packages/react-vapor/src/components/color-picker/ColorPicker.tsx
```diff
@@ -0,0 +1,81 @@
...
+        dispatch(changeInputValue(ownProps.id, value, true));
```
gdostie: I don't understand why we need this either. We already set the `defaultValue` to the `InputConnected` so its supposed to have the right value from the start.

olamothe:
I've had to revert some UT to use `mount` instead of `shallow` since those particular tests are interested in looking at the actual "state" being changed, which requires full component lifecycle.

PR #1278, opened 2019-09-26
popover doc

Charles-ErickBG:
### Proposed Changes

popover documentation

none

### Acceptance Criteria

none


otaillon:
> [Live demo available here](https://coveo.github.io/react-vapor/4a19c50c68427d8cc89565ea8d3d5e75/index.html).

La démo fonctionne pas pour moi 😢

packages/react-vapor/src/components/popover/examples/PopoverConnectedExamples.tsx
```diff
@@ -11,3 +11,7 @@ export const PopoverConnectedExamples = () => (
         </PopoverConnected>
     </div>
 );
+
+export const PopoverExample: ExampleComponent = PopoverConnectedExamples;
```
gdostie: Example files must import only one component :grimacing: 
So you can just remove export on [line 5](https://github.com/coveo/react-vapor/pull/1278/files#diff-6f917e48ba25fcbda1f6b4ef361f2031R5)

packages/react-vapor/src/components/popover/examples/PopoverConnectedExamples.1.use.md
```diff
@@ -0,0 +1,12 @@
+## Guidelines
+
+### Do Not
+
+-   Do not add a title when doing so would not add value to the interaction (e.g., if a popover is triggered from an icon button without a label, you need to explain the context in the title).
+-   Do not display more than one Popover at a time.
```
kpoulincoveo: What I think is missing. Not sure this is exactly the cases, but we should discuss.
- What interactive element should we use
- Not sure I understand the point about the title (probably because the example isn't complete enough)
- ONE Popover can be triggered open by default to act as a little tutorial when a user visits a page for the first time. 
- About having only one at a time: since they appear on click, it would technically be impossible to have two at once (unless you have two open by default)
- Should we explain how this related/is different from the technical documentation?

otaillon:
Agree with @kpoulincoveo's comments.

packages/react-vapor/src/components/popover/examples/PopoverConnectedExamples.tsx
```diff
@@ -1,8 +1,9 @@
 import * as React from 'react';
 
+import {ExampleComponent} from '../../../../docs/src/components/ComponentsInterface';
 import {PopoverConnected} from '../PopoverConnected';
 
-export const PopoverConnectedExamples = () => (
+const PopoverConnectedExamples = () => (
     <div className="mt2">
         <h1 className="text-blue mb1 bold">Popover Connected</h1>
         <PopoverConnected id="popover-connected-example1" attachment="top left" targetAttachment="bottom left">
```
gdostie: The example is quite weird, I see some things that could be done to improve it:
- Add the `inline-block` class to the button
- Remove the title `<h1 className="text-blue mb1 bold">Popover Connected</h1>` (this information is now in the header)
- Add a margin after the button so that the popover doesn't opend in a sneaky way (grey on grey) above the source code block. 
image

packages/react-vapor/src/components/popover/examples/PopoverConnectedExamples.tsx
```diff
@@ -11,3 +12,7 @@ export const PopoverConnectedExamples = () => (
         </PopoverConnected>
     </div>
 );
+
+export const PopoverExample: ExampleComponent = PopoverConnectedExamples;
+PopoverConnectedExamples.description =
+    'Popovers appear when users click an interactive element. Popovers allow you to display various types of information under different layouts (header, footer, split layout, etc.) without leaving the page. Their design also allows users to know where to click in order to reopen the Popover, if needed.';
```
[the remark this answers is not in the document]
Charles-ErickBG: I agree that it is long. However, I don't know how to fix it exactly.

IMO, the description should guide users toward their choice of component without having to read the rest of the documentation (this part should be used once the choice has been made). If we remove that part of the description, we remove one of the elements that could be key in the decision making process.  

As for the code, I agree. I'll see what I can do and we might do it in an upcoming PR.

packages/react-vapor/src/components/popover/examples/PopoverConnectedExamples.1.use.md
Charles-ErickBG: Here is my two cent on these points:

1. Any
2. I'll see what I can come up with
3. You are describing the rich popover component that is used within the Admin Console. This will not necessarily be used elsewhere and Vapor is meant to remain Product agnistic.
4. I see that this is not clear since I believe that the idea is to avoid having 2 popovers opened in a screen (on different elements). It does not necessarily refer to multiple popovers on the same element.  I'll work to clarify this point.
5. Not sure I understand your point here. Sorry :/

kpoulincoveo: I think I missed interpreted the popover for the rich popover actually. Probably because you refer to having various information layout including splits, etc.

packages/react-vapor/src/components/popover/examples/PopoverConnectedExamples.tsx
```diff
@@ -1,13 +1,77 @@
...
+        <div className="display-block w100 mb5">
+            <p className="h3 mb2">{loremIpsum({count: 5, sentenceUpperBound: 15})}</p>
+            <p className="h3 mb2">{loremIpsum({count: 5, sentenceUpperBound: 15})}</p>
+        </div>
+    </div>
+);
+
+export const PopoverExample: ExampleComponent = PopoverConnectedExamples;
+PopoverConnectedExamples.description =
+    'Popovers provide additional details or interactions. They are visually attached to an interactive element and support various content and layouts.';
```
jpmarceau: They are visually attached to interactive elements and support various contents and layouts.

PR #1279, opened 2019-09-26
add Defaults DropRoot as default parentSelector for a SelectConnected

pochretien:
### Proposed Changes

Every SelectConnect use the default drop root as parent Selector to avoid bad rendering when a single select is used inside a small container "position: relative"

I not sure yet but I dont think its a bad idea to always use the default root as parent selector to have the maximum space to render the dropdown

### Acceptance Criteria


GermainBergeron:
If you open a dropdown in the tableHOC demo it crashes

PR #1284, opened 2019-09-30
Fix Demo Always Reload the all the Components

GermainBergeron:
### Proposed Changes

We never need to fetch the full list of components once loaded

None

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/docs/src/components/index.tsx
+++ b/packages/react-vapor/docs/src/components/index.tsx
@@ -26,7 +26,7 @@ const Components: React.FunctionComponent<RouteComponentProps> = ({match}) => {
             return Promise.all(componentFiles.keys().map((path) => load(path, componentFiles)));
         };
         loadAll().then((all) => setComponents(all.filter(Boolean)));
-    }, [match]);
+    }, []);
     const routes = components
         .sort((a: IComponent, b: IComponent) => a.name.localeCompare(b.name))
         .map(({path, ...rest}: IComponent) => (


packages/react-vapor/docs/src/components/index.tsx
```diff
@@ -26,7 +26,7 @@ const Components: React.FunctionComponent<RouteComponentProps> = ({match}) => {
             return Promise.all(componentFiles.keys().map((path) => load(path, componentFiles)));
         };
         loadAll().then((all) => setComponents(all.filter(Boolean)));
-    }, [match]);
+    }, []);
```
pochretien: can we remove the second parameter if the array contains no conditions ?

gdostie:
Demo not working :thinking:

packages/react-vapor/docs/src/components/index.tsx
GermainBergeron: No, if I remove the argument it will be called at every render. Here I want to load the components only once 

From the docs:
>If you want to run an effect and clean it up only once (on mount and unmount), you can pass an empty array ([]) as a second argument. This tells React that your effect doesn’t depend on any values from props or state, so it never needs to re-run. This isn’t handled as a special case — it follows directly from how the dependencies array always works.

GermainBergeron:
> Demo not working 🤔
Try https://coveo.github.io/react-vapor/474f03f92c22cef7d049a8db20639f38/

PR #1289, opened 2019-10-03
Overhaul of FormGroup + new components to better support mock ups

francoislg:
### Proposed Changes

I had a look at some implementations that tried to use `Form` and `FormGroup` and it did not quite support what was required by the mock-ups.

The original premise of using only two levels (Form > FormGroup) is not consistent with what UX has in mind. For instance, they start from the bottom input and define levels to represent those inputs as groups.

I still wanted to provide some "structure" components that would provide props for most of UXs use cases, so here is the new design:

```
<Form>
  <FormGroup title="" description="">
     <LabeledInput title="" description="">
           <Any Input />
     </LabeledInput>
  </FormGroup>
  <FormGroup>
       <FormGroup level={2}>
           ...
       </FormGroup>
  </FormGroup>
</Form>
```

`FormGroup` is used to add a *header* and *delimits* a section. It provides an optional title and description with the appropriate styling. `level` is introduced to properly output `h2/h3/h4` depending on the level.

`LabeledInput` is introduced to *wrap* inputs with a label and/or description. 

For instance: 
* `Input` already provides its own `label`, so you don't need to put one. But `description` might still be useful.
* `SingleSelectConnected` does not have a label, so wrap it and define `title` and/or `description`

So in theory, using those two components together will ensure that you have consistent:

* Font color and size
* Spacing
* Content location

The FormExample has been updated with two real-life mock-ups, and they pretty much look the same :)

Let me know if you have any ideas to improve this, and if there should be changes in the styling

All of the changed attributes are still supported

### Acceptance Criteria


gdostie:
Is there something we can do about the accumulated spacing top?

image

packages/react-vapor/src/components/form/FormGroup.tsx
```diff
@@ -1,16 +1,34 @@
...
+    description,
+    className,
+    mods,
+    level,
+}) => {
+    const titleProps = {
+        className: 'text-medium-blue mb2',
+        children: title,
+    };
+    const levelElement = [() => <h2 {...titleProps} />, () => <h3 {...titleProps} />, () => <h4 {...titleProps} />];
```
gdostie: I think something like this would be easier to read.
```tsx
const H = level ? `h${level}` : 'h1';
return (
    ...
    {title && <H {...titleProps} />}
);
```

francoislg: > Is there something we can do about the accumulated spacing top?

You're right, I had a wrong CSS rule check the demo back with the latest commit

francoislg: And TIL about dynamic name tag, amazing stuff there

francoislg:
Bon ça a brassé pas mal avec les quelques dernières recommandations de UX :D

On attends la prochaine démo et ils vont regarder ce que ça donne

Charles-ErickBG:
It looks like some very very nice stuff!!!
**GG**
I love it :D

francoislg:
In my opinion, permitting everything in `title` and `label` goes against the purpose of this component.

`<FormGroup title={<Input/>}` makes no sense, nor `<FormGroup title={<Description />}`. 

I would rather add props for missing features, fronted with an *intent*. (For instance, `helpText` and `helpLink` instead of `tooltip`. 

I understand the concern for flexibility, but I would rather limit the options to ensure that everyone using those components are in-line with UX's recommendations

francoislg:
Coverage is now increased 🎉 :discoball:

francoislg:
I think it would be ready for another round of review (or a merge) :)

Charles-ErickBG:
Very nice! GG!

Eventually, I'd see a couple of improvements:
1. Like discussed earlier, call this SECTION to apply it to non-form contents
2. Change the props name for
     Title (no change)
     **Help text** (instead of description, since people might think they don't have to add a description, since the label already describes the input. Calling it "Help text" makes it more obvious that it is meant to help user enter their value)
     **Option information** (instead of information, because we should not use tooltips for non-optional info.)

francoislg:
@Charles-ErickBG I applied your proposed changes. Since the component had to be renamed, I opted to revert back the original `FormGroup` component to its original behavior and remove the example so that it doesn't appear in the documentation

Charles-ErickBG:
> @Charles-ErickBG I applied your proposed changes. Since the component had to be renamed, I opted to revert back the original `FormGroup` component to its original behavior and remove the example so that it doesn't appear in the documentation
---
That is just amazing stuff to me! I love it!!! GG.

packages/react-vapor/src/components/input/LabeledInput.tsx
```diff
@@ -0,0 +1,44 @@
...
+                ) : null}
+            </header>
+        ) : null;
+
+    return (
+        <>
+            {header}
+            {children}
+            {!!message && <InputDescription>{message}</InputDescription>}
+            {!!description && <InputDescription>{description}</InputDescription>}
```
GermainBergeron: What is the difference between the message and the description?

francoislg: Message => To tell the user something specific about the input
Description => To describe the input and its purpose

packages/react-vapor/src/components/input/LabeledInput.tsx
```diff
@@ -0,0 +1,44 @@
+import * as classNames from 'classnames';
+import * as React from 'react';
+import {TooltipPlacement} from '../../utils/TooltipUtils';
+import {Svg} from '../svg/Svg';
+import {Tooltip} from '../tooltip/Tooltip';
+import {InputDescription} from './InputDescription';
+
+export interface ILabeledInputProps {
+    label?: string;
```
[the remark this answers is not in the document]
pochretien: use ReactNode instead of string, ReactNode already contains string and we use this props like a reactNode. Don't restrain the user possibility.

PR #1297, opened 2019-10-08
text input documentation

Charles-ErickBG:
### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/input/examples/InputExamples.1.use.md
+++ b/packages/react-vapor/src/components/input/examples/InputExamples.1.use.md
@@ -2,14 +2,14 @@
 
 ### Restrictions & Size
 
--   Users should be informed of any input restrictions such as character types or string lenght.
--   The text input width should give a hint on the expected size of the content to enter. Try to forsee the usual string length so as to avoid hiding parts of the entered strings. The text input field should be large enough to see the whole value in most cases.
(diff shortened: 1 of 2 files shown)


packages/react-vapor/src/components/input/examples/InputExamples.1.use.md
```diff
@@ -0,0 +1,20 @@
+## Guidelines
+
+### Restrictions & Size
+
+-   Users should be informed of any input restrictions such as character types or string lenght.
```
francoislg:
```suggestion
-   Users should be informed of any input restrictions such as character types or string length.
```

packages/react-vapor/src/components/input/examples/InputExamples.1.use.md
```diff
@@ -0,0 +1,20 @@
...
+-   The text input width should give a hint on the expected size of the content to enter. Try to forsee the usual string length so as to avoid hiding parts of the entered strings. The text input field should be large enough to see the whole value in most cases.
+-   Avoid setting a character limit or make sure to set it high enough to avoid constraining most user entry. If this is not possible, let users know how many characters are left.
```
francoislg: I'm curious where we should put this restriction and whether we should only show it after a certain threshold (like 10 or less characters left)

Charles-ErickBG: Showing it after a certain threshold does not seem like a bad idea to me at all :)

PR #1303, opened 2019-10-11
single select documentation and description

Charles-ErickBG:
### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/select/examples/SingleSelectExamples.1.use.md
+++ b/packages/react-vapor/src/components/select/examples/SingleSelectExamples.1.use.md
@@ -1,0 +1,24 @@
+## Guidelines
+
+### General
+
+-   If space allows it and there are 7 or less options to choose from, consider using [radio buttons](#/components/Radio) instead.
+-   Use the shortest strings possible as option label so as to limit the control's width to a minimum.
+
+### Options Sorting
+
+-   Always put the most common items at the top of the list. If no options is to be selected more often than the others, use alphabetical or chronological order.
+
+### Label
+
+-   In most cases, there should be a label to identify the control.
+
+---
+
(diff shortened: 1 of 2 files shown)


packages/react-vapor/src/components/select/examples/SingleSelectExamples.tsx
```diff
@@ -35,6 +35,8 @@ export interface ISingleSelectExamplesState {
 }
 
 export class SingleSelectExamples extends React.PureComponent<{}, ISingleSelectExamplesState> {
+    static description = 'Single Selects allow to choose one single element among a list of two or more options.';
```
gdostie: > a list of two or more options

What about custom values? Isn't it legit to use a single select with custom values when there are no options initially but you can add new options gradually in the life span of the application?

packages/react-vapor/src/components/select/examples/SingleSelectExamples.1.use.md
```diff
@@ -0,0 +1,24 @@
+## Guidelines
+
+### General
+
+-   If space allows it and there are 7 or less options to choose from, consider using [radio buttons](#/components/Radio) instead.
```
gdostie: I think it would be worth mentioning the ability to add custom values.

otaillon:
I'm having a problem with the name of this component. In all the design systems I've checked, it is called a dropdown. 

Shouldn't _single select_ be a prop of dropdown ?

Charles-ErickBG:
> I'm having a problem with the name of this component. In all the design systems I've checked, it is called a dropdown.
> 
> Shouldn't _single select_ be a prop of dropdown ?

Yep, we will work on grouping them pretty soon.

PR #1307, opened 2019-10-21
feat[demo new Vapor] grouped checkboxes and radio plus loading

dmgauthier:
### Proposed Changes

Grouped the Radio and checkbox examples. Plus added loading instead of empty array to prevent a rerender that removes the inputs from the store.

affect only the demo

### Acceptance Criteria


packages/react-vapor/docs/src/OneDemoToRuleThemAll.tsx
```diff
@@ -1,3 +1,4 @@
+/* And to Vapor bind them */
```
francoislg: 👌👌👌👌👌👌👌

best change we didn't need

packages/react-vapor/src/components/radio/examples/RadioExamples.tsx
```diff
@@ -1,21 +1,102 @@
...
+                <Radio id="Option3" name="enabledOptions" value="brown">
+                    <Label>Brown</Label>
+                </Radio>
+            </RadioSelectConnected>
+        </LabeledInput>
+
+        <Button
+            enabled
+            name="disable blue option"
+            onClick={() => ReactVaporStore.dispatch(setRadioSelect(radioSelectConnectedId, {disabledValues: ['blue']}))}
```
gdostie: The examples component should be connected. I know you copy pasted the code from the previous file, but it would still be cleaner to connect the examples component while at it :rabbit:

packages/react-vapor/docs/src/components/ComponentPage.tsx
```diff
@@ -41,20 +42,23 @@ const ComponentPage: React.FunctionComponent<ComponentPageProps> = (props) => {
...
         id: getTabId(tabName),
         title: tabName,
     });
 
-    const tabProps: ITabProps[] = hasMarkdownTabs && [
-        {id: getTabId('development'), title: component.firstTabLabel || 'Develop'},
-        ...tabs.sort((tabA, tabB) => tabA.order - tabB.order).map(mapTabConfigToProps),
-    ];
+    const tabProps: ITabProps[] = hasMarkdownTabs &&
+        tabs && [
```
GermainBergeron: no need to check for tabs here since `hasMarkdownTabs` will be false if you don't have any tabs

packages/react-vapor/src/components/checkbox/examples/CheckboxExamples.tsx
```diff
@@ -1,65 +1,183 @@
...
+            <Section title="Checkbox set">
+                <Checkboxset />
+                <CheckboxConnectedExamples />
+                <GroupableCheckboxConnectedExamples />
+            </Section>
+        );
+    }
+}
+
+const Checkboxset: React.FunctionComponent<ICheckboxWithState> = () => {
```
GermainBergeron: The first generic is the type of the `Props`, but here you don't have any props

GermainBergeron: I'm wondering if we should discard the non connected examples 🤔 Or at least those that don't have any behaviour

packages/react-vapor/docs/src/components/ComponentPage.tsx
dmgauthier: you are right ! :+1:

packages/react-vapor/src/components/checkbox/examples/CheckboxExamples.tsx
dmgauthier: :+1:

packages/vapor/scss/forms/block-form.scss
```diff
@@ -70,17 +70,22 @@ form {
...
                 margin-top: $spacing * 2;
                 &.input-field {
                     margin-top: $form-group-margin + $input-margin-top;
                 }
-                &.form-group {
-                    margin-top: $form-group-margin;
-                }
+            }
+
+            > *:not(:first-of-type).form-group {
```
francoislg: Those changes break the default spacing for the Sections

image

I think we need to see why you needed to change that for your use case and see whether we need to add something here or change the layout in your Examples

packages/react-vapor/src/components/checkbox/examples/CheckboxExamples.tsx
```diff
@@ -1,65 +1,142 @@
...
+                        <Label classes={['label']}>A parent checkbox connected</Label>
+                    </GroupableCheckboxConnected>
+                    <br />
+                    <GroupableCheckboxConnected
+                        id={parentId2 + 1}
+                        parentId={parentId2}
+                        defaultChecked={true}
+                        defaultDisabled
+                    >
+                        <Label classes={['label', 'mr1']}>Checkbox child 1</Label>
```
francoislg: There is kind of a pattern here.

* All checkbox labels have the `label` class
* All checkbox followed by another checkbox have `mr1`

Could it be put in CSS rules directly..?

packages/vapor/scss/forms/block-form.scss
dmgauthier: Indeed before merging I would like us three [Chuck, you, me] to meet and check if we keep the CSS as it is and modify my layout or go with Chuck Idea to wrap everything in Section(s)

packages/react-vapor/src/components/checkbox/examples/CheckboxExamples.tsx
dmgauthier: As it is It would more be a `mr4` and you are quite right about the pattern. I'll check that out.

packages/react-vapor/src/components/input/LabeledInput.tsx
```diff
@@ -34,11 +35,11 @@ export const LabeledInput: React.FunctionComponent<ILabeledInputProps> = ({
         ) : null;
 
     return (
-        <>
+        <Section level={3}>
```
francoislg: UX previously recommended against this change since it could be wrapper on level 3 or 4, are inputs always on level 3 now...?

packages/react-vapor/src/components/section/Section.tsx
```diff
@@ -25,7 +25,7 @@ export const Section: React.FunctionComponent<ISectionProps> = ({
     };
     const H = `h${(level || 1) + 1}`;
     return (
-        <fieldset className={classNames(className, mods, 'form-group mod-padding-children')}>
+        <fieldset className={classNames(className, mods, `level-${level || 1} form-group mod-padding-children`)}>
```
francoislg: I didn't know back then, but I think you could integrate the default value directly in the deconstruction:

```
{ 
  children, 
  ...
  level = 1
}
```

It would remove both `|| 1`

packages/vapor/scss/forms/block-form.scss
```diff
@@ -64,23 +64,30 @@ form {
             }
         }
 
-        &.mod-padding-children {
+        @mixin section-mixin($level) {
```
francoislg: Awesome stuff

packages/react-vapor/src/components/input/LabeledInput.tsx
dmgauthier: since @Charles-ErickBG want almost everything to fall in the right place from the get go, it seemed an easy way.
Do you think you should do it otherwise ?

packages/vapor/scss/forms/block-form.scss
dmgauthier: thx !

packages/react-vapor/src/components/input/LabeledInput.tsx
francoislg: I mean... my initial design suggested three levels:

`Form > FormGroup > Input`

Then it was changed to `Section` to allow having `Section (1) > Section (2) > Section (3) > Input` as well as `Section (1) > Section (2) > Input` as well as `Section (1) > Input`.

Having a level 3 hardcoded into the `LabeledInput` kind of bring it back to the original design and enforces 3 levels.

I mean.. I'm fine with it, but I would have kept the `Form > FormGroup > Input` structure if that's the use case now 😄

dmgauthier: When you put it this way, I think that you designed it like this for more flexibility. We should respect that, I totally agree. It is important that we all learn the pattern correctly to use the components in the good way so it will be easier for everyone. and maybe this is what I missed with the labeled Input.

packages/react-vapor/src/components/toast/examples/Toast2Examples.tsx
```diff
@@ -0,0 +1,23 @@
...
+// import {Toast} from '../Toast'
+// import {ToastContainer} from '../ToastConainer'
+import {ExampleComponent} from '../../../../docs/src/components/ComponentsInterface';
+import {Button} from '../../button/Button';
+import {Section} from '../../section/Section';
+
+export const ToastsExamples: ExampleComponent = () => <ToastsWithLocalState />;
+ToastsExamples.title = 'toasts demo';
+const ToastsWithLocalState: React.FunctionComponent = () => {
+    // const [toastLocation, setToastLocation] = React.useState({
```
francoislg: 👁

dmgauthier: :see_no_evil:

packages/react-vapor/docs/src/components/ComponentPage.tsx
```diff
@@ -50,11 +51,13 @@ const ComponentPage: React.FunctionComponent<ComponentPageProps> = (props) => {
 
     const tabProps: ITabProps[] = hasMarkdownTabs && [
         {id: getTabId('development'), title: component.firstTabLabel || 'Develop'},
-        ...tabs.sort((tabA, tabB) => tabA.order - tabB.order).map(mapTabConfigToProps),
+        ...tabs.sort((tabA: any, tabB: any) => tabA.order - tabB.order).map(mapTabConfigToProps),
```
gdostie:
```suggestion
        ...tabs.sort((tabA: TabConfig, tabB: TabConfig) => tabA.order - tabB.order).map(mapTabConfigToProps),
```

dmgauthier: :heavy_check_mark:

PR #1311, opened 2019-10-22
Add getSearchParams util in UrlUtil

gdostie:
### Proposed Changes

Added a small util to that parses the current url and returns the search params as an object. I added a new lib at the same time that handles query-strings in a graceful way. I will need this lib in some incoming PR.

I didn't do any tests for that one because its simply impossible to mock `window.location` in any way with jasmine :hankey: 

None

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/utils/UrlUtils.ts
+++ b/packages/react-vapor/src/utils/UrlUtils.ts
@@ -1,0 +1,13 @@
+import * as QueryString from 'query-string';
+
+/* istanbul ignore next */
+function getSearchParams() {
+    const currentUrl = window.location.href;
+    const qPos = currentUrl.lastIndexOf('?');
+    const query = qPos >= 0 ? currentUrl.substring(qPos, currentUrl.length) : '';
+    return {...QueryString.parse(query, {parseBooleans: true, parseNumbers: true})};
+}
+
+export const UrlUtils = {
+    getSearchParams,
+};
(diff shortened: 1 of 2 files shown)


packages/react-vapor/src/utils/UrlUtils.ts
```diff
@@ -0,0 +1,13 @@
+import * as QueryString from 'query-string';
+
+/* istanbul ignore next */
+function getSearchParams() {
+    const currentUrl = window.location.href;
+    const qPos = currentUrl.indexOf('?');
```
GermainBergeron: I think this won't work if you have a real querystring and a querystring in the hash (e.g., `http://some.website.com/?dev#path/to/page?param=5`)

You might have to do a 
```suggestion
    const qPos = currentUrl.lastIndexOf('?');
```

packages/react-vapor/src/utils/UrlUtils.ts
```diff
@@ -0,0 +1,13 @@
...
+    const query = qPos >= 0 ? currentUrl.substring(qPos, currentUrl.length) : '';
+    return Object.assign({}, QueryString.parse(query, {parseBooleans: true, parseNumbers: true}));
```
GermainBergeron:
```suggestion
    return {...QueryString.parse(query, {parseBooleans: true, parseNumbers: true})};
```

PR #1317, opened 2019-10-29
[COM-104] add product and channel icons

dlafreniere:
image


dlafreniere:
Cleaned up SVGs with https://jakearchibald.github.io/svgomg/

PR #1324, opened 2019-11-08
Remove turnOnLoading in the navigation components

rosalie-liu:
### Proposed Changes

The `turnOnLoading` in the `tableWithPagination` imposes the parent to turn off the loading after changing page, which causes issues for client-side tables. The client-side table does not require a loader while changing page, and for server-side tables, we can handle the loader with the fetch action.   

The loader won't be triggered by the onPageClick and onPerPageClick events.

### Acceptance Criteria


rosalie-liu:
> The code looks good to me. Did you try linking the project in the admin and having a check on other tables?

Yes, I also compared with the demo link, table row actions works correctly after changing page. This will fix the bug in the API Keys table as well.

PR #1353, opened 2019-11-26
fix(charts): quick fix to avoid undefined throwing

dmgauthier:
### Proposed Changes

Quickly fixed the `cannot apply method of undefined` errors that were throwing in the demo.
I did not took the time to really understand the components or the methods that were used, but I could check deeper if it worth the time.

plus adding some unit tests if need be.

**Edit:** After more research. It was the tooltip rendering that was at fault when not finding the parent component.

I did not change the implementation. Only avoided any undefined in the workflow.

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/drop/DropPod.tsx
+++ b/packages/react-vapor/src/components/drop/DropPod.tsx
@@ -122,9 +122,9 @@ class RDropPod extends React.PureComponent<IRDropPodProps, IDropPodState> {
     }
 
     private updateOffset() {
-        if (this.props.buttonRef && this.props.buttonRef.current) {
+        if (this.props.buttonRef?.current) {
             this.setState({
-                offset: this.props.buttonRef.current.getBoundingClientRect(),
+                offset: this.props.buttonRef?.current?.getBoundingClientRect(),
             });
         }
     }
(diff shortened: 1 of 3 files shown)


packages/react-vapor/src/components/drop/DropPod.tsx
```diff
@@ -83,8 +83,8 @@ class RDropPod extends React.PureComponent<IRDropPodProps, IDropPodState> {
             };
         } else {
             this.lastPosition = {
-                position: DropPodPosition.bottom,
-                orientation: DropPodPosition.left,
+                position: DropPodPosition?.bottom,
+                orientation: DropPodPosition?.left,
```
gdostie: You can remove the `?` here, `DropPodPosition` is a `const` with the following keys (which will always be defined).
```ts
export const DropPodPosition = {
    bottom: 'bottom',
    top: 'top',
    left: 'left',
    right: 'right',
};
```

GermainBergeron:
I want to point out that the page don't throw anymore but the tooltip is nowhere to be found

packages/react-vapor/src/components/drop/DropPod.tsx
dmgauthier: quite true

dmgauthier:
> I want to point out that the page don't throw anymore but the tooltip is nowhere to be found

Good Catch, I will look into that.

PR #1357, opened 2019-11-28
Add utils to facilitate applying filters

gdostie:
### Proposed Changes

When building custom match filter functions for the `tableWithFilter` or the `selectWithFilter`, we often do the same logic to compare the filter value with different types of data. So I extracted this logic in some util functions that we can reuse when needed.

- `defaultMatchFilter` function that was used for filtering list box was renamed to `defaultListBoxMatchFilter`.

### Acceptance Criteria


packages/react-vapor/src/components/filterBox/tests/FilterBoxSelectors.spec.ts
```diff
@@ -38,4 +40,64 @@ describe('FilterBox', () => {
             });
         });
     });
+
+    describe('defaultListBoxMatchFilter', () => {
```
gdostie: This whole described was only moved from another file

gdostie:
Real usage example:
```ts
match(filterValue: string, {displayName, id, members, realms}: GlobalGroupModel): boolean {
    return (
        FilterUtils.matchesString(filterValue, displayName) ||
        FilterUtils.matchesArrayLength(filterValue, members) ||
        FilterUtils.matchesArrayLength(filterValue, realms) ||
        filterValue === id
    );
}
```

packages/react-vapor/src/components/filterBox/FilterBoxSelectors.ts
```diff
@@ -1,12 +1,22 @@
...
+import {IItemBoxProps} from '../itemBox/ItemBox';
 import {IFilterState} from './FilterBoxReducers';
 
 export interface GetFilterTextProps {
     id: string;
 }
 
+export type MatchFilter = (filterValue: string, item: IItemBoxProps) => boolean;
+
+export const defaultListBoxMatchFilter = (filterValue: string, item: IItemBoxProps) => {
```
GermainBergeron: Why did you move it in the selector? It make more sense in the Utils, no?

gdostie: Because its bound to `IItemBoxProps`, so it belongs somewhere in the `ItemBox` related files. I did not want it to be mistaken for the defaultMatchFilter of the table, which doesn't use `IItemBoxProps`. 
The selectors might not be the right file tho.

pochretien: I think we can scope these filter with something like 
```
FilterUtils. defaultMatchFilter
TableUtils.defaultMatchFilter
```
what do you think ?

GermainBergeron: I don't mind the name, but the location of the code seems really unusual to me

gdostie: FilterUtils are also relevant for the table

gdostie: Moved the defaultListBoxFilter to a new FilterBoxUtils file

PR #1358, opened 2019-11-29
feat(middle slider): create a slider that starts from the middle

dmgauthier:
### Proposed Changes

Still Work in Progress

Adapted the `Range` slider from the `RC-Slider` library to act like a slider that starts from the middle.

Since the `Range` has normally two handes, what I did is to catch the good handle to go on the slider location, then put the other slider in the middle to respect the track from the center. The handle in the middle should be invisible to give the impression of an unique handle.
~The `range ` props for the `MiddleSlider` will give the user the ability to tell if the slider goes from a value to another. For example, a range of 10,000 will permit the handle to go from -10,000 to 10,000~
the `range` props of the slider is an array of two number [minimum, maximum]. The minimum should not be above 0 and the maximum not under 0. The crossing point of the slider will always be at 0.

I connected this component to the react-vapor store.

~TODO: a disabled state, make the tooltip work and UTS~

Bumped the RC-slider version. 
That broke the style at least.


packages/react-vapor/src/components/slider/MiddleSlider.tsx
```diff
@@ -0,0 +1,97 @@
...
+    const handlesAreNotInTheCenter = (handleIndex: number) =>
+        (handleIndex === 0 && lowRange <= 50 && highRange === 50) ||
+        (handleIndex === 1 && highRange >= 50 && lowRange === 50);
+
+    React.useEffect(() => {
+        const rangeOutputValue = getComputedRangeValue();
+        setOutputValue(rangeOutputValue);
+    });
+
+    const setRange = (value: number[]) => {
```
dmgauthier: SetHandlesPositions would probably be a bette name

packages/react-vapor/src/components/slider/MiddleSlider.tsx
```diff
@@ -0,0 +1,97 @@
...
+    tabIndex?: number[];
+}
+
+export const mapDispatchToProps = (dispatch: IDispatch, ownProps: MiddleSliderOwnProps) => ({
+    setOutputValue: (value: number) => dispatch(SliderActions.setValue(ownProps.id, value)),
+});
+
+export const MiddleSliderDisconnected: React.FunctionComponent<MiddleSliderOwnProps &
+    ReturnType<typeof mapDispatchToProps>> = ({range, setOutputValue, hasTooltip}) => {
+    const [highRange, setHighRange] = React.useState(50);
```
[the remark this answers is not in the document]
dmgauthier: If I take your suggestions. 50 would be the starting point for the slider. Putting that value in a `const` will indeed be mandatory

packages/react-vapor/src/components/slider/MiddleSlider.tsx
```diff
@@ -0,0 +1,97 @@
...
+import * as React from 'react';
+import {connect} from 'react-redux';
+
+import {IDispatch} from '../../utils/ReduxUtils';
+import {SliderActions} from './SliderActions';
+// import {Tooltip} from '../tooltip/Tooltip';
+
+export interface MiddleSliderOwnProps extends SliderProps {
+    id: string;
+    range: number;
```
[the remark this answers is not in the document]
dmgauthier: line 54. Could work that name out tough

packages/react-vapor/src/components/slider/MiddleSlider.tsx
```diff
@@ -0,0 +1,97 @@
...
+        setLowRange(value[0]);
+    };
+
+    const jumpValueFromLowToHighRange = (value: number[]) => {
+        setLowRange(50);
+        setHighRange(value[1]);
+    };
+
+    const setValueInTheLowRange = (value: number[]) => {
+        const correctLowRangeValue = value[1] < 50 ? value[1] : value[0];
```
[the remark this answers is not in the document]
dmgauthier: Would love to... 
but look. The ternary checks of the value index 1 is lower than 50. If the already chosen value is... lets say 10. When another value chosen is under the initial value (50) but bugger than 10. If we do not check this and we click nearer to the middle than the lowest handle(like 45), the initial range slider will want to put the handle to the right (index 1) on the wanted position. but since this handle value is forced on the original position (50). The handles will not move.

packages/react-vapor/src/components/slider/MiddleSlider.tsx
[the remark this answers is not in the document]
dmgauthier: Since I will apply your suggestions. This value represent the starting point of the slider. Putting this value in a const will indeed be mandatory.

dmgauthier:
> The example without tooltip works really well, this looks promising 
> 
> I think the current implementation lacks a bit of customization, here are some improvement ideas:
> 
> * **Asymmetric middle sliders**: I'm afraid the crossing point won't always be straight in the middle in real use cases.
> * **Custom increments**: We often want sliders to increment using steps (ex: 5 ticks at a time). The other slider we have supports this.
> * **Markers**: Having a mark at a given value is often requested by UX to indicate a recommendation or a threshold. The other slider we have supports this.

I Love all this.

dmgauthier:
On hold until UTs are ready

dmgauthier:
Just Realised I will have to add UTs for the SliderActions , the SliderReducers and the SliderSelectors.

packages/react-vapor/src/components/slider/MiddleSlider.tsx
```diff
@@ -0,0 +1,133 @@
...
+    onChange,
+}) => {
+    const crossingPoint = getCrossingPoint(range);
+    const [highRange, setHighRange] = React.useState(crossingPoint);
+    const [lowRange, setLowRange] = React.useState(crossingPoint);
+    const rangeOutputValue = getComputedRangeValue(lowRange, highRange, range, crossingPoint);
+
+    if (onChange) {
+        onChange(rangeOutputValue);
+    }
```
gdostie: Do we really want onChange to be called each time the component renders? I think it should be added as a side effect

```ts
React.useEffect(() => {
    onChange?.(rangeOutputValue);
}, [onChange, rangeOutputValue]);
```

packages/react-vapor/src/components/slider/MiddleSlider.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+    React.useEffect(() => {
+        propsValidator(range, initialValue);
+        if (initialValue) {
+            const initialRangeValue: number = convertInitialValuetoRangeValue(range, initialValue);
+            setHandlePosition([initialRangeValue, initialRangeValue]);
+        }
+    }, []);
```
gdostie:
```suggestion
    }, [range, initialValue]);
```

packages/react-vapor/src/components/slider/MiddleSlider.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+    React.useEffect(() => {
+        setOutputValue(rangeOutputValue);
+    }, [rangeOutputValue]);
```
gdostie:
```suggestion
    }, [setOutputValue, rangeOutputValue]);
```

packages/react-vapor/src/components/slider/SliderHandle.tsx
```diff
@@ -0,0 +1,34 @@
...
+    value: number;
+}
+
+const SliderHandle: React.FunctionComponent<{handleProps: HandleProps; handleCustomProps: CustomHandleProps}> = ({
+    handleProps,
+    handleCustomProps,
+}) => (
+    <Tooltip
+        prefixCls="rc-slider-tooltip"
+        overlay={handleCustomProps.customTooltip ? handleCustomProps.customTooltip : handleCustomProps.rangeOutput}
```
gdostie:
```suggestion
        overlay={handleCustomProps.customTooltip ?? handleCustomProps.rangeOutput}
```

packages/react-vapor/src/components/slider/SliderSelectors.ts
```diff
@@ -0,0 +1,14 @@
...
+import {IReactVaporState} from '../../ReactVapor';
+import {ISliderState} from './SliderReducers';
+
+const getSlider = (state: IReactVaporState, props: {id: string}): ISliderState =>
+    _.findWhere(state.sliders, {id: props.id});
+
+const getSliderValue = createSelector(getSlider, (slider: ISliderState): number => slider?.value);
+
+export const SliderSelectors = {
+    getSlider,
```
gdostie: no need to export getSlider, if it isn't used IMO

packages/react-vapor/src/components/slider/SliderUtils.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export const getValuesPositionOnRange = (value: number[], crossingPoint: number): string => {
+    if (value.every((singleValue) => singleValue < crossingPoint)) {
+        return valuesPositionOnRange.lower;
+    } else if (value.every((singleValue) => singleValue > crossingPoint)) {
+        return valuesPositionOnRange.higher;
+    } else if (value.every((singleValue) => singleValue !== crossingPoint)) {
+        return valuesPositionOnRange.both;
+    }
+    return null;
```
gdostie: what if every single value is at the crossing point?

packages/react-vapor/src/components/slider/MiddleSlider.tsx
dmgauthier: freaking awesome. this code feels good.

packages/react-vapor/src/components/slider/MiddleSlider.tsx
dmgauthier: I just want this to execute when the component mounts. So I don't think this is relevant.

packages/react-vapor/src/components/slider/MiddleSlider.tsx
dmgauthier: :+1: yeah

packages/react-vapor/src/components/slider/SliderHandle.tsx
dmgauthier: :soNullishCoalescingAdvocate: that is powerful stuff. :heart:

packages/react-vapor/src/components/slider/SliderSelectors.ts
dmgauthier: right :+1:

packages/react-vapor/src/components/slider/SliderUtils.ts
dmgauthier: both handles should render and overlaps thus giving the feel of a single handle.

packages/react-vapor/src/components/slider/MiddleSlider.tsx
gdostie: https://overreacted.io/a-complete-guide-to-useeffect/#dont-lie-to-react-about-dependencies
Written by the inventor of the react hooks :arrow_up: :slightly_smiling_face:

dmgauthier: ok you win 🤺

PR #1365, opened 2019-12-11
Feature adui 4263 time in iso format tooltip

FelixBlaisThon:
### Proposed Changes

Date Tooltip component that takes a date (moment object or string) and converts it in the given format (if defined) to be displayed in the table. It also displays a tooltip that can have its own tooltipFormat also (ISO-8601 by default).

### Acceptance Criteria


dmgauthier:
Could we use the actual tooltip implementation to acheve the same result ?

packages/react-vapor/src/components/tooltip/DateTooltip.tsx
```diff
@@ -0,0 +1,30 @@
...
+        }
+    } else if (moment.isMoment(props.date)) {
+        if (_.isFunction(props.format)) {
+            content = props.format(props.date);
+        } else if (_.isString(props.format)) {
+            content = props.date.format(props.format);
+        } else {
+            content = props.date.format();
+        }
+    }
```
gdostie: There are a lot of nested conditions. You can remove some of them by declaring some `const`, it will remove duplicated logic.

packages/react-vapor/src/components/tooltip/DateTooltip.tsx
```diff
@@ -0,0 +1,30 @@
+import * as moment from 'moment';
+import * as React from 'react';
+import * as _ from 'underscore';
+import {Tooltip} from './Tooltip';
+
```
pochretien: you can create a type like `export type DateTooltips = moment.Moment | string`

packages/react-vapor/src/components/tooltip/DateTooltip.tsx
```diff
@@ -0,0 +1,30 @@
...
+export interface DateTooltipsProps {
+    date: moment.Moment | string;
+    format?: ((date: moment.Moment) => string) | string;
+    tooltipFormat?: string;
+}
+
+export const DateTooltip: React.FunctionComponent<DateTooltipsProps> = (props) => {
```
pochretien:
```suggestion
export const DateTooltip: React.FunctionComponent<DateTooltipsProps> = ({tooltipFormat, date, format}) => {
```

packages/react-vapor/src/components/tooltip/examples/TooltipExamples.tsx
```diff
@@ -48,6 +50,25 @@ export class TooltipExamples extends React.Component<{}, {}> {
                         </Tooltip>
                     </div>
                 </div>
+                <div className="form-group">
+                    <label className="form-control-label">Time Tooltip example</label>
+                    <div className="form-control">
+                        <button type="button" className="btn">
```
pochretien:
```suggestion
                        <Button enabled>
```

packages/react-vapor/src/components/tooltip/DateTooltip.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export interface DateTooltipsProps {
+    date: DateTooltips;
+    format?: ((date: moment.Moment) => string) | string;
+    tooltipFormat?: string;
+}
+
+export const DateTooltip: React.FunctionComponent<DateTooltipsProps> = ({date, format, tooltipFormat}) => {
+    let content: string;
+    const title = tooltipFormat ? moment(date).format(tooltipFormat) : moment(date).format('LLL');
```
gdostie:
```suggestion
    const title = moment(date).format(tooltipFormat ?? 'LLL');
```

packages/react-vapor/src/components/tooltip/DateTooltip.tsx
```diff
@@ -0,0 +1,24 @@
...
+    const momentDate = _.isString(date) ? moment(date) : date;
```
gdostie: If `date` is a string, `momentDate` will be a string too right?

In the next `if-else` block, it will throw an error when calling `.format` on a string.
image

There should be a unit test for this case.

packages/react-vapor/src/components/tooltip/tests/DateTooltip.spec.tsx
```diff
@@ -0,0 +1,42 @@
...
+    let tooltipWrapper: ShallowWrapper<DateTooltipsProps>;
+    const TOOLTIP_PROPS: DateTooltipsProps = {
+        date: moment(),
+    };
+    const isoDate = moment().format();
+    const stringFormat = 'LLL';
+
+    describe('<DateTooltip />', () => {
+        beforeEach(() => {
+            tooltipWrapper = shallowWithState(<DateTooltip {...TOOLTIP_PROPS} />, {});
```
gdostie: The DateTooltip is not connected to redux, so no need to shallow with a state.
```suggestion
            tooltipWrapper = shallow(<DateTooltip {...TOOLTIP_PROPS} />);
```

packages/react-vapor/src/components/tooltip/tests/DateTooltip.spec.tsx
```diff
@@ -0,0 +1,42 @@
...
+            expect(tooltipWrapper.find(Tooltip).length).toBe(1);
+        });
+
+        it('should display content in ISO-8601 if no format is given on moment date', () => {
+            expect(tooltipWrapper.find(Tooltip).props().title).toEqual(isoDate);
+        });
+
+        it('should display the tooltip in the given tooltipformat', () => {
+            const newProps: DateTooltipsProps = _.extend({}, TOOLTIP_PROPS, {tooltipFormat: stringFormat});
+            tooltipWrapper.setProps(newProps);
```
gdostie: I think it would be clearer & cleaner if we shallow the DateTooltip directly in each test instead of calling `setProps` on the wrapper.

packages/react-vapor/src/components/tooltip/DateTooltip.tsx
FelixBlaisThon: IIRC moment(date) will change string type to object type. In this case, moment object

packages/react-vapor/src/components/tooltip/DateTooltip.tsx
```diff
@@ -0,0 +1,24 @@
...
+import * as React from 'react';
+import * as _ from 'underscore';
+import {Tooltip} from './Tooltip';
+
+export type DateTooltips = moment.Moment | string;
+
+export interface DateTooltipsProps {
+    date: DateTooltips;
+    format?: ((date: moment.Moment) => string) | string;
+    tooltipFormat?: string;
```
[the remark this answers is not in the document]
FelixBlaisThon: Because we thought that the element containing the date could be more custom, but the tooltip would be a more formal type of format (which are covered with .format('style') of moment). But if we see he need to make it possible to use a function, i don't mind changing it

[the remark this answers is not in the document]
gdostie: Makes sense to me, I was wondering if it would be better for consistency to have the same logic for the 2 formats.

PR #1370, opened 2019-12-17
+dataId as an ActionableItemPropd interface attribute

FelixBlaisThon:
### Proposed Changes

+dataId attribute in IActionableItemProps interface

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/actionable-item/ActionableItem.tsx
+++ b/packages/react-vapor/src/components/actionable-item/ActionableItem.tsx
@@ -3,7 +3,7 @@ import * as React from 'react';
 import * as _ from 'underscore';
 
 import {DropPodPosition} from '../drop/DomPositionCalculator';
-import {Drop} from '../drop/Drop';
+import {Drop, IDropOwnProps} from '../drop/Drop';
 import {IItemBoxProps} from '../itemBox/ItemBox';
 import {ListBox} from '../listBox/ListBox';
 import {Svg} from '../svg/Svg';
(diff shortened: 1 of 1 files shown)


packages/react-vapor/src/components/actionable-item/ActionableItem.tsx
```diff
@@ -11,6 +11,7 @@ import {actionableItemContainer, actionableItemContent, actionableItemDots} from
 
 export interface IActionableItemProps {
     id: string;
+    dataId?: string;
```
GermainBergeron: With the current name I don't really understand what it does except when I look at the code. Maybe you could instead name it `listContainerProps` and use it like this: `<ActionableItem listContainerProps={{id: 'some-string'}} />`

pochretien: I think the best way is to add the posibility to override all Drop props

in your interface
```
 dropProps : IDropOwnProps;
```


```
<Drop
                        id={this.props.id}
                        positions={[DropPodPosition.bottom, DropPodPosition.top]}
                        buttonContainerProps={{className: 'inline-block'}}
                        parentSelector={'body'}
                        renderOpenButton={(onClick: () => void) => (
                            <div
                                onClick={onClick}
                                className={classNames(
                                    'actionable-item-dots cursor-pointer inline-block mod-border-top mod-border-right border-color-medium-grey mod-border-bottom bg-pure-white',
                                    actionableItemDots,
                                    actionableItemContainer
                                )}
                            >
                                <Svg svgName="more-append" svgClass="icon mod-12 fill-medium-blue" />
                            </div>
                        )}
                        {...this.props.dropProps || {}}
                    >
                        <ListBox items={this.props.actions} />
                    </Drop>
```

PR #1371, opened 2019-12-18
#1 Table Loading: Update table server examples with the new workflow

pochretien:
Review only the code for this PR
Needs more component to have a functional demo

1. Remove the reducer and the connected component, use a internal state for data and isLoading
2. Move useful function 
3. Update functions with the new pagination (incoming)


packages/react-vapor/src/components/table-hoc/examples/TableHOCServerExample.tsx
```diff
@@ -0,0 +1,193 @@
...
+import {tableWithUrlState} from '../TableWithUrlState';
+import {TableHOCExampleUtils} from '../utils/TableHOCExampleUtils';
+import {ITableHOCCompositeState, TableHOCUtils} from '../utils/TableHOCUtils';
+
+type TableHOCServerProps = RouteComponentProps & ReturnType<typeof mapDispatchToProps>;
+
+export interface IExampleRowData {
+    city: string;
+    email: string;
+
```
gdostie:
```suggestion
```

packages/react-vapor/src/components/table-hoc/examples/TableHOCServerExample.tsx
```diff
@@ -0,0 +1,193 @@
...
+TableHOCServerExample.title = 'TableHOC server';
+
+// start-print
+export const TableHOCServerExampleId = 'complex-example';
+
+const renderHeader = (isLoading: boolean) => (
+    <thead>
+        <tr>
+            <TableRowNumberHeader isLoading={isLoading} />
+            <TableHeaderWithSort id="address.city" tableId={TableHOCServerExampleId} isLoading={isLoading}>
```
gdostie: Maybe its out of scope for now but I'm wondering if we could create a react context for the table in which we would put `tableId` and `isLoading`. Could be a future improvement

packages/react-vapor/src/components/table-hoc/examples/TableHOCServerExample.tsx
```diff
@@ -0,0 +1,193 @@
...
+            <TableRowHeader isLoading={isLoading} />
+        </tr>
+    </thead>
+);
+
+const mapDispatchToProps = (dispatch: IDispatch) => ({
+    fetch: () => dispatch(TableHOCServerActions.fetchData()),
+});
+
+class TableExampleDisconnected extends React.PureComponent<TableHOCServerProps, TableHOCServerExamplesState> {
```
dmgauthier: why reverting it to a class component ?

packages/react-vapor/src/components/table-hoc/utils/TableHOCExampleUtils.tsx
```diff
@@ -0,0 +1,74 @@
+import * as moment from 'moment';
```
dmgauthier: I understand the point of doing this in real development. But doesn't it defeat the purpose of showing how the table works in the example page ?

pochretien: I dont want to add noise in this demo. IMO its not important to see how to add a row and other stuff when you look at the server table example but how to use the table to integrate a server side behaviour. What do you think ?

I dont like file with a lot of codes :)

packages/react-vapor/src/components/table-hoc/examples/TableHOCServerExample.tsx
pochretien: to use the internal state, I can switch with a functional component with hooks but I prefer class when I have more than 1 value in the state

packages/react-vapor/src/components/table-hoc/examples/TableHOCServerExample.tsx
pochretien: added

packages/react-vapor/src/components/table-hoc/utils/TableHOCExampleUtils.tsx
dmgauthier: fair enough

PR #1375, opened 2019-12-19
Add the new pagination with the new loading

pochretien:
New loading for the Table HOC

image


packages/react-vapor/src/components/pagination/PaginationPagesNumber.tsx
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    render() {
+        const currentPage: number = this.props.currentPage || 0;
+
+        const showXPages: number = Math.abs((this.props.numberOfPagesToShow || NUMBER_OF_PAGES_SHOWING) - 1);
+        const previousLabel: string = this.props.previousLabel || PREVIOUS_LABEL;
+        const nextLabel: string = this.props.nextLabel || NEXT_LABEL;
+        let start: number = 0;
+        let end: number = showXPages;
```
dmgauthier: `Math.abs((this.props.numberOfPagesToShow || NUMBER_OF_PAGES_SHOWING) - 1);` equals `showXpages` equals `end`
Is it because you feel it is clearer that way ?

packages/react-vapor/src/components/pagination/PaginationPagesNumber.tsx
```diff
@@ -0,0 +1,156 @@
...
+            if (currentPage + showXPages / 2 > lastPage) {
+                end = lastPage;
+                start = Math.max(lastPage - showXPages, 0);
+            } else {
+                start = Math.max(Math.floor(currentPage - showXPages / 2), 0);
+                end = Math.min(start + showXPages, lastPage);
+            }
+
+            _.each(_.range(start, end + 1), (nbr: number): void => {
+                pageSelects.push(
```
dmgauthier: I didn't know you could push item in a `const` ! I would have used a `let` declaration. nice ! :+1:

packages/react-vapor/src/components/table-hoc/TableWithNewPagination.tsx
```diff
@@ -0,0 +1,132 @@
...
+        render() {
+            const newProps = _.omit(this.props, [...TableWithPaginationProps]);
+            return (
+                <Component {...newProps}>
+                    <TablePagination
+                        id={this.props.id}
+                        disabled={this.props.isLoading}
+                        totalPages={this.props.totalPages}
+                        defaultPerPageSelected={
+                            config?.perPageNumbers?.length > 1
```
dmgauthier: `const config = HocUtils.supplyConfig(supplier);`  config can be undefined ?

packages/react-vapor/src/components/table-hoc/TableWithNewPagination.tsx
```diff
@@ -0,0 +1,132 @@
...
+                                ? config.perPageNumbers[1]
+                                : config.perPageNumbers[0] ?? PER_PAGE_NUMBERS[1]
```
dmgauthier: If the `config` of the condition has an optionnal chaining operator, you should put one there too.

dmgauthier:
Can't wait to see the example !! ;)

packages/react-vapor/src/components/pagination/PaginationPagesNumber.tsx
pochretien: showXpages is used below I think. I only copy paste this section from the old pagination without trying to rework because I have already alot of changes :)

packages/react-vapor/src/components/table-hoc/TableWithNewPagination.tsx
pochretien: not really but its a reflex to add a safety just to be sure

packages/react-vapor/src/components/pagination/PaginationPagesNumber.tsx
pochretien: its because when you push in a array, you keep the same reference so for tslint, you dont reassign the value.

PR #1380, opened 2020-01-08
fixed the dateTooltip test where compared date were not the same

FelixBlaisThon:
### Proposed Changes

the date of reference was few minutes late from the date in the test since they were not made in the same scope. Fixed it

### Acceptance Criteria

-   [x ] The proposed changes are covered by unit tests

### DIFF
--- a/packages/react-vapor/src/components/tooltip/tests/DateTooltip.spec.tsx
+++ b/packages/react-vapor/src/components/tooltip/tests/DateTooltip.spec.tsx
@@ -7,7 +7,8 @@ import {Tooltip} from '../Tooltip';
 
 describe('Tooltip', () => {
     let tooltipWrapper: ShallowWrapper<DateTooltipsProps>;
-    let defaultFormat: string;
+    const defaultDate = '1995-12-25';
+    const defaultFormat = moment(defaultDate).format('LLL');
     const stringFormat = 'ddd, hA';
 
     describe('<DateTooltip />', () => {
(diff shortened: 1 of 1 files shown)


packages/react-vapor/src/components/tooltip/tests/DateTooltip.spec.tsx
```diff
@@ -17,16 +17,19 @@ describe('Tooltip', () => {
         });
 
         it('should display content in LLL if no format is given on moment date', () => {
+            defaultFormat = moment().format('LLL');
             tooltipWrapper = shallow(<DateTooltip date={moment()} />);
```
gdostie: I think it would be best to use a constant predefined moment instead of `moment()` which is equivalent to "now" and therefore always changes.

gdostie: Same logic applies for every other test

packages/react-vapor/src/components/tooltip/tests/DateTooltip.spec.tsx
```diff
@@ -7,7 +7,7 @@ import {Tooltip} from '../Tooltip';
 
 describe('Tooltip', () => {
     let tooltipWrapper: ShallowWrapper<DateTooltipsProps>;
-    const defaultFormat = moment().format('LLL');
+    let defaultFormat: string;
     const stringFormat = 'ddd, hA';
 
     describe('<DateTooltip />', () => {
```
pochretien: you can use a beforeEach and create your defaultFormat for each test

packages/react-vapor/src/components/tooltip/tests/DateTooltip.spec.tsx
FelixBlaisThon: I changed it for a constant so i can re-use it everywhere, poke me if that's not ok (@pochretien )

PR #1381, opened 2020-01-09
feat(singleSelect): added a footer props to the singleSelectConnected

dmgauthier:
for ADUI-4973

### Proposed Changes

Added a footer to the SingleSelectConnected component

~added a props that shouldn't break anything~ Broke a UT and fixed it.

### Acceptance Criteria


packages/react-vapor/src/components/select/tests/SingleSelectConnected.spec.tsx
```diff
@@ -254,5 +255,25 @@ describe('Select', () => {
...
+                expect(
+                    shallowedSingleSelect
+                        .dive()
+                        .dive()
+                        .dive()
+                        .dive()
+                        .childAt(0)
+                        .children()
+                        .prop('footer')
+                ).toEqual(footer);
```
dmgauthier: Is there a better way to do this? somehow could get anything with .find(selector) since it's shallow mounted.

packages/react-vapor/src/components/listBox/ListBox.tsx
```diff
@@ -107,9 +108,12 @@ export class ListBox extends React.Component<IListBoxProps, {}> {
...
     render() {
         return (
-            <ul className={this.getClasses()} id={this.props.id}>
-                {this.props.wrapItems(this.getItems())}
-            </ul>
+            <>
+                <ul className={this.getClasses()} id={this.props.id}>
+                    {this.props.wrapItems(this.getItems())}
+                </ul>
+                {this.props.footer}
```
gdostie: This could simply be `this.props.children`, that way you wouldn't need to add a new prop.
Also, it adds more of a "react" feel to it.
```tsx
<ListBox 
    {...props} 
    footer={<span>I am the footer</span>}
/>

// The markup is a little bit more readable (IMO) this way:
<ListBox {...props}>
    <span>I am the footer</span>
</ListBox>
```

PR #1400, opened 2020-01-17
fix(SecondaryActions): separator filtering

gdostie:
### Proposed Changes

Moved the ACTION_SEPARATORS filtering logic one level into the `SecondaryActions` component (instead of `ActionsDropdown` component) so that we know earlier how many actions to display there is.

After filtering out disabled actions and unneeded separators:
- if there is no actions to display: `<SecondaryActions />` renders nothing
image
- if there is only one action: `<SecondaryActions />` renders it as a primary action directly into the action bar
image
- if there are more than one action: `<SecondaryActions />` renders an `<ActionDropdown />` component as it used to. 
image

### Acceptance Criteria


dmgauthier:
Very nice update to the actionBar component GG

PR #1416, opened 2020-01-27
improvement(inputconnected.tsx): added toggle dispatch on inputconnected

FelixBlaisThon:
### Proposed Changes

Added a isdirty toggle dispatch onChange for inputConnected so that the input gets dirty if not empty and displays error message if empty

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/input/tests/InputConnected.spec.tsx
+++ b/packages/react-vapor/src/components/input/tests/InputConnected.spec.tsx
@@ -175,7 +175,7 @@ describe('<InputConnected />', () => {
             expect(validate(newInputState.value)).toBe(false);
             expect(newInputState.valid).toBe(validate(newInputState.value));
         });
-        it('should call changeToDirty is set as props', () => {
+        it('should call changeToDirty if set as props', () => {
             const changeDirtyStateSpy = jasmine.createSpy();
             const wrapperInputConnected = shallowWithState(<InputConnected />, {});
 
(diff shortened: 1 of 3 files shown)


dmgauthier:
> I'm not sure about the change here, can you explain why you want the input to not be dirty when the value is invalid?
> 
> I think it needs to be dirty as soon as the value changes, even if it's an invalid value

I maybe missed a conversation about it, but I tend to agree with that :point_up:

packages/react-vapor/src/components/input/tests/InputConnected.spec.tsx
```diff
@@ -172,5 +175,18 @@ describe('<InputConnected />', () => {
             expect(validate(newInputState.value)).toBe(false);
             expect(newInputState.valid).toBe(validate(newInputState.value));
         });
+        it('should call changeToDirty is set as props', () => {
+            const changeToDirtySpy = jasmine.createSpy();
+            const wrapperInputConnected = shallowWithState(<InputConnected />, {});
```
pochretien: I think you can use the .dive() on shallowWithState
```
wrapperInputConnected.find(input).simulate('onchange');
```
and validate if the props is called with the full loop with the input element

packages/react-vapor/src/components/input/tests/InputConnected.spec.tsx
```diff
@@ -20,8 +21,10 @@ describe('<InputConnected />', () => {
     });
 
     afterEach(() => {
-        store.dispatch(clearState());
-        wrapper.detach();
+        if (wrapper) {
+            store.dispatch(clearState());
+            wrapper.detach();
+        }
```
gdostie: We probably want to clear the state even if we don't mount during the test?

```ts
store.dispatch(clearState());
wrapper?.detach();
```

gdostie:
There is a failing unit test that would need to be fixed, other than that it looks good.

PR #1420, opened 2020-01-28
feat(validation): backbone for generic error, warning and dirty components

francoislg:
### Proposed Changes

The idea here is to have actions and selectors to easily manage errors, warnings and dirty state for components.

Then, we can add HOCs to wrap existing components with features related to this state. 

Some ground rules for those HOCs, which follow React's doc about it:

* Ensure that you pass all the original props to the child component
* Ensure that you pass down props that might be used in multiple HOCs
    * For instance, wrapping a component that does not have an initial value with a new `initialValue` could pretty well be used across multiple validation HOC. Hence it should be propagated
* Do not always reset the state on unmount. The component might be dirty, but unmounted temporarily, for some reason.

Then, there would be Components that read those states and output something. I have one `ValidationMessage` component that simply outputs errors, and if there are no errors, warnings.

As an alternative to HOCs, some components could implement this directly in their code.

Next steps would be to create some more HOCs. For instance, one for a "Save Button" that is disabled when there are errors, or if nothing is dirty.

Let me know what you think of it! :)

None

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/validation/ValidationActions.ts
+++ b/packages/react-vapor/src/components/validation/ValidationActions.ts
@@ -1,0 +1,40 @@
+import {ValidationTypes} from './ValidationTypes';
+
+export const ValidationActionsTypes = {
+    updateError: 'UPDATE_VALIDATION_ERROR',
+    updateWarning: 'UPDATE_VALIDATION_WARNING',
+    updateDirty: 'UPDATE_VALIDATION_DIRTY',
+};
+
+export const ValidationActions = {
+    setError: (id: string, error: string, validationType: string = ValidationTypes.default) => ({
+        type: ValidationActionsTypes.updateError,
+        payload: {
+            id,
+            validationType,
+            value: error,
+        },
+    }),
+    clearError: (id: string, validationType: string = ValidationTypes.default) =>
+        ValidationActions.setError(id, null, validationType),
+    setWarning: (id: string, warning: string, validationType: string = ValidationTypes.default) => ({
+        type: ValidationActionsTypes.updateWarning,
+        payload: {
+            id,
+            validationType,
+            value: warning,
+        },
+    }),
+    clearWarning: (id: string, validationType: string = ValidationTypes.default) =>
+        ValidationActions.setWarning(id, null, validationType),
+    setDirty: (id: string, isDirty: boolean, validationType: string = ValidationTypes.default) => ({
+        type: ValidationActionsTypes.updateDirty,
+        payload: {
+            id,
+            validationType,
+            value: isDirty,
+        },
+    }),
+    clearDirty: (id: string, validationType: string = ValidationTypes.default) =>
+        ValidationActions.setDirty(id, null, validationType),
+};
--- a/packages/react-vapor/src/components/validation/ValidationSelectors.ts
+++ b/packages/react-vapor/src/components/validation/ValidationSelectors.ts
@@ -20,6 +20,10 @@ const getAnyDirty = (ids: string[]) => (state: IReactVaporState) =>
         .reduce((all, id) => all.concat(getIsDirty(id)(state)), [] as ValidationState['isDirty'])
         .filter((dirty) => dirty.value);
 
+const isInError = (ids: string[]) => (state: IReactVaporState) => getAnyError(ids)(state).length > 0;
+const isInWarning = (ids: string[]) => (state: IReactVaporState) => getAnyWarning(ids)(state).length > 0;
+const isDirty = (ids: string[]) => (state: IReactVaporState) => getAnyDirty(ids)(state).length > 0;
+
 export const ValidationSelectors = {
     getErrors,
     getWarnings,
@@ -27,4 +31,7 @@ export const ValidationSelectors = {
     getAnyError,
     getAnyWarning,
     getAnyDirty,
+    isInError,
+    isInWarning,
+    isDirty,
 };
--- a/packages/react-vapor/src/components/validation/ValidationSelectors.ts
+++ b/packages/react-vapor/src/components/validation/ValidationSelectors.ts
@@ -1,0 +1,30 @@
+import {IReactVaporState} from '../../ReactVapor';
+import {ValidationState} from './ValidationState';
+
+const getErrors = (id: string) => (state: IReactVaporState) =>
+    state.validation?.[id]?.error?.filter((error) => !!error.value) || [];
+const getWarnings = (id: string) => (state: IReactVaporState) =>
+    state.validation?.[id]?.warning?.filter((warning) => !!warning.value) || [];
+const getIsDirty = (id: string) => (state: IReactVaporState) => state.validation?.[id]?.isDirty || [];
+
+const getAnyError = (ids: string[]) => (state: IReactVaporState) =>
+    ids
+        .reduce((all, id) => all.concat(getErrors(id)(state)), [] as ValidationState['error'])
+        .filter((error) => !!error.value);
+const getAnyWarning = (ids: string[]) => (state: IReactVaporState) =>
+    ids
+        .reduce((all, id) => all.concat(getWarnings(id)(state)), [] as ValidationState['warning'])
+        .filter((warning) => !!warning.value);
+const getAnyDirty = (ids: string[]) => (state: IReactVaporState) =>
+    ids
+        .reduce((all, id) => all.concat(getIsDirty(id)(state)), [] as ValidationState['isDirty'])
+        .filter((dirty) => dirty.value);
+
+export const ValidationSelectors = {
+    getErrors,
+    getWarnings,
+    getIsDirty,
+    getAnyError,
+    getAnyWarning,
+    getAnyDirty,
+};
--- a/packages/react-vapor/src/components/validation/ValidationState.ts
+++ b/packages/react-vapor/src/components/validation/ValidationState.ts
@@ -1,7 +1,12 @@
+export type ISingleValidation<T> = {
+    validationType: string;
+    value: T;
+};
+
 export type ValidationState = {
-    isDirty: Array<{validationType: string; value: boolean}>;
-    error: Array<{validationType: string; value: string}>;
-    warning: Array<{validationType: string; value: string}>;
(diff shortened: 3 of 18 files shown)


francoislg:
I am currently coding tests for selectors and reducers

packages/react-vapor/src/components/validation/ValidationIds.ts
```diff
@@ -0,0 +1,4 @@
+export const ValidationIds = {
```
GermainBergeron: I think I would prefer ValidationTypes (and update the actions to use `type` instead of `subId`). It would be clearer

packages/react-vapor/src/components/validation/ValidationState.ts
```diff
@@ -0,0 +1,7 @@
+export type ValidationState = {
+    isDirty: Array<{subId: string; value: boolean}>;
+    error: Array<{subId: string; value: string}>;
+    warning: Array<{subId: string; value: string}>;
+};
+
+export type ValidationsState = {[id: string]: ValidationState};
```
GermainBergeron:
```suggestion
export type ValidationsState = Record<string, ValidationState>
```

packages/react-vapor/src/components/validation/components/ValidationMessage.tsx
```diff
@@ -0,0 +1,41 @@
...
+                          {value}
+                      </span>
+                  ))
+                : warnings
+                ? warnings.map(({subId, value}) => (
+                      <span key={subId} className={ValidationMessageClasses.warning}>
+                          {value}
+                      </span>
+                  ))
+                : null}
```
GermainBergeron: You really should avoid the nested ternaries it's hard to read and can be simplified with the classNames library

packages/react-vapor/src/components/validation/hoc/NonEmptyInputValidationHOC.tsx
```diff
@@ -0,0 +1,54 @@
...
+    validationMessage?: string;
+    resetErrorOnUnmount?: boolean;
+}
+
+const mapDispatchToProps = (dispatch: IDispatch) => ({
+    setError: (id: string, error: string) => dispatch(ValidationActions.setError(id, error, ValidationIds.nonEmpty)),
+    clearError: (id: string) => dispatch(ValidationActions.clearError(id)),
+});
+
+export const nonEmptyInputValidationHOC = <T extends IInputOwnProps>(
```
GermainBergeron: I would suggest changing the name for `withValueInput` or something in the same vein, the `with` makes it clear that we compose behavior

GermainBergeron:
It's really nice and seems simple to use, congrats!

packages/react-vapor/src/components/validation/ValidationIds.ts
francoislg: I'm not sure about that, you would have:

```
{ type: 'UPDATE_WARNING', payload: { id: 'x', value: 'x', type: 'something' }}
```

instead of

```
{ type: 'UPDATE_WARNING', payload: { id: 'x', value: 'x', subId: 'something' }}
```

Seems a little less clear what "type" is about to me since we already have `type` in the action 🤔

What about `componentType:` or `validationType:` ?

GermainBergeron: You would trigger ```typescript
ValidationActions.setWarning('x', 'some empty warning', ValidationTypes.nonEmpty)
```
which would translate to ```typescript
{ type: 'UPDATE_WARNING', payload: { id: 'x', value: 'some empty warning', type: 'non-empty' }}
```
It seems a little bit more easy to understand for the developper IMO

I do feel fine with either of your suggestions,  I just think subId is a bit confusing

packages/react-vapor/src/components/validation/components/ValidationMessage.tsx
francoislg: I don't know, this looks weird to me:

```
const hasErrors = errors.length > 0;
const eitherErrorsOrWarnings = hasErrors ? errors : warnings;
return eitherErrorsOrWarnings.map((...) => <div className={classNames({
   "text-red": hasErrors
   "text-yellow": !hasErrors
])></div>);
```

It feels wrong to assume `errors` and `warnings` have the exact same signature (even though they do)

Any other idea to make it look good?

packages/react-vapor/src/components/validation/ValidationActions.ts
```diff
@@ -0,0 +1,40 @@
...
+    setDirty: (id: string, isDirty: boolean, validationType: string = ValidationTypes.default) => ({
+        type: ValidationActionsTypes.updateDirty,
+        payload: {
+            id,
+            validationType,
+            value: isDirty,
+        },
+    }),
+    clearDirty: (id: string, validationType: string = ValidationTypes.default) =>
+        ValidationActions.setDirty(id, null, validationType),
```
dmgauthier: I like the idea of a clearing action :+1:

packages/react-vapor/src/components/validation/ValidationSelectors.ts
```diff
@@ -0,0 +1,30 @@
+import {IReactVaporState} from '../../ReactVapor';
+import {ValidationState} from './ValidationState';
+
+const getErrors = (id: string) => (state: IReactVaporState) =>
+    state?.validation?.[id]?.error?.filter((error) => !!error.value) || [];
```
dmgauthier: How can the state be not defined here ?

packages/react-vapor/src/components/validation/ValidationSelectors.ts
```diff
@@ -0,0 +1,30 @@
...
+        .reduce((all, id) => all.concat(getErrors(id)(state)), [] as ValidationState['error'])
+        .filter((error) => !!error.value);
+const getAnyWarning = (ids: string[]) => (state: IReactVaporState) =>
+    ids
+        .reduce((all, id) => all.concat(getWarnings(id)(state)), [] as ValidationState['warning'])
+        .filter((warning) => !!warning.value);
+const getAnyDirty = (ids: string[]) => (state: IReactVaporState) =>
+    ids
+        .reduce((all, id) => all.concat(getIsDirty(id)(state)), [] as ValidationState['isDirty'])
+        .filter((dirty) => dirty.value);
```
dmgauthier: very nice code

packages/react-vapor/src/components/validation/components/ValidationMessage.tsx
dmgauthier: I personally prefer the quote you did with `ClassNames`

packages/react-vapor/src/components/validation/hoc/WithDirtyInputHOC.tsx
```diff
@@ -0,0 +1,49 @@
...
+}
+
+const mapDispatchToProps = (dispatch: IDispatch) => ({
+    setIsDirty: (id: string, isDirty: boolean) =>
+        dispatch(ValidationActions.setDirty(id, isDirty, ValidationTypes.wrongInitialValue)),
+    clearIsDirty: (id: string) => dispatch(ValidationActions.clearDirty(id, ValidationTypes.wrongInitialValue)),
+});
+
+export const withDirtyInputHOC = <T extends IInputOwnProps, R = any>(
+    Component: React.ComponentClass<T, R> | React.FunctionComponent<T>
```
dmgauthier: https://flow.org/en/docs/react/types/#toc-react-componenttype

packages/react-vapor/src/components/validation/ValidationSelectors.ts
francoislg: 🤷‍♂

francoislg:
Any other feedback for this? Anyone else wants to have a look?

packages/react-vapor/src/components/validation/hoc/WithNonEmptyValueInputValidationHOC.tsx
```diff
@@ -0,0 +1,54 @@
...
+    const WrappedInput: React.FunctionComponent<NewOwnProps & DispatchProps> = ({
+        setError,
+        clearError,
+        validationMessage = 'Input is empty and should not be empty',
+        resetErrorOnUnmount,
+        validate,
+        ...props
+    }) => {
+        React.useEffect(() => {
+            setError(props.id, '');
```
pochretien: why we set the error with empty string ?

packages/react-vapor/src/components/validation/hoc/WithNonEmptyValueInputValidationHOC.tsx
```diff
@@ -0,0 +1,54 @@
...
+                validate={(value: string) => {
+                    const isEmpty = _.isEmpty(value);
+                    setError(props.id, isEmpty ? validationMessage : '');
+                    return !isEmpty || (validate ? validate(value) : true);
+                }}
+            />
+        );
+    };
+
+    const enhance = connect(null, mapDispatchToProps) as InferableComponentEnhancer<DispatchProps>;
```
pochretien: why casting the connect ? if you type the mapDispatchToProps, you can use the connect in 1 line without cast ?

packages/react-vapor/src/components/validation/hoc/WithNonEmptyValueInputValidationHOC.tsx
francoislg: it should be `clearError`

packages/react-vapor/src/components/validation/hoc/WithNonEmptyValueInputValidationHOC.tsx
francoislg: There is an issue with `WeakValidationMap` when you enhance the props.

There are more details in `connectHOC.d.ts`, which points to: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/31363

francoislg:
I replaced the Input with dirty management with the new HOC, way simpler code now 😄 

@GermainBergeron or @gdostie, could you have a quick look at the last commit and merge this?

PR #1428, opened 2020-02-04
feat(table-hoc): blankslate table with filter

pochretien:
Add a blankSlate if the filter is defined but the data is empty
Add uts

### Acceptance Criteria


packages/react-vapor/src/components/table-hoc/TableWithFilter.tsx
```diff
@@ -83,6 +83,10 @@ export const tableWithFilter = (
             const newActions = [...(this.props.actions || []), filterAction];
             const newProps = {
                 ..._.omit(this.props, [...TableWithFilterPropsToOmit]),
+                renderBody:
+                    _.isEmpty(this.props.data) && this.props.filter !== ''
```
rosalie-liu: _.isEmpty(this.props.data) && this.props.filter
Does this work? :thinking:

pochretien: Probably but I prefer to valid not empty to be sure I have the right behaviour. I know the selector always return a string if not defined

PR #1435, opened 2020-02-10
feat(validation): add withdirtysavebutton hoc

francoislg:
### Proposed Changes

The idea of this HOC is to use the generic validation to wrap a button that has all the required behavior according to the components it must validate.

The Button is bound to one or multiple "validationIds". Then, for all the ID states:

If there are errors: Button disabled, shows the error tooltip
If there are warnings: Button enabled, shows the warning tooltip
If there is **no** dirty component: Button disabled, shows the dirty tooltip.

So the button is only enabled when a component is dirty and there are no errors.

`skipDirty` allows you to skip the dirty check, so only errors could disable the button.

None

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/button/examples/ButtonExamples.tsx
+++ b/packages/react-vapor/src/components/button/examples/ButtonExamples.tsx
@@ -1,8 +1,19 @@
 import * as VaporSVG from 'coveo-styleguide';
 import * as React from 'react';
+import {connect} from 'react-redux';
+import {IDispatch} from '../../../utils';
+import {CheckboxConnected} from '../../checkbox';
+import {Label} from '../../input/Label';
+import {LabeledInput} from '../../input/LabeledInput';
 import {Section} from '../../section/Section';
 import {Svg} from '../../svg/Svg';
+import {withDirtySaveButtonHOC} from '../../validation/hoc/WithDirtySaveButtonHOC';
+import {ValidationActions} from '../../validation/ValidationActions';
 import {Button} from '../Button';
+
+// start-print
+
+const SaveButton = withDirtySaveButtonHOC(Button);
 
 export class ButtonExamples extends React.Component<any, any> {
     static description = 'Buttons communicate actions, and, when clicked, initialize those actions.';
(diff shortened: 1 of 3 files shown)


francoislg:
[Added an example usage in the `Button` page](https://coveo.github.io/react-vapor/86da1e13582d0e89e57720b74bd3ed74/index.html#/components/Button)!

Usually, there should be other components (such as Inputs) that would expose their "validation state" to the world and the Save Button would use those components' ids.

But for the sake of the example, I felt the three checkboxes allow more testing flexibility.

packages/react-vapor/src/components/button/examples/ButtonExamples.tsx
```diff
@@ -173,7 +185,54 @@ export class ButtonExamples extends React.Component<any, any> {
...
+
+                <div className="form-group mb5">
+                    <Section level={2} title="Save button with validation">
+                        <DirtyCheckboxesForSaveButton />
+                        <SaveButton
+                            enabled
+                            name="Save Button"
+                            validationIds={['inputId']}
+                            onClick={() => alert('Saving!')}
+                        ></SaveButton>
```
gdostie:
```suggestion
                        />
```

packages/react-vapor/src/components/validation/hoc/WithDirtySaveButtonHOC.tsx
```diff
@@ -0,0 +1,57 @@
...
+    }: T & IWithDirtySaveButtonHOCProps & StateProps) => {
+        const hasErrors = errors.length > 0;
+        const hasWarnings = warnings.length > 0;
+        const isDirty = dirty.length > 0;
+        const canSaveWhenDirty = skipDirty ? true : isDirty;
+        const generatedTooltip = hasErrors
+            ? errorMessage(errors.map((error) => error.value))
+            : hasWarnings && canSaveWhenDirty
+            ? warningMessage(warnings.map((warning) => warning.value))
+            : dirtyMessage(dirty.map((d) => d.value));
```
gdostie: This dark sorcery is forbidden outside Hogwarts walls. It could be split in order to avoid using nested ternaries.

francoislg: Is this allowed, professor Dumbledostie?

```
const generatedTooltip =
            (hasErrors && errorMessage(errors.map((error) => error.value))) ||
            (hasWarnings && canSaveWhenDirty && warningMessage(warnings.map((warning) => warning.value))) ||
            dirtyMessage(dirty.map((d) => d.value));
```

gdostie: yes

francoislg:
Feel free to merge this when you want :)

PR #1446, opened 2020-02-17
Create a separate package for the demo

gdostie:
### Proposed Changes

- Created the new demo package with a minimalist demo. I will move the existing demo files in the new package in a second step.
- Deleted per-package gitignore files, to have one single ignore file at the root
- Split the lint-staged commit hook so that every package can specify its unique way of linting the files

None.

### Acceptance Criteria


packages/demo/package.json
```diff
@@ -0,0 +1,67 @@
...
+    "version": "5.7.5",
+    "description": "Vapor Design System",
+    "main": "src/Index.ts",
+    "scripts": {
+        "start": "webpack-dev-server",
+        "build": "webpack",
+        "precommit": "lint-staged"
+    },
+    "author": "Coveo",
+    "license": "Apache-2.0",
```
GermainBergeron: MIT?

packages/demo/package.json
```diff
@@ -0,0 +1,67 @@
+{
+    "name": "react-vapor-demo",
+    "version": "5.7.5",
```
GermainBergeron: I'd add `"private": true,` to avoid publishing the demo in npm

packages/demo/package.json
```diff
@@ -0,0 +1,67 @@
...
+    },
+    "author": "Coveo",
+    "license": "Apache-2.0",
+    "homepage": "https://react-vapor.prod.cloud.coveo.com/",
+    "repository": {
+        "type": "git",
+        "url": "git+https://github.com/coveo/react-vapor.git"
+    },
+    "dependencies": {
+        "codemirror": "^5.51.0",
```
GermainBergeron: You have a funny combinaison of `^version` and `version` in both dependencies and devDependencies

packages/demo/webpack.config.js
```diff
@@ -0,0 +1,78 @@
...
+    stats: 'minimal',
+    module: {
+        rules: [
+            {
+                test: /\.tsx?$/,
+                use: [
+                    {
+                        loader: 'ts-loader',
+                        options: {
+                            projectReferences: true,
```
GermainBergeron: You should use `transpileOnly` with the fork-ts-checker since we don't need the d.ts to be generated

You should also add `experimentalWatchApi: true,`

packages/react-vapor/tsconfig.build.json
```diff
@@ -1,13 +1,14 @@
 {
-    "extends": "./tsconfig",
+    "extends": "./tsconfig.json",
```
GermainBergeron: Do you really need the extensions?

https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#configuration-inheritance-with-extends

packages/react-vapor/tsconfig.build.json
```diff
@@ -1,13 +1,14 @@
...
-        "declarationDir": "dist/definitions"
+        "declarationDir": "dist/definitions",
+        "outDir": "dist",
+        "composite": true
     },
-    "include": ["types/**/*.d.ts"],
-    "files": ["src/Entry.ts"]
+    "typeRoots": ["types", "node_modules/@types", "../../node_modules/@types"],
+    "include": ["src", "docs/*.ts", "docs/src/components/ComponentsInterface.ts", "types/react-dnd/index.d.ts"],
+    "enclude": ["**/*.spec.ts?"]
```
GermainBergeron:
```suggestion
    "exclude": ["**/*.spec.ts?"]

```

packages/demo/package.json
gdostie: I thought we were supposed to use the Apache-2.0 license for our open source repos now :thinking:

packages/react-vapor/tsconfig.build.json
gdostie: I don't think so, it is what was used in the example I was following https://github.com/RyanCavanaugh/learn-a

packages/demo/webpack.config.js
gdostie: experimentalWatchApi is there by default now https://github.com/TypeStrong/ts-loader/issues/757

PR #1448, opened 2020-02-19
Rename setto icon

oliviertaillon:
### Proposed Changes

I want to rename the _setto_ icon to _pricetag_, to better represent what it actually is.

I don't think there is.

### Acceptance Criteria

-   [X] [README.md](https://github.com/coveo/react-vapor/blob/master/README.md) is adjusted to reflect the proposed changes (if relevant)


francoislg:
Technically, this *is* a breaking change 😄 Any project using this will have to change the name, otherwise their code will not compile.

I am pretty sure we can handle that, though 😄

PR #1451, opened 2020-02-20
feat(demo): move all example files into the demo package

gdostie:
### Proposed Changes

- :tada: Moved the demo in a separate package called `react-vapor-demo`, the folder name is `demo`.  
- :unicorn: Successfully removed the utilization of `ttypescript` , while keeping the interfaces extraction into arrays of strings. This is still done in a precompile step, we are simply not using ttypescript as a compiler anymore. We use the standard typescript compiler, but specify a custom program in the `ts-loader` options.
- :call_me_hand: added `--hoist` option to `lerna bootstrap` and synced the deps versions. I had to this so that react-vapor and the demo would share the exact same instance of `react-dom` lib, the demo would break at run-time without this.
- :books: Updated the READMEs

### Acceptance Criteria


package.json
```diff
@@ -35,10 +35,10 @@
         "not dead"
     ],
     "scripts": {
-        "postinstall": "lerna bootstrap --no-ci",
-        "start": "lerna run start --parallel --scope=\"react-vapor\" --scope=\"coveo-styleguide\"",
+        "postinstall": "lerna bootstrap --no-ci --hoist",
```
GermainBergeron: I think we should use a `setup` task the same way we do it in the admin, otherwise we might have the same problem we had with Windows

```suggestion
        "setup": "lerna bootstrap --no-ci --hoist",
```

GermainBergeron:
Good job 🦄 🎉

package.json
gdostie: Since we already had more than one package and it was working that way for windows users, I will leave it be like this for now, but I'll keep it in mind.

PR #1470, opened 2020-02-28
fix(codemirror): add missing import of codemirror style files

GermainBergeron:


packages/vapor/scss/dependencies.scss
```diff
@@ -4,3 +4,8 @@
 @import 'lib/datepicker';
 @import '~coveo-slider/scss/slider';
 @import '~rc-slider/dist/rc-slider';
+@import '~codemirror/lib/codemirror';
+@import '~codemirror/addon/fold/foldgutter';
+@import '~codemirror/addon/lint/lint';
+@import '~codemirror/addon/dialog/dialog';
+@import '~codemirror/addon/hint/show-hint';
```
gdostie: We should add codemirror in the dependencies of vapor

PR #1478, opened 2020-03-02
fix(partialstringmatch): allow rendering redux Provider inside it

gdostie:
### Proposed Changes

The PartialStringMatch is throwing an error when rendering a redux Provider inside it.

None.

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/tsconfig.json
+++ b/packages/react-vapor/tsconfig.json
@@ -11,5 +11,5 @@
         "noUnusedLocals": true,
         "lib": ["es2015", "dom"]
     },
-    "exclude": ["node_modules", "dist", "coverage", "*.js"]
+    "exclude": ["**/node_modules", "**/dist", "**/coverage", "*.js", "**/target"]
 }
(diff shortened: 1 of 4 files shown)


packages/demo/webpack.config.js
```diff
@@ -39,6 +39,11 @@ module.exports = {
     stats: 'minimal',
     module: {
         rules: [
+            {
+                test: /\.js$/,
+                use: ['source-map-loader'],
+                enforce: 'pre',
```
gdostie: I will better scope this as it is too slow

packages/react-vapor/tsconfig.json
```diff
@@ -11,5 +11,5 @@
         "noUnusedLocals": true,
         "lib": ["es2015", "dom"]
     },
-    "exclude": ["node_modules", "dist", "coverage", "*.js"]
+    "exclude": ["**/node_modules", "**/dist", "**/coverage", "*.js", "**/target"]
```
dmgauthier: is that in the PR scope ?
just wondering

gdostie: Probably not, if we were being 100% strict on the scope of PRs.

PR #1479, opened 2020-03-02
Clean utility section

FelixBlaisThon:
### Proposed Changes

clean the utility section and relocate some of the items in it to other locations like shadow, tranparency, borders. I also added isLink prop to SideNavigationMenuSection component so that a section with no item can also be a MenuSection to keep the same style (behavior expected for Home section in the admin)

### Acceptance Criteria


packages/demo/src/demo-building-blocs/NavigationSection.tsx
```diff
@@ -11,23 +12,14 @@ export interface NavSectionProps extends NavSectionControlledProps {
...
-            onClick={handleClick}
-        >
+    return rest.isLink ? (
+        <Route path={baseUrl}>
+            <Link to={baseUrl}>
+                <SideNavigationMenuSection {...rest}></SideNavigationMenuSection>
+            </Link>
+        </Route>
+    ) : (
+        <SideNavigationMenuSection {...rest} expandable={false}>
```
GermainBergeron: Why is expandable false?

GermainBergeron:
We're missing a selected state for the section links

image

GermainBergeron:
We also have a pointer (like we can click) on section headers but the click does nothing

packages/demo/src/demo-building-blocs/NavigationSection.tsx
FelixBlaisThon: otaillon asked me to make them not expandable, so that they are expanded all the time

packages/react-vapor/src/components/sideNavigation/SideNavigationMenuSection.tsx
```diff
@@ -12,6 +12,7 @@ export interface SideNavigationHeaderProps {
     svgClass?: string;
     customIcon?: React.ReactNode;
     onClick?: (event: React.MouseEvent) => void;
+    isLink?: boolean;
```
GermainBergeron: Remove the prop, it's not used

packages/demo/src/demo-building-blocs/NavigationSection.tsx
GermainBergeron: I'm good with having them opened at first, but I think we should keep the functionality to close them

oliviertaillon: I did ask him to make them non-expandable. 
The rationale is we'd like to have a content page for the section (eventually). We won't be able to have that if clicking on the section title just expand/collapse the menu.

This change should only affect the demo page, **not the admin console**

gdostie: We could still have the same behavior as the collapsible rows perhaps? Aka expand/collapse the sections only when clicking on the arrow. But still navigate to somewhere when clicking on the section header? Maybe it could be done in a later step thought.

oliviertaillon: Why do you need the collapse functionality?

gdostie: for reasons

GermainBergeron: I really like the solution proposed by Gael. I feel like we should stay as close as possible to what we have in the admin.

In the admin the expand/collapse state of the sections are kept in the users preferences. Meaning that someone never using a section can collapse it and never see its content if he doesn't want it

oliviertaillon: > I feel like we should stay as close as possible to what we have in the admin.
Curious as to why you feel this way @GermainBergeron ?

PR #1482, opened 2020-03-02
feat(multiselectconnected.tsx): +toggleClasses prop

FelixBlaisThon:
added the toggleClasses prop as in singleSelect so that we can add classes on multiselect toggle
drop

Sorry, the branch name does not represent the feature, I tried it with spread syntax, didn't work really well, made the fix but forgot to change the branch name

### Acceptance Criteria


packages/react-vapor/src/components/select/MultiSelectConnected.tsx
```diff
@@ -21,6 +21,7 @@ import {ISelectButtonProps, ISelectOwnProps, ISelectProps, SelectConnected} from
 import {SelectSelector} from './SelectSelector';
 
 export interface IMultiSelectOwnProps extends ISelectProps, IDropTargetProps {
+    toggleClasses?: string;
```
GermainBergeron: Can we move this to `ISelectProps` instead? We're using it in both SingleSelect & MultiSelect

PR #1483, opened 2020-03-03
fix(drop pod): the Defaults.DROP_ROOT was not properly set

dmgauthier:
The default props in the drop pod are set before the Defaults variables are loaded, thus defaulting
to 'body' all the time.

### Proposed Changes

passing the Defaults.DROP_ROOT and the Defaults.DROP_PARENT_ROOT directly where they are used instead of relying on defaultProps

none

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/drop/DropPod.tsx
+++ b/packages/react-vapor/src/components/drop/DropPod.tsx
@@ -53,8 +53,6 @@ class RDropPod extends React.PureComponent<IRDropPodProps, IDropPodState> {
         minWidth: 0,
         minHeight: 0,
         hasSameWidth: false,
-        selector: Defaults.DROP_ROOT,
-        parentSelector: Defaults.DROP_PARENT_ROOT,
     };
     private parentMutationObserver: MutationObserver;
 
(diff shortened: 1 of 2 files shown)


packages/react-vapor/src/components/drop/DropPod.tsx
```diff
@@ -135,16 +133,21 @@ class RDropPod extends React.PureComponent<IRDropPodProps, IDropPodState> {
         return this.dropRef.current && this.props.isOpen && !!this.props.positions.length;
     }
 
+    private getRelativeParent = () => {
+        const closestCurrentRef = this.props.buttonRef.current?.closest(
+            this.props.parentSelector ?? Defaults.DROP_PARENT_ROOT
+        );
+        return closestCurrentRef ? closestCurrentRef : this.props.buttonRef.current?.parentElement;
```
gdostie:
```suggestion
        return closestCurrentRef ?? this.props.buttonRef.current?.parentElement;

```

pochretien:
Add uts :)

packages/react-vapor/src/components/drop/DropPod.tsx
dmgauthier: thx !

PR #1484, opened 2020-03-03
fix(modalheader): add redirect on close prop

rosalie-liu:
### Proposed Changes

Add redirectOnClose prop to modal header, this function will be called when clicking on close icon in modal header. 

No

### Acceptance Criteria


packages/react-vapor/src/components/modal/ModalComposite.tsx
```diff
@@ -28,6 +28,7 @@ export interface IModalCompositeOwnProps
     modalFooterClasses?: IClassName;
     isPrompt?: boolean;
     validateShouldNavigate?: (isDirty: boolean) => boolean;
+    redirectOnClose?: () => void;
```
dmgauthier: I guess that you want to passe the redirection logic as a function in the props but the name of the props feels like it will receive a boolean value...

PR #1485, opened 2020-03-03
feat(tablewithfilter): allow specifying config through props

gdostie:
### Proposed Changes

Allow specifying the `tableWithFilter` related options through the props:

#### Usage
```tsx
// before
const Table = _.compose(
    tableWithFilter({
        matchFilter: (filterText, row) => true,
        blankslate: filterBlankslateProps
    }),
    tableWithActions()
)(TableHOC)

<Table
    id='table-id'
    className='table'
    data={myData}
    renderBody={renderMyRows}
/>

// now you also can do:
const Table = _.compose(
    tableWithFilter(),
    tableWithActions()
)(TableHOC)

<Table
    id='table-id'
    className='table'
    data={myData}
    renderBody={renderMyRows}
    matchFilter={(filterText, row) => true}
    blankslate={filterBlankslateProps}
/>
```

This new way of doing things is more convenient because usually you have the dispatch props in the scope when setting the props on the `Table`.

Deleted some useless interfaces:

- `ITableWithFilterStateProps`
- `ITableWithFilterDispatchProps`
- `ITableWithFilterProps`

### Acceptance Criteria

### DIFF
--- a/packages/demo/src/components/examples/TableHOCwithBlankSlateExamples.tsx
+++ b/packages/demo/src/components/examples/TableHOCwithBlankSlateExamples.tsx
@@ -25,19 +25,19 @@ export interface IExampleRowData {
 const TableWithBlankSlateExample: React.FunctionComponent = () => (
     <Section>
         <TableWithBlankSlateComposed
-            id={'tableWithBlankSlate'}
+            id="tableWithBlankSlate"
             className="table"
             data={generateDataWithFacker(0)}
-            renderBody={generateTableRow}
+            renderBody={(data) => generateTableRow(data, 'tableWithBlankSlate')}
+            filterMatcher={(filter: string, data: IExampleRowData) =>
+                data.username.toLowerCase().indexOf(filter.toLowerCase()) !== -1
+            }
         />
     </Section>
 );
 
 const TableWithBlankSlateComposed = _.compose(
     tableWithBlankSlate({title: 'No data caused the table to be empty'}),
-    tableWithFilter({
-        matchFilter: (filter: string, data: IExampleRowData) =>
-            data.username.toLowerCase().indexOf(filter.toLowerCase()) !== -1,
-    }),
+    tableWithFilter(),
     tableWithBlankSlate({title: 'Filter caused the table to be empty'})
-)(TableHOC);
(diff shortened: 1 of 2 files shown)


packages/demo/src/components/examples/TableHOCwithBlankSlateExamples.tsx
```diff
@@ -25,19 +25,19 @@ export interface IExampleRowData {
...
 const TableWithBlankSlateComposed = _.compose(
     tableWithBlankSlate({title: 'No data caused the table to be empty'}),
-    tableWithFilter({
-        matchFilter: (filter: string, data: IExampleRowData) =>
-            data.username.toLowerCase().indexOf(filter.toLowerCase()) !== -1,
-    }),
+    tableWithFilter(),
     tableWithBlankSlate({title: 'Filter caused the table to be empty'})
-)(TableHOC);
+)(TableHOC) as ReturnType<ReturnType<typeof tableWithFilter>>;
```
GermainBergeron: Is the cast necessary here?

gdostie: Its just to have autocompletion on the `TableWithBlankSlateComposed`, otherwise `_.compose` returns `any`.

rosalie-liu:
The props `matchFilter` and `blankslate` in `<Table>` will be only applied to the blankslate in `tableWithFilter` right? If we need a blankslate for the page no results found, we can still configure its proper `matchFilter` and `blankslate` in `tableWithBlankSlate`?

gdostie:
@rosalie-liu Yes exactly, I kept the naming because it allowed me to reuse the same interface, but perhaps we could rename the props `filterBlankslate`, `filterMatcher`, and `filterPlaceholder` ?

rosalie-liu:
Great idea! Btw, as a prop of `<Table>`, `filterPlaceholder` makes me think of placeholder of filterbox, but I'm still ok with it.

gdostie:
It is exactly that, `filterPlaceholder` is the placeholder of the filter box :tada:

rosalie-liu:
oops, I mixed it with title in blankslate, my bad. So these 3 are just perfect :+1:

gdostie:
Great, I'll do the change

PR #1488, opened 2020-03-05
feat(input): add multi values input to replace multiline input

GermainBergeron:
### Proposed Changes

Add new multi-values input to replace non-connected one

None

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/multiValueInputs/MultiValuesInput.tsx
+++ b/packages/react-vapor/src/components/multiValueInputs/MultiValuesInput.tsx
@@ -1,0 +1,65 @@
+import * as React from 'react';
+import * as _ from 'underscore';
+
+import {IButtonProps} from '../button/Button';
+import {IInputOwnProps} from '../input/Input';
+import {InputConnected} from '../input/InputConnected';
+import {multilineBoxWithRemoveButton} from '../multilineBox/hoc/MultilineBoxWithRemoveButton';
+import {IMultilineParentProps, IMultilineSingleBoxProps, MultilineBox} from '../multilineBox/MultilineBox';
+
+const MultilineBoxWithRemoveButton = _.compose(
+    multilineBoxWithRemoveButton({
+        containerNode: (
+            child: React.ReactNode,
+            getRemoveButton: (props?: Partial<IButtonProps>) => React.ReactNode,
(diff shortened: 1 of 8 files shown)


packages/react-vapor/src/components/multiValueInputs/MultiValuesInput.tsx
```diff
@@ -0,0 +1,60 @@
...
+                            const value = e.currentTarget.value;
+                            if (value !== '' && cData.isLast) {
+                                parentProps.addNewBox();
+                            }
+                        }}
+                        onBlur={(value: string) => {
+                            if (value === '' && !cData.isLast) {
+                                parentProps.removeBox(cData.id);
+                            }
+                        }}
```
rosalie-liu: Do we need `validate` and `validateOnChange`, also `labelTitle` and `labelProps` for this `InputConnected`?

packages/react-vapor/src/components/multiValueInputs/MultiValuesInput.tsx
```diff
@@ -0,0 +1,60 @@
...
+            </div>
+        ),
+    })
+)(MultilineBox);
+
+export interface MultiValuesInputProps {
+    id: string;
+    data: string[];
+}
+
```
rosalie-liu: I'm not sure if it's out of scope. `RemoveButton` appears if the input is not the last, but do we need an `AddButton` for the last one? 
image

GermainBergeron: The input is added as soon as the value is not empty so it's ok to not have the add button

packages/react-vapor/src/components/multiValueInputs/MultiValuesInput.tsx
GermainBergeron: I will add an `inputProps` to add everything that is missing

dmgauthier:
I think we should we remove the part that use the multiline box directly in the input examples, since your new component seem to remove the need of it. :clap: :clap: :+1:

GermainBergeron:
> I think we should we remove the part that use the multiline box directly in the input examples, since your new component seems to remove the need for it. 👏 👏 👍

It's a bit different though, my component only allows one input per "line" but the box exemple has 3 inputs per "line"

PR #1491, opened 2020-03-06
fix(filterbox): remove input text when reset filter value

rosalie-liu:
### Proposed Changes

Changed the condition in `componentDidUpdate`. Input text value in filter box should be updated to empty string, when resetting filter box by clicking on Clear Filter button in table blank slate.

No

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/filterBox/FilterBox.tsx
+++ b/packages/react-vapor/src/components/filterBox/FilterBox.tsx
@@ -68,7 +68,7 @@ export class FilterBox extends React.Component<IFilterBoxProps, any> {
     }
 
     componentDidUpdate(prevProps: IFilterBoxProps) {
-        if (this.props.filterText !== prevProps.filterText && this.filterInput?.value !== prevProps.filterText) {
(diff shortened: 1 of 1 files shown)


packages/react-vapor/src/components/filterBox/FilterBox.tsx
```diff
@@ -68,7 +68,7 @@ export class FilterBox extends React.Component<IFilterBoxProps, any> {
     }
 
     componentDidUpdate(prevProps: IFilterBoxProps) {
-        if (this.props.filterText !== prevProps.filterText && this.filterInput?.value !== prevProps.filterText) {
+        if (this.props.filterText !== prevProps.filterText && this.filterInput?.value !== this.props.filterText) {
```
rosalie-liu: Can we simply use `if (this.props.filterText !== prevProps.filterText)` here? I'm not sure if the second condition is necessary :thinking:

GermainBergeron: I think it's required by the FilterBox (not the connected one), you can do some testing if you need but to avoid any unnecessary changes I would keep the proposed conditions

PR #1495, opened 2020-03-11
feat(invisible-collapsible): invisible collapsible

pochretien:
set collapsible invisible if the content is null
add invisible class
add uts

### Acceptance Criteria


packages/react-vapor/src/components/table-hoc/TableRowConnected.tsx
```diff
@@ -112,28 +115,31 @@ class TableRowConnected extends React.PureComponent<
 
     render() {
```
GermainBergeron: I'm wondering if it would be easier to understand like this: 
```typescript
render() {
  if (_.isNull(props.collapsible?.content)) {
    return <td />
  }
  // old logic
}
```

It would remove the need for the custom style

PR #1497, opened 2020-03-12
feat(tableloading): added table of loading cards

FelixBlaisThon:
### Proposed Changes

added a  isCard prop in table loading so we can render loading cards in the same logic as loading table. One addition, is CardSubRow, which is the number of rows inside the card.

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/loading/components/TableLoading.tsx
+++ b/packages/react-vapor/src/components/loading/components/TableLoading.tsx
@@ -40,43 +40,43 @@ const Body = ({
 }) => {
     return (
         <tbody>
-            {isCard
-                ? _.times(numberOfRow, (nColumn: number) => (
-                      <tr key={`table-card-row-loading-${nColumn}`} className="no-hover">
-                          {_.times(numberOfColumns, (nRow: number) => (
-                              <Row
-                                  isCard={isCard}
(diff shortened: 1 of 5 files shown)


FelixBlaisThon:
Forgot UTs..

packages/react-vapor/src/components/loading/components/TableLoading.tsx
```diff
@@ -2,36 +2,96 @@ import * as classNames from 'classnames';
...
+}: {
+    isCard?: boolean;
+    numberOfColumns?: number;
+    numberOfRow?: number;
+    numberOfSubRow?: number;
+}) => {
+    return (
+        <tbody>
+            {isCard
+                ? _.times(numberOfRow, (nColumn: number) => (
```
dmgauthier: is a ternary the best thing to use here :thinking:  ?

FelixBlaisThon: isCard && would work, but did you have something else in mind?

rosalie-liu: I'd probably extract the two components, then in <tbody> for instance:
`{_.times(numberOfRow, (nColumn: number) => { isCard ? (<CardLoading />) : <RowLoading /> })}`

PR #1507, opened 2020-03-17
Vertical line component

FelixBlaisThon:
### Proposed Changes

Simple vertical line component since we want to use it in multiple section of the admin (right now its a span in source for example)

added style according to `separator` class and UTs

### Acceptance Criteria


packages/react-vapor/src/components/verticalLine/styles/VerticalLine.scss
```diff
@@ -0,0 +1,9 @@
+@import '~coveo-styleguide/scss/common/palette.scss';
+@import '~coveo-styleguide/scss/variables.scss';
+
+.separator {
```
gdostie: Why is it called sepator here but the component is called `VerticalLine` ? I would tend to rename one or the other so that both the css class and the react component have the same name.

PR #1515, opened 2020-03-26
feat(form): add noMargin prop

gdostie:
### Proposed Changes

I needed a Form without those nasty default margin top and bottom.

None

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/form/Form.tsx
+++ b/packages/react-vapor/src/components/form/Form.tsx
@@ -7,11 +7,12 @@ export interface IFormProps {
     title?: string;
     className?: string;
     mods?: FormMods | FormMods[];
+    noMargin?: boolean;
 }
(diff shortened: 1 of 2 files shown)


packages/react-vapor/src/components/form/Form.tsx
```diff
@@ -7,11 +7,12 @@ export interface IFormProps {
     title?: string;
     className?: string;
     mods?: FormMods | FormMods[];
+    noMargin?: boolean;
```
GermainBergeron: At first I thought it would be better to add this in the `mods`, but you'd need a bit of refactor

gdostie: I thought about it too. I don't have any preference so I went with the option that keeps it simple.

PR #1525, opened 2020-04-07
IE-757 Generate ESM React-Vapor bundle with Rollup

samisayegh:
### Background and Context
The Coveo Interface Editor uses a subset of react-vapor components for certain menus to have a uniform look with the AdminUI. However, we regularly face issues when we upgrade the react-vapor dependency:

1. Our build time and processing memory requirements increase, leading to javascript heap out of memory errors. These issues often take days to resolve. We look for optimizations by upgrading our webpack loaders, but lately these are becoming harder to find. Instead, we add memory to the node process running webpack. This only defers the problem, and does not reduce our build time.

2. A second effect is our final bundle is growing significantly. With the latest react-vapor version, our largest chunk sits at 4 MB minified; react-vapor constitutes 50% of the payload. In addition to load-time, bundle size is important for one of our project consumers, the Salesforce Integration team. The Salesforce platform imposes a max upload limit of 20 MB per zip. The Interface Editor zip, with all chunks and static assets, regularly goes above the limit when we upgrade react-vapor. We have found ways to cut in the past, but these are becoming harder to find.

### Proposed Changes

This PR uses Rollup to bundle and export the react-vapor library using ecmascript modules (ESM). ESM allows consumer projects to tree-shake the react-vapor dependency, keeping only the code they are using in their final bundle.

With commonjs, if a project is using only one react-vapor component, the entire library is bundled. With ESM, only the code related to that component makes it into the final bundle. Everything else is discarded.

I am not aware of any breaking changes at the moment. I caught a few differences between the two bundlers with the help of the compiler. A notable one is the rollup `sass-variable-loader` does not support the `preserveVariableNames` option. I fix it by adding code to post-process at run-time.

I tried importing the ESM bundle into the Interface Editor. **Everything worked**, but we use only a subset of components. I would be surprised if there isn't at least one breaking change hiding somewhere, e.g. an issue caused by a dependency being incorrectly loaded, a CSS file that is not being processed, etc. **I will need your help to find these**.

While breaking changes are scary, I believe it's important for the react-vapor library to take this step. If we delay, things will continue to get worse for developers on consuming projects, and end-users will have slower and slower experiences. If we work as a team, I think we can catch most, if not all of the issues without external clients being affected.

### Acceptance Criteria

I was not able to test the logic for converting the colors because I could not set up a failing test. e.g.
```
expect(true).toBe(false)
```
I think the issue was Chrome Headless was not starting on my Windows machine?

There is potential, but I am not aware of any at the moment.

Not applicable


dmgauthier:
I find that really interesting. But with some searching I did not find any insightfull documentation about the differences you explained between commonjs and ecmascript modules loading. Since those are kinda significant changes, can you link some doc plz ?

packages/react-vapor/package.json
```diff
@@ -14,14 +14,16 @@
...
+    "module": "dist/react-vapor.esm.js",
     "types": "dist/definitions/src/Entry.d.ts",
     "repository": {
         "type": "git",
         "url": "git+https://github.com/coveo/react-vapor.git"
     },
     "scripts": {
         "start": "webpack --watch",
-        "build": "gulp clean && webpack",
+        "build": "gulp clean && webpack && npm run build:esm",
```
gdostie, line 25: webpack and rollup could be run concurrently

samisayegh:
Hi @dmgauthier, that is a good article. There are others that should pop-up when searching for keywords such as `ESM` and `treeshaking`.

@gdostie the build is now passing and the demo site is working. The link is just above. Seems good at first glance, but let me know if you spot anything.

gdostie:
Everything looks good from the react-vapor side of thing. However it is breaking the Admin UI build. I am currently linking both packages together locally and trying to find a solution.

gdostie:
@samisayegh I made [a PR](https://github.com/coveo/react-vapor/pull/1541) that adds to your branch if you want to check it out.

With those changes, the admin-ui runs successfully using the build made using Rollup :tada:

samisayegh:
Hey @gdostie, I am not authorized to merge this PR. I will fix the conflicts though, and maybe we can merge shortly afterwards?

gdostie:
@samisayegh Yes no problem. I can merge it as soon as the build is green. I will squash those commits thought if you don't mind, because changelogs will be generated out of those. So the less, the better for a single feature.

PR #1530, opened 2020-04-27
Minor fixes of table style

rosalie-liu:
### Proposed Changes

1. When we use the class `.mod-card` in a table row, the background color should not be changed to light grey on hover. This will fix the conflict of background colors in the ML blacklists configuration cards.

2. Add `.no-border-bottom` to allow removing the border bottom in a table row.

Nope. The first one is only used in Blacklists, and the second one is a new class.

### Acceptance Criteria

### DIFF
--- a/packages/vapor/scss/tables/table.scss
+++ b/packages/vapor/scss/tables/table.scss
@@ -101,7 +101,7 @@
             opacity: 0.4;
         }
 
-        &:not(.no-hover):not(.collapsible-row):not(.empty):hover {
+        &:not(.no-hover):not(.collapsible-row):not(.empty):not(.mod-card):hover {
             background-color: $light-grey;
             td {
                 background-color: $light-grey;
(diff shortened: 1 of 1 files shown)


packages/vapor/scss/tables/table.scss
```diff
@@ -137,7 +137,10 @@
     td {
         padding: $table-row-top-bottom-padding $table-cell-left-right-padding;
         text-align: left;
-        border-bottom: $table-border;
+
+        &:not(.no-border-bottom) {
```
gdostie, line 141: We already have the `mod-no-border-bottom` class.
https://github.com/coveo/react-vapor/blob/4d62e60a5a1c2e66fa7326f87f0619e535baad15/packages/vapor/scss/utility/border.scss#L23
Can it be used instead ?

rosalie-liu: image
It seems to be overrode because of the precedence :thinking:

gdostie: :+1: Then perhaps the css rule could be
```suggestion
        &:not(.mod-no-border-bottom) {
```
This way we won't have to use another new class (`no-border-bottom`)

rosalie-liu: Yes!

PR #1542, opened 2020-05-05
feat(middle slider): can append a value on both side

dmgauthier:
### Proposed Changes

we can now append the slider values to both side of the slider. The to side can be formatted with the `appendedValueFormatter` props

I created an union type for the `appendedValue` props. So if we simply pass the `appendedValue` to `true`, the appended value will normaly appear to the right of the slider. But we can pass it with the `AppendedValueSide` enum that has the `left` `right` `both` keys and the value will appear accordingly.

The Initial implementation of the `appendedValueFormatter` prop still work as intended, but depending of the side it formats, the `left` or `right` will be passed in the params to grant the developer the possibility to customize the format differently if need be. 

no breaking change intended

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/slider/MiddleSlider.tsx
+++ b/packages/react-vapor/src/components/slider/MiddleSlider.tsx
@@ -5,6 +5,7 @@ import {Range, SliderProps} from 'rc-slider';
 import {RCTooltip} from 'rc-tooltip';
 import * as React from 'react';
 import {connect} from 'react-redux';
+import {isBoolean} from 'underscore';
 
 import {IDispatch} from '../../utils/ReduxUtils';
 import {SliderActions} from './SliderActions';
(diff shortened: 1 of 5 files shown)


packages/react-vapor/src/components/slider/MiddleSlider.tsx
```diff
@@ -138,9 +147,18 @@ const MiddleSliderDisconnected: React.FunctionComponent<MiddleSliderOwnProps &
     };
 
     const computedStep = computeStep(step, min, max);
+    const isAppendedLeft = appendValue === AppendedValueSide.both || appendValue === AppendedValueSide.left;
+    const isAppendedRight = appendValue && appendValue !== AppendedValueSide.left;
```
pochretien, line 151:
``` 
const isAppendedRight = (_.boolean(appendValue) && appendValue) || appendValue === AppendedValueSide.right || appendValue === AppendedValueSide.both
```

PR #1544, opened 2020-05-06
feat(datepicker): use drop

pochretien:
Add drop as option for the datePicker
Add 2 examples to use the datePicker with drop
add uts for the new drop component

### Proposed Changes

User drop for the Date Picker to be rendered everywhere without parent container problems

none

image


pochretien:
Juste to be sure to avoid any style breaking change. We can remove the old wait but I dont know if somewhere the style is custom with the button dropdown

PR #1548, opened 2020-05-12
fix(selects): cleanup the selects type declarations

gdostie:
### Proposed Changes

- Cleaned up the types of multiple selects components and HOCs

Some missused selects might break

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/select/SelectConnected.tsx
+++ b/packages/react-vapor/src/components/select/SelectConnected.tsx
@@ -4,11 +4,10 @@ import {createStructuredSelector} from 'reselect';
 import {keys} from 'ts-transformer-keys';
 import * as _ from 'underscore';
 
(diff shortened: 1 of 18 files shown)


packages/react-vapor/src/components/select/SelectConnected.tsx
```diff
@@ -119,28 +101,14 @@ export class SelectConnected extends React.PureComponent<ISelectProps & ISelectS
             open: this.props.isOpened,
             'mod-multi': this.props.multi,
         });
-        const minWidth = (this.dropdown && this.dropdown.clientWidth) || undefined;
+        const minWidth = this.dropdown.current?.clientWidth || undefined;
```
pochretien, line 104: you dont use the same object than before. Are you sure. current is available ?

gdostie: https://reactjs.org/docs/refs-and-the-dom.html#creating-refs

PR #1552, opened 2020-05-14
SingleSelectConnected custom button

gdostie:
### Proposed Changes

- Added the `customButton` prop on the `SingleSelectConnected` component. It allows to render a custom toggle for the SingleSelectConnected instead of the default white toggle.
- Added 2 new icons: `flagUs` and `flagEu`. SVGs were taken from Wikipedia
image
- Adjusted the `blue-purple-2` color so that it matches the color we use for active side navigation section and opened dropdowns.

None

### Acceptance Criteria

### DIFF
--- a/packages/vapor/scss/common/palette.scss
+++ b/packages/vapor/scss/common/palette.scss
@@ -70,7 +70,7 @@ $onahau-1: #b6eeff;
 $blue-purple-palette: category;
 $coveo-blue-purple: $mirage;
 $blue-purple-1: #3e5485;
-$blue-purple-2: #354878;
+$blue-purple-2: #354667;
 $blue-purple-3: #2d385e;
 $blue-purple-4: $coveo-blue-purple;
 $blue-purple-5: #101425;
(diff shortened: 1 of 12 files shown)


packages/vapor/scss/common/palette.scss
```diff
@@ -70,7 +70,7 @@ $onahau-1: #b6eeff;
 $blue-purple-palette: category;
 $coveo-blue-purple: $mirage;
 $blue-purple-1: #3e5485;
-$blue-purple-2: #354878;
+$blue-purple-2: #354667;
```
pochretien, line 73: why this color has changed ?

gdostie: The color (`#354667`) was used in multiple places, but wasn't even in the color palette. So I replaced an existing color by it instead of creating yet another color (:wink: :wink: @oliviertaillon). The previous color (`#354878`) was only used for [the NumericInput](https://vapor.cloud.coveo.com/singleselect-custom-button/index.html#/components/NumericInput) [+] and [-] buttons.

PR #1577, opened 2020-06-05
fix(multiselect): fix the custom value selected option box from demo

sallainCoveo:
After the selection of a custom value, using a multi select filter, the selected option box now
appears in the selected option area.

### Proposed Changes

Initially, only the default items, present in the dropdown menu selection, are considered on the creation of the selected option box. However, the previous method was necessary as it would take consideration of the display value, and not only the selected value.  Nevertheless, because the custom values are not included in the items props, I suggested that the method would fetch the remaining selected option (from the selected props) and included in the returned array of IItemBoxProps. In doing so, the custom values may be included in the selection option area.

### Acceptance Criteria

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,14 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [5.50.1](https://github.com/coveo/react-vapor/compare/v5.50.0...v5.50.1) (2020-06-04)
+
+**Note:** Version bump only for package root
+
+
+
+
+
 # [5.50.0](https://github.com/coveo/react-vapor/compare/v5.49.1...v5.50.0) (2020-06-03)
 
 
--- a/lerna.json
+++ b/lerna.json
@@ -2,7 +2,7 @@
     "packages": [
         "packages/*"
     ],
-    "version": "5.50.0",
+    "version": "5.50.1",
     "command": {
         "publish": {
             "conventionalCommits": true,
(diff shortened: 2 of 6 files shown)


FelixBlaisThon:
Nice! a question remains for me. With these changes, is it still necessary to keep the + button on the right? it doesn't seem to add the value to the selected ones @oliviertaillon

oliviertaillon:
@FelixBlaisThon Good point, I couldn't find what it is doing. @sallainCoveo what is supposed to happen when clicking on the ( + ) icon ? 
(Maybe out of scope for this PR)

FelixBlaisThon:
It triggers the same action 'ADD_VALUE_STRING_LIST' as the one when you click on the custom item on the dropdown, but it doesn't add it :/

sallainCoveo:
@oliviertaillon ,

Honestly, I'm not sure. It was already present when I was working on this bug. I suppose that the initial purpose of this icon would be to add a custom item (an alternative action of clicking on the
option on the drop down menu). I noticed that this button didn't work, but it would seems that this
issue is out of the scope of the initial bug. Do you think it would be interesting to have both options for the end-user experience?

oliviertaillon:
> Do you think it would be interesting to have both options for the end-user experience?

I don't think the + icon add much value, we could remove it in a different PR. The part that is more confusing is that you have to first enter a filter, then you are able to add this as a custom value. But this is a different story not relevant to this current PR.

rosalie-liu:
hmm it seems there is a little problem with the format?

sallainCoveo:
@rosalie-liu,

Thanks for the observation! I did use the command npm run lintfix to format my file, but I might have created a few issues along the way. I'll try to add the Prettier plugin to correct these strangely formatted files.

PR #1582, opened 2020-06-08
fix(multiselect): fix the selected custom value option box from demo

sallainCoveo:
Fix the selected custom value option box that was not visible, as the custom value is not included
in the array of items availables from the previous method

### Proposed Changes
(Due to some format issue, I had to create a new pull request),

Before this modification, all displayed selected option box are items that would normally be suggested from the filter. However, as a custom value might be suggested to be added for the end-user and because it is not part of the items' list,  this selected option is not displayed in the selected option box section. To resolve this issue, all the selected items are now checked and included in the returned list, which includes the custom values. In doing, we first checked the selected items from the "items" list, so we can get the displayValue from them, then we included the selected custom values.
 

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/select/MultiSelectConnected.tsx
+++ b/packages/react-vapor/src/components/select/MultiSelectConnected.tsx
@@ -5,6 +5,7 @@ import {createStructuredSelector} from 'reselect';
 import {keys} from 'ts-transformer-keys';
 import * as _ from 'underscore';
 
+import {convertItemsBoxToStringList, convertStringListToItemsBox} from '../../reusableState';
 import {IDispatch, ReduxConnect} from '../../utils/ReduxUtils';
 import {DnDUtils} from '../dragAndDrop/DnDUtils';
 import {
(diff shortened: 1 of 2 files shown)


sallainCoveo:
For the sake of not losing any important information from our previous comments:

As the ( + ) icon doesn't seems to add much value to the end-user experience, the latter could be remove in a next PR.

dmgauthier:
multiSelect

Is the + sign supposed to work ?

sallainCoveo:
@dmgauthier,

No, this button didn't work before I started to work on this fix, and it would seems that this 
action is out of the scope of this case. However, I agree that having this icon available is misleading,
and it should be removed (as mentioned in my previous message). 

Do you think that this change might be relevant to this case, or should it be report on another one?

gdostie:
The [+] button is intended to work for adding custom values. I don't think it should be removed.

sallainCoveo:
As mentioned in this comment:
https://github.com/coveo/react-vapor/pull/1577#issuecomment-639581989

The ( + ) icon might not be a plus value for the end-user experience. However, as it is still present, it could be beneficial that I simply delete it (because it cause confusion on the in the intended usage). 
So, I will delete it, unless someone thinks that it would be an asset to keep for the demo. If so, then I will correct this feature. @coveo/admin-console

packages/react-vapor/src/components/select/MultiSelectConnected.tsx
```diff
@@ -183,7 +184,15 @@ class MultiSelect extends React.PureComponent<IMultiSelectProps> {
...
+        const selectedItemsWithoutCustom: IItemBoxProps[] = this.props.items.filter((option: IItemBoxProps) =>
+            _.contains(this.props.selected, option.value)
+        );
+        const selectedItemsWithoutCustomItems: string[] = convertItemsBoxToStringList(selectedItemsWithoutCustom);
+        const customItemsValues: string[] = _.difference(this.props.selected, selectedItemsWithoutCustomItems);
+        const customItems: IItemBoxProps[] = convertStringListToItemsBox(customItemsValues);
+        const selectedItems: IItemBoxProps[] = selectedItemsWithoutCustom.concat(customItems);
+
+        return selectedItems;
     }
```
j-mcgregor, line 196: Can we get a test for `getSelectedOptions`?

sallainCoveo: Good point, I added two tests to evaluate that if a custom item is stored in the selected, would it be displayed as a SelectedOption.

pochretien:
Approved

PR #1589, opened 2020-06-10
Conversion "limit" style to "limit" React component

sallainCoveo:
### Proposed Changes

This PR represent the first draft of the conversion of the "limit" style to a React component. For now, the "limit" style is still present, so you could compare the two features.  

### Acceptance Criteria


packages/demo/src/components/examples/LimitExamples.tsx
```diff
@@ -0,0 +1,83 @@
...
+    render() {
+        const title: string = 'Limit example';
+        const usageNumber: number = 42;
+        const usageNumberNearLimit: number = 80;
+        const usageNumberEqualToLimit: number = Limit.defaultProps.limitNumber;
+        const customLimitNumber: number = 130;
+        const customLimitTitle: string = 'Throttling limit';
+        const isLimitCanBeChanged: boolean = true;
+        const isLimitIsTheGoalToReach: boolean = true;
+        const isHistoryIsIncluded: boolean = true;
```
j-mcgregor, line 14: I may be wrong as I'm fairly new to Typescript, but I don't think specifying types here is necessary, since you're using `const` and setting the values.

packages/demo/src/components/examples/LimitExamples.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+        return (
+            <div className="coveo-form">
+                <div className="form-group">
+                    <label className="form-control-label"> A limit with an usage value</label>
```
j-mcgregor, line 19: `>A limit...` (space)

packages/demo/src/components/examples/LimitExamples.tsx
```diff
@@ -0,0 +1,83 @@
...
+                        <Limit
+                            title={title}
+                            usageNumber={usageNumber}
+                            limitNumber={customLimitNumber}
+                            limitLabel={customLimitTitle}
+                        />
+                    </div>
+                </div>
+                <div className="form-group">
+                    <label className="form-control-label"> A limit with a modifiable limit value </label>
```
j-mcgregor, line 69: Same here

packages/demo/src/components/examples/LimitExamples.tsx
```diff
@@ -0,0 +1,83 @@
...
+                    <div className="form-control">
+                        <Limit title={title} usageNumber={usageNumber} isLimitModifiable={isLimitCanBeChanged} />
+                    </div>
+                </div>
+                <div className="form-group">
+                    <label className="form-control-label"> A limit without an usage value</label>
```
j-mcgregor, line 75: And here

packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -0,0 +1,142 @@
...
+    isLimitTheGoalToReach?: boolean;
+    isHistoryIncluded?: boolean;
+}
+
+export interface ILimitDispatchProps {
+    onRender?: () => void;
+    onDestroy?: () => void;
+}
+
+export interface ILimitProps extends ILimitOwnProps, ILimitDispatchProps {}
```
j-mcgregor, line 20: Interfaces shouldn't be prefixed with `I` - see [docs](https://github.com/coveord/tsjs#interface-no-i)

packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -0,0 +1,142 @@
...
+        );
+    }
+
+    private getHistoryIcon() {
+        const iconClasses = classNames('icon', 'mod-lg', 'limit-history-button');
+        return (
+            this.props.isHistoryIncluded && (
+                <span className={iconClasses}>
+                    <span className="fill-medium-blue">
+                        <svg className="" viewBox="0 0 16 16">
```
j-mcgregor, line 69: is `className=""` necessary?

j-mcgregor, line 69: There an SVG component [here](https://github.com/coveo/react-vapor/tree/master/packages/react-vapor/src/components/svg)

packages/vapor/scss/components/limit-box.scss
```diff
@@ -12,6 +12,7 @@
     }
 }
 
+//addition of a class
```
j-mcgregor, line 15: ?

packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -0,0 +1,142 @@
...
+}
+
+export interface ILimitDispatchProps {
+    onRender?: () => void;
+    onDestroy?: () => void;
+}
+
+export interface ILimitProps extends ILimitOwnProps, ILimitDispatchProps {}
+
+export class Limit extends React.Component<ILimitProps, {}> {
```
pochretien, line 22: you can use a PureComponent instead of a Component. Its not a big deal but the PureComponent its a bit more faster
`React.Component<ILimitProps>`

If you want too, you can convert the component with a Functional component. Its faster than the PureComponent and give the possibility to use the useEffect.
nice doc: https://www.robinwieruch.de/react-function-component

packages/react-vapor/src/components/limit/Limit.tsx
pochretien: Use the `<Svg />` component

packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -0,0 +1,142 @@
...
+                            <path d="M7.3 0v8.1c0 .2 0 .4.2.5l5.7 5.5C11.7 15.3 9.9 16 8 16c-4.4 0-8-3.6-8-8C0 3.8 3.2.4 7.3 0zm8 4.7c.4 1 .7 2.1.7 3.3 0 2-.8 3.9-2 5.3l-5.1-5zM8.4 0c2.6.1 4.9 1.6 6.3 3.6L8.4 7.3z" />
+                        </svg>
+                    </span>
+                </span>
+            )
+        );
+    }
+
+    private getContentDivision() {
+        const contentClasses = classNames('limit-box-numbers', 'pt1', 'flex');
```
pochretien, line 79: you can set the class direct in the div element. classNames is useful when we want to add conditional class or combine list of classes

packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -0,0 +1,142 @@
...
+                <div className="limit-box-limit">
+                    <label className="form-control-label">{this.props.limitLabel}</label>
+                    <span className="limit-box-limit-value">{this.props.limitNumber}</span>
+                </div>
+            );
+        }
+    }
+
+    private getUsageDivision() {
+        const isUsageRequired: boolean = this.props.usageNumber ? true : false;
```
pochretien, line 111: you can use the `!!this.props.usageNumber` to convert your number into a boolean

packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -0,0 +1,142 @@
...
+                <div className="limit-box-footer">
+                    <div className={progressClasses}></div>
+                </div>
+            )
+        );
+    }
+
+    private getProgressClass() {
+        const progress: number = Math.round((this.props.usageNumber / this.props.limitNumber) * 100);
+        return 'progress-' + progress.toString();
```
pochretien, line 140: if you want, you can use this syntaxe to build your string 
```
`progress-${progress.toString()}`
```

package.json
```diff
@@ -55,5 +55,9 @@
         "extends": [
             "@commitlint/config-conventional"
         ]
+    },
```
pochretien, line 58: Im not sure why we need that 🤔

package.json
```diff
@@ -10,7 +10,7 @@
         "axios": "0.19.0",
         "cz-conventional-changelog": "3.0.2",
         "husky": "3.0.9",
-        "lerna": "3.18.4",
+        "lerna": "^3.18.4",
```
pochretien, line 13: remove ^

packages/react-vapor/src/components/limit/Limit.tsx
FelixBlaisThon: Functional component :heart:

packages/demo/src/components/examples/LimitExamples.tsx
sallainCoveo: To make sure I understand your comment, you suggest that my line would look like this:
` <label className="form-control-label"> A limit with an usage value </label>`
or like this:
`<label className="form-control-label"> 
        A limit with an usage value
</label>`
?

j-mcgregor: Sorry, I should have been clearer - you should remove the unnecessary space between the closing bracket and the letter A

packages/react-vapor/src/components/limit/Limit.tsx
j-mcgregor: Gotta love those template literals

package.json
sallainCoveo: Normally, I don't we would. However, if I don't include this extension, I cannot commit (which I'm not sure why).

j-mcgregor: which node version are you running?

sallainCoveo: v12.17.0

package.json
```diff
@@ -55,5 +55,9 @@
         "extends": [
             "@commitlint/config-conventional"
         ]
+    },
+    "dependencies": {
+        "coveo-styleguide": "5.48.0",
+        "react-vapor": "5.49.0"
```
gdostie, line 61: You can revert those, as our package is not dependent on itself.

packages/demo/src/components/examples/LimitExamples.tsx
```diff
@@ -0,0 +1,82 @@
+import * as React from 'react';
+import {Limit} from 'react-vapor';
+
+export class LimitExamples extends React.Component {
+    render() {
```
gdostie, line 5: We tend to go with functional components instead of class components.
```tsx
export const LimitExamples: React.FunctionComponent = () => {
    // consts here
    return (
        <div>Markup here</div>
    );
};
```

packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -0,0 +1,155 @@
+import classNames from 'classnames';
+import * as VaporSVG from 'coveo-styleguide';
+import * as React from 'react';
+import {Input, Svg} from 'react-vapor';
+
+export interface LimitProps {
+    title: string;
+    limitLabel?: string;
+    usageNumber?: number;
+    limitNumber?: number;
```
gdostie, line 10: As those 2 props are typed as `number` I don't think its necessary to repeat this information in the props name.
```suggestion
    usage?: number;
    limit?: number;
```

packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -0,0 +1,155 @@
...
+export const Limit: React.FunctionComponent<LimitProps> = ({
+    title,
+    limitLabel,
+    usageNumber,
+    limitNumber,
+    isLimitModifiable,
+    isLimitTheGoalToReach,
+    isHistoryIncluded,
+}) => {
+    const HeaderDivision = ({limitTitle, isHistory}: {limitTitle: string; isHistory: boolean}) => {
```
gdostie, line 25: This component could be defined outside of the `Limit` function component

packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -0,0 +1,155 @@
...
+        return (
+            <div className="flex space-between">
+                <label className="form-control-label"> {limitTitle}</label>
+                <HistoryIcon isHistory={isHistory} />
+            </div>
+        );
+    };
+
+    const HistoryIcon = ({isHistory}: {isHistory: boolean}) => {
```
gdostie, line 34: Same here

packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -0,0 +1,155 @@
...
+        return (
+            isHistory && (
+                <span className="icon mod-lg limit-history-button">
+                    <Svg svgName={VaporSVG.svg.menuAnalytics.name} className="fill-medium-blue" />
+                </span>
+            )
+        );
+    };
+
+    const ContentDivision = ({
```
gdostie, line 44: same here

packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -0,0 +1,155 @@
...
+    }) => {
+        return (
+            <div className="limit-box-numbers pt1 flex">
+                <UsageDivision usageInt={usageInt} />
+                <LimitDivision limitInt={limitInt} isModifiable={isModifiable} limitTitle={limitTitle} />
+            </div>
+        );
+    };
+
+    const UsageDivision = ({usageInt}: {usageInt: number}) => {
```
gdostie, line 63: Same here

packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -0,0 +1,155 @@
...
+            isUsageRequired && (
+                <div className="limit-box-usage">
+                    <label className="form-control-label">Usage</label>
+                    <span className="limit-box-usage-value">{usageInt}</span>
+                </div>
+            )
+        );
+    };
+
+    const LimitDivision = ({
```
gdostie, line 75: Same here

packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -0,0 +1,155 @@
...
+            return (
+                <div className="limit-box-limit">
+                    <label className="form-control-label">{limitTitle}</label>
+                    <span className="limit-box-limit-value">{limitInt}</span>
+                </div>
+            );
+        }
+    };
+
+    const ProgressBar = ({
```
gdostie, line 104: Same here

packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -0,0 +1,155 @@
...
+                    <label className="form-control-label">Usage</label>
+                    <span className="limit-box-usage-value">{usageInt}</span>
+                </div>
+            )
+        );
+    };
+
+    const LimitDivision = ({
+        limitInt,
+        isModifiable,
```
gdostie, line 77:
```suggestion
        limitValue,
        isEditable,
```

packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -0,0 +1,210 @@
...
+                <ContentDivision
+                    id={id}
+                    usageValue={usage}
+                    limitValue={limit}
+                    isEditable={isLimitEditable}
+                    limitTitle={limitLabel}
+                    onChangeLimit={onChangeLimit}
+                />
+            </div>
+            <ProgressBar usageValue={usage} isLimitTheGoal={isLimitTheGoalToReach} limitValue={+currentLimit} />
```
FelixBlaisThon, line 66: is the `+currentLimit` a typo or something i never saw?

sallainCoveo: It's something you never saw. Now that the limit component is connected, this variable represent the prop that mapped the state of the input. This value is mainly used to change the progress bar class and move the position of the latter based on the user's limit choice. 
As the input value is a string, I had to convert currentLimit to a number (which explains the + in front).

FelixBlaisThon: today i learned ! :D

packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -0,0 +1,210 @@
...
+    limitLabel?: string;
+    usage?: number;
+    limit?: number;
+    isLimitEditable?: boolean;
+    isLimitTheGoalToReach?: boolean;
+    isHistoryIncluded?: boolean;
+}
+
+const mapStateToProps = (state: IReactVaporState, ownProps: LimitOwnProps) => {
+    const ownLimit: string = ownProps.limit.toString() || Limit.defaultProps.limit.toString();
```
dmgauthier, line 23: `const ownLimit: string = ownProps.limit.toString() ?? Limit.defaultProps.limit.toString();`
`const currentLimit: string = InputSelectors.getValue(state, {id: ownProps.id}) || ownLimit;`

[](https://www.typescriptlang.org/play/#example/nullish-coalescing)

packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -0,0 +1,210 @@
...
+    title: string;
+    limitLabel?: string;
+    usage?: number;
+    limit?: number;
+    isLimitEditable?: boolean;
+    isLimitTheGoalToReach?: boolean;
+    isHistoryIncluded?: boolean;
+}
+
+const mapStateToProps = (state: IReactVaporState, ownProps: LimitOwnProps) => {
```
gdostie, line 22: For even more street creds, `ownProps` could be deconstructed
```suggestion
const mapStateToProps = (state: IReactVaporState, {id, limit}: LimitOwnProps) => {
```

packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -0,0 +1,210 @@
...
+    id,
+    title,
+    limitLabel,
+    usage,
+    limit,
+    currentLimit,
+    isLimitEditable,
+    isLimitTheGoalToReach,
+    isHistoryIncluded,
+    onChangeLimit: onChangeLimit,
```
dmgauthier, line 51: `onChangeLimit: onChangeLimit` => `onChangeLimit`

packages/react-vapor/src/components/limit/Limit.tsx
gdostie, line 23: Not sure I understand why we are converting the limit to a `string`, because later in the component we are reconverting it to a number with the `+` sign. Wouldn't it be simpler to convert the return value of `InputSelectors.getValue` to a number instead?

packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -0,0 +1,210 @@
...
+        dispatch(changeInputValue(ownProps.id, limitValue, true));
+    },
+});
+
+export interface LimitProps
+    extends LimitOwnProps,
+        ReturnType<typeof mapDispatchToProps>,
+        ReturnType<typeof mapStateToProps> {}
+
+export const LimitDisconnect: React.FunctionComponent<LimitProps> = ({
```
gdostie, line 41: We generally don't export the unconnected version.
```suggestion
const LimitDisconnect: React.FunctionComponent<LimitProps> = ({
```

packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -0,0 +1,210 @@
...
+    isEditable,
+    limitTitle,
+    onChangeLimit,
+}: {
+    id: string;
+    usageValue: number;
+    limitValue: number;
+    isEditable: boolean;
+    limitTitle: string;
+    onChangeLimit: (limit: string) => void;
```
dmgauthier, line 103: I would do an interface here in order to extract the logic. 
But even better, in order to reuse logic. Since the props you are passing here are mostly the same than the `LimitDisconnect` props you could create a type by picking the props from the `LimitProps` like this:
`type NewType = Pick<LimitProps, 'Id', 'usage' [...so on] >`  But of course you would have to keep the name of your props the same and not modify them like you did (ex: limit => limitValue).

Im not 100% sure if you can pick type properties since they came from a `returnType<typeof mapStateToProps>` but this sure worth the shot to try.

worst case extract all those properties in a distinct interface.

packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -0,0 +1,210 @@
...
+    return (
+        isHistory && (
+            <span className="icon mod-lg limit-history-button">
+                <Svg svgName={VaporSVG.svg.menuAnalytics.name} className="fill-medium-blue" />
+            </span>
+        )
+    );
+};
+
+const ContentDivision = ({
```
dmgauthier, line 90: this should be a React.FunctionComponent.  `const ContentDivision: React.FunctionComponent<TYPE OF THE PROPS> = (PROPS) => {}`

packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -0,0 +1,210 @@
...
+    onChangeLimit,
+}: {
+    id: string;
+    usageValue: number;
+    limitValue: number;
+    isEditable: boolean;
+    limitTitle: string;
+    onChangeLimit: (limit: string) => void;
+}) => {
+    return (
```
dmgauthier, line 105: This component is only rendering jsx with no logic. So you can skip  the return statement like this: 

`const MyComponent: React.FunctionComponent = () => (<div>Bonjour !</div>)`

packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -0,0 +1,210 @@
...
+                usageValue={usageValue}
+                limitValue={limitValue}
+                isEditable={isEditable}
+                limitTitle={limitTitle}
+            />
+        </div>
+    );
+};
+
+const UsageDivision = ({usageValue}: {usageValue: number}) => {
```
dmgauthier, line 120: `const UsageDivision: React.FunctionComponent<{usageValue:number}> = ({usageValue}) => (!!usageValue && ([...]))`

packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -0,0 +1,210 @@
...
+    limitTitle,
+    onChangeLimit,
+}: {
+    id: string;
+    limitValue: number;
+    usageValue: number;
+    isEditable: boolean;
+    limitTitle: string;
+    onChangeLimit: (limit: string) => void;
+}) => {
```
dmgauthier, line 146: Same here with the `React.FunctionComponent` and the props type management

packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -0,0 +1,210 @@
...
+    const limitValueString: string = limitValue.toString();
+    const minLimitValue: number = usageValue || 0;
```
dmgauthier, line 148: `usageValue ?? 0`

packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -0,0 +1,210 @@
...
+        return (
+            <div className="limit-box-limit">
+                <label className="form-control-label">{limitTitle}</label>
+                <span className="limit-box-limit-value">{limitValue}</span>
+            </div>
+        );
+    }
+};
+
+const ProgressBar = ({
```
dmgauthier, line 171: Same here

packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -0,0 +1,210 @@
...
+    usageValue,
+    isLimitTheGoal,
+    limitValue,
+}: {
+    usageValue: number;
+    isLimitTheGoal: boolean;
+    limitValue: number;
+}) => {
+    const isUsageRequired: boolean = !!usageValue;
+    const progressClass: string = getProgressBarClass({usageValue, limitValue});
```
dmgauthier, line 181: I think you could avoid structuring/destructuring you object even if it is fun ;) since you have only two args in the params.

`getProgressBarClass(usageValue, limitValue)`

`const getProgressBarClass = (usageValue: number, limitValue:number): string => {}`

packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -0,0 +1,210 @@
...
+
+LimitDisconnect.defaultProps = {
+    limitLabel: 'Limit',
+    limit: 100,
+    isLimitEditable: false,
+    isLimitTheGoalToReach: false,
+    isHistoryIncluded: false,
+};
+
+Limit.defaultProps = {...LimitDisconnect.defaultProps};
```
dmgauthier, line 210: This is interesting. but I don't get why you did that here, attaching the default props to the LimitDisconnect this way and then spreading it in the limit components looks over engineering. 
Why no a simple `const limitDefaultProps`  ?

packages/react-vapor/src/components/limit/Limit.tsx
dmgauthier: good point

sallainCoveo: Nice catch! I didn't realize that I could only do one conversion: convert the input (return from the Input Selector), to a number. Doing so, I don't have to convert the limit props, or the currentLimit on the ProgressBar component.

packages/react-vapor/src/components/limit/Limit.tsx
sallainCoveo: Thanks for this tip! Actually, would it be better that I set all my props with Pick? For instance, if I set my ContentProps this way: 
`type ContentProps = type ContentProps = Pick<LimitProps, 'id' | 'usage' | 'limit' | 'isLimitEditable' | 'limitLabel' | 'onChangeLimit'>; `
Then I set my UsageProps this way:
`type UsageProps = Pick<ContentProps, 'usage'>`
Not only I make sure that every props from Limit is passed on through its children, but it will also be easier to read. What do you think?

packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -0,0 +1,183 @@
...
+    usage?: number;
+    limit?: number;
+    isLimitEditable?: boolean;
+    isLimitTheGoalToReach?: boolean;
+    isHistoryIncluded?: boolean;
+}
+
+const mapStateToProps = (state: IReactVaporState, {id, limit}: LimitOwnProps) => {
+    const ownLimit: number = limit ?? Limit.defaultProps.limit;
+    const currentLimit: number = +InputSelectors.getValue(state, {id: id}) || ownLimit;
```
gdostie, line 24:
```suggestion
    const currentLimit: number = +InputSelectors.getValue(state, {id}) || ownLimit;
```

packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -0,0 +1,183 @@
...
+    return isLimitEditable ? (
+        <InputConnected
+            id={id}
+            type="number"
+            labelTitle={limitLabel}
+            defaultValue={limitValueString}
+            min={minLimitValue}
+            classes="limit-box-limit form-group input-field validate"
+            onChange={(limitValue) => onChangeLimit(limitValue)}
+        ></InputConnected>
```
gdostie, line 145:
```suggestion
        />
```

PR #1594, opened 2020-06-16
feat(multiselect): com-435 create non empty validation for multi select

cnguyen-coveo:
https://coveord.atlassian.net/browse/COM-435

### Proposed Changes

Created a validation for non empty multi select

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/validation/hoc/WithNonEmptyValueMultiSelectValidationHOC.tsx
+++ b/packages/react-vapor/src/components/validation/hoc/WithNonEmptyValueMultiSelectValidationHOC.tsx
@@ -1,0 +1,57 @@
+import * as React from 'react';
+import {connect} from 'react-redux';
+import {createStructuredSelector} from 'reselect';
+import {IDispatch} from '../../../utils/ReduxUtils';
+import {IMultiSelectOwnProps} from '../../select/MultiSelectConnected';
+import {SelectSelector} from '../../select/SelectSelector';
+import {ValidationActions} from '../ValidationActions';
+import {ValidationTypes} from '../ValidationTypes';
+import {InferableComponentEnhancer} from './connectHOC';
+
+const mapStateToProps = createStructuredSelector({
+    selectedValues: SelectSelector.getMultiSelectSelectedValues,
+});
+
+const mapDispatchToProps = (dispatch: IDispatch) => ({
+    setError: (id: string, error: string) => dispatch(ValidationActions.setError(id, error, ValidationTypes.nonEmpty)),
+    clearError: (id: string) => dispatch(ValidationActions.clearError(id, ValidationTypes.nonEmpty)),
+});
+
+export interface WithNonEmptyValueMultiSelectValidationProps {
+    nonEmptyValidationMessage?: string;
+    resetNonEmptyMultiSelectErrorOnUnmount?: boolean;
+}
+
+export const withNonEmptyMultiSelectHOC = <T extends IMultiSelectOwnProps>(Component: React.ComponentType<T>) => {
+    type NewOwnProps = T & WithNonEmptyValueMultiSelectValidationProps;
+    type StateProps = ReturnType<typeof mapStateToProps>;
+    type DispatchProps = ReturnType<typeof mapDispatchToProps>;
+    const WrappedMultiSelect: React.FunctionComponent<NewOwnProps & StateProps & DispatchProps> = ({
+        selectedValues,
+        setError,
+        clearError,
+        nonEmptyValidationMessage = 'Selection required',
+        resetNonEmptyMultiSelectErrorOnUnmount,
+        ...props
+    }) => {
+        const hasValuesSelected = selectedValues.length > 0;
+
+        React.useEffect(() => {
+            clearError(props.id);
+            return () => {
+                resetNonEmptyMultiSelectErrorOnUnmount && clearError(props.id);
+            };
+        }, [props.id, resetNonEmptyMultiSelectErrorOnUnmount]);
+
+        React.useEffect(() => {
+            setError(props.id, !hasValuesSelected ? nonEmptyValidationMessage : '');
+        }, [hasValuesSelected, props.id, nonEmptyValidationMessage]);
+
+        return <Component {...(props as T)} />;
+    };
+
+    const enhance = connect(mapStateToProps, mapDispatchToProps) as InferableComponentEnhancer<
+        StateProps & DispatchProps
+    >;
+    return enhance(WrappedMultiSelect);
+};
--- a/packages/react-vapor/src/components/validation/hoc/tests/WithNonEmptyValueMultiSelectValidationHOC.spec.tsx
+++ b/packages/react-vapor/src/components/validation/hoc/tests/WithNonEmptyValueMultiSelectValidationHOC.spec.tsx
@@ -1,0 +1,76 @@
+import {ShallowWrapper} from 'enzyme';
+import {mountWithStore, shallowWithStore} from 'enzyme-redux';
+import React from 'react';
+import {act} from 'react-dom/test-utils';
+import {composeMockStore, getStoreMock, withSelectedValues} from '../../../../utils/tests/TestUtils';
+import {IMultiSelectOwnProps, MultiSelectConnected} from '../../../select/MultiSelectConnected';
+import {ValidationActions} from '../../ValidationActions';
+import {ValidationTypes} from '../../ValidationTypes';
+import {
+    withNonEmptyMultiSelectHOC,
+    WithNonEmptyValueMultiSelectValidationProps,
+} from '../WithNonEmptyValueMultiSelectValidationHOC';
+
+const MultiSelectWithNonEmpty = withNonEmptyMultiSelectHOC(MultiSelectConnected);
+let multiSelectWrapper: ShallowWrapper<WithNonEmptyValueMultiSelectValidationProps & IMultiSelectOwnProps>;
+
+describe('MultiSelectWithNonEmpty', () => {
+    const defaultProps: IMultiSelectOwnProps & WithNonEmptyValueMultiSelectValidationProps = {
+        id: 'SOME_ID',
+        nonEmptyValidationMessage: 'ohno',
+        items: [],
+    };
+
+    let store: ReturnType<typeof getStoreMock>;
+
+    afterEach(() => {
+        store.clearActions();
+    });
+
+    const ONE_VALUE = '🐟';
+
+    it('should dispatch a setError action on mount if there are no values selected', () => {
+        store = composeMockStore(withSelectedValues(defaultProps.id));
+        const component = mountWithStore(<MultiSelectWithNonEmpty {...defaultProps} />, store);
+
+        act(() => {
+            component.mount();
+        });
+
+        expect(store.getActions()).toContain(
+            ValidationActions.setError(
+                defaultProps.id,
+                defaultProps.nonEmptyValidationMessage,
+                ValidationTypes.nonEmpty
+            )
+        );
+    });
+
+    it('should not dispatch a setError action on mount if there is a value selected', () => {
+        store = composeMockStore(withSelectedValues(defaultProps.id, ONE_VALUE));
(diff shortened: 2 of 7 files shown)


packages/react-vapor/src/components/validation/hoc/WithNonEmptyValueMultiSelectValidationHOC.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+const mapDispatchToProps = (dispatch: IDispatch) => ({
+    setError: (id: string, error: string) => dispatch(ValidationActions.setError(id, error, ValidationTypes.nonEmpty)),
+    clearError: (id: string) => dispatch(ValidationActions.clearError(id, ValidationTypes.nonEmpty)),
+});
+
+export interface IWithNonEmptyValueMultiSelectValidationProps {
+    validationMessage?: string;
+    resetErrorOnUnmount?: boolean;
+    isException?: boolean;
```
francoislg, line 23: What is the use case of `isException`? If set to true, the component won't do anything 🤔 Shouldn't it always be `false`, thus not really useful?

packages/react-vapor/src/components/validation/hoc/WithNonEmptyValueMultiSelectValidationHOC.tsx
```diff
@@ -0,0 +1,61 @@
...
+    selectedValues: SelectSelector.getMultiSelectSelectedValues,
+});
+
+const mapDispatchToProps = (dispatch: IDispatch) => ({
+    setError: (id: string, error: string) => dispatch(ValidationActions.setError(id, error, ValidationTypes.nonEmpty)),
+    clearError: (id: string) => dispatch(ValidationActions.clearError(id, ValidationTypes.nonEmpty)),
+});
+
+export interface IWithNonEmptyValueMultiSelectValidationProps {
+    validationMessage?: string;
```
francoislg, line 21: I think I would call it `nonEmptyErrorMessage` instead, so that it doesn't clash with other HOCs

packages/react-vapor/src/components/validation/hoc/WithNonEmptyValueMultiSelectValidationHOC.tsx
```diff
@@ -0,0 +1,61 @@
...
+            return () => {
+                resetErrorOnUnmount && clearError(props.id);
+            };
+        }, []);
+
+        React.useEffect(() => {
+            if (!isException) {
+                setError(props.id, !hasValuesSelected ? validationMessage : '');
+            }
+        }, [hasValuesSelected]);
```
francoislg, line 52: Not sure about the usefulness of `isException` (see previous comments), but if you keep it, it should also be included in the dependencies here

```suggestion
        }, [isException, hasValuesSelected]);
```

packages/react-vapor/src/components/validation/hoc/tests/WithNonEmptyValueMultiSelectValidationHOC.spec.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+    let store: ReturnType<typeof getStoreMock>;
+
+    afterEach(() => {
+        store.clearActions();
+    });
+
+    const ONE_VALUE = '🐟';
+
+    it('should through an error is there are no selection', () => {
```
francoislg, line 30: "should through an error is there" 🤔 two typos here 😄  

Maybe:

```suggestion
    it('should dispatch a setError action on mount if there are no values selected', () => {
```

You need to revise the other test names as well :P

francoislg:
You could also add the component in the Multi Select Examples ?

packages/react-vapor/src/components/validation/hoc/WithNonEmptyValueMultiSelectValidationHOC.tsx
cnguyen-coveo: I added this since we use the same form to create and update a catalog. This means when someone updates a catalog that has no sources, they would not be able to do so. In the admin-ui part the only case where they would not have the error would be if the catalog exists && there were no initial source selected

packages/react-vapor/src/components/validation/hoc/WithNonEmptyValueMultiSelectValidationHOC.tsx
cnguyen-coveo: The other validations i saw used this naming convention `IWithNonEmptyValueInputValidationProps` and `IMultiSelectWithDirtyOwnProps`. not sure if i should but the `NonEmpty` or the `MultiSelect` first?

FelixBlaisThon: I would remove the I before, we have few of them here and there that shouldn't be there, but our guidelines tells not to https://github.com/coveord/tsjs#interface

francoislg: 😕 if we use both `WithNonEmpty` and `WithDirty` (and we will), the `validationMessage` will need to be different for both cases, so we really need a different prop.

This is kind of a flaw in the current HOCs...

Maybe we should always include the purpose of the message from now on?

packages/react-vapor/src/components/validation/hoc/WithNonEmptyValueMultiSelectValidationHOC.tsx
francoislg: Still not keen on adding a boolean to disable the whole feature of the HOC. I feel like we should just not include the HOC if we don't need its validation.

I think it might be done by creating a simple wrapping component that either use the HOC or not. Something like the following (not accurate, but should explain the idea :P ) 

```
const MultiSelectWithValidation = nonEmpty(withDirty(MultiSelect));
const MultiSelectWithout = withDirty(MutliSelect);
const SourceIdsMultiSelect = ({ isEditing, ...props }) => isEditing ? <MultiSelectWithout {...props} /> : <MultiSelectWithValidation {...props} />;
```

packages/react-vapor/src/components/validation/hoc/tests/WithNonEmptyValueMultiSelectValidationHOC.spec.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+        expect(store.getActions()).not.toContain(
+            ValidationActions.setError(
+                defaultProps.id,
+                defaultProps.nonEmptyValidationMessage,
+                ValidationTypes.nonEmpty
+            )
+        );
+    });
+});
```
dlafreniere, line 63: missing some tests for clearError and unmount

packages/demo/src/components/examples/MultiSelectHOCExamples.tsx
```diff
@@ -0,0 +1,55 @@
...
+];
+
+const WithNonEmptyMultiSelect = withNonEmptyMultiSelectHOC(withValidationMessage(MultiSelectConnected));
+
+export interface IMultiSelectExamplesState {
+    first: IItemBoxProps[];
+    id: string;
+}
+
+export class MultiSelectHOCExamples extends React.Component<{}, IMultiSelectExamplesState> {
```
francoislg, line 32: 🤔 I don't think this example brings much to the table compared to what is already `MultiSelectExamples` 

Why did you think we needed another section?

Any opinion from the admin-ui team on this?

packages/demo/src/components/examples/MultiSelectHOCExamples.tsx
```diff
@@ -0,0 +1,55 @@
...
+    {displayValue: 'Four', value: '4'},
+    {displayValue: 'Five', value: '5'},
+    {displayValue: 'Six', value: '6'},
+    {displayValue: 'Seven', value: '7'},
+];
+
+const WithNonEmptyMultiSelect = withNonEmptyMultiSelectHOC(withValidationMessage(MultiSelectConnected));
+
+export interface IMultiSelectExamplesState {
+    first: IItemBoxProps[];
```
dlafreniere, line 28: first?

packages/demo/src/components/examples/MultiSelectHOCExamples.tsx
cnguyen-coveo: I used it to test out an example on the side (it was meant to be deleted 😅 )

packages/react-vapor/src/components/validation/hoc/WithNonEmptyValueMultiSelectValidationHOC.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+export const withNonEmptyMultiSelectHOC = <T extends IMultiSelectOwnProps>(Component: React.ComponentType<T>) => {
+    type NewOwnProps = T & WithNonEmptyValueMultiSelectValidationProps;
+    type StateProps = ReturnType<typeof mapStateToProps>;
+    type DispatchProps = ReturnType<typeof mapDispatchToProps>;
+    const WrappedMultiSelect: React.FunctionComponent<NewOwnProps & StateProps & DispatchProps> = ({
+        selectedValues,
+        setError,
+        clearError,
+        nonEmptyValidationMessage = 'Selection is empty and should not be empty',
```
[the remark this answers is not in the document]
cnguyen-coveo: Gino will get back to me after his meeting

packages/react-vapor/src/components/validation/hoc/WithNonEmptyValueMultiSelectValidationHOC.tsx
```diff
@@ -0,0 +1,57 @@
...
+        ...props
+    }) => {
+        const hasValuesSelected = selectedValues.length > 0;
+
+        React.useEffect(() => {
+            clearError(props.id);
+            return () => {
+                resetErrorOnUnmount && clearError(props.id);
+            };
+        }, []);
```
gdostie, line 44: The effect actually depends on `props.id` and `resetErrorOnUnmount`, therefore it should be added to the dependencies array.
```suggestion
        }, [props.id, resetErrorOnUnmount]);
```

packages/react-vapor/src/components/validation/hoc/WithNonEmptyValueMultiSelectValidationHOC.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+        React.useEffect(() => {
+            setError(props.id, !hasValuesSelected ? nonEmptyValidationMessage : '');
+        }, [hasValuesSelected]);
```
gdostie, line 48: We shouldn't lie about the dependencies :smile_cat: 
```suggestion
        }, [hasValuesSelected, props.id, nonEmptyValidationMessage]);
```

packages/react-vapor/src/components/validation/hoc/WithNonEmptyValueMultiSelectValidationHOC.tsx
```diff
@@ -0,0 +1,57 @@
...
+});
+
+const mapDispatchToProps = (dispatch: IDispatch) => ({
+    setError: (id: string, error: string) => dispatch(ValidationActions.setError(id, error, ValidationTypes.nonEmpty)),
+    clearError: (id: string) => dispatch(ValidationActions.clearError(id, ValidationTypes.nonEmpty)),
+});
+
+export interface WithNonEmptyValueMultiSelectValidationProps {
+    nonEmptyValidationMessage?: string;
+    resetErrorOnUnmount?: boolean;
```
gdostie, line 22: Should the name of this props (`resetErrorOnUnmount`) reflect the fact that it will reset only the `nonEmptyValidationMessage`? In the case where we would create a component with multiple validations, the name of this prop would lead towards confusion as to what error it will actually reset.

lerna.json
```diff
@@ -6,22 +6,15 @@
     "command": {
         "publish": {
             "conventionalCommits": true,
-            "allowBranch": [
```
francoislg, line 9: This should probably be reverted

PR #1595, opened 2020-06-16
feat(multiselect): validate initial values hoc

francoislg:
[COM-450]

### Proposed Changes

The purpose of this one is to show the user when an initial value is not present in the items.

For instance, our "Associated Sources" multi-select could contain an initial value for a now-deleted source. This would previously be completely ignored, so we wanted to provide some feedback on that.

Some details:
* The message is a "warning" so that saving is not prevented.
* I also hooked the `selected` state of the items to the initial value, just to prevent some potential mismatch between `items` and `initialValues`.

image

None

### Acceptance Criteria

[COM-450]: https://coveord.atlassian.net/browse/COM-450


packages/react-vapor/src/components/validation/hoc/tests/WithInitialValuesMultiSelectHOC.spec.tsx
```diff
@@ -0,0 +1,113 @@
...
+                initialValues={[ONE_VALUE]}
+            />,
+            emptyStore
+        ).dive();
+
+        const items = component.find(MultiSelectConnected).prop('items');
+
+        expect(items[0].selected).toBe(true);
+    });
+});
```
dlafreniere, line 113: test unmount logic?

packages/react-vapor/src/components/validation/hoc/WithInitialValuesMultiSelectHOC.tsx
```diff
@@ -0,0 +1,63 @@
...
+const mapDispatchToProps = (dispatch: IDispatch, {id}: IMultiSelectOwnProps) => ({
+    setWarning: (warning: string) =>
+        dispatch(ValidationActions.setWarning(id, warning, ValidationTypes.wrongInitialValue)),
+    clearWarning: () => dispatch(ValidationActions.clearWarning(id, ValidationTypes.wrongInitialValue)),
+});
+
+export type IMultiSelectWithInitialValuesOwnProps = {
+    initialValues: string[];
+    invalidInitialValuesMessage?: (invalidValues: string[]) => string;
+    resetWarningOnUnmount?: boolean;
```
gdostie, line 19: Should the name of this props (`resetWarningOnUnmount`) reflect the fact that it will reset only the `invalidInitialValuesMessage`? In the case where we would create a component with multiple validations, the name of this prop would lead towards confusion as to what warning it will actually reset.

packages/react-vapor/src/components/validation/hoc/WithInitialValuesMultiSelectHOC.tsx
```diff
@@ -0,0 +1,63 @@
...
+    const WrappedMultiSelect: React.FunctionComponent<T & IMultiSelectWithInitialValuesOwnProps & DispatchProps> = ({
+        setWarning,
+        clearWarning,
+        invalidInitialValuesMessage = defaultInvalidInitialValuesMessageGenerator,
+        ...props
+    }) => {
+        const {initialValues = [] as string[], items = []} = props;
+
+        const notFoundInitialValues =
+            initialValues &&
```
gdostie, line 36: `initialValues` will always be truthy because it is set to an empty array by default on line 33. Perhaps it would be better to check the length?
```suggestion
            initialValues?.length &&
```

francoislg:
Can anyone review and/or merge this?

PR #1601, opened 2020-06-18
fix(middleslider): accept min greater than 0 + use more intuitive marks

gdostie:
The middle slider didn't accept a min greater than 0 which is a use case we must support. By fixing
this, it made marks more intuitive but created a breaking change. Since there was already a breaking change, I took the liberty of making other changes that the component easier to use.

BREAKING CHANGES: 

- Marks keys must now be actual values instead of ranging from 0 to 100.
- `enabled` was removed in favor of `disabled` props. 
- The tooltip overlay style was added by default to all handles with tooltips.


packages/react-vapor/tsconfig.json
```diff
@@ -10,7 +10,8 @@
         "removeComments": true,
         "noUnusedLocals": true,
         "lib": ["es2015", "dom"],
-        "allowSyntheticDefaultImports": true
```
j-mcgregor, line 13: How do you know that this won't introduce breaking changes? Or is the fact that because you've introduced breaking changes in the slider anyway means you're more comfortable doing this?

packages/react-vapor/src/components/slider/MiddleSlider.tsx
```diff
@@ -100,59 +78,66 @@ const MiddleSliderDisconnected: React.FunctionComponent<MiddleSliderOwnProps &
...
-                lowRange === crossingPoint ? jumpValueFromHighToLowRange(value) : jumpValueFromLowToHighRange(value);
+                leftHandlePosition === crossingPoint
+                    ? jumpValueFromHighToLowRange(value)
+                    : jumpValueFromLowToHighRange(value);
                 break;
             default:
-                setLowRange(value[0]);
-                setHighRange(value[1]);
+                setLeftHandlePosition(value[0]);
+                setRightHandlePosition(value[1]);
```
j-mcgregor, line 87: 👍  for renaming. makes sense

packages/react-vapor/src/components/slider/SliderUtils.ts
```diff
@@ -15,55 +17,48 @@ export const getValuesPositionOnRange = (value: number[], crossingPoint: number)
...
+    leftHandlePosition: number,
+    rightHandlePosition: number,
     crossingPoint: number
 ): number => {
-    if (lowRange < crossingPoint) {
-        return Math.round(((crossingPoint - lowRange) * min) / crossingPoint);
-    } else if (highRange > crossingPoint) {
-        return Math.round(((highRange - crossingPoint) * max) / (100 - crossingPoint));
+    if (leftHandlePosition < crossingPoint) {
+        return leftHandlePosition;
```
j-mcgregor, line 35: Why remove the Math.round out of interest?

gdostie: Its not about the `Math.round` as much as removing all the calculations. By setting a `min` and a `max` directly on the `Range` component, it scales itself so don't have any scaling calculations to do anymore.

packages/react-vapor/tsconfig.json
gdostie: I didn't remove the line, only added a comma at the end of the line. GitHub shows the diff in a weird way :stuck_out_tongue:

packages/react-vapor/src/components/slider/SliderUtils.ts
j-mcgregor: ohh that makes more sense

packages/react-vapor/tsconfig.json
j-mcgregor: I meant the next line, sorry. Put the comment in the wrong place

gdostie: Oh :see_no_evil: `strictBindCallApply` is only for a stricker type checking inside the package. It shouldn't have any impact on the distributed code or types.
image
https://www.typescriptlang.org/docs/handbook/compiler-options.html

PR #1609, opened 2020-06-25
fix(sub navigation): fixed the link prop and small refactor

dmgauthier:
### Proposed Changes

The onclick html property was in conflict with the link property. in order to access the link, the function passed to onlclick must return true. 

I also refactored the component to get rid of the deprecated class component lifeCycle methods.

Fixed the broken uts.

No breaking change intented

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/subNavigation/SubNavigation.tsx
+++ b/packages/react-vapor/src/components/subNavigation/SubNavigation.tsx
@@ -3,7 +3,7 @@ import * as React from 'react';
 import {keys} from 'ts-transformer-keys';
 import {map, omit} from 'underscore';
 
-export interface ISubNavigationOwnProps extends React.ClassAttributes<SubNavigation> {
+export interface ISubNavigationOwnProps {
     id?: string;
     items: ISubNavigationItem[];
     defaultSelected?: string;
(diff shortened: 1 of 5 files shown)


packages/react-vapor/src/components/subNavigation/SubNavigation.tsx
```diff
@@ -33,39 +33,45 @@ export interface ISubNavigationProps
...
-        return (
-            <nav {...navProps} className={classNames('sub-navigation', navProps.className)}>
-                <ul className="sub-navigation-items">{items}</ul>
-            </nav>
-        );
-    }
+    const navItems = map(items, ({id, link, label, disabled}: ISubNavigationItem) => (
+        <li
+            key={id}
+            className={classNames('sub-navigation-item', {'mod-selected': id === selected ?? defaultSelected})}
```
rosalie-liu, line 60: hmm a little confused, the type of `defaultSelected` is string, so we should probably use `{'mod-selected': id === selected || !!defaultSelected}` here

packages/react-vapor/src/components/subNavigation/SubNavigation.tsx
```diff
@@ -33,39 +33,45 @@ export interface ISubNavigationProps
...
+        >
+            <a
+                href={link || '#'}
```
rosalie-liu, line 63: `href={link ?? '#'}` maybe? not sure

rosalie-liu:
In the demo link, when I click on a menu item in the sub navigation, it redirects to https://vapor.cloud.coveo.com/fix-adui-5664-sub-nav-link-props-is-broken/index.html#/components/ActionableItem
image

dmgauthier:
> In the demo link, when I click on a menu item in the sub navigation, it redirects to https://vapor.cloud.coveo.com/fix-adui-5664-sub-nav-link-props-is-broken/index.html#/components/ActionableItem

This is the intended behavior of the component with the `href={link ?? '#'}`.  This was coded like that at first but it never worked. I don't have any objection in changing that if you think that causes problem.

packages/react-vapor/src/components/subNavigation/SubNavigation.tsx
dmgauthier: `href={link}` is preferable in order to not redirect if no link is defined

packages/react-vapor/src/components/subNavigation/SubNavigation.tsx
dmgauthier: Your solution dont work while`id === selected ?? defaultSelected` does work. But If you are confused I changed for the more verbose `id === selected || id === defaultSelected`

PR #1616, opened 2020-06-29
feat(radio-button): add an optional wrapper for radio buttons

sallainCoveo:
### Proposed Changes

A radio button can now be wrapped inside a div container or a fragment container. Which means that an element could be in the same node level in the DOM as the radio button and a CSS class could affect both components. For instance the following image display a radio button that share the same CSS class with an image, and it's consider inside the same row. 

partOfTheSameContainer

### Acceptance Criteria

### DIFF
--- a/packages/demo/src/components/examples/RadioExamples.tsx
+++ b/packages/demo/src/components/examples/RadioExamples.tsx
@@ -42,20 +42,22 @@ const radioSelectConnectedId = 'radioselectconnected';
 
 // start-print
 
-const blueRadioButtonProps = {
-    id: 'blue',
+const radioButtonProps = {
     name: 'rankingResult',
-    value: 'blue',
     outerContainerClass: 'modal-radio-button',
     outerElementInContainer: <img src="https://via.placeholder.com/150x100" />,
 };
 
+const blueRadioButtonProps = {
+    id: 'blue',
+    value: 'blue',
+    ...radioButtonProps,
+};
+
 const redRadioButtonProps = {
     id: 'red',
-    name: 'rankingResult',
     value: 'red',
-    outerContainerClass: 'modal-radio-button',
-    outerElementInContainer: <img src="https://via.placeholder.com/150x100" />,
+    ...radioButtonProps,
 };
 
 const RadioSelectExample: React.FunctionComponent = () => (
@@ -122,7 +124,7 @@ const RadioSelectWithDivWrapAroundRadioButtonExample: React.FunctionComponent = 
         <RadioSelectConnected id="addRankingResult" valueOnMount={'blue'}>
             <Radio {...blueRadioButtonProps}>
                 <Label>
-                    <span style={{fontWeight: 'bold'}}>{'Blue color'}</span>
+                    <span className="bold">{'Blue color'}</span>
                 </Label>
                 <InputDescription>
                     <div style={{...paragraphStyle}}>{'Blue is the best color.'}</div>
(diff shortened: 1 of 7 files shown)


packages/demo/src/components/examples/RadioExamples.tsx
```diff
@@ -25,15 +26,38 @@ export const RadioButtonExamples: ExampleComponent = () => (
...
+};
+
+const redRadioButtonProps = {
+    id: 'red',
+    name: 'rankingResult',
+    value: 'red',
+    outerContainerClass: 'modal-radio-button',
+    outerElementInContainer: <img src="https://via.placeholder.com/150x100" />,
+};
+
```
dmgauthier, line 60:
```
const radioButtonProps = {
    name: 'rankingResult',
    outerContainerClass: 'modal-radio-button',
    outerElementInContainer: <img src="https://via.placeholder.com/150x100" />,
}

const redRadioButtonProps = {
    id: 'red',
    value: 'red',
   ...radioButtonProps,
}

const blueRadioButtonProps = {
    id: 'blue',
    value: 'blue',
   ...radioButtonProps,
}

File: packages/demo/src/components/examples/RadioExamples.tsx [hunk 2]
```diff
@@ -92,3 +116,26 @@ const RadioSelectDisabledExample: React.FunctionComponent = () => (
         </LabeledInput>
     </Section>
 );
+
+const RadioSelectWithDivWrapAroundRadioButtonExample: React.FunctionComponent = () => (
+    <Section level={3} title="A radio select with a wrapped radio button.">
+        <RadioSelectConnected id="addRankingResult" valueOnMount={'blue'}>
+            <Radio {...blueRadioButtonProps}>
+                <Label>
+                    <span style={{fontWeight: 'bold'}}>{'Blue color'}</span>
```
dmgauthier, line 125: can you use `classname="bold"` instead of the inline styling ?

packages/react-vapor/src/components/radio/Radio.tsx
```diff
@@ -2,15 +2,39 @@ import classNames from 'classnames';
...
-export class Radio extends Input {
-    static defaultProps: Partial<IInputProps> = {
+export interface RadioOwnProps {
+    outerContainerClass?: string;
+    outerElementInContainer?: React.ReactNode;
+}
+
+export interface RadioProps extends RadioOwnProps, IInputProps {}
+
+export class Radio extends React.PureComponent<RadioProps> {
```
dmgauthier, line 12: I know you probably took example on other components for your `pureComponent` here but I think that would be cleaner and faster with a function component instead than a class component. And since it's stateless, you will not have any difficulty doing so.

You can do something like that 
```
export const Radio: React.FunctionComponent<RadioProps> = (props) => {
[... then here you will have consts instead of class static and private elements and instead of a render() method you are going to return the jsx directly]
}
```

packages/demo/src/components/examples/RadioExamples.tsx
```diff
@@ -25,15 +26,40 @@ export const RadioButtonExamples: ExampleComponent = () => (
     <Section title="Radio set">
         <RadioSelectExample />
         <RadioSelectDisabledExample />
+        <RadioSelectWithDivWrapAroundRadioButtonExample />
```
j-mcgregor, line 29: Seems quite verbose - maybe something like `<RadioSelectWrappedExample />`

packages/demo/src/components/examples/RadioExamples.tsx
```diff
@@ -25,15 +26,40 @@ export const RadioButtonExamples: ExampleComponent = () => (
...
+    marginLeft: 36,
+    marginTop: 10,
+    marginRight: 36,
+};
+
 RadioButtonExamples.description = 'Radio Buttons allow for the selection of a single option among a set of options.';
 
 const radioSelectConnectedId = 'radioselectconnected';
 
 // start-print
```
j-mcgregor, line 43: is this still necessary?

dmgauthier: This is used to scope the demo code.

PR #1625, opened 2020-07-10
feat(jsoneditor): create jsoneditorconnected component

rosalie-liu:
### Proposed Changes

- Converted JSONEditor to FunctionComponent
- Created JSONEditorConnected, updated store and reducers
- Added actions, reducers and selectors to JSONEditorConnected

None

### Acceptance Criteria


packages/react-vapor/src/components/editor/EditorConstants.ts
```diff
@@ -10,3 +10,5 @@ export const CodeMirrorModes = {
     Python: 'python',
     Text: 'text/plain',
 };
+
+export const DEFAULT_JSON_ERROR_MESSAGE: string = 'The JSON configuration is syntactically invalid.';
```
j-mcgregor, line 14: is the `string` necessary? I thought it was implied and immutable since it's a const?

packages/react-vapor/src/components/editor/JSONEditor.tsx
```diff
@@ -5,73 +5,76 @@ import * as React from 'react';
...
     onChange?: (json: string, inError: boolean) => void;
     errorMessage?: string;
+    containerClasses?: string[];
+    className?: string;
+    ref?: React.Ref<any>;
 }
 
-export interface IJSONEditorState {
-    isInError: boolean;
+export interface IJSONEditorStateProps {
```
j-mcgregor, line 22: Are we not refraining from prefixing interfaces with `I`? I know it breaks the pattern in this case, so not sure what's best practice 🤔

packages/react-vapor/src/components/editor/EditorConstants.ts
rosalie-liu: Nice catch! I simply moved it from another file :see_no_evil:

packages/react-vapor/src/components/editor/JSONEditor.tsx
rosalie-liu: I personally have no preference, it seems that using `I` for the naming of interfaces is the pattern existed in react-vapor :thinking:

dmgauthier: [Link](https://til.hashrocket.com/posts/cfnlyzxcrc-do-not-prefix-typescript-interface-names) I think we should just drop the I

packages/react-vapor/src/components/editor/JSONEditor.tsx
```diff
@@ -5,73 +5,76 @@ import * as React from 'react';
...
-                    value={this.props.value}
-                    onChange={(json: string) => this.handleChange(json)}
-                    mode={CodeMirrorModes.JSON}
-                    readOnly={this.props.readOnly}
-                />
-                {this.getValidationDetails()}
-            </div>
-        );
-    }
+        return onUnmount;
```
dmgauthier, line 50: Can a useEffect hook can return `undefined` as an escape function ? I wonder since your `onUnmount` prop is optionnal.

packages/react-vapor/src/components/editor/JSONEditor.tsx
```diff
@@ -5,73 +5,76 @@ import * as React from 'react';
...
+    }, []);
```
dmgauthier, line 51: `[onMount, onUnmount]` ?

packages/react-vapor/src/components/editor/JSONEditorUtils.ts
```diff
@@ -0,0 +1,12 @@
+const validateValue = (jsonValue: string) => {
+    try {
+        JSON.parse(jsonValue);
+        return true;
+    } catch (error) {
```
dmgauthier, line 5: is it usefull to pass the `error` if you don't use it ?

packages/react-vapor/src/components/editor/JSONEditor.tsx
rosalie-liu: Yes, this is about how `return` works. If `onUnmount` is `undefined`, then `return () => undefined();`, `return undefined;` or `return;`, they are the same thing.

packages/react-vapor/src/components/editor/JSONEditor.tsx
rosalie-liu: I think that it works correctly here without dependencies, since we only need to call `onMount` and `onUnmount` once when rendering and destroying. If these props change, then it'll be another new component. Also, we've already implemented `onMount` and `onUnmount` in this way several times, e.g. https://github.com/coveo/react-vapor/blob/master/packages/react-vapor/src/components/textarea/TextArea.tsx#L78

packages/react-vapor/src/components/editor/JSONEditorUtils.ts
rosalie-liu: Oops, I'll remove it!

PR #1627, opened 2020-07-13
test(tabselectors.spec.tsx): change getTabSelected

FelixBlaisThon:
### Proposed Changes

the selector now returns the selected tab (or false) instead of a boolea isSelected

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/tab/TabSelectors.ts
+++ b/packages/react-vapor/src/components/tab/TabSelectors.ts
@@ -9,9 +9,9 @@ const getTabGroup = (state: IReactVaporState, ownProps: ITabPaneOwnProps) => {
     return _.findWhere(state.tabs, {id: ownProps.groupId ?? DEFAULT_GROUP_ID});
 };
 
-const getTabSelected = () =>
-    createSelector(getTabGroup, (tabGroup: ITabGroupState) => _.findWhere(tabGroup.tabs, {isSelected: true}) ?? false);
+const getSelectedTab = () =>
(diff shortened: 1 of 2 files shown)


packages/react-vapor/src/components/tab/TabSelectors.ts
```diff
@@ -9,11 +9,8 @@ const getTabGroup = (state: IReactVaporState, ownProps: ITabPaneOwnProps) => {
     return _.findWhere(state.tabs, {id: ownProps.groupId ?? DEFAULT_GROUP_ID});
 };
 
-const getTabSelected = (tabId: string) =>
-    createSelector(
-        getTabGroup,
-        (tabGroup: ITabGroupState): boolean => !!_.findWhere(tabGroup.tabs, {id: tabId})?.isSelected
-    );
+const getTabSelected = () =>
```
gdostie, line 12:
```suggestion
const getSelectedTab = () =>
```

packages/react-vapor/src/components/tab/TabSelectors.ts
```diff
@@ -9,11 +9,8 @@ const getTabGroup = (state: IReactVaporState, ownProps: ITabPaneOwnProps) => {
...
+    createSelector(getTabGroup, (tabGroup: ITabGroupState) => _.findWhere(tabGroup.tabs, {isSelected: true}) ?? false);
```
gdostie, line 13: Since now we return an object instead of previously a boolean, I think it would be best to return `undefined` when no tab is selected. I will be easier to get the typing right
```suggestion
    createSelector(getTabGroup, (tabGroup: ITabGroupState) => _.findWhere(tabGroup.tabs, {isSelected: true}));
```

PR #1631, opened 2020-07-14
fix(table hoc): missing filterBlankslate in the ownProps interface

dmgauthier:
### Proposed Changes

**What is going on:**

- The filterBlankslate was missing in the table hoc interface

none

### Acceptance Criteria


gdostie:
If you want to use a TableHOC that has the table with filter props, you can use something like that:
```ts
const Table: ReturnType<ReturnType<typeof tableWithFilter>> = _.compose(
    tableWithFilter(),
    tableWithActions()
)(TableHOC);
```

PR #1632, opened 2020-07-15
fix: linewrapping for react-vapor code editor

aben-coveo:
### Proposed Changes

Exposed line wrapping options for code editor and json editor.

Should not break anything since its a not required exposed prop.

### Acceptance Criteria

-   The proposed changes are covered by unit tests
-   No breaking changes

### DIFF
--- a/packages/react-vapor/src/components/editor/CodeEditor.tsx
+++ b/packages/react-vapor/src/components/editor/CodeEditor.tsx
@@ -15,6 +15,7 @@ import {CodeMirrorGutters} from './EditorConstants';
 export interface ICodeEditorProps {
     value: string;
     mode: any;
+    lineWrapping?: boolean;
     readOnly?: boolean;
     onChange?: (code: string) => void;
     onMount?: (codemirror: ReactCodeMirror.Controlled) => void;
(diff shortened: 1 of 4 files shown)


packages/react-vapor/src/components/editor/CodeEditor.tsx
```diff
@@ -81,6 +82,7 @@ export class CodeEditor extends React.Component<ICodeEditorProps, CodeEditorStat
                 onChange={(editor, data, value: string) => this.props.onChange?.(value)}
                 options={{
                     ...CodeEditor.defaultOptions,
+                    lineWrapping: this.props.lineWrapping,
```
gdostie, line 85: Wasn't already available through `this.props.options`? 🤔

aben-coveo: not for the json editor

gdostie: Oh I see, then perhaps we could have added the options prop to the JSONEditor instead. That way it would've stayed uniform and scaled better if we need other options than line wrapping.

aben-coveo: yeah i could do that, but I didn't know if we wanted to expose the whole editor configuration. Otherwise I can just expose linewrapping on jsoneditor and then pass in options object with linewrapping flag set to whatever user specifies to code editor. I was trying to make a quick fix for another pr in admin-ui :sweat_smile:

gdostie: Its fine as it is 👍  I was just thinking "out loud" 🙂

PR #1638, opened 2020-07-22
Add both result ranking icons

sallainCoveo:
### Proposed Changes

Addition of the featured result and the ranking rules icons. These will be used in the result ranking modal.

Please note that these icons are quite large. If you want to take a better look on these icons, you will have to go directly in the new .svg file implemented in this branch.

### Acceptance Criteria


rosalie-liu:
It looks like this in the demo, is it expected? 
image
Also we get 404 when clicking on it https://github.com/coveo/react-vapor/blob/master/packages/vapor/resources/icons/svg/coveo-search-ui-filetypes/featured-result.svg

sallainCoveo:
@rosalie-liu, 
I think it's normal that these icons are displayed this way, because they are larger then most of the other icons (I supposed that there's is a max-width). 

As for the 404 files: I'm not sure why. Correct me if I'm wrong, but it seems that the file is opened from the master branch (if I look to the url), but this file doesn't exist in the latter (which might explain why these pages are unfound...?)

PR #1639, opened 2020-07-28
feat(labeledinput): change default tooltip placement

rosalie-liu:
### Proposed Changes

The tooltip placement was set to `top`, if we want to display information/helper in the label title. Since we use this component mostly in forms, it could be changed to `right` according to current design.

None

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/input/LabeledInput.tsx
+++ b/packages/react-vapor/src/components/input/LabeledInput.tsx
@@ -12,7 +12,6 @@ export interface ILabeledInputProps {
     helpText?: React.ReactNode;
     headerClassName?: string;
     optionalInformation?: string;
-    informationPlacement?: TooltipPlacement;
 }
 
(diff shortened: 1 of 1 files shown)


FelixBlaisThon:
do you have any example in vapor? i don't know what this prop would add more than placement Top/Left/Right/bottom

francoislg:
The nice thing about `information` is that it completely obscures the implementation (tooltip) and if we ever want to, instead, show the `information` prop as (for instance) a full-width panel, then the placement does not make sense.

So my question is, should the default placement be changed instead? Is there *really* a use case for *each* placement?

rosalie-liu:
> do you have any example in vapor? i don't know what this prop would add more than placement Top/Left/Right/bottom
@FelixBlaisThon There is an example in `Section`, _Search panel -> Main button -> Location on screen_

PR #1641, opened 2020-07-31
feat(multi-values-input): add dataLimit and placeholders optional props

sallainCoveo:
### Proposed Changes

Add an optional prop "dataLimit", which indicates how many inputs that could be added and edited in a MultiValuesInput component. Also, addition of optional placeholders props. 

### Acceptance Criteria

-   [X ] The proposed changes are covered by unit tests

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -6,7 +6,9 @@
     "command": {
         "publish": {
             "conventionalCommits": true,
-            "allowBranch": ["master"]
+            "allowBranch": [
+                "master"
+            ]
         },
         "version": {
             "conventionalCommits": true,
@@ -14,7 +16,12 @@
         },
         "bootstrap": {
             "hoist": true,
-            "nohoist": ["unidiff", "query-string", "strict-uri-encode", "split-on-first"]
+            "nohoist": [
+                "unidiff",
+                "query-string",
+                "strict-uri-encode",
+                "split-on-first"
+            ]
         }
     }
 }
--- a/lerna.json
+++ b/lerna.json
@@ -6,9 +6,7 @@
     "command": {
         "publish": {
             "conventionalCommits": true,
-            "allowBranch": [
-                "master"
-            ]
+            "allowBranch": ["master"]
         },
         "version": {
             "conventionalCommits": true,
@@ -16,12 +14,7 @@
         },
         "bootstrap": {
             "hoist": true,
-            "nohoist": [
-                "unidiff",
-                "query-string",
-                "strict-uri-encode",
-                "split-on-first"
-            ]
+            "nohoist": ["unidiff", "query-string", "strict-uri-encode", "split-on-first"]
         }
     }
 }
--- a/packages/react-vapor/src/components/multiValueInputs/MultiValuesInput.tsx
+++ b/packages/react-vapor/src/components/multiValueInputs/MultiValuesInput.tsx
@@ -18,7 +18,7 @@ const MultilineBoxWithRemoveButton = _.compose(
             <div key={`${data[index].id}`} className="flex">
                 <div className="flex-auto">{child}</div>
                 {getRemoveButton({
-                    classes: ['mod-no-border', 'bg-transparent', 'align-end'],
+                    classes: ['mod-no-border', 'bg-transparent', 'align-end', 'mt2'],
                 })}
             </div>
         ),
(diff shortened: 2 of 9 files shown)


packages/react-vapor/src/components/multiValueInputs/MultiValuesInput.tsx
```diff
@@ -56,6 +69,10 @@ export const MultiValuesInput: React.FunctionComponent<MultiValuesInputProps> =
                                 parentProps.removeBox(cData.id);
                             }
                         }}
+                        placeholder={index >= limit ? reachedLimitPlaceholder : placeholder}
+                        disabled={index >= limit}
+                        innerInputClasses={index >= limit && 'mod-no-border py1 input-wider-text-box'}
+                        disabledTooltip={index >= limit ? "this input can't be edited." : undefined}
```
rosalie-liu, line 75: `"This input can't be edited."` Should we confirm this string with someone from doc team, or allow passing an optional prop for `disabledTooltip`?

packages/react-vapor/src/components/multiValueInputs/MultiValuesInput.tsx
```diff
@@ -29,15 +29,28 @@ export interface MultiValuesInputProps {
...
+    id,
+    data,
+    inputProps,
+    dataLimit,
+    placeholder,
+    reachedLimitPlaceholder,
+}) => {
+    const limit = !!dataLimit ? dataLimit : Infinity;
+    placeholder = !!placeholder ? placeholder : '';
+    reachedLimitPlaceholder = !!reachedLimitPlaceholder ? reachedLimitPlaceholder : '';
```
rosalie-liu, line 47: `reachedLimitPlaceholder = reachedLimitPlaceholder ?? '';` Does this work?

packages/react-vapor/src/components/multiValueInputs/MultiValuesInput.tsx
```diff
@@ -56,6 +69,10 @@ export const MultiValuesInput: React.FunctionComponent<MultiValuesInputProps> =
                                 parentProps.removeBox(cData.id);
                             }
                         }}
+                        placeholder={index >= limit ? reachedLimitPlaceholder : placeholder}
```
rosalie-liu, line 72: Should we add someone from UX to review the style? It seems that placeholder has a large space on the top and left :thinking: 
image

packages/react-vapor/src/components/multiValueInputs/MultiValuesInput.tsx
sallainCoveo: Good point, I think it would be interesting to allows this feature to be customizable  .

packages/react-vapor/src/components/multiValueInputs/MultiValuesInput.tsx
sallainCoveo: Yes! I'm not even sure that this placeholder would be relevant, so an UX opinion would help answer these questions. Summoning @oliviertaillon

FelixBlaisThon:
This looks pretty good, good job!

packages/react-vapor/src/components/multiValueInputs/MultiValuesInput.tsx
oliviertaillon: Can you make the styling the same as the one when there is an error on validation?

dmgauthier: I tend to agree with Sylvie here.  Do we really want a warning as soon as the user has begun to fill the last available field?  All in All I would put this behavior optionnal, the warning appearing only if it is passed in the props.

lerna.json
```diff
@@ -6,22 +6,15 @@
...
         },
         "bootstrap": {
             "hoist": true,
-            "nohoist": [
-                "unidiff",
-                "query-string",
-                "strict-uri-encode",
-                "split-on-first"
-            ]
+            "nohoist": ["unidiff", "query-string", "strict-uri-encode", "split-on-first"]
```
dmgauthier, line 17: Those are wanted changes ?

packages/react-vapor/src/components/multiValueInputs/MultiValuesInput.tsx
sallainCoveo: Before I commit my changes, I would like to confirm that the new style of the placeholder is what you expected (which is looks like the invalid-message of the input). All things consider, if we all agree with this configuration, I suggest that this props is still an optional one (but with this new style). 
exampleOfReachedPlaceholder

oliviertaillon: You can remove the grey background but after this I confirm the styling ✅

lerna.json
sallainCoveo: Good question, I think that my IDE tends to format this file that way (and if I use prettier it does the same thing). When you change and push this file, does it usually look differently?

packages/react-vapor/src/components/multiValueInputs/MultiValuesInput.tsx
```diff
@@ -29,35 +29,55 @@ export interface MultiValuesInputProps {
...
+                                parentProps.addNewBox();
+                            }
+                        }}
+                        onBlur={(value: string) => {
+                            inputProps?.onBlur?.(value);
+                            if (value === '' && !cData.isLast) {
+                                parentProps.removeBox(cData.id);
+                            }
+                        }}
+                        placeholder={index >= limit ? reachedLimitPlaceholder : inputProps?.placeholder}
```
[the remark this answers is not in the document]
sallainCoveo: If one or neither of these props are set, then they will be consider as 'undefined', which lead to an non-existing placeholder attribute (it seems that it doesn't generate an error in the console either, so I think it's fine). However, I didn't realize, but if the reachedLimitPlaceholder is empty, the tooltip that indicate that the input cannot be edited still appears...Which is strange, so I'll add a condition that won't allow this

packages/react-vapor/src/components/multiValueInputs/MultiValuesInput.tsx
```diff
@@ -29,35 +29,55 @@ export interface MultiValuesInputProps {
...
-);
+export const MultiValuesInput: React.FunctionComponent<MultiValuesInputProps> = ({
+    id,
+    data,
+    inputProps,
+    dataLimit,
+    reachedLimitPlaceholder,
+    disabledTooltipTitle,
+}) => {
+    const limit = !!dataLimit ? dataLimit : Infinity;
```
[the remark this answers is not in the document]
sallainCoveo: Yeah, good idea! On it!

lerna.json
dmgauthier: Personnaly I would not commit this file in the PR since it is clearly out of scope :man_shrugging:

PR #1644, opened 2020-08-10
improvement: added classes prop to Limit component

j-mcgregor:
### Proposed Changes

As part of https://coveord.atlassian.net/browse/ADUI-5749 and the larger https://coveord.atlassian.net/browse/ADUI-5819

In order to have the same look I need to apply classes to the outer container of the Limit component

none

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/limit/Limit.tsx
+++ b/packages/react-vapor/src/components/limit/Limit.tsx
@@ -13,6 +13,7 @@ export interface LimitOwnProps {
     id?: string;
     title: string;
     limitLabel?: string;
+    classes?: string;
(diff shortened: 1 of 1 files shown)


packages/react-vapor/src/components/limit/Limit.tsx
```diff
@@ -13,6 +13,7 @@ export interface LimitOwnProps {
     id?: string;
     title: string;
     limitLabel?: string;
+    classes?: string;
```
GermainBergeron, line 16: Can you name the property `className` instead? I think it would be a better fit with React

PR #1652, opened 2020-08-13
fix(infinite-loop): use effect bad watch

pochretien:
Update every variable to watch on useEffect. we cant watch on a function from mapDispatchToProps redefine every time.

### Acceptance Criteria


packages/react-vapor/src/components/select/hoc/SelectWithPredicate.tsx
```diff
@@ -46,7 +46,7 @@ export const selectWithPredicate = <P extends Omit<ISelectOwnProps, 'button'> &
         const {onUpdate, predicate} = props;
         React.useEffect(() => {
             onUpdate?.();
-        }, [predicate, onUpdate]);
+        }, [predicate]);
```
j-mcgregor, line 49: are you removing all values from the initiator array that may be undefined?

pochretien: every function from mapDispatchToProps

PR #1655, opened 2020-08-17
feat(multi-select): supporting the selectDisplayValue prop

kdang-coveo:
Adding the support for the selectDisplayValue prop for the multi-select component, similar to the
SingleSelect component

Test for this feature seems to be completed in `MultiSelectConnected.spec.tsx` lines 87-92 with `selectedDisplayValue`

### Proposed Changes

The value to show `selectedDisplayValue` if it is passed in as a prop. This will allow flexibility for users who want to display a different value when an item is selected versus when the item is shown on the dropdown

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/select/MultiSelectConnected.tsx
+++ b/packages/react-vapor/src/components/select/MultiSelectConnected.tsx
@@ -92,7 +92,8 @@ class MultiSelect extends React.PureComponent<IMultiSelectProps> {
     }
 
     private renderOption(item: IItemBoxProps): JSX.Element {
-        const displayValue = item.displayValue || item.value;
+        const displayValue = item.selectedDisplayValue || item.displayValue || item.value;
+
         return (
             <SelectedOption
                 label={displayValue}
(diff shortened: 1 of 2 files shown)


packages/react-vapor/src/components/select/MultiSelectConnected.tsx
```diff
@@ -110,7 +111,7 @@ class MultiSelect extends React.PureComponent<IMultiSelectProps> {
             <div className="flex flex-row flex-center sortable-selected-option" key={item.value}>
                 <span className="mr1 text-medium-grey">{index + 1}</span>
                 <DraggableSelectedOption
-                    label={item.displayValue || item.value}
+                    label={item.selectedDisplayValue || item.displayValue || item.value}
```
sallainCoveo, line 114: It might not be relevant in that case, but it could be interesting to use nullish coalescing operator instead of the "OR" logical operator. It would look like this `item.selectedDisplayValue ?? item.displayValue ?? item.value`
Here's some [info](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) explaining this suggestion.

kdang-coveo: Very interesting, thank you for the suggestion!

PR #1658, opened 2020-08-19
feat(multi-select): allowing flexibility in tooltip value in multiselect

kdang-coveo:
This change adds a new prop 'hoverTooltipValue' in IItemBoxProps interface to allow users to change
the tooltip value without having the component to rely on the displayValue or value for selected
item. The DraggableSelectedOption has been refactored to accommodate for the new feature with a new
prop.

### Proposed Changes

- Add hoverTooltipValue (need feedback on the naming of this prop to see if it made sense) to the IItemBoxProps
- The SelectedOption component label prop takes the hoverTooltipValue first and if its undefined will default to the displayValue prop
- Add new prop "hoverTooltipValue" to DraggableSelectedOption and refactored to use this value first before defaulting back to the displayValue

- Unit test added for both MultiSelectedConnected and DraggableSelectedOption.

### Acceptance Criteria


packages/react-vapor/src/components/dropdownSearch/MultiSelectDropdownSearch/DraggableSelectedOption.tsx
```diff
@@ -93,7 +94,7 @@ export class DraggableSelectedOption extends React.PureComponent<IDraggableSelec
                 <div className="selected-option-wrapper" style={{opacity}}>
                     <SelectedOption
                         {..._.omit(this.props, DraggableSelectedOptionPropsToOmit)}
-                        label={this.props.isDragging ? null : this.props.label}
+                        label={this.props.isDragging ? null : this.props.hoverTooltipValue || this.props.label}
```
dmgauthier, line 97: `this.props.isDragging ? null : this.props.hoverTooltipValue ?? this.props.label` 
The `||`(or) operator certainly works here but the nullish coalesing operator is more appropriate for this use case.

packages/react-vapor/src/components/select/MultiSelectConnected.tsx
```diff
@@ -95,7 +95,7 @@ class MultiSelect extends React.PureComponent<IMultiSelectProps> {
         const displayValue = item.displayValue || item.value;
         return (
             <SelectedOption
-                label={displayValue}
+                label={item.hoverTooltipValue || displayValue}
```
dmgauthier, line 98: `item.hoverTooltipValue ?? displayValue`

dmgauthier:
If at the line 20 of the multiSelect example file I write this:

```
const defaultItems: IItemBoxProps[] = [
    {displayValue: 'Test', value: '0', tooltip: {placement: 'top', title: 'Hello There'}},
    {displayValue: 'Test One', value: '1'},
[...]
```

I get this if I hover the value in the drop down:
tooltip_top

But if I select that same value and hover its box I get this:
hardCodedToolTipWithValue

IMO that does not make a lot of sense. And would prefer that the tooltip prop I decide to put on the value at the first place be taken into consideration once the value is selected instead of an hardcoded tooltip displaying the value itself. That would remove the need for a another prop that feels kinda duplicated.

kdang-coveo:
> IMO that does not make a lot of sense. And would prefer that the tooltip prop I decide to put on the value at the first place be taken into consideration once the value is selected instead of an hardcoded tooltip displaying the value itself. That would remove the need for a another prop that feels kinda duplicated.

Thanks for the input! You are correct, it does seem to be duplicated... I will confirm if that is the desired result. I was under the impression that we needed a new prop for the IItemBoxProps based on the description of our ticket since it seems like the tooltip only applies to the dropdown.

e.g
```
const defaultItems: IItemBoxProps[] = [
    {displayValue: 'Test', value: '0', tooltip: {placement: 'left', title: 'Hello There'}},
    {displayValue: 'Test One', value: '1'},
[...]
```
The placement property changed from `left` will result in something below for the dropdown
image

But for the selected result it will still remain defaulted to the value and placement of `top`:
image

packages/react-vapor/src/components/itemBox/ItemBox.tsx
```diff
@@ -20,6 +20,7 @@ export interface IItemBoxProps {
     append?: IContentProps;
     onOptionClick?: (option?: IItemBoxProps, index?: number) => void;
     selectedDisplayValue?: string;
+    hoverTooltipValue?: string;
```
gdostie, line 23: - Our tooltips always appear on hover so I would drop that part from the prop name.
- This prop will intrinsically be linked to a selected item so I would specify that in the name.
- Our other tooltip prop is of type `ITooltipProps`, I think this one should be the same.

```suggestion
    selectedTooltip: ITooltipProps;
```

packages/react-vapor/src/components/select/MultiSelectConnected.tsx
```diff
@@ -1,200 +1,201 @@
-import classNames from 'classnames';
```
gdostie, line 1: Weird diff

packages/react-vapor/src/components/dropdownSearch/MultiSelectDropdownSearch/tests/DraggableSelectedOption.spec.tsx
```diff
@@ -44,6 +45,20 @@ describe('DraggableSelectedOption', () => {
             expect(selectedOption.find('.selected-option-value').first().text()).toBe(label);
         });
 
+        it('should render the a custom value for the tooltip in the selectedOption', () => {
```
FelixBlaisThon, line 48: render the a custom

packages/react-vapor/src/components/dropdownSearch/MultiSelectDropdownSearch/tests/DraggableSelectedOption.spec.tsx
```diff
@@ -44,6 +45,20 @@ describe('DraggableSelectedOption', () => {
...
+            expect(selectedOption.find(Tooltip).prop('title')).toBe(customValue);
+        });
+
+        it('should render a custom value instead of the label prop in the selectedOption', () => {
+            const customValue: string = 'Bananas';
+            mountOption({label: 'helloworld', selectedTooltip: {title: customValue}});
+
+            expect(selectedOption.find(Tooltip).prop('title')).toBe(customValue);
+        });
+
```
FelixBlaisThon, line 61: I think you should add a test in the case where no selectedTooltip is provided (so you go pass through the ?? operator and it's the props.label that is used instead) ? i'm not familiar with that component though

kdang-coveo: Thanks, after looking at the test files I found that lines 41 to 46 has covered this test case scenario where no selectedTooltip is provided and only label will be shown.

kdang-coveo:
After discussion with Gael, @dmgauthier, this is the expected solution since we are adding more control over the selected items tooltip (versus the dropdown tooltip). A similar logic was applied to selectedDisplay prop we had initially for the SingleSelect and for the MultiSelect component. Thus, the additional prop is required for this component.

Hopefully this clarifies the implementation for this component.

packages/react-vapor/src/components/dropdownSearch/MultiSelectDropdownSearch/DraggableSelectedOption.tsx
```diff
@@ -93,7 +95,7 @@ export class DraggableSelectedOption extends React.PureComponent<IDraggableSelec
                 <div className="selected-option-wrapper" style={{opacity}}>
                     <SelectedOption
                         {..._.omit(this.props, DraggableSelectedOptionPropsToOmit)}
-                        label={this.props.isDragging ? null : this.props.label}
+                        label={this.props.isDragging ? null : this.props.selectedTooltip?.title ?? this.props.label}
```
gdostie, line 98: Considering `props.selectedTooltip` is now of type `ITooltipProps`, I think we should modify our implementation of `SelectedOption` component a bit, so that it respects what we set in the `selectedTooltip` prop.

For example, if my item has the following props:
```ts
selectedTooltip: {
    title: 'whatever',
    placement: TooltipPlacement.Bottom,
}
```
I would expect the tooltip of the selected value to appear on the bottom, while with the current implementation it will appear on top.

dmgauthier (approved):

PR #1687, opened 2020-09-16
feat(form-validation): new form validation poc

pochretien:
POC to test a Form Validation

The idea of this POC is to create a wrapper hover your component to valid if all is valid and initialValue

### DIFF
--- a/packages/demo/src/components/examples/FormValidationExamples.tsx
+++ b/packages/demo/src/components/examples/FormValidationExamples.tsx
@@ -1,0 +1,69 @@
+import * as React from 'react';
+import {connect} from 'react-redux';
+import {
+    FormValidation,
+    InputConnected,
+    InputLabelWithTooltip,
+    FormValidationSelectors,
+    Button,
+    Section,
+    StickyFooter,
+    IReactVaporState,
+} from 'react-vapor';
+
+const validate = (value: any) => !!value;
+
+const mapStateToProps = (state: IReactVaporState) => ({
+    isValid: FormValidationSelectors.isValid(state, 'FormValidation1'),
+    isAllInitialValue: FormValidationSelectors.isAllInitialValue(state, 'FormValidation1'),
+});
+
+const StickyFooterExampleDisconnected: React.FunctionComponent<ReturnType<typeof mapStateToProps>> = ({
+    isValid,
+    isAllInitialValue,
+}) => (
+    <Section title="A sticky footer example">
+        <StickyFooter className="sticky-footer-mod-header" isOpened>
+            {isAllInitialValue && <Button primary name="Nothing has changed!" />}
+            {isValid && <Button primary name="All Valid, you can save" />}
+        </StickyFooter>
+    </Section>
+);
+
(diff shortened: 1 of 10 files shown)


coveobot (commented):
[Live demo available here](https://vaporqa.cloud.coveo.com/feature/POC-validation-form/index.html).

packages/demo/src/components/examples/FormValidationExamples.tsx
```diff
@@ -0,0 +1,69 @@
...
+            {isValid && <Button primary name="All Valid, you can save" />}
+        </StickyFooter>
+    </Section>
+);
+
+const StickyFooterExample = connect(mapStateToProps)(StickyFooterExampleDisconnected);
+
+export const FormValidationExamples = () => (
+    <Section level={3} title="A simple input connected to the store">
+        <FormValidation id="FormValidation1">
```
gdostie, line 37: I think it should be named `Validator` because it doesn't render a `<form>`, in fact it doesn't render anything else than its children.

gdostie (commented):
Couldn't test the demo, something seems to be broken
image

Even if I like not having to know which selectors to call when I want to validate, I am not convinced by the solution, because it will likely required to modify every component that we would want to validate and adjust the FormValidation component in consequence. Also, the similar validation could've been implemented using a simple selector without having to wrap them inside anything. Using a custom selector also allows more control over the validation. 

```ts
// something like this
isValid: InputSelectors.getIsValid(state, {id: 'test1'}) && InputSelectors.getIsValid(state, {id: 'test2'})
```

packages/demo/src/components/examples/FormValidationExamples.tsx
pochretien: good point, Im not sure too if FormValidation is good because like you said, I render nothing. Juste bind the children with redux

packages/demo/src/components/examples/FormValidationExamples.tsx
```diff
@@ -0,0 +1,69 @@
...
+                    <InputLabelWithTooltip
+                        label="I am a connected input, and validated in real time."
+                        tooltip="This input is connected."
+                        invalidMessage="Do not leave me empty"
+                    />
+                }
+                validateOnChange
+            />
+        </FormValidation>
+        <FormValidation id="FormValidation1">
```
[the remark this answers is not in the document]
pochretien: the id for the FormValidation is to create a group of validation. so you can have more than 1 in the same page and valid both in your footer if you want. I think we will se that with collapsible with step.

I use the first children only in my case to avoid problem in the futur. With the actual design, we need to wrap each input with the Validator.

francoislg:
:thinking: I feel there is a small overlap with the existing [Validation](https://github.com/coveo/react-vapor/tree/master/packages/react-vapor/src/components/validation) component.

Technically, the `isValid` state could be one of the generic "error" from the Validation state, and the `isDirty` already handles the initial state comparison.

Is the idea to deprecate the Validation HOCs and use this at-large instead?

PR #1690, opened 2020-09-21
fix(validation): should not display tooltip with skipDirty

francoislg:
[COM-575]

### Proposed Changes

The `skipDirty` prop allows bypassing the dirty check for Save Buttons. 

This is useful if you know the initial values can be valid and thus saveable without requiring the user to change any value.

However, the dirty tooltip was still displayed!

This removes the tooltip if `skipDirty` is set to true

None

### Acceptance Criteria

[COM-575]: https://coveord.atlassian.net/browse/COM-575

### DIFF
--- a/packages/react-vapor/src/components/validation/hoc/WithDirtySaveButtonHOC.tsx
+++ b/packages/react-vapor/src/components/validation/hoc/WithDirtySaveButtonHOC.tsx
@@ -47,7 +47,7 @@ export const withDirtySaveButtonHOC = <T extends IButtonProps>(Component: React.
         const generatedTooltip =
             (hasErrors && errorMessage(errors.map((error) => error.value))) ||
             (hasWarnings && canSaveWhenDirty && warningMessage(warnings.map((warning) => warning.value))) ||
-            dirtyMessage(dirty.map((d) => d.value));
+            (skipDirty ? '' : dirtyMessage(dirty.map((d) => d.value)));
 
         return (
             <Component
--- a/packages/react-vapor/src/components/validation/hoc/tests/WithDirtySaveButtonHOC.spec.tsx
+++ b/packages/react-vapor/src/components/validation/hoc/tests/WithDirtySaveButtonHOC.spec.tsx
@@ -195,6 +195,21 @@ describe('WithDirtySaveButtonHOC', () => {
                 expect(tooltip).toEqual(message);
             });
 
+            it('should not change the tooltip to the dirty message formatter if the skipDirty prop is set to true', () => {
+                const message = 'you must change stuff';
+                const buttonWrapper = shallowButton(
+                    {
+                        skipDirty: true,
+                        dirtyMessage: () => message,
+                    },
+                    getStoreMock()
+                );
+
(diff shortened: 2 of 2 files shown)


coveobot (commented):
[Live demo available here](https://vaporqa.cloud.coveo.com/feature/fix-com-575-skipdirty-savebutton/index.html).

packages/react-vapor/src/components/validation/hoc/WithDirtySaveButtonHOC.tsx
```diff
@@ -47,7 +47,8 @@ export const withDirtySaveButtonHOC = <T extends IButtonProps>(Component: React.
         const generatedTooltip =
             (hasErrors && errorMessage(errors.map((error) => error.value))) ||
             (hasWarnings && canSaveWhenDirty && warningMessage(warnings.map((warning) => warning.value))) ||
-            dirtyMessage(dirty.map((d) => d.value));
+            (!skipDirty && dirtyMessage(dirty.map((d) => d.value))) ||
```
gdostie, line 50: I think a ternary would be easier on the eye here.
```suggestion
            skipDirty ? '' : dirtyMessage(dirty.map((d) => d.value)));
```

gdostie (approved):

packages/react-vapor/src/components/validation/hoc/WithDirtySaveButtonHOC.tsx
francoislg: It auto-formats like this though:

image

what are your eyes' opinion on that? :smile:

toussaints (approved):

Spuffynism (approved):

dlafreniere (approved):

rosalie-liu (approved):

packages/react-vapor/src/components/validation/hoc/WithDirtySaveButtonHOC.tsx
Spuffynism: Using a ternary changes the behavior and could result in `generatedTooltip` having a value of whatever `dirtyMessage` returns when `skipDirty` is false, as opposed to always having a fallback on `''`. I don't know if this is actually a problem however.

Also, we could use the ternary + a fallback, but this feels like going back to @francoislg 's solution.

```suggestion
(skipDirty 
    ? '' 
    : dirtyMessage(dirty.map((d) => d.value))) ||
    '';
```

agagnon-coveo: I understand that using a "not" condition breaks the flow here considering that it's always `condition && ...` above but I think a ternary also breaks that flow.

Maybe create a variable above that would be `const showDirty = !skipDirty`? Not sure if that really adds value but I think it's an option and would allow for a smooth `(showDirty && dirtyMessage(...`

francoislg: :point_up: not needed because of this line down there: `generatedTooltip || tooltip`

PR #1693, opened 2020-09-28
Adui 5371 investigate move to tsc/rollup instead of webpack PART 1

j-mcgregor:
This PR will fail. This is expected.

### Proposed Changes

Part 1 - importing Jest config, removing Jasmine, Karma, Webpack

Too many to list - Jest uses Jasmine behind the scenes but everything else is different
This PR will point to Master. A separate PR containing the test fixes will be made and point to this branch. There will be hundreds of files in that one running around 3,500 UTs. 

Most of those change will be slight variations (eg `jasmine.spy() -> jest.spyOn()`, but many won't be 

### Acceptance Criteria


packages/react-vapor/jest.config.js
```diff
@@ -0,0 +1,141 @@
...
+        'ts-jest': {
+            // relative path to the ts-jest-keys-transformer.js file
+            astTransformers: ['<rootDir>/setupTests/ts-jest-keys-transformer.js'],
+        },
+        _: 'underscore',
+    },
+    testEnvironment: 'jest-environment-jsdom-fourteen',
+};
+
+// start
```
GermainBergeron, line 27: What are all those comments?

packages/react-vapor/src/components/listBox/ListBox.tsx
```diff
@@ -91,7 +91,7 @@ export class ListBox extends React.Component<IListBoxProps> {
                     key={item.value}
                     {...item}
                     onOptionClick={(option: IItemBoxProps) => this.onSelectItem(item, item.index)}
-                    selected={_.contains(this.props.selected, item.value)}
+                    active={_.contains(this.props.selected, item.value)} // here
```
GermainBergeron, line 94: I think you should revert this change

packages/react-vapor/package.json
```diff
@@ -37,6 +35,10 @@
     },
     "dependencies": {
         "diff2html": "3.1.11",
+        "karma-rollup-plugin": "^0.2.4",
```
GermainBergeron, line 38: are we still using karma?

packages/react-vapor/package.json
```diff
@@ -63,6 +72,8 @@
         "underscore.string": "^3.3.5"
     },
     "devDependencies": {
+        "@babel/preset-react": "^7.10.4",
+        "@immowelt/jest-preset-react": "^3.1.0",
```
GermainBergeron, line 76: babel? immowelt? are those used somewhere?

GermainBergeron (commented):
You can probably do a cleanup at the same time (removing karma related files, removing packages that we don't use in the package.json, etc)

packages/react-vapor/setupTests/assetsTransformer.js
```diff
@@ -0,0 +1,7 @@
+const path = require('path');
+
+module.exports = {
+    process(src, filename, config, options) {
+        return 'module.exports = ' + JSON.stringify(path.basename(filename)) + ';';
```
rosalie-liu, line 5: Would it be easier to read `return `module.exports = ${JSON.stringify(path.basename(filename))};`;` ?

packages/react-vapor/jest.config.js
FelixBlaisThon: Changes he made and how it fixed tests, but i dont know why it's in this pr and not the other

PR #1701, opened 2020-10-15
feat(multivalue): can disabled

pochretien:
update input to handle readOnly to avoid confusion with the disabled from the state
Update multiValueInputs to use disabled as a props
Update MultilineBoxWithRemoveButton to use the disabled as a props
add input and multiValue examples

** Uts next PR

### Proposed Changes

Add "disabled" as a props for the MultiValuesInput.
The first problem was with the Input. we dont have any way to use the ownProps to set the component in disabled. We use the state to do that. I added a new props readOnly to avoid problem with the disabled from the state. I tried to use disabled in both case but it's to dangerous to get weird behaviour in the futur so I created the readOnly to set the component on render ou later in disabled state with his own props.

No breaking changes. The new props are optional and can modify the component only if we set the props manualy.

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/input/Input.tsx
+++ b/packages/react-vapor/src/components/input/Input.tsx
@@ -32,6 +32,7 @@ export interface IInputAdditionalOwnProps {
     maximum?: number /* @deprecated use max instead */;
     onBlur?: (value: string) => void;
     defaultValue?: string;
+    isReadOnly?: boolean;
 }
 
 export interface IInputNativeTagStateProps {
@@ -79,6 +80,7 @@ export class Input extends React.Component<IInputProps, IInputComponentState> {
         valid: true,
         labelTitle: '',
         required: true,
+        isReadOnly: false,
     };
 
     constructor(props: IInputProps, state: IInputState) {
@@ -182,6 +184,11 @@ export class Input extends React.Component<IInputProps, IInputComponentState> {
             this.props.innerInputClasses
         );
 
+        let additionalProps: {disabled?: boolean} = {};
+        if (this.props.isReadOnly) {
+            additionalProps = {disabled: true};
+        }
+
         const inputElements = [
             <input
                 key={this.props.id}
@@ -195,6 +202,7 @@ export class Input extends React.Component<IInputProps, IInputComponentState> {
                 min={this.props.minimum}
                 max={this.props.maximum}
                 {..._.omit(this.props, [...PropsToOmitUtils.input, ...inputPropsToOmit])}
+                {...additionalProps}
             />,
             this.getLabel(),
             this.props.children,
--- a/packages/react-vapor/src/components/multiValueInputs/MultiValuesInput.tsx
+++ b/packages/react-vapor/src/components/multiValueInputs/MultiValuesInput.tsx
@@ -36,6 +36,7 @@ export interface MultiValuesInputProps {
     disabledTooltipTitle?: string;
     disabledInputInnerClasses?: IClassName;
     disabledInputClasses?: IClassName;
+    disabled?: boolean;
 }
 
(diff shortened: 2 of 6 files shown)


coveobot (commented):
[Live demo available here](https://vaporqa.cloud.coveo.com/feature/toggle-multi-values-disabled/index.html).

pochretien:
image
image

packages/react-vapor/src/components/input/Input.tsx
```diff
@@ -195,6 +202,7 @@ export class Input extends React.Component<IInputProps, IInputComponentState> {
                 min={this.props.minimum}
                 max={this.props.maximum}
                 {..._.omit(this.props, [...PropsToOmitUtils.input, ...inputPropsToOmit])}
+                {...additionalProps}
```
rosalie-liu, line 205: Does it work if we use directly `disabled={this.props.isReadOnly}` here?

packages/react-vapor/src/components/multilineBox/hoc/MultilineBoxWithRemoveButton.tsx
```diff
@@ -113,5 +116,5 @@ export const multilineBoxWithRemoveButton = (
         }
     }
 
-    return MultilineBoxWithRemoveButton;
+    return connect(undefined, mapDispatchToProps)(MultilineBoxWithRemoveButton);
```
rosalie-liu, line 119: Is it better to use `connect(null, mapDispatchToProps)`?

packages/react-vapor/src/components/multilineBox/hoc/MultilineBoxWithRemoveButton.tsx
```diff
@@ -73,8 +76,8 @@ export const multilineBoxWithRemoveButton = (
                     style={{
                         visibility: !data.isLast ? 'visible' : 'hidden',
                     }}
-                    onClick={() => !data.isLast && this.props.removeBox(data.id)}
-                    enabled={!data.isLast}
+                    onClick={() => !data.isLast && !this.props.disabled && this.props.removeBox(data.id)}
+                    enabled={!(this.props.disabled || data.isLast)}
```
rosalie-liu, line 80: I'm a little confused about `!(this.props.disabled || data.isLast)` :joy: Could we extract `this.props.disabled && data.isLast` into a variable then we can use it in both line 65 and line 80? 
Or simply remove `={!(this.props.disabled || data.isLast)}`, I think `enabled` could be always `true` here, otherwise it returns already `null` at the line 66.

rosalie-liu (approved):

packages/react-vapor/src/components/multiValueInputs/MultiValuesInput.tsx
```diff
@@ -63,6 +65,11 @@ export const MultiValuesInput: React.FunctionComponent<MultiValuesInputProps> =
                 const classes = isInputLimitReached
                     ? classNames(inputProps?.classes, disabledInputClasses)
                     : inputProps?.classes;
+
+                if (cData.isLast && disabled) {
```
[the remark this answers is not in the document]
pochretien: its the name used in multilineBox so I only reused the same name to understand its the same variable

packages/react-vapor/src/components/input/Input.tsx
pochretien: I cant because we already have the prop disabled so if I want to set disabled on the input in case of readOnly, I need to override the previous disabled set at line 204. This disabled is from the state so to avoid any problem with the input,  we need to set the additionalProps only when the props is set to true.

dmgauthier (approved):

PR #1715, opened 2020-11-05
Adui 6097 multifacet is so pale that it s looking disabled

FelixBlaisThon:
### Proposed Changes

We will let the disabled component as it is, but we made some upgrade on it. We now make a distinction between a multiselect disabled and in readonly. I've added an example to the multiselectexamples file and added UTs for draggable component and selectedoption component

### Acceptance Criteria

### DIFF
--- a/packages/demo/src/components/examples/MultiSelectExamples.tsx
+++ b/packages/demo/src/components/examples/MultiSelectExamples.tsx
@@ -146,7 +146,6 @@ export class MultiSelectExamples extends React.Component<null, IMultiSelectExamp
                         placeholder="Select something"
                         deselectAllTooltipText="Remove all"
                         sortable
-                        readOnly
                     />
                 </div>
                 <div className="form-group">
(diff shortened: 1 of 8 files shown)


coveobot (commented):
[Live demo available here](https://vaporqa.cloud.coveo.com/feature/ADUI-6097-multifacet-is-so-pale-that-it-s-looking-disabled/index.html).

aiheon:
Would it be possible to have a demo for the multi-select in read-only mode with no selected value? 
Or if would just be an empty box?

FelixBlaisThon:
> Would it be possible to have a demo for the multi-select in read-only mode with no selected value?
> Or if would just be an empty box?
> 
> Something like this?
> image

image

aiheon (approved):

rosalie-liu (approved):

dmgauthier (approved):

PR #1718, opened 2020-11-09
fix(action row): if disabled no double click

dmgauthier:
### Proposed Changes

The row call on double click should not be triggered if the row is disabled.

The American gouvernement administration change. Although probably not.

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/table-hoc/TableRowConnected.tsx
+++ b/packages/react-vapor/src/components/table-hoc/TableRowConnected.tsx
@@ -5,6 +5,7 @@ import * as _ from 'underscore';
 
 import {SlideY} from '../../animations/SlideY';
 import {IReactVaporState} from '../../ReactVapor';
+import {UrlUtils} from '../../utils';
 import {EventUtils} from '../../utils/EventUtils';
 import {IDispatch, ReduxConnect} from '../../utils/ReduxUtils';
 import {IActionOptions} from '../actions/Action';
@@ -197,11 +198,12 @@ class TableRowConnected extends React.PureComponent<
     private handleDoubleClick = () => {
         _.chain(this.props.actions)
             .filter((action: IActionOptions) => action.callOnDoubleClick)
+            .filter((action: IActionOptions) => action.enabled)
             .forEach((action: IActionOptions) => {
                 if (action.link) {
-                    window.location.href = action.link;
-                } else if (action.trigger) {
-                    action.trigger();
+                    UrlUtils.redirectToUrl(action.link);
+                } else {
+                    action.trigger?.();
                 }
             });
     };
--- a/packages/react-vapor/src/components/table-hoc/tests/TableRowConnected.spec.tsx
+++ b/packages/react-vapor/src/components/table-hoc/tests/TableRowConnected.spec.tsx
@@ -2,6 +2,7 @@ import {ShallowWrapper} from 'enzyme';
 import {mountWithStore, shallowWithStore} from 'enzyme-redux';
 import * as React from 'react';
 
+import {UrlUtils} from '../../../utils';
 import {getStoreMock, ReactVaporMockStore} from '../../../utils/tests/TestUtils';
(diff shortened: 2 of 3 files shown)


dmgauthier:
Should I test this ?

coveobot (commented):
[Live demo available here](https://vaporqa.cloud.coveo.com/feature/fix-ADUI-6200-disable-double-click-action-row/index.html).

aiheon (approved):

rosalie-liu (approved):

pochretien (approved):
Yep, you can fix the old test if they exist or add to valid the call on link if defined or not

packages/react-vapor/src/components/table-hoc/TableRowConnected.tsx
```diff
@@ -198,10 +198,8 @@ class TableRowConnected extends React.PureComponent<
         _.chain(this.props.actions)
             .filter((action: IActionOptions) => action.callOnDoubleClick)
             .forEach((action: IActionOptions) => {
-                if (action.link) {
-                    window.location.href = action.link;
-                } else if (action.trigger) {
-                    action.trigger();
+                if (action.enabled) {
+                    action?.link ? (window.location.href = action.link) : action?.trigger();
```
GermainBergeron, line 202: Personally I find it way easier to read with an if/else. Also you don't need the `?.` after action because it will always be defined. `trigger` might not though:

```suggestion
                    if (action.link) {
                        window.location.href = action.link
                    } else {
                        action.trigger?.()
                    }
```

GermainBergeron (commented):
I would also add a UT

packages/react-vapor/src/components/table-hoc/TableRowConnected.tsx
```diff
@@ -198,10 +198,8 @@ class TableRowConnected extends React.PureComponent<
         _.chain(this.props.actions)
             .filter((action: IActionOptions) => action.callOnDoubleClick)
             .forEach((action: IActionOptions) => {
```
gdostie, line 200: We could add another filter here instead of the if block down there?
```suggestion
            .fitler((action: IActionOptions) => action.enabled)
            .forEach((action: IActionOptions) => {
```

packages/react-vapor/src/components/table-hoc/TableRowConnected.tsx
dmgauthier: > Personally I find it way easier to read with an if/else. Also you don't need the `?.` after action because it will always be defined. `trigger` might not though:

The easier to read argument always wins in my book. I'll modify it accordingly

packages/react-vapor/src/components/table-hoc/TableRowConnected.tsx
dmgauthier: so obvious :see_no_evil:

gdostie (commented):
Add unit test and you have my approval 🏆

dmgauthier:
> Add unit test and you have my approval

Added !

packages/react-vapor/src/components/table-hoc/utils/TableHOCUtils.ts
```diff
@@ -88,13 +88,18 @@ const getTablePredicates = (tableId: string, state: IReactVaporState): ITableHOC
 
 const getDatePickerId = (tableId: string) => `${tableId}-date-range`;
 
+const redirectToUrl = (link: string) => {
```
gdostie, line 91: This should rather go in the [`UrlUtils`](https://github.com/coveo/react-vapor/blob/fcd398a8b29252804be8d569b0d05f9655c97220/packages/react-vapor/src/utils/UrlUtils.ts)

gdostie (approved):

packages/react-vapor/src/components/table-hoc/utils/TableHOCUtils.ts
dmgauthier: will move it thx !

PR #1729, opened 2020-11-17
fix(btn): fix extra margin when using tooltip

francoislg:
[COM-700]

I put two buttons side-by-side, one with a tooltip, the other without, and there was a difference in margin!

I found out that the `vertical-align: middle` that is *only* on the `btn` class and not on `btn-container`  was responsible for that! See here:

Peek 2020-11-17 11-16

The issue only occurs when the button is included **inline** and another component is in the same block.

It should not cause any issue, if anything, it will repair margins in some places

### Acceptance Criteria

[COM-700]: https://coveord.atlassian.net/browse/COM-700

### DIFF
--- a/packages/vapor/scss/elements/btn.scss
+++ b/packages/vapor/scss/elements/btn.scss
@@ -193,6 +193,7 @@
 .btn-container {
     display: inline-block;
     margin: 0;
+    vertical-align: middle;
 
     .btn[disabled] {
         pointer-events: none;


rosalie-liu (approved):

coveobot (commented):
[Live demo available here](https://vaporqa.cloud.coveo.com/feature/fix-com-700-btn-alignment-with-tooltip/index.html).

gdostie (approved):

francoislg:
Can one of you merge it? :smile:

PR #1736, opened 2020-11-23
fix(modal composite): openmodals array no duplication and clear

dmgauthier:
### Proposed Changes

The openModals array in the store was emptying only if the modal was explicitaly closed. The modal Id will now be removed from the openModals array if the modal is symply removed.

Also, a condition in the reducer will prevent the same modal from being added in the openModals array many times

none

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/modal/ModalReducers.ts
+++ b/packages/react-vapor/src/components/modal/ModalReducers.ts
@@ -60,9 +60,14 @@ export const modalsReducer = (
 export const openModalsReducer = (state: string[] = [], action: IReduxAction<IModalActionPayload>): string[] => {
     switch (action.type) {
         case ModalAction.openModal:
-            return [...state, action.payload.id];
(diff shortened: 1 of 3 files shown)


dmgauthier:
To test.

You can go in the modalWindow examples. And there you open a modal.  

With the modal still open, you change the url  ex `/#/components/logocard`   and you should see the modal id removed from the openModals array in the store. It was not before.

Also, you should see that the modal id is still added to the openModals array, but the modal id cannot be added many times. I tested this behavior before fixing the first issue, so I saw that if the Id was already in the array, the modal was opening but the same Id was not pushed in the array over and over again.

coveobot (commented):
[Live demo available here](https://vaporqa.cloud.coveo.com/feature/fix-modal-composite-openmodals-array/index.html).

rosalie-liu (approved):

gdostie (approved):

PR #1744, opened 2020-12-07
feat(color): generated

pochretien:
override color generated
create script to generate color faster


coveobot (commented):
[Live demo available here](https://vaporqa.cloud.coveo.com/feature/new-color-generated-override/index.html).

gdostie (approved):

GermainBergeron (changes requested):
As discussed via Slack I think we can use the logic we already have while changing it just a bit

GermainBergeron (approved):

rosalie-liu (approved):

PR #1749, opened 2020-12-07
feat(tab): add the possibility to include a svg to a tab

sallainCoveo:
### Proposed Changes
I added the possibility to include a Svg icon to an Tab component. You could see an example in the Tab section.
new_tab_with_icon

### Acceptance Criteria

-   [X ] The proposed changes are covered by unit tests

### DIFF
--- a/packages/demo/src/components/examples/TabConnectedExample.tsx
+++ b/packages/demo/src/components/examples/TabConnectedExample.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import {TabConnected, TabContent, TabNavigation, TabPaneConnected} from 'react-vapor';
+import {Svg, TabConnected, TabContent, TabNavigation, TabPaneConnected} from 'react-vapor';
 
 const TAB1_ID = 'tab1';
 const TAB2_ID = 'tab2';
@@ -23,8 +23,7 @@ export class TabsExamples extends React.Component<any, any> {
                             <TabConnected
                                 id={TAB3_ID}
                                 title="Tab with an icon"
-                                svgName="help"
-                                svgClass="icon fill-orange mod-16 mr1"
+                                children={<Svg svgName={'help'} svgClass={'icon fill-orange mod-16 mr1'} />}
                             />
                             <TabConnected id={TAB4_ID} title="A Disabled Tab" tooltip="I am a disabled tab" disabled />
                         </TabNavigation>
--- a/packages/demo/src/components/examples/TabConnectedExample.tsx
+++ b/packages/demo/src/components/examples/TabConnectedExample.tsx
@@ -4,6 +4,7 @@ import {TabConnected, TabContent, TabNavigation, TabPaneConnected} from 'react-v
 const TAB1_ID = 'tab1';
 const TAB2_ID = 'tab2';
 const TAB3_ID = 'tab3';
+const TAB4_ID = 'tab4';
 
 const TAB11_ID = 'tab11';
 const TAB22_ID = 'tab22';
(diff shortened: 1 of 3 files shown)


coveobot (commented):
[Live demo available here](https://vaporqa.cloud.coveo.com/feature/feat-ADUI-6301-a-tab-component-could-include-a-svg-component-in-it/index.html).

packages/react-vapor/src/components/tab/Tab.tsx
```diff
@@ -34,6 +37,7 @@ export class Tab extends React.Component<ITabProps, any> {
 
         return (
             <div className={className} onClick={this.handleSelect}>
+                {!!this.props.svgName && <Svg svgName={this.props.svgName} svgClass={this.props.svgClass} />}
```
GermainBergeron, line 40: Should we allow any children instead?

It could be useful if we want a different tooltip on the svg for instance

sallainCoveo: Good point! It will be a more flexible option that way (so, I should set a prop that is a React.ReactNode instead).

packages/demo/src/components/examples/TabConnectedExample.tsx
```diff
@@ -19,7 +20,12 @@ export class TabsExamples extends React.Component<any, any> {
                         <TabNavigation>
                             <TabConnected id={TAB1_ID} title="A Tab" />
                             <TabConnected id={TAB2_ID} title="Another Tab" tooltip="I am an enabled tab" />
-                            <TabConnected id={TAB3_ID} title="A Disabled Tab" tooltip="I am a disabled tab" disabled />
+                            <TabConnected
+                                id={TAB3_ID}
+                                title="Tab with an icon"
+                                children={<Svg svgName={'help'} svgClass={'icon fill-orange mod-16 mr1'} />}
+                            />
```
GermainBergeron, line 27:
```suggestion
                            <TabConnected id={TAB3_ID} title="Tab with an icon">
                                <Svg svgName="help" svgClass="icon fill-orange mod-16 mr1" />}
                            </TabConnected>
```

packages/react-vapor/src/components/tab/Tab.tsx
```diff
@@ -10,6 +10,7 @@ export interface ITabOwnProps {
     title: string;
     disabled?: boolean;
     tooltip?: string;
+    children?: React.ReactNode;
```
GermainBergeron, line 13: I think you don't need to add it to the interface since React.Component already adds it to the props object

packages/react-vapor/src/components/tab/Tab.tsx
```diff
@@ -34,6 +35,7 @@ export class Tab extends React.Component<ITabProps, any> {
 
         return (
             <div className={className} onClick={this.handleSelect}>
+                {!!this.props.children && this.props.children}
```
GermainBergeron, line 38: You can simply use {this.props.children} since the default value is undefined which doesn't get rendered

packages/react-vapor/src/components/tab/tests/Tab.spec.tsx
```diff
@@ -71,4 +72,13 @@ describe('Tab', () => {
         expect(tab.find(Tooltip).exists()).toBe(true);
         expect(tab.find(Tooltip).props().title).toBe(expectedTooltipText);
     });
+
+    it('should render a children component if it is not empty', () => {
+        const tab = shallow(
+            <Tab {...basicProps} children={<Svg svgName={'help'} svgClass={'icon fill-orange mod-16 mr1'} />} />
```
GermainBergeron, line 78:
```suggestion
            <Tab {...basicProps}>
                <Svg svgName="help" svgClass="icon fill-orange mod-16 mr1" />
            </Tab>
```

GermainBergeron (approved):

gdostie (approved):

PR #1787, opened 2021-01-04
fix(code-editor): add the possibility to refresh the editor

sallainCoveo:
### Proposed Changes

This fix add the possibility to refresh the editor when the `isRefreshRequired` prop is true. An example is included in the collapsible section in which the editor is display properly when the collapsible is opened. 
collapsibleWithJsonEditor

### Acceptance Criteria

-   [ X] The proposed changes are covered by unit tests

### DIFF
--- a/packages/react-vapor/src/components/editor/CodeEditor.tsx
+++ b/packages/react-vapor/src/components/editor/CodeEditor.tsx
@@ -11,9 +11,12 @@ import * as React from 'react';
 import * as ReactCodeMirror from 'react-codemirror2';
 
 import classNames from 'classnames';
+import {CollapsibleSelectors} from '../collapsible/CollapsibleSelectors';
 import {CodeMirrorGutters} from './EditorConstants';
+import {IReactVaporState} from '../../ReactVapor';
+import {ReduxConnect} from '../../utils';
 
-export interface ICodeEditorProps {
+export interface ICodeEditorOwnProps {
     value: string;
     mode: any;
     readOnly?: boolean;
@@ -23,14 +26,25 @@ export interface ICodeEditorProps {
     extraKeywords?: string[];
     className?: string;
     options?: CodeMirror.EditorConfiguration;
-    isRefreshRequired?: boolean;
+    collapsibleId?: string;
+}
+
+export interface CodeEditorMappedState {
+    isCollapsibleExpanded?: boolean;
 }
 
 export interface CodeEditorState {
     value: string;
 }
 
-export class CodeEditor extends React.Component<ICodeEditorProps, CodeEditorState> {
+const mapStateToProps = (state: IReactVaporState, {collapsibleId}: ICodeEditorOwnProps) => ({
+    isCollapsibleExpanded: CollapsibleSelectors.isExpanded(state, collapsibleId),
+});
+
+export interface ICodeEditorProps extends ICodeEditorOwnProps, Partial<ReturnType<typeof mapStateToProps>> {}
+
+@ReduxConnect(mapStateToProps)
+export class CodeEditor extends React.Component<ICodeEditorProps, CodeEditorState, CodeEditorMappedState> {
     static defaultProps: Partial<ICodeEditorProps> = {
         className: 'mod-border',
         value: '{}',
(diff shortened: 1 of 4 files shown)


packages/react-vapor/src/components/editor/CodeEditor.tsx
```diff
@@ -67,27 +73,33 @@ export class CodeEditor extends React.Component<ICodeEditorProps, CodeEditorStat
...
-                onChange={(editor, data, value: string) => this.props.onChange?.(value)}
-                options={{
-                    ...CodeEditor.defaultOptions,
-                    readOnly: this.props.readOnly,
-                    mode: this.props.mode,
-                    ...this.props.options,
-                }}
-                className={classNames(this.props.className, {'code-editor-no-cursor': this.props.readOnly})}
-            />
+            <>
```
FelixBlaisThon, line 79: Remove fragment if unecessary

coveobot (commented):
[Live demo available here](https://vaporqa.cloud.coveo.com/feature/fix-ADUI-6347-jsoneditor-does-not-show-its-value-on-first-render-if-inside-a-collapsible-that-has-not-the-openonmount-props/index.html).

packages/react-vapor/src/components/editor/CodeEditor.tsx
```diff
@@ -59,6 +60,11 @@ export class CodeEditor extends React.Component<ICodeEditorProps, CodeEditorStat
 
     componentDidMount() {
         this.props.onMount?.(this.codemirror.current);
+        this.props.isRefreshRequired && setInterval(this.timer, 300);
```
gdostie, line 63: Do we really want to refresh the editor every 0,3 s ? Maybe a setTimeout would be more appropriate ?

sallainCoveo: The setTimeout wouldn't work in that case, because the editor is mounted about the same time as the CollapsibleContainer component. So, if you don't open the collapsible before the time is up, the editor's content won't be shown. However, I do agree it's not an ideal solution. It should trigger a setTimeout only when the collapsible is opened.

coveobot (commented):
[Live demo available here](https://vaporqa.cloud.coveo.com/feature/PR-1787/index.html).

dmgauthier (approved):

packages/react-vapor/src/components/editor/CodeEditor.tsx
```diff
@@ -62,11 +73,13 @@ export class CodeEditor extends React.Component<ICodeEditorProps, CodeEditorStat
     }
 
     componentDidUpdate(prevProps: ICodeEditorProps) {
+        this.props.isCollapsibleExpanded && this.editor.refresh();
```
GermainBergeron, line 76: Won't that trigger some useless refresh when the value changes? I think it would be better to do it this way:

```suggestion
        if (!prevProps.isCollapsibleExpanded && this.props.isCollapsibleExpanded) {
            this.editor?.refresh()
        }
```

packages/react-vapor/src/components/editor/tests/CodeEditor.spec.tsx
```diff
@@ -81,30 +80,34 @@ describe('CodeEditor', () => {
...
+                .find(ReactCodeMirror.Controlled)
+                .first()
+                .props()
+                .onChange({} as CodeMirror.Editor, undefined, expectedValue);
 
             expect(onChangeSpy).toHaveBeenCalledTimes(1);
             expect(onChangeSpy).toHaveBeenCalledWith(expectedValue);
         });
 
-        it(`should clear codemirror's history if we set a new value`, () => {
```
GermainBergeron, line 94: Why did you delete this test?

GermainBergeron (approved):

packages/react-vapor/src/components/editor/tests/CodeEditor.spec.tsx
sallainCoveo: Honestly, not for a good reason. Because this component is now connected to Redux, it broke every tests and I couldn't reached the private properties as it was done before. Finally I figured out a way to pass this test.

packages/react-vapor/src/components/editor/CodeEditor.tsx
sallainCoveo: It wouldn't work, because the componentDidUpdate is called twice when we open the collapsible, and the second time is the moment the editor has to be refreshed. However, I found an alternative solution, in which I check if the codeEditor was refresh less then twice.

j-mcgregor (approved):

PR #1817, opened 2021-01-21
fix(option-picker): change the hover and active option picked

sallainCoveo:
### Proposed Changes

The odd colors noticed in the section mentioned in the Jira's case, is due to color set to the option-picker class (used for the OptionPicker component). The .gif below shows when a cursor hovers an option and when an user select an option. The design when the option is selected is based on my understanding of the coveo library canvas (in Figma). However, I'm not really sure is what is expected (would surely need the approval of UX!).

Edit: the example can be found in the DatePicker section.
new_color_option_picker

### Acceptance Criteria

### DIFF
--- a/packages/vapor/scss/components/calendar.scss
+++ b/packages/vapor/scss/components/calendar.scss
@@ -251,7 +251,7 @@
 
                 &.active {
                     color: var(--white);
-                    background-color: var(--blue-purple-3);
+                    background-color: var(--digital-blue-60);
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


coveobot (commented):
[Live demo available here](https://vaporqa.cloud.coveo.com/feature/PR-1817/index.html).

aiheon:
I think the "new " date picker don't have a hover color at all :monocle_face: 

https://www.figma.com/file/A0wfZiuNFuDRp7zN3KOyA2/Platform-Light-Reskin?node-id=253%3A9468

aiheon (approved):

guscoveo (approved):
If this is a quick re-skin, I would just address the colour of the text to move it away from the old Coveo blue (and bring #282829, 100%) I think the interaction makes sense. That said, this component is yet to be properly addressed, revised and updated on our end. We have new iconography + colours to be considered still.

GermainBergeron (approved):

PR #1825, opened 2021-01-22
fix(multiline-field.scss): when disabled, hide add, change remove icons

sallainCoveo:
### Proposed Changes
These changes are for an older multiline inputs configuration/classes. However, these changes are based of what seems to be the norm : the 'minus' icon is grey out when the input is disabled, but the 'plus' icon will be hidden instead.

### Acceptance Criteria

### DIFF
--- a/packages/vapor/scss/controls/multiline-field.scss
+++ b/packages/vapor/scss/controls/multiline-field.scss
@@ -83,30 +83,28 @@
             }
         }
 
-        &:disabled {
-            i {
-                position: relative;
-                display: block;
-                width: 18px;
-                height: 18px;
-                border: 2px solid var(--button-secondary-disabled-icon-color);
-                border-radius: 18px;
+        &:disabled i {
+            position: relative;
+            display: block;
+            width: 18px;
+            height: 18px;
+            border: 2px solid var(--button-secondary-disabled-icon-color);
+            border-radius: 18px;
 
-                &.add-action {
-                    visibility: hidden;
-                }
+            &.add-action {
+                visibility: hidden;
+            }
 
-                &.delete-action {
-                    &:before {
-                        position: relative;
-                        top: 6px;
-                        left: 2px;
-                        display: block;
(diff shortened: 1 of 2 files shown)


coveobot (commented):
[Live demo available here](https://vaporqa.cloud.coveo.com/feature/PR-1825/index.html).

aiheon:
Should we have a demo of the disabled add and remove icon? :thinking:

sallainCoveo:
> Should we have a demo of the disabled add and remove icon? 
> (so that way the wanted behaviour would be explicit)

Yeah...Because most of the component that use the css classes that need that change are pretty old, I wasn't sure it would be a good idea to implement those in react vapor. However, I think you're right. If we still use them, it should have an example (I'll edit my description as soon that I implement one)!

 Thanks!

sallainCoveo:
An example is now available...But it's a little bit hacky! Here's how you can see the effect of these changes:
1. Access the "TextInput" Section in react-vapor.
2. Scroll down to the "Multiline component (deprecated component)" example.
3. Open your devtool, and choose the 'Elements' tab.
4. Add the 'disabled' attribute on both <button> tag in which it includes the class="js-add-value-button".

Please take note that the MultilineInput is actually a deprecated component. So, after this PR will be approved, I will remove the example before I merge this branch (I don't want to motivate the usage of this component by showing as an example).
multiline_disabled_icons
disabled_in_devtool

packages/vapor/scss/controls/multiline-field.scss
```diff
@@ -82,6 +82,34 @@
                 box-shadow: 0 0 6px 1px var(--medium-blue);
             }
         }
+
+        &:disabled {
```
rosalie-liu, line 86: You could try `&:disabled i {` here instead of creating another scope

rosalie-liu (approved):

aiheon (approved):

PR #1832, opened 2021-01-25
build(next): avoid bumping a premajor on every commit

GermainBergeron:
The premajor was "bumping" a major on every merge, which is not what we want. I saw that we can use `--conventional-prerelease` instead and do a breaking change in one of the commits to bump the major (only need to do it once).

I also reverted to version 10.0.0-next.0


dmgauthier (approved):

kdang-coveo (approved):

rosalie-liu (approved):

gdostie (approved):

coveobot (commented):
[Live demo available here](https://vaporqa.cloud.coveo.com/feature/PR-1832/index.html).

PR #1838, opened 2021-01-26
fix(single-select): buggy infinite scroll when zoom is 110% in chrome

rosalie-liu:
### Proposed Changes

[ADUI-6482](https://coveord.atlassian.net/browse/ADUI-6482)

By default, the `scrollThreshold` is `1` (100%), so the `next` function (`fetchNextPage`) will be triggered only when scrolling down to the bottom. However, after investigating, the height of the list box item changes randomly when zooming. For instance, the height is 54px when 100%, but 53.631px when 110%, 53.2px when 120%, 53.333px when 150%.
image

Therefore, this issue was mainly caused by the unfixed height of list box items. So far, I thought about 2 options, give a fixed height, or slightly reduce `scrollThreshold` value to, for example, `10px` to compensate the height lost. Then we will fetch the next page earlier once there is `10px` to the bottom of the scroll bar.

In this PR, I fixed the height for the example, but still, I think it would be worth having an optional prop scrollThreshold just in case we can’t have a fixed height for some reason.

None

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/select/hoc/SelectWithInfiniteScroll.tsx
+++ b/packages/react-vapor/src/components/select/hoc/SelectWithInfiniteScroll.tsx
@@ -8,7 +8,6 @@ import {ISelectOwnProps} from '../SelectConnected';
 export interface SelectWithInfiniteScrollProps {
     totalEntries: number;
     next: () => void;
-    scrollThreshold?: React.ReactText;
 }
 
 export const selectWithInfiniteScroll = <P extends Omit<ISelectOwnProps, 'button'>>(
@@ -25,7 +24,7 @@ export const selectWithInfiniteScroll = <P extends Omit<ISelectOwnProps, 'button
                 loader={<Loading className="p2 full-content-x" />}
                 next={props.next}
                 scrollableTarget={props.id}
-                scrollThreshold={props.scrollThreshold ?? 1}
+                scrollThreshold={0.9}
                 style={{overflow: 'initial'}}
                 hasChildren={items.length > 0 || props.isLoading}
             >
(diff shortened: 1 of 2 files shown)


jmprovencher-coveo (approved):
looks good to me

coveobot (commented):
[Live demo available here](https://vaporqa.cloud.coveo.com/feature/PR-1838/index.html).

packages/react-vapor/src/components/select/hoc/SelectWithInfiniteScroll.tsx
```diff
@@ -8,6 +8,7 @@ import {ISelectOwnProps} from '../SelectConnected';
 export interface SelectWithInfiniteScrollProps {
     totalEntries: number;
     next: () => void;
+    scrollThreshold?: React.ReactText;
```
gdostie, line 11: This is pretty technical and users of the component will have to go [`react-infinit-scroll-component`'s documentation](https://www.npmjs.com/package/react-infinite-scroll-component#props) to understand what it does. I suggest we remove this prop and change the value to something more forgiving instead. Have you tried `0.9`? or not specifying it entirely to use the default value of `0.8`?

aiheon (approved):

FelixBlaisThon (approved):
Looks good to me, did you notice that the drop container was expanding after the first scroll? I don't know if it's expected or not

rosalie-liu:
> Looks good to me, did you notice that the drop container was expanding after the first scroll? I don't know if it's expected or not

Yes, it was like that before, it expands since there is a looong item.
image

PR #1846, opened 2021-01-27
refactor(font): implement gibson and update default text colors

rosalie-liu:
### Proposed Changes

- Implemented Gibson into react-vapor
- As discussed with UX, created `--default-text-color` #626971 and `--title-text-color` #282829. In General, `--title-text-color` will replace medium-blue that we currently use for labels and headings. And `--default-text-color` is for paragraph and description. We can still use different text colors for certain components, these two are default values for most use cases.
- Other small fixes of style

Label colors and heading colors should be replaced by `--title-text-color`. We will need more adjustments for text colors of specific components, but nothing changed to the functionalities.

### Acceptance Criteria

### DIFF
--- a/packages/vapor/scss/components/blankslate.scss
+++ b/packages/vapor/scss/components/blankslate.scss
@@ -1,6 +1,8 @@
 .blankslate {
+    --color: var(--title-text-color);
+
     padding: $blankslate-padding;
-    color: var(--blankslate-text-color);
+    color: var(--color);
     font-weight: var(--main-font-regular);
     text-align: center;
     background-color: var(--blankslate-background-color);
@@ -19,7 +21,7 @@
     p {
         max-width: $blankslate-description-width;
         margin: 0 auto $blankslate-paragraph-margin-bottom auto;
-        color: var(--blankslate-text-color);
+        color: var(--color);
     }
 
     &.mod-small {
--- a/packages/vapor/scss/components/facet.scss
+++ b/packages/vapor/scss/components/facet.scss
@@ -56,9 +56,7 @@
 }
 
 .facet-header-title {
-    --color: var(--title-text-color);
-
-    color: var(--color);
+    color: var(--title-text-color);
     font-weight: var(--facet-header-font-weight);
     font-size: var(--facet-header-font-size);
     font-family: var(--main-font);
--- a/packages/vapor/scss/components/facet.scss
+++ b/packages/vapor/scss/components/facet.scss
@@ -56,7 +56,9 @@
 }
 
 .facet-header-title {
-    color: var(--facet-header-text-color);
+    --color: var(--title-text-color);
+
+    color: var(--color);
     font-weight: var(--facet-header-font-weight);
     font-size: var(--facet-header-font-size);
     font-family: var(--main-font);
--- a/packages/vapor/scss/components/labeled-value.scss
+++ b/packages/vapor/scss/components/labeled-value.scss
@@ -1,4 +1,6 @@
 .box {
+    --color: var(--default-text-color);
+
     &.padded {
         padding: 5px $collapsible-box-padding;
     }
@@ -10,14 +12,16 @@
     }
 
     .label {
+        --color: var(--title-text-color);
+
         padding: 5px 0px 5px 5px;
-        color: var(--title-text-color);
+        color: var(--color);
         font-size: var(--small-font-size);
     }
 
     .value {
         padding: 5px;
-        color: var(--default-text-color);
+        color: var(--color);
         font-size: var(--default-font-size);
         line-height: $default-line-height;
     }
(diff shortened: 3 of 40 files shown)


coveobot (commented):
[Live demo available here](https://vaporqa.cloud.coveo.com/feature/PR-1846/index.html).

packages/vapor/scss/components/blankslate.scss
```diff
@@ -1,6 +1,8 @@
 .blankslate {
+    --color: var(--title-text-color);
```
aiheon, line 2: I'm wondering what'S the benefit to declare a var for `--color ` instead of using the `--title-text-color` directly in the classes? :eyes:

packages/vapor/scss/components/facet.scss
```diff
@@ -56,7 +56,9 @@
 }
 
 .facet-header-title {
-    color: var(--facet-header-text-color);
+    --color: var(--title-text-color);
```
aiheon, line 59: same here

packages/vapor/scss/components/labeled-value.scss
```diff
@@ -11,17 +11,22 @@
 
     .label {
         padding: 5px 0px 5px 5px;
-        color: var(--label-text-color);
+        color: var(--title-text-color);
```
aiheon, line 14: here you are using it directly in the classes :eyes:

aiheon (approved):

packages/vapor/scss/components/blankslate.scss
rosalie-liu: If I understand correctly, the new pattern is to use fewer variables and keep them as simple as possible. For instance:
.class {
    --color: var(--title-text-color);
    color: var(--color);
}
Then we could override --color inside the scope. By using --color, we can avoid too much repetition, --title-text-color was defined as variable because it's a default value and it will be reused a lot.

packages/vapor/scss/components/facet.scss
rosalie-liu: I think here we can use directly var(--title-text-color) since there is no need to override in such a small scope.

packages/vapor/scss/components/blankslate.scss
GermainBergeron: To add on what Rosalie said, we should define variables when they are overriden in a `mod`. This way, the mod only has to redefine the variable instead of having to re-declare the color.

It doesn't seem like much but if your component has some nested components then the new variable will impact all the children

packages/vapor/scss/components/facet.scss
GermainBergeron: We should only define `--color` if it's overriden by a mod

packages/vapor/scss/components/tab.scss
```diff
@@ -21,12 +23,14 @@
         transition: all 0.2s ease;
 
         &.active {
-            color: var(--tab-text-color);
+            color: var(--color);
```
GermainBergeron, line 26: You don't need this rule since it's defined in .tab

packages/vapor/scss/components/tab.scss
```diff
@@ -21,12 +23,14 @@
...
             border-bottom: $tab-bottom-border-height solid var(--tab-border-color);
         }
 
         &.disabled {
-            color: var(--medium-grey);
+            --color: var(--medium-grey);
+
+            color: var(--color);
```
GermainBergeron, line 33: You should only override the variable, no need to redefine the same rule that is already in .tab
```suggestion
            --color: var(--medium-grey);
```

packages/vapor/scss/controls/flat-select.scss
```diff
@@ -62,7 +64,7 @@
             }
 
             .icon {
-                fill: var(--flat-select-text-color);
+                fill: var(--color);
```
GermainBergeron, line 67: We should probably use `fill: currentColor;`
```suggestion
                fill: currentColor;
```

packages/vapor/scss/lib/_mixins.scss
```diff
@@ -23,7 +23,7 @@
     }
 }
 
-@mixin placeholder($text-color: var(--placeholder-text-color), $font-family: var(--main-font)) {
+@mixin placeholder($text-color: var(--default-text-color), $font-family: var(--main-font)) {
```
GermainBergeron, line 26: The placeholder should probably keep its color, no?

GermainBergeron (approved):

packages/vapor/scss/lib/_mixins.scss
rosalie-liu: I removed `--placeholder-text-color: var(--grey-80);`since these two are duplicate. 
https://github.com/coveo/react-vapor/pull/1846/files/36467a8f38acdd9b72c136752a10a5c505bd9f92#diff-9cbf75f3c25011738aa63f67ae9d3987443f2923120f40d4e21bd1753412d80dL229

PR #1859, opened 2021-02-04
Ad UI 6476/remove border color border x color from border scss

aiheon:
### Proposed Changes

I've removed the border-X-color util from Vapor. 

I've also made some changes in StatusCard component to limit the choice of possible color for the border-left-color.

To avoid breaking changes with the snapshot feature in the admin, i'm using hard coded value for now. 
image

image

At some point I would like the `color` prop to be replace by a `status` prop to be consistent.  Or maybe there's a better way to deal with this :thinking: 

None

### Acceptance Criteria


coveobot (commented):
[Live demo available here](https://vaporqa.cloud.coveo.com/feature/PR-1859/index.html).

packages/react-vapor/src/components/statusCard/styles/StatusCard.scss
```diff
@@ -21,6 +21,22 @@
     }
 }
 
+.isSuccess {
+    border-left-color: #2ba488; // var(--success-60)
```
gdostie, line 25: Why not use the variable?

rosalie-liu: Also, I'm wondering if it would be better to create classes like `.mod-success` and scope them into `.statusCard`. In this way, we could try something like `border-left-color: var(--border-left-color);` then `--border-left-color: var(--success-60)` :thinking:

aiheon: @gdostie  :no_mouth: ... be right back :sweat_smile: 

@rosalie-liu  Mmmh :thinking: Where would you put the .mod ? In the Border.scss? or in the statusCard.scss? If we can get rid of the styles.something thingy to use the classes i'm all for it :eyes:
I just don't know how to do it, I'll need to look into it and learn :nerd_face:

coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/PR-1859/index.html).

rosalie-liu (approved):

gdostie (approved):

PR #1860, opened 2021-02-08
fix(drop): revert the revert of the fix for drop

GermainBergeron:
### Proposed Changes

The fix is actually working, it simply need some tweaking in the admin to make sure the focus don't shift under the body

Requires coveo/admin-ui#1748 to be merged first

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/drop/tests/DropPod.spec.tsx
+++ b/packages/react-vapor/src/components/drop/tests/DropPod.spec.tsx
@@ -162,26 +162,56 @@ describe('DropPod', () => {
                         Defaults.DROP_ROOT = 'body';
                     });
 
-                    it('should create a portal with the Defaults.DROP_ROOT if no selector is passed in props', () => {
+                    it('should create a div element and create a portal with it', () => {
+                        const divElement = document.createElement('div');
+                        jest.spyOn(document, 'querySelector');
+                        const portalSpy = jest.spyOn(ReactDOM, 'createPortal');
+
+                        shallow(<DropPod renderDrop={() => '🍟'} ref={buttonRef} />, {}).dive();
+
+                        expect(portalSpy).toHaveBeenCalledWith('🍟', divElement);
+                    });
+
+                    it('should append the div element to the Defaults.DROP_ROOT as the portal root if no selector is passed in props', () => {
+                        const divElement = document.createElement('div');
                         const expectedElement = document.querySelector('head');
                         Defaults.DROP_ROOT = 'head';
 
                         jest.spyOn(document, 'querySelector');
-                        const portalSpy = jest.spyOn(ReactDOM, 'createPortal');
+                        const appendChildSpy = jest.spyOn(expectedElement, 'appendChild');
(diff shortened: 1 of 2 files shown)


coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/PR-1860/index.html).

gdostie:
Should it be tagged as a breaking change? / merged in next?

GermainBergeron:
> Should it be tagged as a breaking change? / merged in next?

It's not breaking, we're only wrapping the drop inside a div which is removed when the drop component is unmounted. We're avoiding zombie elements that aren't controlled by anything

In the admin-ui case the issue is that the body has an height of 100%, no scrollbar and we render an input in some Drop that has the autoFocus=true property. This cause the body to scroll even if it shouldn't

gdostie (approved):

packages/react-vapor/src/components/drop/tests/DropPod.spec.tsx
```diff
@@ -162,27 +162,59 @@ describe('DropPod', () => {
...
-                        const portalSpy = jest.spyOn(ReactDOM, 'createPortal');
+                        const appendChildSpy = jest.spyOn(expectedElement, 'appendChild');
                         shallow(<DropPod renderDrop={() => '🍟'} ref={buttonRef} />, {}).dive();
 
-                        expect(portalSpy).toHaveBeenCalledWith('🍟', expectedElement);
+                        expect(appendChildSpy).toHaveBeenCalledWith(divElement);
                     });
 
-                    it('should create a portal with the selector prop if passed to the dropPod', () => {
+                    it('should append the div element to with the selector prop as the portal root if passed to the dropPod', () => {
```
aiheon, line 187:
```suggestion
                    it('should append the div element to the selector prop as the portal root if passed to the dropPod', () => {

```
i think?

packages/react-vapor/src/components/drop/tests/DropPod.spec.tsx
```diff
@@ -162,27 +162,59 @@ describe('DropPod', () => {
...
+                            <DropPod renderDrop={() => '🍟'} ref={buttonRef} />,
+                            {}
+                        ).dive();
+
+                        componentToUnmount.unmount();
+
+                        expect(removeChildSpy).toHaveBeenCalledWith(divElement);
+                    });
+
+                    // expect(ReactDOM.createPortal).toHaveBeenCalledWith('🍟', expectedElement);
```
aiheon, line 217: was this forgotten? :yum:

aiheon (approved):

packages/react-vapor/src/components/drop/tests/DropPod.spec.tsx
GermainBergeron: @sallainCoveo should I remove the comment or remove the line?

sallainCoveo: I think so, I don't remember what this line was supposed to be used for.

PR #1885, opened 2021-02-25
feat(button): center all the button/flatselect/popovers content

sallainCoveo:
### Proposed Changes
I (think) that now all the button/flatselect/popovers has its text is centered.

Here's some examples:

Menu with the svg aligned:
menu_button

FlatSelect:
flat_select

Popovers:
new_popover

Buttons with different sizes (please note that I added another word for the "small" button, I felt the current example didn't show cased the alignment):
button_sizes

The only one example I feel the alignment is not right, is the flatselect with a prepended or a appended svg. 
### Acceptance Criteria


coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/feat-ADUI-6646-fix-the-button-css-to-center-the-text/index.html).

FelixBlaisThon:
I've added @oliviertaillon to your pr so he can have his say on this. As for the FlatSelect, it's overlapping with my issue (color), so I will let you work with the alignement since it's more in the scope of yours, but it feels like the google icon is still too high, and the text a bit too, other than the flatSelect, it seems to be good overall :)

dmgauthier:
Screenshot from 2021-02-26 12-35-50
is the small button text vertical alignment one or two pixel too high  ?

sallainCoveo:
> [...] FlatSelect, [...] it feels like the google icon is still too high, and the text a bit too, other than the flatSelect, it seems to be good overall :)

I think I fix it, but let me know if the solution could be improved!
flatselect_svg_centered

sallainCoveo:
> Screenshot from 2021-02-26 12-35-50
> is the small button text vertical alignment one or two pixel too high ?

I think that even if we centered align, it always seems too high, so I changed the padding, here's the result:
small_button

However, let me know if another solution would be better in that case!

rosalie-liu:
> > is the small button text vertical alignment one or two pixels too high?
> I think that even if we centered align, it always seems too high, so I changed the padding, here's the result:
> However, let me know if another solution would be better in that case!

@sallainCoveo The alignment issue of the small button is caused by the line-height. I have a fix on it in this PR https://github.com/coveo/react-vapor/pull/1892 Basically, there is a conflict between the old line-height and the override value for rebranding.

kdang-coveo (approved):

rosalie-liu (approved):

PR #1890, opened 2021-02-26
fix(SFINT-3779): detect changes after clearing dirty flag

lbergeron:
### Proposed Changes

An issue was found regarding `withDirtyInputHOC` where a component would stop detecting user changes after clearing the `isDirty` flag. The root cause is that `withDirtyInputHOC` kept the last `isDirty` value in a local state and was not aware of the dispatched clear action. So, when user makes new changes to the input, the component thinks it is still dirty and does not dispatch the set dirty action.

In this PR, `withDirtyInputHOC` relies on the store to retrieve the last `isDirty` value instead of the local state. This way it stays synchronized.

None

### Acceptance Criteria

-   [ ] The potential breaking changes are clearly identified. Does not apply
-   [ ] [README.md](https://github.com/coveo/react-vapor/blob/master/README.md) is adjusted to reflect the proposed changes (if relevant) Does not apply

### DIFF
--- a/packages/react-vapor/src/components/validation/hoc/WithDirtyInputHOC.tsx
+++ b/packages/react-vapor/src/components/validation/hoc/WithDirtyInputHOC.tsx
@@ -1,15 +1,22 @@
 import * as React from 'react';
 import {connect} from 'react-redux';
+import {IReactVaporState} from 'src/ReactVapor';
 
 import {IDispatch} from '../../../utils/ReduxUtils';
 import {IInputOwnProps} from '../../input/Input';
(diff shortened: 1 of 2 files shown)


coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/fix/SFINT-3779/index.html).

francoislg (approved):
:+1: makes sense

louis-bompart (approved):
lgtm 👍

rosalie-liu (approved):

nathanlb (approved):

PR #1900, opened 2021-03-04
fix(infinite-scroll): use the infinite scroll only when the select is open

GermainBergeron:
### Proposed Changes

Use the infinite scroll only when the select is open
Removed all the tests that only get the props. Tested the actual behaviour instead

None that I can think of

### Acceptance Criteria


coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/fix-ADUI-6673-infinite-scroll/index.html).

rosalie-liu (approved):

FelixBlaisThon (approved):

PR #1901, opened 2021-03-04
fix(singleselect): change the selected value if the items change

francoislg:
[COM-891]

### Proposed Changes

I am really not sure about this fix, so I'm opening this PR as a discussion point :sweat: 

But first, let me explain the exact issue we got:

We use the `withDirtySingleSelectHOC` which exposes the `initialValue` prop. The `items` prop is iterated over to set the `selected` value according to the `initialValue`. So far, so good.

The issue happens when there is some asynchronicity in the thing. The `initialValue` is fetched asynchronously and the `items` are also fetched asynchronously.

So if we update either `initialValue` or `items`, the SingleSelect *stays unselected* because it only uses its *initial items* when rendering to update the state. Regardless of which comes back first, there will never be a selected value. (Note that the issue also occurs if you *update* wither `initialValue` or `items` for some reason)

So this here is an attempt to synchronize any `items` change.

When we explore the `React` component with the debugger, we see the `items` prop with the proper `selected: true` when there was nothing in the Select. I'm finding the new behavior less confusing overall, but I'm still not sure this is the right way :man_shrugging:  

----

Another solution could be to dispatch a single `setValue` to update the state when the `initialValue` changes, but it does not cover the case where you get the initialValue *first* and update the items afterward.

If anything used to rely on `items` update not doing anything, it could "clear" the currently selected value.

### Acceptance Criteria

[COM-891]: https://coveord.atlassian.net/browse/COM-891

### DIFF
--- a/packages/demo/src/components/examples/SingleSelectExamples.tsx
+++ b/packages/demo/src/components/examples/SingleSelectExamples.tsx
@@ -24,6 +24,7 @@ import {
     withDirtySingleSelectHOC,
     withNonEmptySingleSelectHOC,
     ValidationMessage,
+    Label,
 } from 'react-vapor';
 
 import * as _ from 'underscore';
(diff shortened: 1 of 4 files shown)


coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/fix-com-891-single-select-react-on-items-change/index.html).

GermainBergeron:
I'm curious about your use case. My first thought would have been to hide the dropdown (or the form) until all data is fetched from the backend. 

Imagine a really slow network, won't the experience be weird for the user if we display the dropdown has having no selected value (because the initialValue is not fetched yet) but the list of items is present. He opens the dropdown and now the call comes back with the initial value, what happens then?

The inverse could also be weird for a user, the initial value arrives first and he wants to update the dropdown but there's no item in the list?

It's one of those case where I think UX input is required before we even look at the code

francoislg:
Too risky of a change, will not do.

For reference, here is the comment that spawned this experiment:  https://github.com/coveo/admin-ui/pull/1933#issuecomment-791442314

PR #1908, opened 2021-03-15
feat(modal-wizard): adding onFinishAsync prop to modal wizard

kdang-coveo:
image

### Proposed Changes
Added a new prop `onFinishAsync` to ModalWizard component to prevent closing of the modal if the finish method encounters an error.

The reason for this change is because without having this option, when a function errors out, it will close the modal and reset the users progress. Having this `onFinishAsync` will allow the modal to stay open. 

E.g
When a user enters a duplicate name for the Web Source, the original `onFinish` method will throw an error and automatically run the `close()` method from the ModalWizard since the method is async. Having this prop will allow the function to complete and then close the modal upon clicking `Finish`

`onFinishAsync()?.then(close);`

### How to test:
1. Go to the ModalWizard component
2. Test the second modal with validation and onFinishAsync
3. When completing the modal, pressing the `Finish` will open an alert saying that it cannot finish since an error was thrown.

https://user-images.githubusercontent.com/66333175/111213023-b7d22500-85a6-11eb-90dd-5affc87a12fd.mp4

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/modalWizard/ModalWizard.tsx
+++ b/packages/react-vapor/src/components/modalWizard/ModalWizard.tsx
@@ -2,6 +2,7 @@ import classNames from 'classnames';
 import * as React from 'react';
 import {connect} from 'react-redux';
 
+import _ from 'underscore';
 import {ConnectedProps, IDispatch, TooltipPlacement} from '../../utils';
 import {Button} from '../button';
 import {ModalActions} from '../modal/ModalActions';
@@ -90,7 +91,7 @@ const ModalWizardDisconneted: React.FunctionComponent<ModalWizardProps & Connect
                                             onFinish?.();
                                             close();
                                         } else if (onFinishAsync) {
-                                            onFinishAsync()?.then(close);
+                                            onFinishAsync().then(close).catch(_.noop);
                                         }
                                     } else {
                                         setCurrentStep(currentStep + 1);
--- a/packages/react-vapor/src/components/modalWizard/ModalWizard.tsx
+++ b/packages/react-vapor/src/components/modalWizard/ModalWizard.tsx
@@ -13,6 +13,7 @@ export interface ModalWizardProps
     extends Omit<IModalCompositeOwnProps, 'modalBodyChildren' | 'validateShouldNavigate'> {
     id: string;
     onFinish?: () => unknown;
+    onFinishAsync?: () => Promise<void>;
     validateStep?: (currentStep: number, isLastStep?: boolean) => {isValid: boolean; message?: string};
     isDirty?: boolean;
     cancelButtonLabel?: string;
@@ -29,6 +30,7 @@ const ModalWizardDisconneted: React.FunctionComponent<ModalWizardProps & Connect
     id,
     close,
     onFinish,
+    onFinishAsync,
     children,
     modalFooterChildren,
     validateStep,
@@ -84,8 +86,12 @@ const ModalWizardDisconneted: React.FunctionComponent<ModalWizardProps & Connect
                                 enabled={isValid}
                                 onClick={() => {
                                     if (isLastStep) {
-                                        onFinish?.();
-                                        close();
+                                        if (onFinish) {
+                                            onFinish?.();
+                                            close();
+                                        } else if (onFinishAsync) {
+                                            onFinishAsync()?.then(close);
+                                        }
                                     } else {
                                         setCurrentStep(currentStep + 1);
                                     }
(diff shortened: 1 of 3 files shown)


coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/modal-wizard-improvement-onFinishAsync/index.html).

aiheon:
the checkbox is broken :troll:
image

aiheon (approved):

kdang-coveo:
> the checkbox is broken :troll:
> image

Fixed :D

sallainCoveo (approved):

packages/react-vapor/src/components/modalWizard/ModalWizard.tsx
```diff
@@ -13,6 +14,7 @@ export interface ModalWizardProps
     extends Omit<IModalCompositeOwnProps, 'modalBodyChildren' | 'validateShouldNavigate'> {
     id: string;
     onFinish?: () => unknown;
+    onFinishAsync?: () => Promise<void>;
```
francoislg, line 17: :thinking: `onFinishAsync` looks like it is something executed on finish that does not require to be synchronous. It does not imply "validating" the modal to me.

Should this be named `onFinishValidation`? In fact, can this feature be supported by the `validateStep` method instead? :thinking:

packages/react-vapor/src/components/modalWizard/ModalWizard.tsx
```diff
@@ -13,6 +14,7 @@ export interface ModalWizardProps
     extends Omit<IModalCompositeOwnProps, 'modalBodyChildren' | 'validateShouldNavigate'> {
     id: string;
     onFinish?: () => unknown;
```
francoislg, line 16: Since `onFinish` is returning `unknown` anyway, could we just allow using a promise for this one?

kdang-coveo: I think this gives the component more flexibility if the user wanted to create a toast after its been completed for an alert and have the modal close instantly.

Another reason for not enforcing the `onFinish` to be a promise is that this would be considered a breaking change as well.

packages/react-vapor/src/components/modalWizard/ModalWizard.tsx
kdang-coveo: Interesting thought! I'm not sure if it can be featured by the `validationStep` method because we would need to trigger the promise to return something to let us know that the last step cannot be completed. Perhaps @gdostie could chime in on this one.

I think the name `onFinishValidation` might be a little misleading as we are not really validating something but to wait for a promise to finish. The hope is that we catch an error that might come from the back end to prevent the modal from closing prematurely and waiting til the `onFinishAsync` method is complete.

packages/react-vapor/src/components/modalWizard/ModalWizard.tsx
francoislg: You don't have to enforce a promise :thinking: 
you can execute `onFinish`, get the return value

francoislg: You don't have to enforce a promise :thinking: 

you can execute `onFinish`, get the return value

packages/react-vapor/src/components/modalWizard/ModalWizard.tsx
francoislg: My main issue with this is the fact that `onFinishAsync` overlaps a bit too much with the other properties and it can confuse our users.

If I end up using the component, I won't know at first glance that:

* `onFinishAsync` can throw an error to cancel the close action
* `onFinishAsync` will **not** be executed if I set `onFinish`
* `onFinish` *can't* cancel the close action like `onFinishAsync`

I really think it should be included in `onFinish`. Having a second attribute would just cause more confusion. I think we can do this in a non-breakable way, but even then, it is a pretty new component we might even be able to afford a breaking change.

gdostie:
I see that having a promise complicates things a lot. Any thoughts on having

```tsx
onFinish: (close: () => void) => void;
```

That way it becomes the responsibility of the ModalWizard user to close the wizard when he sees fit.

kdang-coveo:
> I see that having a promise complicates things a lot. Any thoughts on having
> 
> ```tsx
> onFinish: (close: () => void) => void;
> ```
> 
> That way it becomes the responsibility of the ModalWizard user to close the wizard when he sees fit.

hmm this actually makes a lot of sense... I like this idea!

PR #1913, opened 2021-03-17
refactor(modal-wizard): changing the way close ModalWizard on finish

kdang-coveo:
Removing the closing method in the last step of the modal and passing it as a callback for the onFinish function.
The user will be responsible for calling the callback to close the modal.

BREAKING CHANGE: changing the way we close the modal - the user will need to dispatch a closeModal
action at the end of the onFinish method. An example is updated to show the close method being used
after onFinish (after the alert)

### Proposed Changes

This will allow the user to close the modal after an onFinish method is completed.

How to test:

1. Go to ModalWizard and upon finish, the modal will close. (method close() was added after alert which dispatches the closeModal action)

There are no changes to the behaviour. Modal should close as expected when finishing modal.

- ModalWizard will stay open after the onFinish function is triggered unless the callback `close()` is called.

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/modalWizard/ModalWizard.tsx
+++ b/packages/react-vapor/src/components/modalWizard/ModalWizard.tsx
@@ -12,7 +12,7 @@ import {StepProgressBar} from '../stepProgressBar';
 export interface ModalWizardProps
     extends Omit<IModalCompositeOwnProps, 'modalBodyChildren' | 'validateShouldNavigate'> {
     id: string;
-    onFinish?: () => unknown;
+    onFinish?: (close: () => void) => unknown;
     validateStep?: (currentStep: number, isLastStep?: boolean) => {isValid: boolean; message?: string};
     isDirty?: boolean;
     cancelButtonLabel?: string;
@@ -84,8 +84,7 @@ const ModalWizardDisconneted: React.FunctionComponent<ModalWizardProps & Connect
                                 enabled={isValid}
                                 onClick={() => {
                                     if (isLastStep) {
-                                        onFinish?.();
-                                        close();
+                                        onFinish?.(close);
                                     } else {
                                         setCurrentStep(currentStep + 1);
                                     }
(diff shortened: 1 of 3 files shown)


packages/react-vapor/src/components/modalWizard/ModalWizard.tsx
```diff
@@ -12,7 +12,7 @@ import {StepProgressBar} from '../stepProgressBar';
 export interface ModalWizardProps
     extends Omit<IModalCompositeOwnProps, 'modalBodyChildren' | 'validateShouldNavigate'> {
     id: string;
-    onFinish?: () => unknown;
+    onFinish?: (close: () => void) => void;
```
gdostie, line 15: onFinish could genuinely return any type and we don't really care, so there is no point of enforcing it to be void.

```suggestion
    onFinish?: (close: () => void) => unknown;
```

packages/react-vapor/src/components/modalWizard/tests/ModalWizard.spec.tsx
```diff
@@ -90,7 +90,28 @@ describe('ModalWizard', () => {
         userEvent.click(screen.getByRole('button', {name: 'Finish'}));
 
         expect(finishSpy).toHaveBeenCalledTimes(1);
+        expect(screen.getByRole('dialog')).toBeInTheDocument();
+    });
+
+    it('calls the "onFinish" prop and the modal closes when clicking on the "finish" button', async () => {
```
gdostie, line 96: 👏

gdostie (approved):

coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/modal-wizard-close-refactor/index.html).

francoislg (approved):
Great! I like better this design :D

aiheon (approved):

sallainCoveo (approved):

PR #1918, opened 2021-03-22
Cypress integration setup

pomorinCoveo:
Initial setup for Cypress test runner.

To start cypress run `npm install` then `npx cypress open`


coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/feature/cypress/UITEST-74-Create-Cypress-reposiory/index.html).

gdostie (commented):
Have you looked at https://testing-library.com/docs/cypress-testing-library/intro? I think it could one interesting thing to have since we're going the testing library route for the unit tests.

packages/cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
```
gdostie, line 1: Any thoughts about naming our end to end tests package something like `e2e` instead of the technology we use (`cypress`)

packages/cypress/integration/Vapor/ActionableItem.spec.js
```diff
@@ -0,0 +1,43 @@
...
+    cy.visit('http://192.168.0.143:8080/#/components/ActionableItem')
+  })
+  it('Validate Actionnable item without onItemClick Method',() => {
+    cy.get('#actionable-item-1 div[class*="actionable-item-content"]')
+      .should('not.have.class', 'cursor-pointer')
+  })
+  it('Validate dot menu on Actionnable item without onItemClick', () => {
+    const stub = cy.stub();
+    cy.on('window:alert', stub)
+    cy.get('#actionable-item-1 div[class*="actionable-item-dots"]')
```
gdostie, line 14: Getting elements of the page using such precise selectors will make the tests very prone to breaking. I think using the testing library I linked above will make these much more robust

pomorinCoveo: Will be looking at the testing librarie thanks for the head's up!

packages/cypress/fixtures/example.json
pomorinCoveo: would still have to name a folder `cypress` in order to use the framework but i put it into a e2eTesting folder

pomorinCoveo:
Since i wasn't able to make a good description on my last commit, this is what's new:
- swapped `cypress` folder into `e2eTesting` folder.
- removed initial test files since this PR is meant for the initial setup and not for producing tests.
- imported cypress-testing-librarie inside `command.js` file.

packages/cypress/fixtures/example.json
gdostie: oo I wasn't aware of that, nvm mind my comment then!

dmgauthier (approved):
Configuration jobs are often daunting :clap: , good work !

package.json
```diff
@@ -6,8 +6,10 @@
         "@commitlint/config-conventional": "8.3.4",
         "@commitlint/travis-cli": "8.3.5",
         "@sindresorhus/slugify": "0.11.0",
+        "@testing-library/cypress": "^7.0.4",
         "aws-sdk": "2.576.0",
-        "axios": "0.19.0",
+        "axios": "^0.21.1",
```
gdostie, line 11: What do we need axios for?

gdostie (approved):

package.json
pomorinCoveo: i guess it got updated when i did a `npm install`

PR #1926, opened 2021-03-29
fix(calendar): added condition to default to current year if bad data

j-mcgregor:
### Proposed Changes

The Datepicker works by providing a list of years and a startingYear / selectedYear index, which is used to select the year from the given list. However, if the given year index falls outside the range provided then it would break. I'm guessing this problem wasn't found before because normally pickiing the year is done in the UI, but when the Datepicker gets its data from the API it can be anything.

This PR adds a check to confirm the selected year (from the api or the user) is present in the years array. If it isn't then it defaults to the current year (UX didn't give me an answer on whether this is the best approach so if its flawed let me know)

[ADUI-6789](https://coveord.atlassian.net/browse/ADUI-6789)

None

### Steps to reproduce

This can only really be tested locally as far as I know. I tried changing the values in Redux and React Dev tools in the live site but it didn't update the DOM. What I did was to include the following line in the `DatePickerExamples` in the Demo:

```tsx
<Section level={2} title="DatePickerBox">
    <DatePickerBox id="date-picker-box" withReduxState datesSelectionBoxes={SELECTION_BOXES} />
    <DatePickerBox id="date-picker-box-2" withReduxState startingYear={2038} /> // <-- I messed around with different values here to mimic API data
</Section>
```

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/calendar/Calendar.tsx
+++ b/packages/react-vapor/src/components/calendar/Calendar.tsx
@@ -56,7 +56,7 @@ export interface ICalendarProps extends ICalendarOwnProps, ICalendarStateProps, 
 export const DEFAULT_MONTHS: string[] = moment.months();
 
 export const DEFAULT_YEARS: string[] = [
-    ...DateUtils.getPreviousYears(10),
+    ...DateUtils.getPreviousYears(30),
     DateUtils.currentYear.toString(),
     ...DateUtils.getNextYears(30),
 ];
--- a/packages/react-vapor/src/components/calendar/Calendar.tsx
+++ b/packages/react-vapor/src/components/calendar/Calendar.tsx
@@ -58,7 +58,7 @@ export const DEFAULT_MONTHS: string[] = moment.months();
 export const DEFAULT_YEARS: string[] = [
     ...DateUtils.getPreviousYears(10),
     DateUtils.currentYear.toString(),
-    ...DateUtils.getNextYears(10),
+    ...DateUtils.getNextYears(30),
 ];
 
 export const DEFAULT_DAYS: string[] = moment.weekdaysShort();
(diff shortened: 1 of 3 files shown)


coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/bugfix/ADUI-6789-broken-date-picker-when-provided-with-bad-year/index.html).

aiheon (approved):
I think this is better than nothing ^^ 
Also I think we should not have to handle broken data stored in the backend but that's just my opinion :hankey: 

They could make the API reject invalid date and everyone win :see_no_evil:

packages/react-vapor/src/components/calendar/Calendar.tsx
```diff
@@ -195,9 +195,15 @@ export class Calendar extends React.Component<ICalendarProps, any> {
...
+
+        // this.props.startingYear is an index, so can't be more than length of this.props.years
+        const isStartingYearDefined = !!this.props.years[this.props.startingYear];
+
         const startingYear: number =
-            this.props.startingYear ||
+            (isStartingYearDefined && this.props.startingYear) ||
             (startingYearIndex >= 0 ? startingYearIndex : Math.floor(this.props.years.length / 2));
+
+        // debugger;
```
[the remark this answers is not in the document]
j-mcgregor: god dammit

aiheon (approved):

packages/react-vapor/src/components/calendar/Calendar.tsx
```diff
@@ -58,7 +58,7 @@ export const DEFAULT_MONTHS: string[] = moment.months();
 export const DEFAULT_YEARS: string[] = [
     ...DateUtils.getPreviousYears(10),
     DateUtils.currentYear.toString(),
-    ...DateUtils.getNextYears(10),
+    ...DateUtils.getNextYears(30),
```
dmgauthier, line 61: what is this doing and why is this changed to `30` while the `getPreviousYears` hasn't change ?

j-mcgregor: It extends the future years available to 30 years instead of 10 since the original issue saw a date come from the API of 2038. 10 years in the future wasn't enough but 50 feels like overkill. 30 was the middle ground. I don't know know why I didn't change the previous years but I guess it won't hurt

gdostie (approved):

PR #1928, opened 2021-03-31
fix(react-vapor): remove ts-transformer-keys

gdostie:
### Proposed Changes

Remove custom typescript transformer to generate arrays from interfaces.

Might have changed how some interfaces related to the tableWithNewPagination are exported, but the type of this interface was in error.

### Acceptance Criteria


coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/ADUI-6825/remove-ts-tranformer-keys/index.html).

packages/react-vapor/src/components/numericInput/NumericInputConnected.tsx
```diff
@@ -86,7 +85,20 @@ export class NumericInputConnected extends React.PureComponent<NumericInputProps
...
+                                'invalidMessage',
+                                'max',
+                                'maxLength',
+                                'min',
+                                'mount',
+                                'setValue',
+                                'step',
+                                'unmount',
+                                'value',
+                            ])}
```
j-mcgregor, line 101: For consistency, should be extract this array to `inputPropsToOmit` or something?

packages/react-vapor/src/components/chart/tests/ChartTooltipContent.spec.tsx
```diff
@@ -1,12 +1,27 @@
...
 
+jest.mock('react', () => {
+    const originReact = jest.requireActual('react');
+    return {
+        ...originReact,
+        useContext: jest.fn(),
+    };
+});
+
+const mockedReact = mocked(React);
```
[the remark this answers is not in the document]
gdostie: the first block actually mocks the thing, the second thingy makes typescript understand it is mocked

dmgauthier (approved):

packages/react-vapor/src/components/numericInput/NumericInputConnected.tsx
gdostie: Done

FelixBlaisThon (approved):

FelixBlaisThon:
It will probably sound basic, but why did we use this package before and why are we removing it now? was it only for its utility that we used it? if so, we do we remove it now?

Regards,

Félix Blais

gdostie:
> It will probably sound basic, but why did we use this package before and why are we removing it now? was it only for its utility that we used it? if so, we do we remove it now?
> 
> Regards,
> 
> Félix Blais

Very relevant question. Here are the pros and cons of using this lib I can think of that helps understanding the whys of this change.

#### Pros (or why it was added)

- Ease of use, it requires less effort to list the props we want to omit, because the compiler does it for us
- If an interface changes, we don't have to think about manually changing the array of props to omit

#### Cons (or why we're removing it)

- Adds complexity to the build process. We have to use a custom transformer to the typescript compiler to make it work.
- Slows down the transpilation because the compiler has to run an additional step
- When using this you cannot turn off the type checking to speed up the build (`transpileOnly: true`) or else it won't have information about the interfaces keys.
- Having the arrays of props to omit dynamic makes you less aware of what props are omitted or not (more error prone)

PR #1934, opened 2021-04-07
feat(code editor): value in store

dmgauthier:
### Proposed Changes

I gave the possibility for the `CodeEditor` to have a value in the redux store.

Basically, if you set the `id` prop. An action will be dispatched on change to set the value of the `CodeEditor` in the store.

You can check the Demo, the first `CodeEditor` displayed there should dispatch the update action on change.

Peek 2021-04-07 13-29

I debounced the redux call not to spam the store with new values on each stroke.

EDIT: upon unmount, the component will remove its ID from the store

We also have a selector to get the value from the codeEditor.

none

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/editor/CodeEditor.tsx
+++ b/packages/react-vapor/src/components/editor/CodeEditor.tsx
@@ -6,17 +6,21 @@ import 'codemirror/addon/search/matchesonscrollbar';
 import 'codemirror/addon/search/search';
 import 'codemirror/mode/python/python';
 
+import classNames from 'classnames';
 import * as CodeMirror from 'codemirror';
 import * as React from 'react';
 import * as ReactCodeMirror from 'react-codemirror2';
-
-import classNames from 'classnames';
 import {connect} from 'react-redux';
-import {CollapsibleSelectors} from '../collapsible/CollapsibleSelectors';
-import {CodeMirrorGutters} from './EditorConstants';
+import * as _ from 'underscore';
+
 import {IReactVaporState} from '../../ReactVapor';
+import {IDispatch} from '../../utils';
+import {CollapsibleSelectors} from '../collapsible/CollapsibleSelectors';
+import {CodeEditorActions} from './CodeEditorActions';
+import {CodeMirrorGutters} from './EditorConstants';
 
 export interface ICodeEditorProps {
+    id?: string;
     value: string;
     mode: any;
     readOnly?: boolean;
@@ -38,8 +42,13 @@ const mapStateToProps = (state: IReactVaporState, {collapsibleId}: ICodeEditorPr
     isCollapsibleExpanded: CollapsibleSelectors.isExpanded(state, collapsibleId),
 });
 
+const mapDispatchToProps = (dispatch: IDispatch, {id}: {id?: string}) => ({
+    updateStoreValue: (value: string) => dispatch(CodeEditorActions.updateValue(id, value)),
+    clearCodeEditorFromStore: () => dispatch(CodeEditorActions.remove(id)),
+});
+
 class CodeEditorDisconnect extends React.Component<
-    ICodeEditorProps & Partial<ReturnType<typeof mapStateToProps>>,
+    ICodeEditorProps & Partial<ReturnType<typeof mapStateToProps> & ReturnType<typeof mapDispatchToProps>>,
     CodeEditorState
 > {
     static defaultProps: Partial<ICodeEditorProps> = {
@@ -76,6 +85,9 @@ class CodeEditorDisconnect extends React.Component<
             this.editor.refresh();
             this.setState({numberOfRefresh: this.state.numberOfRefresh + 1});
         }
+        if (this.props.id) {
+            this.props.updateStoreValue(this.props.value);
+        }
     }
 
     componentDidUpdate(prevProps: ICodeEditorProps) {
@@ -88,6 +100,16 @@ class CodeEditorDisconnect extends React.Component<
             this.editor.getDoc().clearHistory();
         }
     }
+
+    componentWillUnmount() {
+        if (this.props.id) {
+            this.props.clearCodeEditorFromStore();
+        }
+    }
+
+    private debouncedUpdateStore = _.debounce((value: string) => {
+        this.props.updateStoreValue(value);
+    }, 500);
 
     render() {
         return (
@@ -103,6 +125,9 @@ class CodeEditorDisconnect extends React.Component<
                 value={this.state.value}
                 onChange={(editor, data, value: string) => {
                     this.props.onChange?.(value);
+                    if (this.props.id) {
+                        this.debouncedUpdateStore(value);
+                    }
                 }}
                 options={{
                     ...CodeEditorDisconnect.defaultOptions,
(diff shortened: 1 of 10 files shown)


packages/react-vapor/src/components/editor/tests/CodeEditor.spec.tsx
```diff
@@ -42,18 +44,6 @@ describe('CodeEditor', () => {
...
-
-            expect(readOnlyProp).toBeUndefined();
-
-            mountWithProps({readOnly: true});
-
-            readOnlyProp = wrapper.props().readOnly;
-
-            expect(readOnlyProp).toBe(true);
-        });
-
```
dmgauthier, line 56: useless

packages/react-vapor/src/components/editor/tests/CodeEditor.spec.tsx
```diff
@@ -133,5 +111,18 @@ describe('CodeEditor', () => {
...
+
+            const component = mountWithStore(<CodeEditor id="anId" value="a value" />, store);
+
+            const inputArea = component.find('div');
+
+            inputArea.simulate('change', {target: {value: 'new value'}});
+
+            expect(store.getActions().length).toBe(1);
+            expect(store.getActions()[0].type).toBe(CodeEditorActionTypes.update);
+        });
```
dmgauthier, line 126: Really wanted to use RTL. But couldn't find a way to get the dispatched store actions

coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/ADUI-6831-code-editor-store-value/index.html).

j-mcgregor (approved):

packages/react-vapor/src/components/editor/tests/CodeEditor.spec.tsx
gdostie: This is an implementation detail, that's why its hard to test it with RTL

packages/react-vapor/src/components/editor/CodeEditorActions.ts
```diff
@@ -0,0 +1,21 @@
+import {IReduxAction} from '../../utils';
+
+export const CodeEditorActionTypes = {
+    add: 'ADD_CODE_EDITOR',
+    remove: 'REMOVE_CODE_EDITOR',
```
gdostie, line 5: `add` and `remove` don't seem to be used, but they probably should (especially `remove` to remove the code editor entry from the store).

packages/react-vapor/src/components/editor/tests/CodeEditor.spec.tsx
dmgauthier: Checking the dispatching of the actions might be an implementation detail, but I dont see another way I could test that the component interacts with the store properly.

packages/react-vapor/src/components/editor/CodeEditorActions.ts
dmgauthier: :+1:  good point I'll add an action to clear the code editor from the store and call this action in the end of life cycle.

packages/react-vapor/src/components/editor/tests/CodeEditor.spec.tsx
aiheon: does the modify store reflect in any way in the ui?

packages/react-vapor/src/components/editor/tests/codeEditorSelectors.spec.ts
```diff
@@ -0,0 +1,21 @@
+import {IReactVaporState} from '../../../ReactVapor';
+import {CodeEditorState} from '../CodeEditorReducers';
+import {CodeEditorSelectors} from '../CodeEditorSelectors';
+
+describe('CodeEditorSelectors', () => {
+    describe('getValue', () => {
+        it('should not throw and return an empty string when passing a falsy id', () => {
```
aiheon, line 7:
```suggestion
        it('should return an empty string when passing a falsy id', () => {
```
nitpick but since you return an empty string, we know it does not throw :yum:

aiheon (approved):

packages/react-vapor/src/components/editor/tests/CodeEditor.spec.tsx
dmgauthier: no. the behavior is the same. But imo its important to know the component is dispatching actions

gdostie: One way to test it would be to create a fixture component conencted to the store that uses the code editor value and render that component next to the CodeEditor in the RTL `render()`. For example a button with an onclick handler. Then in your test you could interact with that fixture component (for example click on the button) to assert that the code editor value is the correct one.

dmgauthier: Isn't that a bit extreme when the reducer and the selector are both tested independently ?.  

I know the confidence in the tests are now the new thing and I agree, but making sure the component is dispatching the actions wouldn't break unless someone refactors the component and changes the action names or doesn't call the actions at all. So IMO just having the capacity to access the store dispatched actions while implenting tests with RTL would give us the capacity to create tests based on the interaction of the component with the store without being flaky. Since the reducers and the selectors must be tested anyway.  

I know that this won't be the case in React vapor, but what if a component has to send a fetch with a thunkAction. I think that having the possibility to access the actions dispatched would be very helpfull then.

gdostie: I'm not saying that we should absolutely use that approach, I'm just explaining what would be the correct way to test it out using the react-testing-library approach. 

Another aspect to consider as well with that approach is that the reducers, actions and selectors wouldn't have to be tested independently because those tests would be redundant. Everything would already be covered by the CodeEditor's tests. In a full 100% RTL mindset, only the CodeEditor component functionality would be tested from a `react-vapor` library user standpoint. That means no store mock, no watching on actions being dispatched, and no assertions on the state directly.

packages/react-vapor/src/components/editor/CodeEditor.tsx
```diff
@@ -38,8 +42,15 @@ const mapStateToProps = (state: IReactVaporState, {collapsibleId}: ICodeEditorPr
     isCollapsibleExpanded: CollapsibleSelectors.isExpanded(state, collapsibleId),
 });
 
+const mapDispatchToProps = (dispatch: IDispatch, {id}: {id?: string}) => ({
+    updateStoreValue: _.debounce((value: string) => {
+        dispatch(CodeEditorActions.updateValue(id, value));
+    }, 500),
```
gdostie, line 48: I think this will create a new debounced function each time the state changes, but I might be mistaken. A safer approach is to debounce the prop from inside the component's implementation.

gdostie (approved):

packages/react-vapor/src/components/editor/tests/CodeEditor.spec.tsx
dmgauthier: Yes I agree that by using this approach. All the tests we do for the reducers and selectors and so on wouldn't be that useful.

 I am willing to chat about that if we intent to completely ditch the black box concept of our test. I feel that a quick strategies sharing and brainstorm could help us save some time in the future.

packages/react-vapor/src/components/editor/CodeEditor.tsx
dmgauthier: ok I'll do that

kdang-coveo (approved):

PR #1945, opened 2021-04-13
feat(radio-card): added radio card variant

nathanlb:
### Proposed Changes
- Added a RadioCard component to enable usage of the RadioSelect component with cards instead of checkboxes

FYI: This has been developed with members of the UX team. 

#### Sample
https://user-images.githubusercontent.com/16785453/114727369-27554480-9d0c-11eb-8597-13a5a78fd448.mov

This is born from the task to create a ModalSelect component to be used when offering a choice between detailed options.
To avoid confining this type of radio button to this modal component I built it separately. 
I thought, for example, that it might be used in a ModalWizard scenario.

Screen Shot 2021-04-13 at 2 40 57 PM

Don't think there are any

### Acceptance Criteria

### DIFF
--- a/packages/demo/src/components/examples/RadioExamples.tsx
+++ b/packages/demo/src/components/examples/RadioExamples.tsx
@@ -152,7 +152,7 @@ const RadioCardSelectExample: React.FunctionComponent = () => (
         <RadioSelectConnected
             id="radio-card-select-example"
             valueOnMount={'blue'}
-            className="flex flex-wrap spaced-boxes-container"
+            className="flex flex-wrap"
             disabledTooltip="you see me because of the disabledTooltip prop"
             disabledValuesOnMount={['red']}
         >
--- a/packages/demo/src/components/examples/RadioExamples.tsx
+++ b/packages/demo/src/components/examples/RadioExamples.tsx
@@ -125,22 +125,22 @@ const RadioSelectDisabledExample: React.FunctionComponent = () => (
 );
 
 const RadioSelectWrappedExample: React.FunctionComponent = () => (
-    <Section level={3} title="A radio select with a wrapped radio button.">
+    <Section level={3} title="A radio select with a wrapped radio button">
         <RadioSelectConnected id="addRankingResult" valueOnMount={'blue'}>
             <Radio {...blueRadioButtonProps}>
                 <Label>
-                    <span className="bold">{'Blue color'}</span>
+                    <span className="bold">Blue color</span>
                 </Label>
                 <InputDescription>
-                    <div style={{...paragraphStyle}}>{'Blue is the best color.'}</div>
+                    <div style={{...paragraphStyle}}>Blue is the best color.</div>
                 </InputDescription>
             </Radio>
             <Radio {...redRadioButtonProps}>
                 <Label>
-                    <span className="bold">{'Red color'}</span>
+                    <span className="bold">Red color</span>
                 </Label>
                 <InputDescription>
-                    <div style={{...paragraphStyle}}>{'Red is the best color.'}</div>
+                    <div style={{...paragraphStyle}}>Red is the best color.</div>
                 </InputDescription>
             </Radio>
         </RadioSelectConnected>
@@ -148,10 +148,10 @@ const RadioSelectWrappedExample: React.FunctionComponent = () => (
 );
 
 const RadioCardSelectExample: React.FunctionComponent = () => (
-    <Section level={3} title="A radio card select.">
+    <Section level={3} title="A radio card select">
         <RadioSelectConnected
             id="radio-card-select-example"
-            valueOnMount={'blue'}
+            valueOnMount="blue"
             className="flex flex-wrap"
             disabledTooltip="you see me because of the disabledTooltip prop"
             disabledValuesOnMount={['red']}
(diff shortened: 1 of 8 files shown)


coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/feat/SFINT-3851-radio-card/index.html).

packages/react-vapor/src/components/radio/RadioCard.tsx
```diff
@@ -0,0 +1,49 @@
...
+    return props.disabled && props.disabledTooltip ? (
+        <label onClick={props.onClick} className={containerClasses}>
+            <Tooltip title={props.disabledTooltip} placement={TooltipPlacement.Bottom}>
+                <RadioCardContent props={props} classes={classes} />
+            </Tooltip>
+        </label>
+    ) : (
+        <label onClick={props.onClick} className={containerClasses}>
+            <RadioCardContent props={props} classes={classes} />
+        </label>
```
gdostie, line 22: Since `props.onClick` is set on the label, it probably shouldn't be passed down to `RadioCardContent`. Otherwise it may end-up in the onClick call back being called more than once per click.

```tsx
// before the return
const {onClick, ...otherProps} = props;

// inside the return
<label onClick={onClick} className={containerClasses}>
    <RadioCardContent props={otherProps} classes={classes} />
</label>

```

packages/react-vapor/src/components/radio/RadioCard.tsx
```diff
@@ -0,0 +1,49 @@
...
+    );
+};
+
+const RadioCardContent: React.FunctionComponent<{props: RadioCardProps; classes: string}> = ({props, classes}) => {
+    // const inputProps = omit(props, ['children', 'classes']);
```
gdostie, line 27:
```suggestion
```

packages/react-vapor/src/components/radio/RadioCard.tsx
```diff
@@ -0,0 +1,49 @@
...
+    const inputProps = {
+        type: 'radio',
+        checked: props.checked,
+        id: props.id,
+        name: props.name,
+        className: 'card-input',
+        disabled: props.disabled,
+        onChange: (event: React.ChangeEvent<HTMLInputElement>) => props.onChange(event.target.value),
+    };
```
gdostie, line 36: You can move those props directly inside the jsx markup, the const doesn't seem to be reused anyway and doing this removes the provided autocompletion for input props.

packages/react-vapor/src/components/radio/RadioCard.tsx
```diff
@@ -0,0 +1,49 @@
...
+            </Tooltip>
+        </label>
+    ) : (
+        <label onClick={props.onClick} className={containerClasses}>
+            <RadioCardContent props={props} classes={classes} />
+        </label>
+    );
+};
+
+const RadioCardContent: React.FunctionComponent<{props: RadioCardProps; classes: string}> = ({props, classes}) => {
```
gdostie, line 26:
```suggestion
const RadioCardContent: React.FunctionComponent<RadioCardProps & {classes: string}> = ({classes, checked, id, name, disabled, onChange, children}) => {
```

packages/vapor/scss/controls/radios.scss
```diff
@@ -122,6 +122,50 @@
...
+
+    .radio-card {
+        display: flex;
+        flex-direction: column;
+        height: 100%;
+        padding: 20px;
+        border: $default-border;
+        border-radius: 8px;
+        box-shadow: 0px 2px 8px var(--shadow-color);
+        transition: all 0.3s ease-in-out;
```
gdostie, line 137: If you add the [`card` class](https://github.com/coveo/react-vapor/blob/master/packages/vapor/scss/components/card.scss#L17-L23) to the RadioCard component, we could ditch some of the redundant style here.

```suggestion
    .radio-card {
        display: flex;
        flex-direction: column;
        height: 100%;
        padding: 20px;
        transition: all 0.3s ease-in-out;
```

nathanlb: That's what i did at first but because of how sass resolves extensions, it was hard to get the order of precedence I wanted for the hover effect to work properly.

nathanlb: I was using @extends though so maybe if i just add the class to the element it would work

nathanlb: It does!

gdostie (approved):

FelixBlaisThon (approved):
good stuff :tada:

packages/demo/src/components/examples/RadioExamples.tsx
```diff
@@ -138,3 +146,43 @@ const RadioSelectWrappedExample: React.FunctionComponent = () => (
         </RadioSelectConnected>
     </Section>
 );
+
+const RadioCardSelectExample: React.FunctionComponent = () => (
+    <Section level={3} title="A radio card select.">
```
FelixBlaisThon, line 151: level={2} to be consistent with the others and remove the dot at the end of the title. (and at the same time, can you do the same thing for the section over this one (line 128, `<Section level={3} title="A radio select with a wrapped radio button.">`, level 2 and remove the dot in the title. Thanks!

PR #1946, opened 2021-04-14
fix(badges): new border colors to make the badge stand out

aiheon:
### Proposed Changes

[Mockup here](https://www.figma.com/file/JbXRfSOwdEuq3VJGpUyyiJ/Coveo-Brand-Library?node-id=1066%3A152)
[Jira here](https://coveord.atlassian.net/browse/ADUI-6814)

Une pierre 2 coups for the badge colors.
UX was ready with the new colors for the badge so I've made all the changes while I was in there.
The new border color will make the badge visible in the selected source row in the source table :tada: 

Before ---> 
image

After --->
image

None

### Acceptance Criteria

### DIFF
--- a/packages/vapor/scss/components/badge.scss
+++ b/packages/vapor/scss/components/badge.scss
@@ -56,8 +56,14 @@
 
     &.mod-success {
         --color: var(--semantic-success-100);
-        --border-color: var(--dark-green-30);
+        --border-color: var(--forest-green-30);
         --background-color: var(--semantic-success-0);
+    }
+
+    &.mod-beta {
+        --color: var(--sienna-orange-70);
+        --border-color: var(--sienna-orange-40);
+        --background-color: var(--sienna-orange-20);
     }
 
     &:last-child {
--- a/packages/vapor/scss/components/badge.scss
+++ b/packages/vapor/scss/components/badge.scss
@@ -14,7 +14,7 @@
     --height: 32px;
     --padding: 6px 16px;
     --font-size: 16px;
-    --line-height: 20px;
+    --line-height: 18px;
     --color: var(--grey-80);
     --border-color: var(--grey-50);
     --background-color: var(--grey-20);
@@ -33,7 +33,7 @@
         --height: 24px;
         --padding: 4px 8px;
         --font-size: 12px;
-        --line-height: 16px;
+        --line-height: 14px;
     }
 
     &.mod-critical {
--- a/packages/vapor/scss/components/badge.scss
+++ b/packages/vapor/scss/components/badge.scss
@@ -16,7 +16,7 @@
     --font-size: 16px;
     --line-height: 20px;
     --color: var(--grey-80);
-    --border-color: var(--grey-20);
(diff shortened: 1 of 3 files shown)


coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/ADUI-6814/fix-default-badge-color/index.html).

packages/vapor/scss/components/badge.scss
```diff
@@ -37,27 +37,27 @@
     }
 
     &.mod-critical {
-        --color: var(--critical-100);
-        --border-color: var(--critical-20);
-        --background-color: var(--critical-20);
+        --color: var(--semantic-critical-100);
```
gdostie, line 40: I think it would be wiser to change the `--critical-*`, `--warning-*`, `--information-*` and `--success-*` colors instead if it makes sense

gdostie (commented):
While at it do you think it would be possible to fix the alignment issue? It seems too close to the bottom.
image

packages/vapor/scss/components/badge.scss
aiheon: > While at it do you think it would be possible to fix the alignment issue? It seems too close to the bottom.
> image

yes :smile: :yum: 
https://coveo.slack.com/archives/GLS6Z9RPD/p1618410483055600?thread_ts=1618337997.050200&cid=GLS6Z9RPD

aiheon: > I think it would be wiser to change the `--critical-*`, `--warning-*`, `--information-*` and `--success-*` colors instead if it makes sense

I did not want to change the colors to much since they are used a lot everywhere :sweat_smile: 
If those are the final* colors, then I think we need a user story to change them all :thinking: 

*_subject to change_  :rofl:

gdostie: Yes I understand the concern. I think it would be worth confirming with the designers that this is what we want, because reverting back to variables with specific color names feels like a regression in terms of scalability. The goal what to have variable names that can stand theme changes.

aiheon: Yeah i'll ask :+1:

aiheon: I've done some tests and for information-XX there's other components that the colors is changing also. I feel like this is getting out of scope :see_no_evil: 

I'll open a Jira to get this sorted out properly, I don't want to do it more than once :laughing:

aiheon: https://coveord.atlassian.net/browse/ADUI-6881

aiheon:
**Note to self**
We can use the new .mod-beta after the RV bump in admin-ui :eyes: 
image

gdostie (approved):

packages/vapor/scss/components/badge.scss
gdostie: Sounds good!

j-mcgregor (approved):

PR #1950, opened 2021-04-16
feat(cypress): make setup more resilient

GermainBergeron:
### Proposed Changes

Added a basic test and some commands (`pnpm cy:open` `pnpm cy:run`)
Moved cypress dependencies to the package
Setup for Typescript (and converted the existing files)

The actual PRs will need to be converted to typescript (just renaming the file from `js` to `ts` should do it)


coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/fix/cypress/UITEST-205-update-setup/index.html).

package.json
```diff
@@ -44,10 +42,12 @@
         "setup": "pnpm install",
         "start": "pnpm build --stream --filter=coveo-styleguide && pnpm -r start --parallel --filter=\\!react-vapor",
         "test": "pnpm -r test --parallel",
+        "cy:open": "pnpm -r cy:open --stream",
+        "cy:run": "pnpm -r cy:run --stream",
         "build": "pnpm -r build --stream",
         "lintfix": "pnpm -r lintfix --parallel",
         "commit-cli": "git-cz",
-        "reconstruct": "npx rimraf node_modules && npm run setup && echo done"
+        "reconstruct": "npx rimraf node_modules && npx rimraf packages/*/node_modules && npm run setup && echo done"
```
gdostie, line 50: I think you can pass a space delimited list of things to delete to rimraf.

```suggestion
        "reconstruct": "npx rimraf node_modules packages/*/node_modules && npm run setup && echo done"
```

package.json
```diff
@@ -44,10 +42,12 @@
         "setup": "pnpm install",
         "start": "pnpm build --stream --filter=coveo-styleguide && pnpm -r start --parallel --filter=\\!react-vapor",
         "test": "pnpm -r test --parallel",
+        "cy:open": "pnpm -r cy:open --stream",
+        "cy:run": "pnpm -r cy:run --stream",
```
gdostie, line 46: Should these really be on the root package.json? Maybe we could have only one of them like `test:e2e`

packages/e2eTesting/cypress.json
```diff
@@ -1 +1,15 @@
...
+  "pluginsFile": "plugins/index.ts",
+  "supportFile": "support/index.ts",
+  "videosFolder": "videos",
+  "screenshotsFolder": "screenshots",
+  "reporter": "junit",
+  "reporterOptions": {
+    "mochaFile": "reports/junit-[hash].xml"
+  },
+  "viewportWidth": 1366,
+  "viewportHeight": 768
```
gdostie, line 14: Do we really want a viewport that small?

packages/e2eTesting/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "e2etesting",
+  "private": true,
+  "version": "15.0.3",
+  "description": "Functionnal tests for react-vapor",
```
gdostie, line 5:
```suggestion
  "description": "End to end tests for react-vapor",
```

gdostie (approved):

packages/e2eTesting/cypress.json
GermainBergeron: I just copied what commerce use in the admin, it looks like a laptop resolution though. I don't mind changing it in the future if Gab or PO have any problem with it

package.json
GermainBergeron: It would be easier for dev/jenkins to start the command if it's on the root package

What do you think of `e2e:open` and `e2e:run` ?

gdostie: What's the use-case for opening but not running?

GermainBergeron: it opens a window where you can choose what you run, you see the browser run, etc. 

Jenkins will run `e2e:run`
a dev working on e2e tests will probably run `e2e:open`
a dev working on react-vapor might run `e2e:run` to make sure they didn't break anything

aiheon (approved):

PR #1952, opened 2021-04-20
Bundle demo from tsc --build output

gdostie:
### Proposed Changes

Since the upgrade to PNPM, webpack ts-loader would compile with multiple type errors which prevented rebuilds. `tsc --build` is working without those errors, so I figured we could just build the demo from the `tsc` output instead of compiling the typescript files through webpack.

Also added a command at root: `clean`. It removes all dist folders.

None

### Acceptance Criteria


packages/demo/webpack.config.js
```diff
@@ -11,7 +10,7 @@ const CopyPlugin = require('copy-webpack-plugin');
  */
 module.exports = {
     entry: {
-        main: './src/Index.tsx',
+        main: './built/Index.js',
```
aiheon, line 13: :thinking: Does this means that the build is done before webpack start instead for webpack doing the building? :eyes:

.gitignore
```diff
@@ -23,6 +23,7 @@ travis-deploy-key
 
 # distribution
 **/dist
+**/built
```
aiheon, line 26: /built is for the demo only?

packages/demo/webpack.config.js
gdostie: Yes, TypeScripts compiler runs before webpack to convert .ts files to .js and outputs the result into `built` folder. Then webpack bundles everything together from those .js file inside `built` folder.

.gitignore
gdostie: yep, the only `built` folder we have is in the demo package

coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/ADUI-6850/bundle-demo-from-tsc-output/index.html).

packages/demo/webpack.config.js
aiheon: Does this slow down the hot reload time a bit? :eyes: Since you have a two step build kinda :thinking: 
Or tsc is so fast anyway it's not an issue? :laughing:

GermainBergeron (approved):

packages/demo/webpack.config.js
gdostie: In theory it should indeed be slower, but in practice when I tried it I didn't notice any difference in speed 🤷

aiheon: Good to know :sweat_smile: :woman_shrugging: :rocket:

aiheon (approved):

PR #1956, opened 2021-04-22
feat: 🎸 Validate Checkbox

pomorinCoveo:
Validate checkbox from http://192.168.0.168:8080/#/components/Checkbox

### DIFF
--- a/packages/e2eTesting/integration/checkbox.spec.ts
+++ b/packages/e2eTesting/integration/checkbox.spec.ts
@@ -1,4 +1,3 @@
-/// <reference types="cypress" />
 context('Actions', () => {
   before(() => {
 	  cy.visit('localhost:8080/#/components/Checkbox')
--- a/packages/e2eTesting/integration/checkbox.spec.ts
+++ b/packages/e2eTesting/integration/checkbox.spec.ts
@@ -1,0 +1,75 @@
+/// <reference types="cypress" />
+context('Actions', () => {
+  before(() => {
+	  cy.visit('localhost:8080/#/components/Checkbox')
+  })
+	it('Validate checkbox is checked', ()=>{
+		cy.get('#checkbox1').should('not.be.checked')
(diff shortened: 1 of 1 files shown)


coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/feature/cypress/UITEST-110-Validate_checkbox/index.html).

packages/e2eTesting/integration/checkbox.spec.ts
```diff
@@ -0,0 +1,75 @@
+/// <reference types="cypress" />
```
gdostie, line 1: Normally typescript should be properly setup so that we don't have to add this annotation to each test file we create.
```suggestion
```

packages/e2eTesting/integration/checkbox.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+context('Actions', () => {
+  before(() => {
+	  cy.visit('localhost:8080/#/components/Checkbox')
+  })
+	it('Validate checkbox is checked', ()=>{
```
gdostie, line 6: The formatting of this file seems a little off. It seems as though Prettier is not setup properly to run automatically.

packages/e2eTesting/integration/checkbox.spec.ts
pomorinCoveo: Got it! will remove it from here and all the other tests.

packages/e2eTesting/integration/checkbox.spec.ts
pomorinCoveo: Yea.. my prettier doesnt work since pretty much the beginning of my internship i didnt remember this was that bad though 🤔

j-mcgregor (approved):

dmgauthier (approved):

PR #1969, opened 2021-04-26
fix(types): include types of dependencies we export

gdostie:
### Proposed Changes

We need to add the @types packages into our dependencies when we export types that directly reference them, and the package is not specified as a peer dep.

For example here when trying to specify the prop `marks` of the Slider component, we get told that this property doesn't exist.
image

When looking at the declarations provided by `react-vapor` we can see that `SliderProps` is `unknown` because indeed the types of rc-slider lib are never installed.
```ts
import 'rc-slider/assets/index.css';
import { SliderProps } from 'rc-slider';  //  <---  Here
import { RCTooltip } from 'rc-tooltip';
export declare enum AppendedValueSide {
    left = "LEFT",
    right = "RIGHT",
    both = "BOTH"
}
export interface SliderOwnProps extends SliderProps {
    id: string;
    initialValue?: number;
    crossingPoint?: number;
    tabIndex?: number[];
    onChange?: (rangeOutputValue: number) => any;
    customTooltip?: (value: any) => JSX.Element;
    appendValue?: boolean | AppendedValueSide;
    appendValueFormatter?: (value: number, side?: Exclude<AppendedValueSide, AppendedValueSide.both>) => React.ReactNode;
    tooltipStyle?: Partial<RCTooltip.Props>;
    hasTooltip?: boolean;
}
export declare const Slider: import("react-redux").ComponentClass<Pick<SliderOwnProps & {
    setOutputValue: (value: number) => import("../../utils/ReduxUtils").IReduxAction<import("./SliderActions").ISetSliderValuePayload>;
}, never> & SliderOwnProps> & {
    WrappedComponent: React.ComponentType<SliderOwnProps & {
        setOutputValue: (value: number) => import("../../utils/ReduxUtils").IReduxAction<import("./SliderActions").ISetSliderValuePayload>;
    }>;
};
//# sourceMappingURL=Slider.d.ts.map
```

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/package.json
+++ b/packages/react-vapor/package.json
@@ -33,6 +33,8 @@
         "lintfix": "../../node_modules/.bin/stylelint --fix \"**/*.scss\" && ../../node_modules/.bin/prettier --write \"**/*.{scss,ts,tsx,js,jsx,json,md,yml,html}\" && ../../node_modules/.bin/eslint -c ./.eslintrc.js \"src/**/*.{ts,tsx}\" --fix"
     },
     "dependencies": {
+        "@types/codemirror": "0.0.62",
+        "@types/rc-slider": "8.6.0",
         "chosen-js": "1.8.7",
         "query-string": "6.13.1",
         "rc-slider": "8.7.1",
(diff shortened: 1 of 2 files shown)


coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/ADUI-6662/provide-types-for-deps-we-export/index.html).

dmgauthier (approved):

aiheon (approved):

PR #1970, opened 2021-04-27
build(rollup): remove weird dependency on 'indexof'

gdostie:
### Proposed Changes

The esm build we produce had a weird require of the `indexof` module which doesn't exist. I created an alias that points to the real module called `component-indexof`.

Anyone using our esm build would get 
```bash
./node_modules/react-vapor/dist/bundles/react-vapor.esm.js
Cannot find module: 'indexof'. Make sure this package is installed.
```

Since the [issue](https://github.com/component/classes/issues/31) isn't resolved yet, I added an alias that points to `component-indexof` instead.

Tested it and it works, but the weird thing is that it was working on master without the alias 😕 

None

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/package.json
+++ b/packages/react-vapor/package.json
@@ -66,6 +66,7 @@
         "underscore.string": "^3.3.5"
     },
     "devDependencies": {
+        "@rollup/plugin-alias": "^3.1.2",
         "@rollup/plugin-commonjs": "11.0.2",
         "@rollup/plugin-inject": "4.0.1",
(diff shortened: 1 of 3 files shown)


dmgauthier (approved):

rosalie-liu (approved):

FelixBlaisThon (approved):
AWESOME!!!!! NICE

coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/ADUI-6662/remove-indexof-require/index.html).

PR #1987, opened 2021-05-11
fix(badge): allow badges with only icons

gdostie:
### Proposed Changes

Badges didn't allow having only an icon and no label. This is a feature that I needed for the IconCard.

### Before

image

### After

image

None

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/badge/Badge.tsx
+++ b/packages/react-vapor/src/components/badge/Badge.tsx
@@ -4,23 +4,43 @@ import {Svg} from '../svg';
 
 export const DEFAULT_BADGE_CLASSNAME = 'badge';
 
-export interface IBadgeProps {
-    label: string;
-    icon?: string;
+interface BadgeBasicProps {
     extraClasses?: string[];
 }
+
+interface BadgeWithLabelProps extends BadgeBasicProps {
+    label: string;
+}
+interface BadgeWithIconProps extends BadgeBasicProps {
+    icon: string;
+}
+
+export type IBadgeProps = BadgeWithLabelProps | BadgeWithIconProps | (BadgeWithLabelProps & BadgeWithIconProps);
 
 export class Badge extends React.Component<IBadgeProps> {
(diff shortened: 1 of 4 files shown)


coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/ADUI-6904/badges-icons/index.html).

packages/react-vapor/src/components/badge/Badge.tsx
```diff
@@ -5,7 +5,7 @@ import {Svg} from '../svg';
 export const DEFAULT_BADGE_CLASSNAME = 'badge';
 
 export interface IBadgeProps {
-    label: string;
+    label?: string;
     icon?: string;
     extraClasses?: string[];
 }
```
GermainBergeron, line 11: We want either:
- a label and an icon
- a label
- an icon

Right now it seems to be possible to have neither?

I would suggest something along those lines:

```typescript
export interface IBadgeWithLabelProps {
  label: string;
  icon?: string;
  extraClasses?: string[];
}
export interface IBadgeWithIconProps {
  icon: string;
  label?: string;
  extraClasses?: string[];
}
export type IBadgeProps = IBadgeWithLabelProps |  IBadgeWithIconProps;
```

GermainBergeron: You don't have any error if you have neither? 😞 Can you have a base interface like in this answer?

https://stackoverflow.com/a/37688375

GermainBergeron (approved):

rosalie-liu (approved):

packages/react-vapor/src/components/badge/Badge.tsx
gdostie: It works! 🎉  But I had to do some type inference inside the badge implementation to be able to use the `icon`and `label` props without errors.

```ts
'label' in this.props ? <span>{this.props.label}</span> : null
```

GermainBergeron (approved):

PR #2021, opened 2021-06-03
refactor(demo): move TextInput examples to Input

gdostie:
### Proposed Changes

The component is called Input so it should be displayed under a section that has that name.

- Moved the input examples from `TextInput` section to `Input` section
- Moved the `InputConnected` code back into the `Input` file.

None

### Acceptance Criteria


packages/react-vapor/src/components/input/Input.tsx
```diff
@@ -243,3 +248,24 @@ export class Input extends React.Component<IInputProps, IInputComponentState> {
...
+    onRender: (value: string = '', valid = true, disabled = false) =>
+        dispatch(addInput(ownProps.id, value, valid, disabled)),
+    onDestroy: () => dispatch(removeInput(ownProps.id)),
+    onChange: (value: string, valid = true) => {
+        dispatch(changeInputValue(ownProps.id, value, valid));
+        ownProps.changeDirtyState?.(value);
+    },
+});
+
+export const InputConnected = connect(mapStateToProps, mapDispatchToProps, ReduxUtils.mergeProps)(Input);
```
j-mcgregor, line 271: Is it out of scope to refactor to redux hooks?

coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/PR-2021/index.html).

packages/react-vapor/src/components/input/Input.tsx
gdostie: Yes, that input is going into the garbage can anyway 😅

sallainCoveo (approved):
I really like that the `Input` and the `InputConnected` components are in the same file.

packages/react-vapor/src/components/input/Input.tsx
sallainCoveo, line 271: I was wondering, what `ReduxUtils.mergeProps` doing exactly? Is it still relevant to have it?

rosalie-liu (approved):

packages/react-vapor/src/components/input/Input.tsx
gdostie: It inverts the order of props overriding. Without it the state props override the own props. With it the own props override the state props. Even if its not ideal, since it was already that way, I don't want to remove it in case something somewhere relies on this behaviour.

PR #2044, opened 2021-06-08
feat(textinput): add new TextInput component

gdostie:
### Proposed Changes

Add a new version of the input component called `TextInput`. 

As you will see, I have tried a new state management pattern that is **free of redux**. The advantage of doing that is that we get rid of unnecessary global redux state updates each time the value of an input changes.

3 things (other than types) related to the `TextInput` are exported:
- `TextInput`: the component. Can only be rendered inside a `FormProvider`
- `FormProvider`: provides the form state. We can have as many of those as we need. Best practice is to have it as low as possible in the DOM tree => less children equals less potential unnecessary rerenders.
- `useTextInput` hook: the only way to query and interact with the text input state. Can only be used within a `FormProvider`.

None, the old input component was preserved so we can phase it out progressively.

### Acceptance Criteria

### DIFF
--- a/packages/demo/src/components/examples/TextInputExamples.tsx
+++ b/packages/demo/src/components/examples/TextInputExamples.tsx
@@ -1,0 +1,100 @@
+import * as React from 'react';
+import {InputValidator, Section, TextInput, FormProvider, useTextInput, Button} from 'react-vapor';
+
+const nonEmptyValidation: InputValidator = (value: string) => {
+    const isEmpty = !value;
+    const isWhiteSpace = !isEmpty && !value.trim();
+    if (isWhiteSpace) {
+        return {status: 'warning', message: 'A white space is not empty, but is not ideal.'};
+    } else if (isEmpty) {
+        return {status: 'invalid', message: 'Cannot be empty.'};
+    } else {
+        return {status: 'valid'};
+    }
+};
+
+export const TextInputExamples: React.FunctionComponent = () => (
+    <FormProvider>
+        <Section>
+            <TextInput
+                required
+                showValidationOnBlur
+                validate={nonEmptyValidation}
+                type="text"
+                label="Label"
+                title="Title"
+                description="Description"
+                helpText="Help text"
+                tooltip="Tooltip"
+            />
+            <Section level={2} title="Disabled">
+                <TextInput type="email" label="Email" disabled />
+                <TextInput type="email" label="Email" defaultValue="123@abc.com" disabled />
+            </Section>
+            <Section level={2} title="useTextInput hook usage demo">
+                <div className="flex space-around">
+                    <TextInput
+                        id="favorite-dish"
+                        type="text"
+                        label="Favorite Dish"
+                        defaultValue="pizza"
+                        validate={(value) =>
+                            /pizza/i.test(value)
+                                ? {status: 'valid'}
+                                : {status: 'warning', message: "Pizza is best but that's okay."}
+                        }
+                        showValidationOnBlur
+                    />
+                    <TextInput
+                        id="favorite-animal"
+                        type="text"
+                        label="Favorite Animal"
+                        defaultValue="Tiger"
+                        required
+                        validate={nonEmptyValidation}
+                        showValidationOnBlur
+                    />
+                </div>
+                <HookUsageDemo />
+            </Section>
+        </Section>
+    </FormProvider>
+);
+
+const HookUsageDemo: React.FunctionComponent = () => {
+    const favoriteDishInput = useTextInput('favorite-dish');
+    const favoriteAnimalInput = useTextInput('favorite-animal');
+
+    return (
+        <>
+            <div className="flex space-around">
+                <p style={{whiteSpace: 'pre-wrap'}}>state = {JSON.stringify(favoriteDishInput.state, null, 4)}</p>
+                <p style={{whiteSpace: 'pre-wrap'}}>state = {JSON.stringify(favoriteAnimalInput.state, null, 4)}</p>
+            </div>
+            <p>
+                The hook gives you full access to the state of the input that has the specified id. As you can see, they
+                are both independent, one doesn't influence the state of the other. The hook also exposes dispatch,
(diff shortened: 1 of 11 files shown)


coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/PR-2044/index.html).

packages/react-vapor/src/components/textInput/TextInputContext.tsx
```diff
@@ -0,0 +1,78 @@
...
+    const newState: TextInputState = oneInputReducer(oldState, action.payload);
+    return oldState === newState
+        ? state
+        : {
+              ...state,
+              [action.type]: newState,
+          };
+};
+
+export const TextInputProvider: React.FunctionComponent = ({children}) => {
```
francoislg, line 64: As discussed in Slack, I find the `TextInputProvider` a bit too specific, I'd instead target something like `FormProvider` that exposes a context that can support a couple of different inputs.

packages/demo/src/components/examples/TextInputExamples.tsx
```diff
@@ -0,0 +1,36 @@
...
+    if (isWhiteSpace) {
+        return {status: 'warning', message: 'A white space is not empty, but is not ideal.'};
+    } else if (isEmpty) {
+        return {status: 'invalid', message: 'Cannot be empty.'};
+    } else {
+        return {status: 'valid'};
+    }
+};
+
+export const TextInputExamples: React.FunctionComponent = () => (
```
francoislg, line 16: Maybe add an example that uses 2 inputs since it's very probable it will happen in a real scenario? :P 

And another example with a default value?

I'd also like to see how I can leverage the state with `useTextInput` to drive something else, a scenario like "This field is only shown when you have entered a value in the first one", or "You can't click Next if the input is empty"

francoislg (commented):
I find it pretty cool :)

I wasn't convinced at first but now I like it :smile:

packages/demo/src/components/examples/TextInputExamples.tsx
```diff
@@ -0,0 +1,36 @@
...
+    <TextInputProvider>
+        <Section>
+            <TextInput
+                required
+                showValidationOnBlur
+                validate={nonEmptyValidation}
+                type="text"
```
j-mcgregor, line 23: If its called `TextInput` I would expect `type="text"` to be default and not need to be added as a prop. Or is this just to make it obvious?

packages/react-vapor/src/TestUtils.tsx
```diff
@@ -40,5 +40,24 @@ const renderModal: typeof customRender = (ui, renderOptions) => {
...
+const expectToThrow = (func: () => unknown, error?: JestToErrorArg): void => {
+    // Even though the error is caught, it still gets printed to the console
+    // so we mock that out to avoid the wall of red text.
+    const spy = jest.spyOn(console, 'error');
+    spy.mockImplementation(() => null);
+
+    expect(func).toThrow(error);
+
+    spy.mockRestore();
+};
```
j-mcgregor, line 58: Nice

packages/demo/src/components/examples/TextInputExamples.tsx
gdostie: The goal was to make it known that there are other types. Otherwise people will just use `text` for emails and urls. It works but its better to specify the right type for accessibility reasons.

j-mcgregor: makes sense, thanks

j-mcgregor (approved):

kdang-coveo:
Looks great!! I hate to be nitpicky, but could we still save the transition effect from the `Label` in the old input to the new one?

I notice it felt a little quick to jump compared to the input
label

Input Label:
inputlabel

packages/react-vapor/src/TestUtils.tsx
```diff
@@ -40,5 +40,24 @@ const renderModal: typeof customRender = (ui, renderOptions) => {
...
+
+type JestToErrorArg = Parameters<jest.Matchers<unknown, () => unknown>['toThrow']>[0];
```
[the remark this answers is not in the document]
gdostie: I'm not sure I understand either tbh I have taken this from somewhere on github or stackoverflow 😬  I just know that this is the right way to type the `toThrow` matcher parameters correctly.

gdostie:
> Looks great!! I hate to be nitpicky, but could we still save the transition effect from the `Label` in the old input to the new one?
> 
> I notice it felt a little quick to jump compared to the input
> label
> 
> Input Label:
> inputlabel

Unfortunately, because of the way the label and the input are positioned in relation to each other, the movement of the label cannot be animated properly. It leverages css properties that are not transitionable like `max-height: 0`. I have tried it but the transition wasn't looking good at all so I though it was better to not have a transition at all on the label position.

I think it could be reworked by using absolute positioning to make it work, but I don't have the time to do it right now. Since it has already been approved by UX this way, I am going to move on with what we have right now and come back to it later if needed.

packages/react-vapor/src/components/textInput/TextInputContext.tsx
gdostie: Implemented the `FormProvider`, it will support multiple different form components.

packages/demo/src/components/examples/TextInputExamples.tsx
gdostie: Added a hook usage demo in the example page, let me know if it doesn't cover something you think should be in there.

francoislg: Amazing! The example really clears things out

image

francoislg (approved):
Really cool :+1:

kdang-coveo (approved):

rosalie-liu (approved):
GG :clap:

FelixBlaisThon (approved):
Ship it bruv it's lit

PR #2070, opened 2021-06-30
fix(css,color-dot): add margin top & bottom auto for center the dot

toofff:
### Proposed Changes

Currently the point is not centered. This allows it to be centered vertically.

### Acceptance Criteria

### DIFF
--- a/packages/vapor/scss/elements/color-dot.scss
+++ b/packages/vapor/scss/elements/color-dot.scss
@@ -7,6 +7,7 @@
     min-width: var(--size);
     height: var(--size);
     min-height: var(--size);
+    margin: auto 0;
     background-color: var(--background-color);
     border-radius: 100%;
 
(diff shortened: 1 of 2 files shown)


coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/PR-2070/index.html).

packages/vapor/scss/elements/color-dot.scss
```diff
@@ -7,6 +7,7 @@
     min-width: var(--size);
     height: var(--size);
     min-height: var(--size);
+    margin: auto 0;
```
gdostie, line 10: Not sure I understand this change. I seems that with the suggested solution, the color dot is lower than what we had before.

#### Problem reported
image

#### Suggested solution
image

#### Before (without `margin: auto 0;`)
image

toofff: This work is only part of the solution, the rest is done in the UI admin and it is nearing completion.

I should have put this PR in draft, a thousand excuses

PR #2077, opened 2021-07-12
fix(option-picker): alignement when used as a button

GermainBergeron:
### Proposed Changes

Change the style of the Option Cycle to match the normal buttons

**Before**
image

**After**
image

None

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/optionsCycle/OptionsCycle.tsx
+++ b/packages/react-vapor/src/components/optionsCycle/OptionsCycle.tsx
@@ -74,7 +74,7 @@ export class OptionsCycle extends React.Component<IOptionsCycleProps> {
                     onClick={() => this.goToPreviousOption()}
                     disabled={!this.props.wrapAround && this.props.currentOption === 0}
                 >
-                    <Svg svgName="arrow-left-rounded" svgClass="icon mod-16" />
+                    <Svg svgName="arrow-left-rounded" className="icon mod-16" />
                 </button>
                 <span className={classNames('options-cycle-option', this.props.buttonClassName)}>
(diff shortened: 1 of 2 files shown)


coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/PR-2077/index.html).

toofff (approved):

aiheon (approved):

packages/react-vapor/src/components/optionsCycle/OptionsCycle.tsx
```diff
@@ -74,7 +74,7 @@ export class OptionsCycle extends React.Component<IOptionsCycleProps> {
                     onClick={() => this.goToPreviousOption()}
                     disabled={!this.props.wrapAround && this.props.currentOption === 0}
                 >
-                    <Svg svgName="arrow-left-rounded" svgClass="icon mod-16" />
+                    <Svg svgName="arrow-left-rounded" className="icon mod-16" />
```
gdostie, line 77: Is this still required? IIRC `svgClass` is more appropriate as it will set those classes on the `svg` tab instead of the wrapping `span`.

gdostie (approved):

packages/react-vapor/src/components/optionsCycle/OptionsCycle.tsx
GermainBergeron: The problem is that the `span` (which is vertically centered) doesn't take the full height of the `svg`. For this simple fix I didn't want to change the alignment of the SVG because I didn't want to impact all of react-vapor

PR #2089, opened 2021-07-23
Chore/ad UI 7296 improve accessibility of select components

j-mcgregor:
### Proposed Changes

https://coveord.atlassian.net/browse/UITOOL-72

Improved the accessibility of select components by adding `listbox` and `option` to ListBox components, and updated the tests

eg BEFORE

```tsx
let lists = screen.getAllByRole('list');

expect(within(lists[1]).queryByText('first')).not.toBeInTheDocument();
```

AFTER

```tsx
expect(
    screen.getByRole('option', {
        name: /first/i,
    })
).toBeInTheDocument();
```

None: functionality is the same, only testing is easier

NOTE: I was unable to update `MultiSelectWithPredicateAndFilter.spec.tsx` to RTL, the input won't change when using RTL and attempts to debug were taking too long. I will open another ticket & PR for documenting debugging Jest tests in React Vapor)

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/filterBox/FilterBox.tsx
+++ b/packages/react-vapor/src/components/filterBox/FilterBox.tsx
@@ -102,6 +102,7 @@ export class FilterBox extends React.Component<IFilterBoxProps, any> {
                         onKeyUp={this.props.onKeyUp}
                         style={inputMaxWidth}
                         autoFocus={this.props.isAutoFocus}
+                        aria-label={`${this.props.id}-name`}
                     />
                     <Svg
                         svgName="clear"
--- a/packages/react-vapor/src/components/filterBox/FilterBox.tsx
+++ b/packages/react-vapor/src/components/filterBox/FilterBox.tsx
@@ -102,7 +102,6 @@ export class FilterBox extends React.Component<IFilterBoxProps, any> {
                         onKeyUp={this.props.onKeyUp}
                         style={inputMaxWidth}
                         autoFocus={this.props.isAutoFocus}
-                        aria-label={`${this.props.id}-name`}
                     />
                     <Svg
                         svgName="clear"
--- a/packages/react-vapor/src/components/listBox/ListBox.tsx
+++ b/packages/react-vapor/src/components/listBox/ListBox.tsx
@@ -127,7 +127,7 @@ export class ListBox extends React.Component<IListBoxProps> {
         const items = this.props.isLoading ? this.getLoadingItems() : this.getItems();
         return (
             <>
-                <ul className={this.getClasses()} id={`${this.props.id}-container`}>
+                <ul className={this.getClasses()} id={`${this.props.id}-container`} role="listbox" tabIndex={0}>
                     {this.props.wrapItems(items)}
                 </ul>
                 {this.props.footer}
--- a/packages/react-vapor/src/components/select/hoc/tests/MultiSelectWithFilter.spec.tsx
+++ b/packages/react-vapor/src/components/select/hoc/tests/MultiSelectWithFilter.spec.tsx
@@ -1,4 +1,5 @@
 import {render, screen} from '@test-utils';
+import userEvent from '@testing-library/user-event';
 import * as React from 'react';
 
 import {MultiSelectWithFilter} from '../SelectComponents';
@@ -8,24 +9,23 @@ describe('Select', () => {
         const id: string = 'multi-select-with-filter';
 
         it('should add a duplicate if the filterValue is already selected', () => {
-            const filterValue: string = 'a';
-            const duplicateText: string = 'duplicate';
+            const duplicateText: string = 'a';
 
             render(
                 <MultiSelectWithFilter
                     id={id}
                     duplicateText={duplicateText}
                     items={[
-                        {value: filterValue, selected: true},
-                        {value: filterValue, selected: true},
+                        {value: duplicateText, selected: true},
+                        {value: duplicateText, selected: true},
                     ]}
-                />,
-                {
-                    initialState: {},
-                }
+                />
             );
 
+            userEvent.click(screen.getByText(/select an option/i));
+
             const listItems = screen.getAllByRole('listitem');
+            screen.logTestingPlaygroundURL();
 
             expect(listItems.length).toBe(2);
             expect(listItems[0]).toHaveTextContent(listItems[1].textContent);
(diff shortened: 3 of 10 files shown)


coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/PR-2089/index.html).

packages/react-vapor/src/components/select/tests/MultiSelectConnected.spec.tsx
```diff
@@ -68,11 +68,21 @@ describe('Select', () => {
...
-            expect(within(lists[1]).getByText(':potato:')).toBeVisible();
-            expect(within(lists[1]).getByText('🍟')).toBeVisible();
+            expect(
+                screen.getByRole('option', {
+                    name: /:seed:/i,
+                })
+            ).toBeVisible();
+            expect(
+                screen.getByRole('option', {
+                    name: /:seed:/i,
```
sallainCoveo, line 78: I think it would be :potato:?
```suggestion
                    name: /:potato:/i,

```

packages/react-vapor/src/components/select/tests/MultiSelectConnected.spec.tsx
```diff
@@ -141,26 +154,47 @@ describe('Select', () => {
...
-            expect(within(lists[1]).getByText('🌱')).toBeVisible();
-            expect(within(lists[1]).queryByText('🥔')).not.toBeInTheDocument();
-            expect(within(lists[1]).getByText('🍟')).toBeVisible();
-
-            userEvent.click(within(lists[1]).getByText('🍟'));
-
-            lists = screen.getAllByRole('list');
-            expect(within(lists[0]).getByText('🥔')).toBeVisible();
-            expect(within(lists[0]).getByText('🍟')).toBeVisible();
+            // pre-selected option is in the listbos
```
sallainCoveo, line 157:
```suggestion
            // pre-selected option is in the listbox

```

sallainCoveo (approved):
Looking good, I like it!

kdang-coveo (approved):

packages/react-vapor/src/components/filterBox/FilterBox.tsx
```diff
@@ -102,6 +102,7 @@ export class FilterBox extends React.Component<IFilterBoxProps, any> {
                         onKeyUp={this.props.onKeyUp}
                         style={inputMaxWidth}
                         autoFocus={this.props.isAutoFocus}
+                        aria-label={`${this.props.id}-name`}
```
GermainBergeron, line 105: The label should be something that labels the current element, here it's more of an id to identify the input?

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute

One thing to note is that you can get the input by its [placeholder with RTL](https://testing-library.com/docs/react-testing-library/cheatsheet/#queries) so I'm not sure we need that here

packages/react-vapor/src/components/listBox/ListBox.tsx
```diff
@@ -127,7 +127,7 @@ export class ListBox extends React.Component<IListBoxProps> {
         const items = this.props.isLoading ? this.getLoadingItems() : this.getItems();
         return (
             <>
-                <ul className={this.getClasses()} id={`${this.props.id}-container`}>
+                <ul className={this.getClasses()} id={`${this.props.id}-container`} role="listbox" tabIndex={0}>
```
GermainBergeron, line 130: Won't the tab index mess with the document?

Edit. doesn't seems to in the demo

packages/react-vapor/src/components/select/hoc/tests/MultiSelectWithPredicateAndFilter.spec.tsx
```diff
@@ -11,69 +11,67 @@ import {IFlatSelectOptionProps} from '../../../flatSelect/FlatSelectOption';
...
+        wrapper = mount(
+            <Provider store={store}>
+                <MultiSelectWithPredicateAndFilter
+                    id={id}
+                    items={[]}
+                    options={defaultFlatSelectOptions}
+                    matchPredicate={matchPredicate}
+                />
+            </Provider>,
+            {attachTo: document.getElementById('App')}
```
GermainBergeron, line 37: You can remove that line, we don't need to attach to the document

packages/react-vapor/src/components/select/hoc/tests/MultiSelectWithFilter.spec.tsx
```diff
@@ -1,88 +1,73 @@
...
+                    id={id}
+                    duplicateText={duplicateText}
+                    items={[
+                        {value: filterValue, selected: true},
+                        {value: filterValue, selected: true},
+                    ]}
+                />,
+                {
+                    initialState: {},
+                }
```
GermainBergeron, line 25: You don't need to send this object if the initialState is empty

Same comment for the other tests

packages/react-vapor/src/components/select/hoc/tests/MultiSelectWithFilter.spec.tsx
```diff
@@ -1,88 +1,73 @@
...
 
-            mountMultiSelect({items: [{value: filterValue, selected: true}], duplicateText});
-            store.dispatch(filterThrough(id, filterValue));
+            render(
+                <MultiSelectWithFilter
+                    id={id}
+                    duplicateText={duplicateText}
+                    items={[
+                        {value: filterValue, selected: true},
+                        {value: filterValue, selected: true},
```
GermainBergeron, line 20: I guess the values should be different from each other?

packages/react-vapor/src/components/select/hoc/tests/MultiSelectWithFilter.spec.tsx
```diff
@@ -1,88 +1,73 @@
...
+            render(<MultiSelectWithFilter id={id} noItemsText={noItemsText} customValues items={[]} />, {
+                initialState: {},
+            });
 
-            expect(itemBox.props().value).toBe(noItemsText);
+            expect(
+                screen.getByRole('option', {
+                    hidden: true,
+                })
+            ).toHaveTextContent(noItemsText);
```
GermainBergeron, line 45: From the test name you should open the dropdown before validating that the option is there. Then you wouldn't need the `hidden: true`

packages/react-vapor/src/components/select/hoc/tests/MultiSelectWithFilter.spec.tsx
```diff
@@ -1,88 +1,73 @@
...
+                    initialState: {},
+                }
+            );
 
-            expect(itemBox.props().value).toBe(noItemsText);
+            expect(
+                screen.getByRole('option', {
+                    hidden: true,
+                })
+            ).toHaveTextContent(noItemsText);
```
GermainBergeron, line 70: Same thing here, open the dropdown

GermainBergeron (commented):
I like the changes to the tests, way simpler to test!

TIL about listbox role

aiheon (approved):

packages/react-vapor/src/components/select/hoc/tests/MultiSelectWithFilter.spec.tsx
j-mcgregor: The test name is "should add a duplicate if the filterValue is already selected"

PR #2103, opened 2021-08-12
fix(date-picker): allow minimal range limit

sallainCoveo:
### Proposed Changes

First part to fix this bug: https://coveord.atlassian.net/browse/SEARCHAPI-6055

This PR includes three changes for the date-picker:
1. If the date picker includes a range, and the "end" date is empty when the user apply the changes, the date is set as the end of today, instead of the current time.

Before:

https://user-images.githubusercontent.com/58052881/129240570-5dc0d1e5-ee69-4068-aea4-a8caac439f21.mp4

After:

https://user-images.githubusercontent.com/58052881/129240587-922c6cf1-5547-40a3-bb4a-4de9e8587e95.mp4

2. A new option (optional) is added, which allows a minimal range limit. If this minimum is not reached, the apply button is disabled. That way, it indicates to the user that an minimum range has to be set to save their selection.

Demo:

https://user-images.githubusercontent.com/58052881/129241340-271d033d-bbdd-4b5e-a0f9-6ed26f413cca.mp4

3. When a minimal range is set, and the user wants the start date as "now", the end date is set to the minimal date that would respect the minimal range limit.

Demo:

https://user-images.githubusercontent.com/58052881/129240737-a8e881b0-b38c-4edf-aa9a-51d983b67a08.mp4

To test it:
1. Access the "DatePicker" section.
2. Play around with the `DatePicker dropdown with a minimal range limit of 5 minutes` component included for this demo.

Because this option is optional, nothing should be breaking.

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/datePicker/DatePicker.tsx
+++ b/packages/react-vapor/src/components/datePicker/DatePicker.tsx
@@ -1,9 +1,11 @@
+import moment from 'moment';
 import classNames from 'classnames';
 import * as React from 'react';
 
 import {DateUtils} from '../../utils/DateUtils';
 import {DateLimits} from './DatePickerActions';
 import {DEFAULT_DATE_PICKER_COLOR} from './DatePickerConstants';
+import {IRangeLimit} from './DatesSelection';
 import {SetToNowButton} from './SetToNowButton';
 
 export interface IDatePickerProps extends React.ClassAttributes<DatePicker> {
@@ -13,6 +15,7 @@ export interface IDatePickerProps extends React.ClassAttributes<DatePicker> {
     withTime?: boolean;
     hasSetToNowButton?: boolean;
     upperLimit?: boolean;
+    minimalRangeLimit?: IRangeLimit;
     date?: Date;
     setToNowTooltip?: string;
     isSelecting?: string;
@@ -39,8 +42,22 @@ export class DatePicker extends React.PureComponent<IDatePickerProps, {isSelecte
 
     private setToToday = () => {
         const date = new Date();
-        this.dateInput.value = this.getStringFromDate(date);
+        this.dateInput.value = this.props.minimalRangeLimit
+            ? this.getStringFromDate(this.updateDateBasedOnMinimalLimit(date))
+            : this.getStringFromDate(date);
         this.handleChangeDate();
+    };
+
+    private updateDateBasedOnMinimalLimit = (date: Date) => {
+        const {weeks, days, hours, minutes} = this.props.minimalRangeLimit;
+        const minimalLimitInMinutes: number =
+            (weeks ? weeks * 10080 : 0) + (days ? days * 1440 : 0) + (hours ? hours * 60 : 0) + (minutes ? minutes : 0);
+        const diff = moment().diff(moment(date), 'minutes');
+        if (Math.abs(diff) < minimalLimitInMinutes) {
+            return moment(date).add(minimalLimitInMinutes, 'minutes').toDate();
+        }
+
+        return date;
     };
 
     private handleChangeDate = () => {
--- a/packages/react-vapor/src/components/datePicker/DatePicker.tsx
+++ b/packages/react-vapor/src/components/datePicker/DatePicker.tsx
@@ -49,9 +49,7 @@ export class DatePicker extends React.PureComponent<IDatePickerProps, {isSelecte
     };
 
     private updateDateBasedOnMinimalLimit = (date: Date) => {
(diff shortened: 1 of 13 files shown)


coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/PR-2103/index.html).

packages/react-vapor/src/components/datePicker/DatePickerDropdown.tsx
```diff
@@ -273,24 +275,50 @@ export class DatePickerDropdown extends React.Component<IDatePickerDropdownProps
             : DateUtils.getSimpleDate(date);
     }
 
+    private getLimitAndDiffInMinutes(range: IRangeLimit): {limit: number; diff: number} {
+        const {weeks, days, hours, minutes} = range;
+        const limit: number =
+            (weeks ? weeks * 10080 : 0) + (days ? days * 1440 : 0) + (hours ? hours * 60 : 0) + (minutes ? minutes : 0);
```
acote-coveo, line 281: FYI: You can use `moment.duration({year: 2, weeks: 3}).asMinutes()` to create a duration and get the duration in a specific unit.

packages/react-vapor/src/components/datePicker/DatePicker.tsx
```diff
@@ -39,10 +42,24 @@ export class DatePicker extends React.PureComponent<IDatePickerProps, {isSelecte
...
+            ? this.getStringFromDate(this.updateDateBasedOnMinimalLimit(date))
+            : this.getStringFromDate(date);
         this.handleChangeDate();
     };
 
+    private updateDateBasedOnMinimalLimit = (date: Date) => {
+        const {weeks, days, hours, minutes} = this.props.minimalRangeLimit;
+        const minimalLimitInMinutes: number =
+            (weeks ? weeks * 10080 : 0) + (days ? days * 1440 : 0) + (hours ? hours * 60 : 0) + (minutes ? minutes : 0);
+        const diff = moment().diff(moment(date), 'minutes');
```
acote-coveo, line 55: This code seem similar with the one in DatePickerDropdown.tsx. 

Maybe that could be extracted into a DatePickerHelper?

acote-coveo (approved):
GG 🏅

packages/react-vapor/src/components/datePicker/tests/DatePickerReducers.spec.ts
```diff
@@ -894,6 +894,22 @@ describe('Date picker', () => {
 
                 expect(newDatePicker.appliedUpperLimit).toBe(oldDatePicker.lowerLimit);
             });
+
+            it('should set the upperLimit to the end of the day value if the state is set as a range and the upperLimit is undefined', () => {
```
aiheon, line 898: I did not followed all the discussion for this change but are we sure it will not cause issues that the upper limit is now set to the end of the day and not now? :thinking:

packages/react-vapor/src/components/datePicker/DatePicker.tsx
```diff
@@ -39,10 +42,22 @@ export class DatePicker extends React.PureComponent<IDatePickerProps, {isSelecte
...
         const date = new Date();
-        this.dateInput.value = this.getStringFromDate(date);
+        this.dateInput.value = this.props.minimalRangeLimit
+            ? this.getStringFromDate(this.updateDateBasedOnMinimalLimit(date))
+            : this.getStringFromDate(date);
         this.handleChangeDate();
     };
 
+    private updateDateBasedOnMinimalLimit = (date: Date) => {
+        const minimalLimitInMinutes: number = DateUtils.convertRangeToMinutes(this.props.minimalRangeLimit);
```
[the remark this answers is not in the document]
sallainCoveo: No, but minutes is the smallest unit. So to evaluate the difference between the limit and the current date, I convert the range to minutes, and I compare the value with the date (in minutes also)).

packages/react-vapor/src/components/datePicker/tests/DatePickerReducers.spec.ts
sallainCoveo: It causes an issue actually that the date was set a now, because you add a validation if the upperLimit and the lowerLimit were equal. The user could set themself as now still though. Do you think there an instance that it could causes an issue (I might have oversight something).

aiheon: I don't know that's why I ask :see_no_evil: 

But Like you already said, if the props is not used it wont be an issue :shrug:

aiheon (approved):

PR #2104, opened 2021-08-19
feat(form): +customhook to get the state of all specific forms

FelixBlaisThon:
### Proposed Changes

added a custom hook that, by passing it a specific type of Form, will return the state of all forms. For example, useTextInput need an id so it can return the state of that specific TextInput. In my case, I want to use a hook that returns all of the TextInput of a component (really useful for validation and also when juggling with TextInput with guid() rather than hardcoded ids.

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/textInput/useForm.ts
+++ b/packages/react-vapor/src/components/textInput/useForm.ts
@@ -1,0 +1,12 @@
+import {useContext, useMemo} from 'react';
+import {TextInputState} from './TextInputReducer';
+import {FormComponent, FormContext} from '../form/FormProvider';
+
+export const useFormComponent = (component: FormComponent): {state: Record<string, TextInputState>} => {
+    const formContext = useContext(FormContext);
+    if (formContext === undefined) {
+        throw new Error('useFormComponent must be used within a FormProvider.');
+    }
+    const state = formContext.state[component] || {};
+    return useMemo(() => ({state}), [state]);
+};
--- a/packages/react-vapor/src/components/form/FormProvider.tsx
+++ b/packages/react-vapor/src/components/form/FormProvider.tsx
@@ -18,7 +18,7 @@ const componentReducers = {
     TextInput: generateRecordReducer(textInputReducer),
 };
 
-type FormComponent = keyof typeof componentReducers;
+export type FormComponent = keyof typeof componentReducers;
 type FormComponentReducer = typeof componentReducers[FormComponent];
 type FormAction = {type: FormComponent} & React.ReducerAction<FormComponentReducer>;
 type FormState = Record<FormComponent, React.ReducerState<FormComponentReducer>>;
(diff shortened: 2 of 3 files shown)


coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/PR-2104/index.html).

packages/react-vapor/src/components/textInput/useForm.ts
```diff
@@ -0,0 +1,12 @@
...
+import {TextInputState} from './TextInputReducer';
+import {FormComponent, FormContext} from '../form/FormProvider';
+
+export const useForm = (form: FormComponent): {state: Record<string, TextInputState>} => {
+    const formContext = useContext(FormContext);
+    if (formContext === undefined) {
+        throw new Error('useForm must be used within a FormProvider.');
+    }
+    const state = formContext.state[form] || {};
+    return useMemo(() => ({state}), [state]);
```
gdostie, line 11: I think it should provide a `dispatch` function just like the `useTextInput`. I would see myself using it to clear the value of multiple text inputs at once for example.

packages/react-vapor/src/components/textInput/useForm.ts
```diff
@@ -0,0 +1,12 @@
+import {useContext, useMemo} from 'react';
+import {TextInputState} from './TextInputReducer';
+import {FormComponent, FormContext} from '../form/FormProvider';
+
+export const useForm = (form: FormComponent): {state: Record<string, TextInputState>} => {
```
gdostie, line 5: I feel the name isn't quite right. We don't use the whole form here. A form typically consists of multiple form components (for example a text input, some checkboxes and a single select). But here the purpose of the hook is to probe and/or interact with multiple instances of the same form component type.

```suggestion
export const useFormComponent = (component: FormComponent): {state: Record<string, TextInputState>} => {
```

packages/react-vapor/src/components/textInput/useForm.ts
```diff
@@ -0,0 +1,15 @@
+import {useContext, useMemo, Dispatch} from 'react';
+import {TextInputAction, TextInputState} from './TextInputReducer';
+import {FormComponent, FormContext} from '../form/FormProvider';
+
+export const useFormComponent = (
+    component: FormComponent
+): {state: Record<string, TextInputState>; dispatch: (id: string, action: TextInputAction) => void} => {
```
gdostie, line 7: I think this dispatch should provide thinks to make it easier to dispatch to multiple inputs at the time.

```ts
dispatch: (ids: string[] | 'All', action) => void
```

gdostie: Here the dispatch should be able to dispatch other actions than TextInputActions

FelixBlaisThon: FYI: since we cannot pinpoint a use case for a dispatch function, we will leave it as it is right now, so with only the state returned.

gdostie (approved):

rosalie-liu (approved):

PR #2119, opened 2021-09-14
style(multiselect): reduced the min-height by 1px

j-mcgregor:
### Proposed Changes

https://coveord.atlassian.net/browse/UITOOL-186

When exploring, the difference didn't look as extreme to me as it did in the screenshot in the ticket, but I made an adjustment of 1px to the line-height anyway

None

### Acceptance Criteria

### DIFF
--- a/packages/vapor/scss/components/selected-option.scss
+++ b/packages/vapor/scss/components/selected-option.scss
@@ -12,7 +12,7 @@
     margin: $dropdown-multiselect-margin;
     color: var(--title-text-color);
     font-size: $button-small-font-size;
-    line-height: $dropdown-multiselect-option-height;
+    line-height: $dropdown-multiselect-option-height - 1px;
     background-color: var(--white);
     border: $default-border;
     border-radius: $border-radius;
--- a/packages/vapor/scss/components/selected-option.scss
+++ b/packages/vapor/scss/components/selected-option.scss
@@ -12,7 +12,7 @@
     margin: $dropdown-multiselect-margin;
     color: var(--title-text-color);
     font-size: $button-small-font-size;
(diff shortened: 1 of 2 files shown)


coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/PR-2119/index.html).

packages/vapor/scss/components/selected-option.scss
```diff
@@ -12,7 +12,7 @@
     margin: $dropdown-multiselect-margin;
     color: var(--title-text-color);
     font-size: $button-small-font-size;
-    line-height: $dropdown-multiselect-option-height;
+    line-height: $dropdown-multiselect-option-height - 1px;
```
toofff, line 15: Why not just put `23px` on the variable `$dropdown-multiselect-option-height`, so that we can have the `min-height` and the `line-height` equal?

By the way, what's the point of `min-height` if we still have text inside?

FelixBlaisThon: I would tend to agree to just change the dropdown-multiselect-option-height to 23px directly if it doesnt break anything at some places else

toofff: if you want passed `min-height` at `26px`, this to broke

j-mcgregor: `min-height` makes sure the dividing line is the full height of the box. I made the line-height 1px smaller because I could see a tiny tiny difference (it sat too high) but the difference really is negligible

aiheon (approved):

toofff (approved):

PR #2131, opened 2021-09-22
chore(form): added line to export context

j-mcgregor:
I need the FormContext in a class component

### Proposed Changes

ADUI-7504-temporaryaccessmodal

Blocking ADUI-7504

A simple export of the FormContext so I can use it in a class component (no hooks allowed in a class component, and refactoring the whole class was way out of scope).

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/form/index.ts
+++ b/packages/react-vapor/src/components/form/index.ts
@@ -1,4 +1,4 @@
 export * from './Form';
 export * from './FormGroup';
-export {FormProvider} from '../form/FormProvider';
+export {FormProvider, FormContext} from '../form/FormProvider';
 export * from './useFormComponent';


coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/PR-2131/index.html).

GermainBergeron (approved):
It's the second time someone needs the context and I think we should go with it

I was curious about the problem and I found [this article](https://infinum.com/the-capsized-eight/how-to-use-react-hooks-in-class-components) that uses a HOC to add a prop on the class component which is nice but don't really help with readability and maintainability IMO

j-mcgregor:
> It's the second time someone needs the context and I think we should go with it
> 
> I was curious about the problem and I found [this article](https://infinum.com/the-capsized-eight/how-to-use-react-hooks-in-class-components) that uses a HOC to add a prop on the class component which is nice but don't really help with readability and maintainability IMO

What was the first time?

gdostie:
> but don't really help with readability and maintainability IMO

I would tend to disagree, because now you will have to reproduce the same logic we see in the `useTextInput` hook, but everywhere you need it in a class component.

I quite like the renderProps solution from the article you referred to @GermainBergeron. I would see something like the following as a cleaner solution.

```tsx
<TextInputConsumer id='input-id'>
    {(state, dispatch) => (
        <Button name="Save" onClick={() => save(state.value)} enabled={state.status === 'valid'} />
    )}
<TextInputConsumer>
```

j-mcgregor:
> > but don't really help with readability and maintainability IMO
> 
> I would tend to disagree, because now you will have to reproduce the same [logic](https://github.com/coveo/react-vapor/blob/master/packages/react-vapor/src/components/textInput/useTextInput.ts#L10-L11) we see in the `useTextInput` hook, each time you need it in a class component.
> 
> I quite like the renderProps solution from the article you referred to @GermainBergeron. I would see something like the following as a cleaner solution.
> 
> ```tsx
> <TextInputConsumer id='input-id'>
>     {(state, dispatch) => (
>         <Button name="Save" onClick={() => save(state.value)} enabled={state.status === 'valid'} />
>     )}
> <TextInputConsumer>
> ```

So instead of doing the following

```tsx
<FormContext.Consumer>
       {({state: {TextInput}}) => {
              const reasonOfTheRequest =
               TextInput[TemporaryAccessConstants.Inputs.Request.reasonOfTheRequest];
                const isInputValid = reasonOfTheRequest && reasonOfTheRequest.status;
                // etc
```

It would be

```tsx
<TextInputConsumer id='input-id'>
      {({state: {TextInput}}) => {
     // etc
```

j-mcgregor:
> Not exactly, it would be
> 
> ```tsx
> <TextInputConsumer id={TemporaryAccessConstants.Inputs.Request.reasonOfTheRequest}>
>     {({state}) => {
>         const isInputValid = state.status === 'valid';
>     }
> </TextInputConsumer>
> ```

Ok I'll have a go at that, thanks

aiheon (approved):

PR #2139, opened 2021-09-28
feat(multiboxwithremovebutton): *remove button

FelixBlaisThon:
### Proposed Changes

remove button is now hidden only when the MultiBox has one child. In any other cases, the remove button will be visible. 

The current behaviour usually expects us to add a method where we call `parentProps.addNewBox()` inside the child component, which can lead to unwanted display like _You just wrote something in the first child (let's say you are rendering inputs). With the expected implementation, a new input will be rendered, but without a remove button_ . I think it would be better to give the option to the user to delete the last child.

it's only a change in visibility/style. No big refactor

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/multilineBox/hoc/MultilineBoxWithRemoveButton.tsx
+++ b/packages/react-vapor/src/components/multilineBox/hoc/MultilineBoxWithRemoveButton.tsx
@@ -59,20 +59,21 @@ export const multilineBoxWithRemoveButton = (
         };
 
         private getRemoveButtonNode(
-            data: Partial<IMultilineSingleBoxProps<T>> = {},
-            props: Partial<IButtonProps> = {}
+            data: Array<Partial<IMultilineSingleBoxProps<T>>>,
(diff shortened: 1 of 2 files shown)


coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/PR-2139/index.html).

dmgauthier:
:smiling_imp:  testing ?

toofff (approved):
LGTM

aiheon (approved):

j-mcgregor:
How simple is it to remove the unnecessary extra remove buttons from the demo?
image

j-mcgregor (approved):

FelixBlaisThon:
> How simple is it to remove the unnecessary extra remove buttons from the demo? image

hmm I think thats just a feature from the component, because it was always there in the demo.. :/ 2 containers to wrap a remove button. in my opinion, i don't like this example lol

PR #2140, opened 2021-09-29
style(sidenavigation): implemented new focus styles

j-mcgregor:
### Proposed Changes

https://coveord.atlassian.net/browse/ADUI-7505

[Slack thread](https://coveo.slack.com/archives/GLS6Z9RPD/p1632768959073000)

This was originally done in AdminUI but better to have the changes here. Also took the liberty to clarify the demo for SideNavigation, though this will eventually all be replaced by Plasma

https://user-images.githubusercontent.com/22773767/135301202-6d650ad9-6561-4a5b-8ac7-1785255fc7ae.mov

EDIT: After discussion with UX the focus will not have rounded corners for now, since it looks terrible with the left bar of the nav when its active. The future restyle will fix this issue by removing the left bar and making all the navs more rounded, and the new mixin has the option to provide a border-radius in the future

None

### Acceptance Criteria

### DIFF
--- a/packages/demo/src/components/examples/TabConnectedExample.tsx
+++ b/packages/demo/src/components/examples/TabConnectedExample.tsx
@@ -11,7 +11,7 @@ const TAB11_ID = 'tab11';
 const TAB22_ID = 'tab22';
 const TAB33_ID = 'tab33';
 
-export const TabsConnectedExamples: React.FC = () => (
+export const TabsConnectedExamples: React.FunctionComponent = () => (
     <Section>
         <Section level={3} title="Basic Tabs">
             <div>
--- a/packages/demo/src/components/examples/TabConnectedExample.tsx
+++ b/packages/demo/src/components/examples/TabConnectedExample.tsx
@@ -11,82 +11,67 @@ const TAB11_ID = 'tab11';
 const TAB22_ID = 'tab22';
 const TAB33_ID = 'tab33';
 
-export class TabsExamples extends React.Component<any, any> {
-    render() {
-        return (
-            <Section>
-                <Section level={3} title="Basic Tabs">
-                    <div>
-                        <TabNavigation>
-                            <TabConnected id={TAB1_ID} title="A Tab" />
-                            <TabConnected id={TAB2_ID} title="Another Tab" tooltip="I am an enabled tab" />
-                            <TabConnected id={TAB3_ID} title="Tab with an icon">
-                                <Svg svgName={'help'} svgClass={'icon documentation-link mod-16 mr1'} />
-                            </TabConnected>
-                            <TabConnected
-                                id={TAB4_ID}
-                                title=" Another Tab with an icon"
-                                children={<Svg svgName={'info'} svgClass={'icon documentation-link mod-16 mr1'} />}
-                            />
-                            <TabConnected id={TAB5_ID} title="A Disabled Tab" tooltip="I am a disabled tab" disabled />
-                        </TabNavigation>
-                        <TabContent>
-                            <TabPaneConnected id={TAB1_ID}>
-                                <div className="mod-header-padding mod-form-top-bottom-padding">
-                                    Content of the first tab.
-                                </div>
-                            </TabPaneConnected>
-                            <TabPaneConnected id={TAB2_ID}>
-                                <div className="mod-header-padding mod-form-top-bottom-padding">
-                                    Content of the second tab.
-                                </div>
-                            </TabPaneConnected>
-                            <TabPaneConnected id={TAB3_ID}>
-                                <div className="mod-header-padding mod-form-top-bottom-padding">
-                                    Content of the tab with an icon.
-                                </div>
-                            </TabPaneConnected>
-                            <TabPaneConnected id={TAB4_ID}>
-                                <div className="mod-header-padding mod-form-top-bottom-padding">
-                                    Content of the other tab with an icon.
-                                </div>
-                            </TabPaneConnected>
-                            <TabPaneConnected id={TAB5_ID}>
-                                <div className="mod-header-padding mod-form-top-bottom-padding">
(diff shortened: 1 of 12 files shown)


coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/PR-2140/index.html).

packages/vapor/scss/components/navigation.scss
```diff
@@ -125,10 +125,20 @@
...
                             font-size: var(--default-font-size);
                             letter-spacing: 0.05em;
+                            border: 2px solid transparent;
+                            border-radius: 8px;
+
+                            &:focus-visible {
+                                background: var(--navy-blue-10);
+                                border: 2px solid var(--cerulean-blue-50);
+                                outline: none;
+                                box-shadow: 0px 0px 0px 3px var(--cerulean-blue-20);
```
toofff, line 140:
```suggestion
                                box-shadow: 0 0 0 3px var(--cerulean-blue-20);
```

toofff (approved):

packages/vapor/scss/components/navigation.scss
```diff
@@ -125,10 +125,20 @@
...
                             margin-right: $navigation-right-margin;
+                            padding-right: 2px;
                             padding-left: $navigation-horizontal-space;
                             color: var(--navy-blue-80);
                             font-size: var(--default-font-size);
                             letter-spacing: 0.05em;
+                            border: 2px solid transparent;
+                            border-radius: 8px;
+
+                            &:focus-visible {
```
gdostie, line 136: `:focus-visible` doesn't seem supported on Safari, can't we use `:focus`?

packages/vapor/scss/components/navigation.scss
```diff
@@ -125,10 +125,20 @@
                             justify-content: space-between;
                             height: $navigation-menu-section-height;
                             margin-right: $navigation-right-margin;
+                            padding-right: 2px;
                             padding-left: $navigation-horizontal-space;
                             color: var(--navy-blue-80);
                             font-size: var(--default-font-size);
                             letter-spacing: 0.05em;
+                            border: 2px solid transparent;
```
gdostie, line 133: Is this one really necessary?
```suggestion
```

packages/vapor/scss/components/navigation.scss
```diff
@@ -125,10 +125,20 @@
...
+                            border-radius: 8px;
+
+                            &:focus-visible {
+                                background: var(--navy-blue-10);
+                                border: 2px solid var(--cerulean-blue-50);
```
gdostie, line 138: IMHO it looks better (more like the other focus styles that we already have on links for example) without those 2
```suggestion
```

gdostie (changes requested):
Have you tried to use the reusable focus mixin?

```scss
:focus {
    @include focus-border(2px);
}
```

packages/vapor/scss/components/navigation.scss
j-mcgregor: Using `:focus` in the dev tools in Brave didn't toggle the effect, only `:focus-visible` did. I didn't try on Safari, but I'll try it and add `:focus`

j-mcgregor:
> Have you tried to use the reusable focus mixin?
> 
> ```css-scss
> :focus {
>     @include focus-border(2px);
> }
> ```

I didn't know we had one

packages/vapor/scss/components/navigation.scss
j-mcgregor: yes, otherwise when the focus is triggered it shifts the content to the right because its adding a 2px border when there was no border before

packages/vapor/scss/components/navigation.scss
j-mcgregor: These colors match the new designs

gdostie:
I guess it would make sense to update the mixin, otherwise we will still have bugs about the focus state not being uniform 🙂

gdostie:
Playing with the mixin affected the focus style of the links:
Before:
image

After:
image

I still think the focus style should be the same both for links and nav items, but the new style looks bad for links

toofff (approved):

GermainBergeron:
The focused style looks a bit weird to me with the corner on the left having a different colour. Is UX aware of that issue?

image

j-mcgregor:
> The focused style looks a bit weird to me with the corner on the left having a different colour. Is UX aware of that issue?
> 
> image

I'm discussing it with them now

packages/vapor/scss/components/navigation.scss
j-mcgregor: Also, Safari doesn't let me inspect the page because it sucks

packages/demo/src/components/examples/TabConnectedExample.tsx
```diff
@@ -11,82 +11,67 @@ const TAB11_ID = 'tab11';
...
-                                <div className="mod-header-padding mod-form-top-bottom-padding">Last tab.</div>
-                            </TabPaneConnected>
-                        </TabContent>
-                    </div>
-                </Section>
-            </Section>
-        );
-    }
-}
+export const TabsConnectedExamples: React.FC = () => (
```
toofff, line 14:
```suggestion
export const TabsConnectedExamples: React.FunctionComponent = () => (

```

j-mcgregor:
> Playing with the mixin affected the focus style of the links. I think the offset parameter is no longer applied, making the border too close.
> 
> Before: image
> 
> After: image
> 
> I still think the focus style should be the same both for links and nav items, but the new style looks bad for links

I made a new mixin and updated the color and default params on the old one, since links look rubbish with the new `focus-border` instead of the old `focus-outline`

gdostie (commented):
we seem to have lost the rounded borders on the focused links though
image

image

PR #2144, opened 2021-09-30
feat(association-rules-to-group.svg): include new svg

sallainCoveo:
### Proposed Changes
Story (part 1): https://coveord.atlassian.net/browse/SEARCHAPI-6152

Implement the SVG required for the improvement on the `Associate Rules to This Group` modal

Nope

### Acceptance Criteria

### DIFF
--- a/packages/vapor/resources/icons/svg/association-rules-to-group.svg
+++ b/packages/vapor/resources/icons/svg/association-rules-to-group.svg
@@ -1,0 +1,137 @@
+<svg width="328" height="188" viewBox="0 0 328 188" fill="none" xmlns="http://www.w3.org/2000/svg">
+<rect x="171" y="5" width="118.794" height="118.794" rx="9.25424" transform="rotate(45 171 5)" fill="url(#paint0_linear)"/>
+<path d="M179.194 127.048L183.454 127M183.525 131.361L183.475 127C179.193 132.674 167.605 134.881 162.088 128.481" stroke="url(#paint1_linear)" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/>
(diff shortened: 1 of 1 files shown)


packages/vapor/resources/icons/svg/association-rules-to-group.svg
```diff
@@ -0,0 +1,137 @@
+<svg width="328" height="188" viewBox="0 0 328 188" fill="none" xmlns="http://www.w3.org/2000/svg">
```
toofff, line 1: I know that for ReactVapor we remove the `width` & `height` attributes from SVGs because it's quite easy to control their size in CSS, for your case I think we could do the same thing but not completely sure.

toofff (approved):

toofff (changes requested):
This is not an icon but an infographic.

I would not see this infographic in this `icons` folder and not in this project either

sallainCoveo:
> This is not an icon but an infographic.
> 
> I would not see this infographic in this `icons` folder and not in this project either

It would make sense. I'm just confused a little bit on that matter, what is the difference with this one and [`emailSent`](https://github.com/coveo/react-vapor/blob/master/packages/vapor/resources/icons/svg/email-sent.svg) or [`inAppTeaser`](https://github.com/coveo/react-vapor/blob/master/packages/vapor/resources/icons/svg/in-app-teaser.svg)? In that sense, are they misplaced also?

sallainCoveo:
After a discussion with @toofff, we came to the conclusion that this image might be better to be included directly in the Admin-UI project instead, as this image is specific for a component that will be only used in that private project.

PR #2151, opened 2021-10-06
fix(form): prevent the default behaviour of submitting on forms

GermainBergeron:
### Proposed Changes

The UTs in admin-ui are running forever because the Form of react-vapor submit action is not handled. Since we have no action on the form, the default one is to reload the page with different query parameters.

I added a unit test to make sure we handle the submit

None

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/form/tests/Form.spec.tsx
+++ b/packages/react-vapor/src/components/form/tests/Form.spec.tsx
@@ -1,7 +1,8 @@
+import userEvent from '@testing-library/user-event';
 import {shallow} from 'enzyme';
 import * as React from 'react';
 
-import {render} from '@test-utils';
+import {render, screen} from '@test-utils';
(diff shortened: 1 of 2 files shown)


packages/react-vapor/src/components/form/tests/Form.spec.tsx
```diff
@@ -61,5 +62,21 @@ describe('Form', () => {
             expect(formCollection.getElementsByClassName('mt2').length).toBe(0);
             expect(formCollection.getElementsByClassName('mb2').length).toBe(0);
         });
+
+        it('does not change the url when the user press enter inside a form', () => {
+            const initialHREF = window.location.href;
+
+            render(
+                <Form>
+                    <input defaultValue="" name="patate" />
```
aiheon, line 71: :potato: :heart:

PR #2155, opened 2021-10-08
fix(form): remove the "block" type display for the form component

toofff:
### Proposed Changes

When updating this component with this [PR](https://github.com/coveo/react-vapor/pull/2143) we now add a `form` tag to solve the autocompletion problem but the `form` tag is of type `display: block` and breaks the design.

So my idea is to change to `display: contents` to let the content react as before

### Acceptance Criteria


coveobot (commented):
[Live demo of your latest successful build available here](https://vaporqa.cloud.coveo.com/feature/PR-2155/index.html).

packages/react-vapor/src/components/form/Form.tsx
```diff
@@ -15,7 +15,7 @@ export const Form: React.FunctionComponent<IFormProps> = ({children, className,
         submitEvent.preventDefault();
     };
     return (
-        <form onSubmit={onSubmit}>
+        <form onSubmit={onSubmit} style={{display: 'contents'}}>
```
gdostie, line 18: Not sure I am willing to remove all forms from the accessibility tree 😬 

> Current implementations in most browsers will remove from the accessibility tree any element with a display value of contents (but descendants will remain). This will cause the element itself to no longer be announced by screen reading technology.

https://developer.mozilla.org/en-US/docs/Web/CSS/display#display_contents

gdostie (approved):

packages/react-vapor/src/components/form/Form.tsx
toofff: OK for used CSS class `full-content`

PR #2156, opened 2021-10-13
fix(countdown): moment fix to display better text

j-mcgregor:
### Proposed Changes

https://coveord.atlassian.net/browse/UITOOL-221

Provide better text if on the last day of the month

Currently, on the last day, depending on the time of day, it will change from 'A few hours left', 'A few minutes left', 'A few seconds left' if you're on the last day of the month. I added a condition to check if the current day is the last day of the month and if so just say 'Last day'. Also cleaned up a bit of CSS that was applying the Date range to the countdown

image

### How to test

A bit tricky without it either being the last day of the month or checking out the branch and testing it locally yourself. I wrote a unit test which mocks the `Date.now` function which moment uses

### Acceptance Criteria

### DIFF
--- a/packages/react-vapor/src/components/calendar/tests/Countdown.spec.tsx
+++ b/packages/react-vapor/src/components/calendar/tests/Countdown.spec.tsx
@@ -14,6 +14,11 @@ describe('Countdown', () => {
 
         expect(container.firstChild).toHaveClass('countdown-calendar');
         expect(today).toHaveClass('todays-date');
+        expect(
+            screen.getByRole('heading', {
+                name: /days left/i,
+            })
+        ).toBeInTheDocument();
     });
 
     it('should display "Last day" on the last day of the month', () => {


packages/react-vapor/src/components/calendar/Calendar.tsx
```diff
@@ -280,13 +280,16 @@ export class Calendar extends React.Component<ICalendarProps, any> {
...
         if (this.props.countdown) {
-            const countdownDays = moment()
-                .date(moment().month(this.props.months[selectedMonth]).daysInMonth())
-                .fromNow(true);
+            const daysInMonth = moment().month(this.props.months[selectedMonth]).daysInMonth();
+            const daysLeftInMonth = moment().date(daysInMonth).fromNow(true);
+            const currentDayOfMonth = moment().date();
+
+            // Datepicker uses below logic to avoid displaying 'A few [hours|minutes|seconds] left'
+            const countdownTitle = currentDayOfMonth === daysInMonth ? 'Last day' : `${daysLeftInMonth} left`;
```
GermainBergeron, line 288: I haven't tested it but could we go with something like that? I find it easier to read
```suggestion
            const endOfMonth = moment().endOf("month");
            const daysLeftInMonth = endOfMonth.diff(moment(), 'days');

            // Datepicker uses below logic to avoid displaying '0 days left'
            const countdownTitle = daysLeftInMonth === 0 ? 'Last day' : `${daysLeftInMonth} days left`;
```

packages/react-vapor/src/components/calendar/tests/Countdown.spec.tsx
```diff
@@ -15,4 +15,20 @@ describe('Countdown', () => {
         expect(container.firstChild).toHaveClass('countdown-calendar');
         expect(today).toHaveClass('todays-date');
     });
+
+    it('should display "Last day" on the last day of the month', () => {
+        const dateNow = Date.now;
+        // @ts-ignore
+        Date.now = jest.fn(() => new Date('2020-01-31T12:00:00.000Z'));
```
GermainBergeron, line 22: Could we go with something like that instead? https://stackoverflow.com/a/63377110

We're using Jest 26 so it should be good
```suggestion
        jest
            .useFakeTimers('modern')
            .setSystemTime(new Date('2020-01-31).getTime());
```

packages/react-vapor/src/components/calendar/tests/Countdown.spec.tsx
```diff
@@ -17,9 +17,7 @@ describe('Countdown', () => {
     });
 
     it('should display "Last day" on the last day of the month', () => {
-        const dateNow = Date.now;
-        // @ts-ignore
-        Date.now = jest.fn(() => new Date('2020-01-31T12:00:00.000Z'));
+        jest.useFakeTimers('modern').setSystemTime(new Date('2020-01-31T12:00:00.000Z').getTime());
```
aiheon, line 20: TIL 
i'm using a Date.now = jest.fn() in UA, i'll go change it :see_no_evil:

packages/react-vapor/src/components/calendar/Calendar.tsx
```diff
@@ -25,6 +25,12 @@ export interface ICalendarSelectionRule {
     isFor: string;
 }
 
+type Languages = 'en';
+
+interface I18nConfig {
+    [key: string]: Record<Languages, string>;
+}
+
```
j-mcgregor, line 33: Hi @toofff is this what you meant?

mikedidomizio (approved):

PR #2196, opened 2021-11-02
fix(datepicker): added valid date check

j-mcgregor:
### Proposed Changes

Initial bug: user could enter an invalid date range
Extra bug: While manually typing an invalid date, the `setState` triggered by `componentDidMount` would trigger an infinte loop as parent components tried to validate it and then update the component

How to test:

Datepicker:

- The bug applies to the datepicker dropdown, but worth testing in every example.
- I made a new example in the demo to mimic the bug in Log Browser (see screenshot)
- For any datepicker with `selectionRules`, if you try and manually enter a date that's outside of what's specified in its rules, then it will stay selected, use the last valid date as display (but not set it), and show a red `invalid` border

image

https://user-images.githubusercontent.com/22773767/139930393-12f55da7-2d0d-499f-b9fb-8f1da5897999.mov

### DIFF
--- a/packages/demo/src/components/examples/DatePickerExamplesCommon.ts
+++ b/packages/demo/src/components/examples/DatePickerExamplesCommon.ts
@@ -154,6 +154,21 @@ export const CALENDAR_SELECTION_RULES: ICalendarSelectionRule[] = [
     },
 ];
 
+export const CALENDAR_ADVANCED_SELECTION_RULES: ICalendarSelectionRule[] = [
+    {
+        test: (date: Date) => date <= new Date() && date >= moment().subtract(2, 'weeks').toDate(), // is the date within the last two weeks
+        isFor: CalendarSelectionRuleType.all,
+    },
+    {
+        test: (date: Date, endDate: Date) => moment(endDate).diff(moment(date), 'day') >= 0, // The end of your selection cannot be before the start of your selection
+        isFor: CalendarSelectionRuleType.upper,
+    },
+    {
+        test: (date: Date, endDate: Date) => moment(endDate).diff(moment(date), 'day') <= 7, // You cannot select more than 7 days at a time
+        isFor: CalendarSelectionRuleType.lower,
+    },
+];
+
 export const SELECTION_BOX: IDatesSelectionBox[] = [
     {
         title: 'Select date',
--- a/packages/react-vapor/src/components/datePicker/DatePicker.tsx
+++ b/packages/react-vapor/src/components/datePicker/DatePicker.tsx
@@ -67,15 +67,17 @@ export class DatePicker extends React.PureComponent<IDatePickerProps, {isSelecte
             // and we need the value immediately to decide whether to to call props.onBlur
             let isDatePermitted = true;
 
-            this.props.selectionRules?.forEach((rule: ICalendarSelectionRule) => {
-                if (
-                    (rule.isFor === CalendarSelectionRuleType.lower && this.props.isSelecting === DateLimits.lower) ||
-                    (rule.isFor === CalendarSelectionRuleType.upper && this.props.isSelecting === DateLimits.upper)
-                ) {
+            this.props.selectionRules
+                ?.filter(
+                    (rule: ICalendarSelectionRule) =>
+                        (rule.isFor === CalendarSelectionRuleType.lower &&
+                            this.props.isSelecting === DateLimits.lower) ||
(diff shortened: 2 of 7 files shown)


coveobot (commented):
Stale review, a new push occured on the branch. Check for the new status :arrow_down:

packages/react-vapor/src/components/datePicker/DatePicker.tsx
```diff
@@ -61,8 +63,21 @@ export class DatePicker extends React.PureComponent<IDatePickerProps, {isSelecte
...
+
+            this.props.selectionRules?.forEach((rule: ICalendarSelectionRule) => {
+                if (
+                    (rule.isFor === CalendarSelectionRuleType.lower && this.props.isSelecting === DateLimits.lower) ||
+                    (rule.isFor === CalendarSelectionRuleType.upper && this.props.isSelecting === DateLimits.upper)
+                ) {
+                    isDatePermitted = rule.test(date);
+                    this.setState({isDatePermitted});
+                }
+            });
```
toofff, line 78: Use `filter` please

```suggestion
            this.props.selectionRules?
                .filter((rule: ICalendarSelectionRule) => (rule.isFor === CalendarSelectionRuleType.lower && this.props.isSelecting === DateLimits.lower) || (rule.isFor === CalendarSelectionRuleType.upper && this.props.isSelecting === DateLimits.upper))
                .forEach((rule: ICalendarSelectionRule) => {
                    isDatePermitted = rule.test(date);
                    this.setState({isDatePermitted});
                });
```

toofff (approved):

packages/demo/src/components/examples/DatePickerExamplesCommon.ts
```diff
@@ -154,6 +154,21 @@ export const CALENDAR_SELECTION_RULES: ICalendarSelectionRule[] = [
     },
 ];
 
+export const CALENDAR_ADVANCED_SELECTION_RULES: ICalendarSelectionRule[] = [
+    {
+        test: (date: Date) => date <= new Date() && date >= moment().subtract(2, 'weeks').toDate(), // is the date within the last two weeks
```
aiheon, line 159: should those rules have the range allowed as parameter?
I feel like we would need to add more rules for use case where the range is 3 days, 1 day, etc :thinking:

aiheon:
image
This demo is broken also for the range :eyes: I can pick a bigger range from the calender directly even :sus:

packages/demo/src/components/examples/DatePickerExamplesCommon.ts
aiheon: Also, I think we should have a description popping at the same time that the red border if we are going down that route :thinking: 
Right now, it turns red but we don't know why (well we do but our users probably won't :see_no_evil: )

aiheon: Maybe mimic this pattern instead? :eyes: 
image

j-mcgregor: > should those rules have the range allowed as parameter? I feel like we would need to add more rules for use case where the range is 3 days, 1 day, etc 🤔

That was my first attempt but using range triggers the infinite loop

> Also, I think we should have a description popping at the same time that the red border if we are going down that route 🤔
Right now, it turns red but we don't know why (well we do but our users probably won't 🙈 )

I thought that too; I didn't attempt it yet as I wasn't sure what I have so far was the right way to go, and adding a message involves a fair bit more complexity

packages/react-vapor/src/components/datePicker/DatePicker.tsx
```diff
@@ -61,8 +63,23 @@ export class DatePicker extends React.PureComponent<IDatePickerProps, {isSelecte
     private handleChangeDate = () => {
         if (this.dateInput.value !== '') {
             const date: Date = this.getDateFromString(this.dateInput.value);
-
-            if (date.getDate()) {
+            // setting isDatePermitted as let and in component state, since setState is async and triggers rerender of component
+            // and we need the value immediately to decide whether to to call props.onBlur
```
gdostie, line 67:
```suggestion
```

gdostie (commented):
Add a unit test to make sure this edge case is covered and you have my approval

gdostie (approved):

GermainBergeron (approved):

PR #2205, opened 2021-11-04
ci: fix the git credentials

GermainBergeron:
### Proposed Changes

There was a rotation of credentials, and it seems the new token cannot publish anymore. I switched to the same credentials we use in admin-ui

None


toofff:
Even if I arrive a little late and the CI has to validate the PRs before us, it's ok for me ;)

PR #2358, opened 2021-12-16
feat(radioselect): allow specific tooltip

nathanlb:
### Context
We need to disable certain model creation flows based on missing privileges as well as license. Each scenario needs a specific disabled tooltip string but right now the RadioSelect component forces a single string to be used for all options.

### Proposed Changes
Only set disabledTooltip value on option clones if none is specified already.

### Acceptance Criteria


GermainBergeron (approved):
LGTM

sallainCoveo (approved):
Looks good, do you think that an example in the demo would be a good addition? (to illustrate this use case).

dmgauthier (approved):

PR #2359, opened 2021-12-17
feat(code-editor): Dynamic code editor in the demo

GermainBergeron:
### Proposed Changes

Required code to have a Sandbox component that accept any code and render the results

https://user-images.githubusercontent.com/260007/146600870-0fdc64e3-65c5-4889-8011-e1a73b68f2c8.mov

None

### Acceptance Criteria

### DIFF
--- a/packages/demo/src/demo-building-blocs/useTypescriptServer.ts
+++ b/packages/demo/src/demo-building-blocs/useTypescriptServer.ts
@@ -1,0 +1,61 @@
+import {createDefaultMapFromCDN} from '@typescript/vfs';
+import lzstring from 'lz-string';
+import * as monaco from 'monaco-editor';
+import React from 'react';
+import * as ts from 'typescript';
+
+monaco.languages.typescript.typescriptDefaults.setCompilerOptions({
+    jsx: monaco.languages.typescript.JsxEmit.React,
+    jsxFactory: 'React.createElement',
+    jsxFragmentFactory: 'React.Fragment',
+    reactNamespace: 'React',
+});
+monaco.languages.typescript.typescriptDefaults.setDiagnosticsOptions({
+    noSemanticValidation: false,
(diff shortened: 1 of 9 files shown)


GermainBergeron:
I had to put a few `as any` in the code since we're on an older version of typescript and the dependencies are resolved in a weird way. `@typescript/vfs` declare its typescript dependency as `*` which resolves to the one installed by the design token package

aiheon:
> I had to put a few `as any` in the code since we're on an older version of typescript and the dependencies are resolved in a weird way. `@typescript/vfs` declare its typescript dependency as `*` which resolves to the one installed by the design token package

I suppose we will be able to remove them soon enough with the bump of typescript for the search bar :smile:

aiheon (approved):

gdostie:
> which resolves to the one installed by the design token package

I got a few errors fixed by changing the typescript version to 3.8.3 in the design-tokens package

packages/demo/src/demo-building-blocs/useTypescriptServer.ts
```diff
@@ -0,0 +1,61 @@
...
+            createDefaultMapFromCDN(compilerOptions as any, ts.version!, true, ts as any, lzstring).then((map) => {
+                mappedTypes.forEach(({path, content}) => {
+                    monaco.languages.typescript.typescriptDefaults.addExtraLib(content, `file://${path}`);
+                    map.set(path, content);
+                });
+                setFsMap(map);
+            });
+        });
+    }, []);
+    return {fsMap, setFsMap, compilerOptions};
```
gdostie, line 60: When implementing the props table, I realized that the setFsMap doesn't need to be exposed. To add stuff to the file system we can just do `fsMap.set`

```suggestion
    return {fsMap, compilerOptions};
```

gdostie (approved):

PR #2362, opened 2021-12-20
feat(demo): implement PropsDoc component

gdostie:
### Proposed Changes

Implemented a table that displays the props of a component exported from react-vapor.

Usage:

```tsx
<PropsDoc componentName='Button' />
```

You can view the result here: https://vapor.coveo.com/feature/PR-2362/index.html#/input/Buttonv2

none

### Acceptance Criteria

### DIFF
--- a/packages/demo/src/demo-building-blocs/ExampleLayout.tsx
+++ b/packages/demo/src/demo-building-blocs/ExampleLayout.tsx
@@ -1,8 +1,11 @@
+import '@demo-styling/example-layout.scss';
+
 import * as React from 'react';
 import {useSelector} from 'react-redux';
 import {TabContent, TabPaneConnected, TabSelectors, TabsHeader, Tile} from 'react-vapor';
+
 import {GuidelinesTab} from './GuidelinesTab';
-import '@demo-styling/example-layout.scss';
+import {PropsDoc} from './PropsDoc';
 import Sandbox from './Sandbox';
 
 export interface ExampleLayoutProps {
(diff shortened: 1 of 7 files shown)


packages/demo/src/demo-building-blocs/PropsDoc.tsx
```diff
@@ -0,0 +1,101 @@
...
+    const {fsMap, compilerOptions} = useTypescriptServer();
+
+    React.useEffect(() => {
+        if (fsMap === null) {
+            return;
+        }
+        const system = createSystem(fsMap);
+        const content = `import * as React from 'react';
+import {${componentName}} from 'react-vapor';
+const props: React.ComponentProps<typeof ${componentName}> = {`;
```
aiheon, line 27:
```suggestion
        const content = `import * as React from 'react';
                                  import {${componentName}} from 'react-vapor';
                                  const props: React.ComponentProps<typeof ${componentName}> = {`;
```

gdostie: This would create useless whitespaces in the string

packages/demo/src/demo-building-blocs/ExampleLayout.tsx
```diff
@@ -43,32 +46,36 @@ export const ExampleLayout: React.FunctionComponent<ExampleLayoutProps> = ({
...
+    id,
+}) => (
     <>
         <div className="example-layout__main-code example-layout__section">
             <Sandbox id="main-code">{code}</Sandbox>
         </div>
         <div className="example-layout__props example-layout__section">
-            <h4 className="h2 mb5">Props</h4>
-            <div>Insert table props here</div>
+            <h2 className="mb1">Props</h2>
```
GermainBergeron, line 66: I used h4 to be semantically correct since the h2 (e.g., inputs) is the section and the h3 the title (e.g., buttons). I think we should use semantically correct tags and override the style with classes

GermainBergeron (approved):

packages/demo/src/demo-building-blocs/ExampleLayout.tsx
gdostie: oh right ok, I will revert back, I was wondering why it was that way

packages/demo/src/demo-building-blocs/PropsDoc.tsx
aiheon: oh i see  :eyes:

aiheon:
I approve but also I think we should find a way to make the "grattage de props" a little less aggressive? :laughing: 

https://user-images.githubusercontent.com/63734941/146932986-3968615e-942e-4ac0-ad62-aebbcb236266.mp4

aiheon (approved):

gdostie:
> I think we should find a way to make the "grattage de props" a little less aggressive? 😆

Haha yes I agree, but I haven't found a way to do so for now.

PR #2370, opened 2021-12-23
UI tool 209/get some result in search bar

aiheon:
### Proposed Changes

https://coveord.atlassian.net/browse/UITOOL-209

Now search and get results! :see_no_evil: 

Should be none

### Acceptance Criteria


FelixBlaisThon:
I don't know how it was locally, but when I type, nothing render as suggestions until I press ENTER. Also, I can't use the UP & DOWN arrows to _scroll_ through the results from the search (should it be possible?)

aiheon:
> I don't know how it was locally, but when I type, nothing render as suggestions until I press ENTER. Also, I can't use the UP & > DOWN arrows to _scroll_ through the results from the search (should it be possible?)

The enter press is normal, we don't have query suggestions yet :see_no_evil: 

The arrows were not working locally for me also, it's something I plan to work on in another pr/jira (that with the ''no result'' behavior)

gdostie (commented):
I seem to always have the 2 same homepage urls when making a search

image
image
image

aiheon:
> I seem to always have the 2 same homepage urls when making a search
> 
> image image image

Yeah I think it's because all the components name are on the home page, I need to find a way to remove it from the result without breaking the source yet :see_no_evil:

gdostie:
If I navigate to a component using the search bar results, the code snippet is not displayed

image

PR #2377, opened 2022-01-13
feat(tile): make title, description, href optional

gdostie:
### Proposed Changes

Making the title, description and href props optional on the tile component. We have a use-case in the website where we only want to display the icon in the tile.

None

### Acceptance Criteria

### DIFF
--- a/packages/react/src/components/tile/Tile.tsx
+++ b/packages/react/src/components/tile/Tile.tsx
@@ -3,20 +3,38 @@ import * as React from 'react';
 import {Svg, SvgNames} from '../svg';
 
 export interface TileProps {
-    title: string;
+    title?: string;
     svgName: SvgNames;
-    description: string;
-    href: string;
+    description?: string;
+    href?: string;
 }
 
-export const Tile: React.FunctionComponent<TileProps> = ({title, description, svgName, href}) => (
-    <a className="tile" href={href}>
+export const Tile: React.FunctionComponent<TileProps> = ({title, description, svgName, href}) => {
+    const tileIcon = (
         <div className="tile-icon">
             <Svg svgName={svgName} svgClass="icon" />
         </div>
+    );
+    const tileInfo = (title || description) && (
         <div className="tile-information">
-            <div className="tile-title h6-subdued">{title}</div>
-            <div className="tile-description body-m-book-subdued">{description}</div>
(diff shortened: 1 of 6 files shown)


aiheon:
I'm not sure why there's no demo link posted here :confused:  ...

https://vapor.coveo.com/feature/PR-2377/

aiheon:
> > I'm not sure why there's no demo link posted here confused ...
> > https://vapor.coveo.com/feature/PR-2377/index.html
> 
> I guess it is because the build failed

Maybe but still, if the demo stage pass, it should post the link IIRC :thinking: 
I'll take a look a it tomorrow

gdostie:
> 

I don't understand how you got a working link 😅 

image

aiheon:
> 

the first build went though he deploy pr demo fine, I got it there :eyes:

aiheon:
OK everything is fine :tada: demo is here!

aiheon (approved):

packages/react/src/components/tile/Tile.tsx
```diff
@@ -3,20 +3,38 @@ import * as React from 'react';
...
 
-export const Tile: React.FunctionComponent<TileProps> = ({title, description, svgName, href}) => (
-    <a className="tile" href={href}>
+export const Tile: React.FunctionComponent<TileProps> = ({title, description, svgName, href}) => {
+    const tileIcon = (
         <div className="tile-icon">
             <Svg svgName={svgName} svgClass="icon" />
         </div>
+    );
+    const tileInfo = title && description && (
```
sallainCoveo, line 18: I think that this condition would mean that you have to have a `title` and a `description` props set, right? Would it be better to check if one of the properties is set, and then check if the expected property is set (like you do below).
```suggestion
    const tileInfo = (title || description) && (
```

aiheon: mmhhh I think we don't want to have tile with only one or the other but always to have both a title and a description? :thinking: 

J'parle à travers mon chapeau, quote moi pas là-dessus :laughing:

sallainCoveo (approved):

packages/react/src/components/tile/Tile.tsx
gdostie: Yea good catch @sallainCoveo 👍

PR #2380, opened 2022-01-14
fix(actionableitem): editing component to hide active on actionableitem component

kdang-coveo:
The `ActionableItem` component contains the `ListBox` component which has the noActive prop but we cannot access this to remove the highlight on the first item in the ActionableItem.

### Proposed Changes

Setting the the ListBox component to have noActive will fix this problem.

Without this change, the item looks like it's already selected on mount (this is not what we want when using the ActionableItem)

**Before (when action clicked)**
image

**After (when action clicked)**
image

The highlighted of the first item will no longer be there

### Acceptance Criteria

### DIFF
--- a/packages/react/src/components/actionable-item/ActionableItem.tsx
+++ b/packages/react/src/components/actionable-item/ActionableItem.tsx
@@ -50,7 +50,7 @@ export class ActionableItem extends React.Component<IActionableItemProps & React
                         )}
                         {...(this.props.dropProps || {})}
                     >
-                        <ListBox items={this.props.actions} />
+                        <ListBox items={this.props.actions} noActive />
                     </Drop>
                 ) : null}
             </div>
--- a/packages/react/src/components/actionable-item/tests/ActionableItem.spec.tsx
+++ b/packages/react/src/components/actionable-item/tests/ActionableItem.spec.tsx
@@ -61,6 +61,13 @@ describe('ActionableItem', () => {
                 expect(screen.getByText('hello world')).toBeVisible();
             });
 
+            it('should not highlight dropdown options', () => {
+                const children = <div className="children">hello world</div>;
+                render(<ActionableItem {...basicProps}>{children}</ActionableItem>);
+
+                expect(screen.getByText('some action')).not.toHaveClass('active');
+            });
+
             it('should render the moreAppend svg in the Drop', () => {
                 shallowWithProps();
 


mikedidomizio:
> Without this change, the item looks like it's already selected on mount.

If I understand correctly, if it appears to be mounted, but isn't, should we instead make it not a prop and just force it to not highlight?  Is there a reason why someone would want it highlighted?

kdang-coveo:
> > Without this change, the item looks like it's already selected on mount.
> 
> If I understand correctly, if it appears to be mounted, but isn't, should we instead make it not a prop and just force it to not highlight? Is there a reason why someone would want it highlighted?

From what I understand, the Drop component is pretty complex and was created quite some time ago. The highlight is initially there to allow users to navigate through the Drop component using arrow key navigation, without the highlight, the user would not know which item is selected using arrow keys.

By default, it always selects the first item of the list and I think passing in an optional prop for now is the best way to handle this without having to introduce a breaking change and forcing it to be true.

The pattern is also repeated in the `ListBox` [component](https://github.com/coveo/plasma/blob/master/packages/react/src/components/listBox/ListBox.tsx#L20)

kdang-coveo:
PR closed - will need to refine the behaviour specifically for the items being shown in the Drop as the highlight and functionality not consistent

packages/react/src/components/actionable-item/ActionableItem.tsx
```diff
@@ -50,7 +50,7 @@ export class ActionableItem extends React.Component<IActionableItemProps & React
                         )}
                         {...(this.props.dropProps || {})}
                     >
-                        <ListBox items={this.props.actions} />
+                        <ListBox items={this.props.actions} noActive={true} />
```
gdostie, line 53:
```suggestion
                        <ListBox items={this.props.actions} noActive />

```

gdostie (approved):

packages/react/src/components/actionable-item/tests/ActionableItem.spec.tsx
```diff
@@ -61,6 +61,13 @@ describe('ActionableItem', () => {
                 expect(screen.getByText('hello world')).toBeVisible();
             });
 
+            it('should remove the active class from the children', () => {
```
mikedidomizio, line 64: Can we change the test name to something like
`should not highlight dropdown options` or `should not apply CSS class to dropdown options`

Just to possibly make it more clear on what the test is trying to prove.

mikedidomizio (approved):

PR #2392, opened 2022-01-21
refactor(website): reorganize and cleanup the side navigation sections

gdostie:
### Proposed Changes

Moved a lot of examples around. Also deleted legacy examples pages that we don't want people to reproduce.

Had to make some corrections to the ReactVaporState interface, it wasn't representing properly the real return type of some reducers.

### Acceptance Criteria


packages/react/src/ReactVaporState.ts
```diff
@@ -75,7 +75,7 @@ export interface IReactVaporState {
     groupableCheckboxes?: IGroupableCheckboxesState[];
     inputs?: IInputState[];
     itemFilters?: IItemFilterState[];
-    codeEditors?: CodeEditorState[];
+    codeEditors?: CodeEditorsState;
```
GermainBergeron, line 78: Is this a breaking change? 🤔

toofff (commented):
image

For:
- Info Box
- Labeled Value
- Limit Card

image
Small problem for the first two examples?

GermainBergeron (approved):

packages/website/src/pages/Home.tsx
```diff
@@ -191,10 +148,9 @@ const Layout: React.FC = () => (
     </Section>
 );
 
-const Input: React.FC = () => (
+const FormPages: React.FC = () => (
```
GermainBergeron, line 151: I think all the link here should be updated, they still target `#/input/something`

packages/react/src/ReactVaporState.ts
```diff
@@ -98,7 +98,7 @@ export interface IReactVaporState {
     searchBars?: ISearchBarState[];
     selects?: ISelectState[];
     selectWithFilter?: ISelectWithFilterCompositeState;
-    sliders?: ISliderState[];
+    sliders?: ISlidersState;
```
aiheon, line 101: same comment as Germain here :eyes:

packages/website/src/pages/foundations/index.tsx
```diff
@@ -2,21 +2,19 @@ import * as React from 'react';
...
         <Route path="SVG" element={<SvgExamples />} />
-        <Route path="Illustration" element={<div />} />
-        <Route path="Palette" element={<div />} />
-        {/* TODO: roll existing Typekit components... */}
-        <Route path="typekit/*" element={<Typekit />} />
-        {/* ...into new Typekit section */}
-        <Route path="Typekit" element={<div />} />
+        <Route path="Headings" element={<Headings />} />
+        <Route path="Links" element={<Links />} />
+        <Route path="Whitespace" element={<Whitespace />} />
```
aiheon, line 17: They are not in alphabetical order :female_detective:

aiheon (approved):

toofff:
> I'm wondering if we should keep all the subNav close by default? Do you guys think it's looking better? eyes image

They are open for me by default

aiheon:
> > I'm wondering if we should keep all the subNav close by default? Do you guys think it's looking better? eyes image
> 
> They are open for me by default

They are for me too, i<m just asking if it would be better looking to close them all and let the user open them as needed :see_no_evil:

toofff:
@aiheon sorry, I didn't understand everything :(

Yes everything is right, the page would be a little lighter with the menu closed by default. I just ran through some documentation site and it's rare to have the basic menu open

gdostie:
> @aiheon sorry, I didn't understand everything :(
> 
> Yes everything is right, the page would be a little lighter with the menu closed by default. I just ran through some documentation site and it's rare to have the basic menu open

Good idea, but out of scope for this PR! I will keep it in mind because it is more complicated than just closing them by default. I need to open the right collapsible in which the active section is.

PR #2403, opened 2022-01-28
feat(website): multiple improvements to the search

gdostie:
### Proposed Changes

- Added a default number of results per page of `1000` (it should never get to that we currently have 76 possible results)
- Excluded `node_modules` from `source-map-loader` to remove a bunch of warnings
- Stripped clickable URI to keep only relative links, this way the search works even in PRs and locally
- Simplified the search route from `#/plasma-search/ResultPage?query=whatever` to `#/search?q=whatever`
- Added key prop to search result tiles to remove warning
- Added the `@crawled-title` css class to identify the component titles both in the old and new layouts. The current web scraping configuration wasn't retrieving titles properly for those that had the new layout like the text input
image (instead of `Plasma Design System`, it should be `Text Input`). With this new `@crawled-title` class it will be a lot easier to get it right.

None

### Acceptance Criteria


packages/website/src/building-blocs/ExampleLayout.tsx
```diff
@@ -56,7 +56,7 @@ export const ExampleLayout: React.FunctionComponent<ExampleLayoutProps> = ({
                 >
                     View source
                 </GithubButton>
-                <h3 className="h1-light normal-white-space">{title}</h3>
+                <h3 className="h1-light normal-white-space @crawled-title">{title}</h3>
```
toofff, line 59: Why use a CSS class starting with a `@`?

packages/website/src/pages/SearchResultPage.tsx
```diff
@@ -31,9 +32,14 @@ const ResultListRenderer: FunctionComponent<ResultListProps> = (props) => {
                     <h2>Search Results:</h2>
                     <div className="body-l-book plasma-description">Patate King!</div>
                     <div className="tile-grid">
-                        {state.results.map((result: Result) =>
-                            result.title !== 'Plasma Design System' ? (
-                                <Tile title={result.title} href={result.clickUri} description={result.excerpt} />
+                        {state.results.map(({title, clickUri, excerpt, uniqueId}: Result) =>
+                            title !== 'Plasma Design System' ? (
```
toofff, line 36: it has nothing to do with your PR but instead of forgetting I'm telling you about it here :D

https://github.com/coveo/plasma/blob/master/packages/website/webpack.config.js#L39

toofff (approved):

packages/website/src/building-blocs/ExampleLayout.tsx
aiheon: This is not to be used by the css but by the Coveo Crawler for the website source :smile:

packages/website/src/pages/SearchResultPage.tsx
aiheon: i'll change it in my own pr, coming soon this morning :smile:

packages/website/src/building-blocs/ExampleLayout.tsx
gdostie: What Ariane says☝️

I just thought that if I don't put a special character here, one could see that its not applying any css rules and think that it can be removed

packages/website/src/pages/SearchResultPage.tsx
toofff: BANCO :D

PR #2405, opened 2022-01-31
fix(website): remove invalid character for css class

gdostie:
### Proposed Changes

The `@` is an invalid character for a css class, the web crawler cannot access it.

None

### Acceptance Criteria

### DIFF
--- a/packages/website/src/building-blocs/ExampleLayout.tsx
+++ b/packages/website/src/building-blocs/ExampleLayout.tsx
@@ -56,7 +56,7 @@ export const ExampleLayout: React.FunctionComponent<ExampleLayoutProps> = ({
                 >
                     View source
                 </GithubButton>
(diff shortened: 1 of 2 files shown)


aiheon (approved):

mikedidomizio:
Is there any other way we can still have something like [this](https://github.com/coveo/plasma/pull/2403#discussion_r795700282) so it doesn't get removed?  Attribute?

gdostie:
> Is there any other way we can still have something like [this](https://github.com/coveo/plasma/pull/2403#discussion_r795700282) so it doesn't get removed? Attribute?

The good way of doing this would be either an id or a data attribute, but the BasicHeader component that is used by the VaporComponent doesn't pass down those to the actual DOM nodes

mikedidomizio (approved):

PR #2456, opened 2022-02-11
UITOOL-402/add analytics

aiheon:
### Proposed Changes

https://coveord.atlassian.net/browse/UITOOL-402

I've added calls to send events to UA to feed our ML models down the line.

Currently, we have here click events for the side nav, the Tile in the Home page, the tile in the SearchResult page and the no results page. 

none

### Acceptance Criteria


packages/website/src/pages/SearchResultPage.tsx
```diff
@@ -45,13 +46,14 @@ const ResultListRenderer: FunctionComponent<ResultListProps> = (props) => {
...
-                                    title={title}
-                                    href={clickUri.slice(clickUri.indexOf('#'))}
-                                    description={raw.description as string}
-                                    thumbnail={raw.thumbnail as TileProps['thumbnail']}
+                                    key={result.uniqueId}
+                                    title={result.title}
+                                    href={result.clickUri.slice(result.clickUri.indexOf('#'))}
+                                    description={result.raw.description as string}
+                                    thumbnail={result.raw.thumbnail as TileProps['thumbnail']}
+                                    sendAnalytics={() => engine.dispatch(logDocumentOpen(result))}
```
gdostie, line 56: Why is here a document open event, and the other ones are custom?

aiheon: This one is taking a Result as param and I don't have them in other places than the SearchResultPage. 
It makes me want to make the Home page just a big result page :see_no_evil: But that would still not work for the sidenav...

mikedidomizio:
Since Plasma is a UI component framework and could be used by anyone, should we be putting analytics capturing into it or should that be put inside of our Platform?

aiheon:
> Since Plasma is a UI component framework and could be used by anyone, should we be putting analytics capturing into it or should that be put inside of our Platform?

I'm not sure to follow what you mean :sweat_smile:

mikedidomizio:
> > Since Plasma is a UI component framework and could be used by anyone, should we be putting analytics capturing into it or should that be put inside of our Platform?
> 
> I'm not sure to follow what you mean 😅

From what I understand this adds some analytics to various spots of components, but wouldn't it be more appropriate for the admin-ui to have the click events to send that data?  I know our primary customer with this is internal (us), but someone using Plasma externally may have no use for this additional logic.

aiheon:
> > > Since Plasma is a UI component framework and could be used by anyone, should we be putting analytics capturing into it or should that be put inside of our Platform?
> > 
> > 
> > I'm not sure to follow what you mean sweat_smile
> 
> From what I understand this adds some analytics to various spots of components, but wouldn't it be more appropriate for the admin-ui to have the click events to send that data? I know our primary customer with this is internal (us), but someone using Plasma externally may have no use for this additional logic.

The analytics is for the plasma website, not the components :see_no_evil: 
We want to track page views, clicks and other events to feed ML and activate the query suggestions in our search bar :smile: 
The TileComponent is not in the plasma react package, it cannot be used outside the website.

PR #2476, opened 2022-02-22
feat(website): automatically open the browser on start

gdostie:
### Proposed Changes

Since the move to nextjs, we had to manually open that localhost url.

None


aiheon (approved):

toofff (approved):
ok but if we might not use the `open-browser` command elsewhere, I think you can put everything in the `start` command

PR #2477, opened 2022-02-22
refactor(website): separate header from example layout

gdostie:
### Proposed Changes

Make it possible to use the new header without having code sandboxes, guidelines, props tables, etc.

It might break the web scraping configuration of our sitemap source, but I will check it out once its merged.


aiheon (approved):

packages/website/src/building-blocs/PageHeader.tsx
```diff
@@ -0,0 +1,44 @@
...
+    componentSourcePath,
+}) => (
+    <div className="plasma-page-header">
+        <h2 className="h5-subdued normal-white-space">{section}</h2>
+        <h1 className="h1-light normal-white-space" data-coveo-field="title">
+            {title}
+        </h1>
+        <h3 className="h4-book-subdued" data-coveo-field="description">
+            {description}
+        </h3>
```
toofff, line 33: There are small issues with the title architecture

**Tab Implementation**
image

**Tab Guidelines**
image

toofff (approved):
Apart from the problem with the title architecture, the rest is OK

packages/website/src/building-blocs/PageHeader.tsx
gdostie: What tool do you use to view those? 👀

gdostie: The same problem exists on master though, so out of scope for this PR

GermainBergeron (approved):

packages/website/src/building-blocs/PageHeader.tsx
toofff: See this chrome plugin https://chrome.google.com/webstore/detail/headingsmap/flbjommegcjonpdmenkdiocclhjacmbi?hl=en

Out of scope ok, but would it be possible for you to write a stain to fix this?

PR #2480, opened 2022-02-23
feat(website): update the iconography page

gdostie:
### Proposed Changes

image

None, the old icons still exist and the `Svg` component too

### Acceptance Criteria


packages/react-icons/bin/index.js
```diff
@@ -7,24 +7,39 @@ const {rmSync} = require('fs-extra');
...
+        const componentName = getComponentName(file);
         const fileContent = await fs.readFile(file);
         const tsCode = await svgr.transform(
             fileContent.toString('utf8'),
-            {typescript: true, exportType: 'named', namedExport: componentName},
+            {
+                typescript: true,
+                exportType: 'named',
+                namedExport: componentName,
+                svgProps: {className: '{"icon mod-stroke " + props.className}', },
```
gdostie, line 38: This means all icons now have `icon` `mod-stroke` classes by default

aiheon:
The demo freeze instantly on my end, I cannot use the website :sweat_smile: 
And it's not just my internet, mes others pages are fine and fluid...

packages/react-icons/bin/index.js
aiheon: won't this create conflicts with the style if we add mod-fill? :eyes:

packages/website/src/pages/foundations/Iconography.tsx
```diff
@@ -1,34 +1,102 @@
...
-                .map((svgName: SvgName) => (
-                    <IconItem key={svgName} svgName={svgName} fileName={Icons[svgName].fileName} />
-                ))}
-        </ul>
-    </PlasmaComponent>
+const {iconsList, ...Icons} = PlasmaReactIcons;
+
+const Sandbox = dynamic(
+    import('../../building-blocs/Sandbox').then((mod) => mod.Sandbox),
+    {ssr: false, loading: () => <PlasmaLoading />}
```
aiheon, line 14: I really like the built-in loading prop in Next :eyes: 
So much cleaner that the if(loading) {return <Loading />} else { return <MyPage />}

packages/react-icons/bin/index.js
gdostie: `mod-fill`? this class doesn't exist 🤔

GermainBergeron (approved):

packages/website/src/pages/foundations/Iconography.tsx
```diff
@@ -1,34 +1,102 @@
...
+
+export const IconographyExamples = () => (
+    <>
+        <PageHeader
+            section="Foundations"
+            title="Iconography"
+            thumbnail="iconography"
+            description="Icons are used to visually represent actions, functionalities, and features."
+        />
+        <div className="palsma-page-layout">
```
GermainBergeron, line 91: I don't know if we have the same typo elsewhere?
```suggestion
        <div className="plasma-page-layout">
```

GermainBergeron (changes requested):
Hmm after looking into the demo I'm not sure we should add `icon mod-stroke` otherwise the `stroke: currentColor` doesn't work. 

image

If I remove the icon class the size is wrong but the colour is correct

image

packages/website/src/pages/foundations/Iconography.tsx
```diff
@@ -1,34 +1,101 @@
...
+    import * as React from 'react';
+    import { InfoSize32Px } from '@coveord/plasma-react-icons';
+
+    
+    // Control the size using "height" or "width" attributes (defaults to 1em)
+    // The icon takes the same color as the text around it
+
+    export default () => (
+        <div style={{color: 'var(--info-60)'}}>
+            <InfoSize32Px height={64} />;
```
aiheon:
```suggestion
            <InfoSize32Px height={64} />

```
I think it's this one :see_no_evil: 
image

aiheon (approved):

GermainBergeron (approved):

PR #2551, opened 2022-03-15
refactor(website): update the InfoBox component demo page

gdostie:
### Proposed Changes

image

None

### Acceptance Criteria


aiheon (approved):

sallainCoveo (approved):

gdostie:
> Are the props relevant for the Infobox? 🤔 I think I would just remove it 👀

I think its best not to remove because if we add new props we won't think about going in the demo page to re-add it

aiheon:
> > Are the props relevant for the Infobox? thinking I think I would just remove it eyes
> 
> I think its best not to remove because if we add new props we won't think about going in the demo page to re-add it

That is a really good point :+1: :smile:

PR #2555, opened 2022-03-16
fix(style): rounded inline-prompt button

mikedidomizio:
### Proposed Changes

Moves the `.prompt-action` outside of `.action` which was overriding `border-radius`.  The rest of `.action` untouched as it may still be used.

| Before | After
| ---- | ---- |
|  |  |

If anything moving the SCSS not being required to be `.action.prompt-action` and now applying the styling to just `.prompt-action`.  

### Acceptance Criteria

### DIFF
--- a/packages/react/src/components/inlinePrompt/tests/InlinePrompt.spec.tsx
+++ b/packages/react/src/components/inlinePrompt/tests/InlinePrompt.spec.tsx
@@ -1,6 +1,5 @@
 import {mount, ReactWrapper, shallow} from 'enzyme';
 import * as React from 'react';
-import * as _ from 'underscore';
 
 import {IInlinePromptOptions, InlinePrompt} from '../InlinePrompt';
 
@@ -48,7 +47,7 @@ describe('InlinePrompt', () => {
         });
 
         it('should have the className as a class if defined', () => {
-            const newOptions: IInlinePromptOptions = _.extend({}, options, {className: undefined});
+            const newOptions: IInlinePromptOptions = Object.assign({}, options, {className: undefined});
 
             expect(inlinePrompt.find('.prompt-' + options.className).length).toBe(1);
 
@@ -59,7 +58,7 @@ describe('InlinePrompt', () => {
         });
 
         it('should have the opened as a class if opened', () => {
-            const newOptions = _.extend({}, options, {isOpened: true});
+            const newOptions = Object.assign({}, options, {isOpened: true});
 
             expect(inlinePrompt.find('.opened').length).toBe(0);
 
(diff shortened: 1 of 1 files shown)


packages/react/src/components/inlinePrompt/InlinePrompt.tsx
```diff
@@ -30,12 +30,7 @@ export const InlinePrompt: React.FunctionComponent<
     IInlinePromptProps & Partial<ReturnType<typeof mapDispatchToProps>>
 > = ({options, onCancel}) => {
     const choices: JSX.Element[] = _.map(options.userChoice.choices, (choice: string) => (
-        <button
-            type="button"
-            className="btn action mod-danger prompt-action enabled"
-            onClick={options.onClick}
-            key={choice}
-        >
+        <button type="button" className="btn mod-danger prompt-action enabled" onClick={options.onClick} key={choice}>
```
gdostie, line 33: Lets use a `Button` component instead

packages/react/src/components/inlinePrompt/InlinePrompt.tsx
```diff
@@ -30,24 +31,19 @@ export const InlinePrompt: React.FunctionComponent<
     IInlinePromptProps & Partial<ReturnType<typeof mapDispatchToProps>>
 > = ({options, onCancel}) => {
     const choices: JSX.Element[] = _.map(options.userChoice.choices, (choice: string) => (
-        <button
-            type="button"
-            className="btn action mod-danger prompt-action enabled"
-            onClick={options.onClick}
-            key={choice}
-        >
+        <Button className="btn mod-danger prompt-action enabled" onClick={options.onClick} key={choice}>
```
gdostie, line 34: Is the `prompt-action` still needed now that we're using the Button component?

```suggestion
        <Button className="mod-danger" onClick={options.onClick} key={choice}>
```

gdostie (approved):

packages/react/src/components/inlinePrompt/InlinePrompt.tsx
mikedidomizio: For this fix, yes it's still needed.  As the first button relies on the `margin-left` of [.prompt-action](https://github.com/coveo/plasma/pull/2555/files#diff-fd9e26d5638b15caa5356b2d413f8588607d199e3bb25810ab762dcc378c6ff0R44)


With no `.prompt-action`
image

gdostie: Lets use an existing class if its only for the margin then, it will avoid having one more css class to maintain.

What about `ml1`?

PR #2562, opened 2022-03-16
refactor(website): update ChildForm component demo page

gdostie:
### Proposed Changes

image

None

### Acceptance Criteria


toofff (approved):

packages/website/src/SideNavigation.tsx
```diff
@@ -88,6 +88,7 @@ export const Navigation: React.FunctionComponent = () => {
                 <NavLink href="/layout/BorderedLine" label="Bordered line" />
                 <NavLink href="/layout/BrowserPreview" label="Browser preview" />
                 <NavLink href="/layout/Chart" label="Chart" />
+                <NavLink href="/layout/ChildForm" label="Child form" />
```
aiheon, line 91: you forgot to change it in the home page :eagle:

aiheon: also would it be overkill to patante something to make the homepage build automatically :see_no_evil:

aiheon: from the home page, breaks because the component is not there's anymore... maybe we should also think of a better 404 page? :eyes: 
image

PR #2599, opened 2022-03-24
MVP guidelines for Multiline box

kpoulincoveo:
Was reviewed by UX and Danny

### Proposed Changes

New guidelines

None I hope

### Acceptance Criteria

### DIFF
--- a/packages/website/docs/MultilineBox.md
+++ b/packages/website/docs/MultilineBox.md
@@ -1,45 +1,46 @@
 # Best Practices
-The Multiline box allows users to **add the same sequence of inputs multiple times**.  
-Each sequence appears on a different line, making it easier to consult.  
 
-Do not use the multiline box to regroup inputs that have no connection to one another.   
-Multiline works best when the sequence of inputs is identical on each line.  
+A multiline box allows users to **add multiple inputs or multiple series of inputs, each on a different line**.
 
-Multiline box can combine multiple inputs together to form a more complex input.  
-They can also combine text elements when an input or sequence requires more explanation.  
+Inputs can be combined into a series to form a more complex input.  
+Text elements can also be included when an input or series requires more explanation.
 
-Only use the drag and drop function if the order of each sequence is important.   
-When doing so, the list must always be presented to the user in the order they have set.
+A multiline box works best when the sequence of inputs is identical on all lines. 
+Don't use the multiline box to group inputs that have no connection to one another.    
+
+
+Only offer the drag and drop feature when the order of the inputs or series is important.
+In such case, the inputs/series must always be presented in the order set by the user.
 
 # Labeling
+
 Keep titles short, preferably **under three words**.  
 
 If the list of inputs is very simple, provide a descriptive title without action verbs.  
 For example, write "Email list" rather than "Provide a list of emails”.
 
-If the sequence of inputs is more complex, consider using a title with a trigger verb to clearly explain that the outcome will apply or consider the entire list.  
+If the series of inputs are complex, consider using a title with an action verb to clarify the outcome of the choice to make.  
 For example, write “Show content that matches" instead of "Filter rules”. 
 
 When using a dialog approach, always consider other inputs in the form so that it is cohesive with other inputs.
 
 # Help Text and Instructions
-Avoid the need for help text or instruction for each individual input that appears in a multiline line. 
 
-Instead, provide short instructions, above the lines, that will briefly explain the type of inputs to be provided in the input sequence.
+Avoid adding help text or instructions for each individual input that appears in a series. 
+Instead, provide brief instructions above the multiline box to explain the inputs in the series.
 
 # Feedback and Validation
+
 Make sure that a first sequence is **completed and valid** before allowing the addition of a new line. 
 
 If the multiline is optional, do not show an empty first line. Users should be required to add it manually if they need it. 
-In the same fashion, if the multiline is mandatory, the first line should be visible and the user should not be allowed to delete it. 
+If users must fill at least one line, the first line should be visible and users should not be allowed to delete it. 
 
-Always let the users **add a new line manually** instead of automating it. This avoids empty lines when the form is submitted. 
+**Let users add new lines manually** rather than automating the process. This pevents submitting forms that contain empty lines.
 
-If each line has a sequence of inputs of variable width, make sure lines width at least allows for delete buttons to appear vertically aligned.
-
+When each line has a series of inputs of variable width, ensure line width allows vertical alignment of the delete buttons.
 
 # Related Components
 
-If your use case doesn't match the guidelines above, consider using one of the following components instead:
-[Multiselect](#/form/multiselect) when the user only has to select options from a single list.
+When the user must select one or more options from a single list, consider using a [multi select](https://plasma.coveo.com/form/MultiSelect) instead.
 If users don’t need to repeat inputs, use a custom form layout instead.
--- a/packages/website/docs/MultilineBox.md
+++ b/packages/website/docs/MultilineBox.md
@@ -8,7 +8,6 @@ Text elements can also be included when an input or series requires more explana
 A multiline box works best when the sequence of inputs is identical on all lines. 
 Don't use the multiline box to group inputs that have no connection to one another.    
 
-
 Only offer the drag and drop feature when the order of the inputs or series is important.
 In such case, the inputs/series must always be presented in the order set by the user.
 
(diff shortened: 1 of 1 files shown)


dmgauthier (approved):

FelixBlaisThon (approved):
LGTM

packages/website/docs/MultilineBox.md
```diff
@@ -0,0 +1,45 @@
+# Best Practices
+The Multiline box allows users to **add the same sequence of inputs multiple times**.
```
jlevesquecoveo, line 2: @kpoulincoveo Before I review this, I would need to clarify the terminology we use. This is the component that allows users to enter multiple URLs when creating a web source, right?

```suggestion
The multiline box allows users to **provide several inputs of the same type**.  
```
or 
`The multiline box allows users to **provide a series of inputs of the same type**.`
Not sure what we mean with "sequence"... 🤔

aiheon: I Like the second option better personally :eyes:

kpoulincoveo: @jlevesquecoveo The multiline also serves to build complex conditions (ex: Filter expression) 

By "sequence", I mean that each line can hold multiple inputs (ex: first name + last name + email). That bundle of different input is a sequence. And each of line can be repeated multiple times (ex: to provide a list of people)

In some cases, the sequence of inputs can vary from one line to another (although it's not recommended to do it but some case allows it such as condition. That means that the first line could have a sequence of (field+comparator+single select) but the second line could have only (field+comparator+numerical input)


I let you decide which version of what you propose works best.

jlevesquecoveo: OK, thanks for these details, @kpoulincoveo! Can the sequence contain only one input, though? Like this:
image
Or is this a different component?

kpoulincoveo: Yes in that case it can (weither or not you use Multiline depend on the type of input, for instance, having a single select only on each line should be using a multi-select instead)

packages/website/docs/MultilineBox.md
```diff
@@ -0,0 +1,45 @@
...
+
+Always let the users **add a new line manually** instead of automating it. This avoids empty lines when the form is submitted. 
+
+If each line has a sequence of inputs of variable width, make sure lines width at least allows for delete buttons to appear vertically aligned.
+
+
+# Related Components
+
+If your use case doesn't match the guidelines above, consider using one of the following components instead:
+[Multiselect](#/form/multiselect) when the user only has to select options from a single list.
```
gdostie, line 44: This link won't work. I think you need to use `https://plasma.coveo.com/form/MultiSelect`

packages/website/docs/MultilineBox.md
```diff
@@ -0,0 +1,45 @@
+# Best Practices
```
gdostie, line 1: Add a new line after each heading please

```suggestion
# Best Practices

```

packages/website/docs/MultilineBox.md
```diff
@@ -0,0 +1,45 @@
...
+
+# Help Text and Instructions
+Avoid the need for help text or instruction for each individual input that appears in a multiline line. 
+
+Instead, provide short instructions, above the lines, that will briefly explain the type of inputs to be provided in the input sequence.
+
+# Feedback and Validation
+Make sure that a first sequence is **completed and valid** before allowing the addition of a new line. 
+
+If the multiline is optional, do not show an empty first line. Users should be required to add it manually if they need it.
```
gdostie, line 33: This is not a behaviour the component currently implements AFAIK

packages/website/docs/MultilineBox.md
jlevesquecoveo, line 44:
```suggestion
When the user must select one or more options from a single list, consider using a [multi select](https://plasma.coveo.com/form/MultiSelect) instead.
```

packages/website/docs/MultilineBox.md
```diff
@@ -0,0 +1,45 @@
...
+Always let the users **add a new line manually** instead of automating it. This avoids empty lines when the form is submitted. 
+
+If each line has a sequence of inputs of variable width, make sure lines width at least allows for delete buttons to appear vertically aligned.
+
+
+# Related Components
+
+If your use case doesn't match the guidelines above, consider using one of the following components instead:
+[Multiselect](#/form/multiselect) when the user only has to select options from a single list.
+If users don’t need to repeat inputs, use a custom form layout instead.
```
jlevesquecoveo, line 45: @kpoulincoveo I don't understand this, can you explain plz? 🙂

jlevesquecoveo (changes requested):

packages/website/docs/MultilineBox.md
dmgauthier: The component indeed renders a new "empty" box by default when it mounts, it is in fact problematic for my current work with the `queryExpressionBuilder` and I though about adding some `doNotAddAnEmptyBoxByDefaultBecauseIDontWantOne` prop down the road.

kpoulincoveo (commented):
@jlevesquecoveo we can move on with proofread. sorry for the late response

packages/website/docs/MultilineBox.md
```diff
@@ -0,0 +1,46 @@
...
+In such case, the inputs/series must always be presented in the order set by the user.
+
+# Labeling
+
+Keep titles short, preferably **under three words**.  
+
+If the list of inputs is very simple, provide a descriptive title without action verbs.  
+For example, write "Email list" rather than "Provide a list of emails”.
+
+If the series of inputs are complex, consider using a title with an action verb to clarify the outcome of the choice to make.
```
jlevesquecoveo, line 22:
```suggestion
If the series of inputs is complex, consider using a title with an action verb to clarify the outcome of the choice to make.  
```

packages/website/docs/MultilineBox.md
```diff
@@ -0,0 +1,46 @@
...
+
+# Labeling
+
+Keep titles short, preferably **under three words**.  
+
+If the list of inputs is very simple, provide a descriptive title without action verbs.  
+For example, write "Email list" rather than "Provide a list of emails”.
+
+If the series of inputs are complex, consider using a title with an action verb to clarify the outcome of the choice to make.  
+For example, write “Show content that matches" instead of "Filter rules”.
```
jlevesquecoveo, line 23:
```suggestion
For example, write “Show content that matches..." instead of "Filter rules”. 
```

packages/website/docs/MultilineBox.md
```diff
@@ -0,0 +1,45 @@
...
+# Labeling
+Keep titles short, preferably **under three words**.  
+
+If the list of inputs is very simple, provide a descriptive title without action verbs.  
+For example, write "Email list" rather than "Provide a list of emails”.
+
+If the sequence of inputs is more complex, consider using a title with a trigger verb to clearly explain that the outcome will apply or consider the entire list.  
+For example, write “Show content that matches" instead of "Filter rules”. 
+
+When using a dialog approach, always consider other inputs in the form so that it is cohesive with other inputs.
```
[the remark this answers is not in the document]
jlevesquecoveo:
```suggestion
You may choose to use a dialog approach with the labels. For example, you could have: "Among all items matching... [condition1] Show only the items with... [condition2]". In such case, consider the other inputs in the form so that your approach is coherent with all inputs.
```

packages/website/docs/MultilineBox.md
```diff
@@ -0,0 +1,45 @@
...
+
+# Help Text and Instructions
+Avoid the need for help text or instruction for each individual input that appears in a multiline line. 
+
+Instead, provide short instructions, above the lines, that will briefly explain the type of inputs to be provided in the input sequence.
+
+# Feedback and Validation
+Make sure that a first sequence is **completed and valid** before allowing the addition of a new line.
```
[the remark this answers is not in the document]
jlevesquecoveo:
```suggestion
Ensure that all inputs on the first line are **complete and valid** before allowing the addition of a new line. 
```

packages/website/docs/MultilineBox.md
jlevesquecoveo:
```suggestion
If the multiline box is optional, don't show an empty first line. Users should add it manually if they need it. 
```

packages/website/docs/MultilineBox.md
jlevesquecoveo:
```suggestion
Don't use a multiline box to group inputs that have no connection to one another. Use a custom form layout instead. For example, if you want a user to provide their name and email address, just provide two text inputs. However, if you want them to provide the name and email address of several people, use a multiline box and show the name and email inputs on one line for each person.
```

jlevesquecoveo (approved):

kpoulincoveo:
@gdostie Ready to Merge

PR #2603, opened 2022-03-29
fix(droppod): keep the original position of the drop after first render

aiheon:
### Proposed Changes

https://coveord.atlassian.net/browse/UITOOL-362

I'm adding a condition to check if the position has already be determine and if so, I skip the loop that was making the position change.

To test, go in the Single select page and open the dropdown when there's room below or on the input. If you scroll down or up after opening it, it should not change position.

None

### Acceptance Criteria

### DIFF
--- a/packages/react/src/components/drop/DropPod.tsx
+++ b/packages/react/src/components/drop/DropPod.tsx
@@ -2,7 +2,6 @@ import * as React from 'react';
 import * as ReactDOM from 'react-dom';
 import * as _ from 'underscore';
 
-import {Defaults} from '../../Defaults';
 import {BrowserUtils} from '../../utils/BrowserUtils';
 import {
     DomPositionCalculator,
@@ -46,6 +45,10 @@ const RDropPod: React.FunctionComponent<IRDropPodProps> = ({
     renderDrop,
     parentSelector,
 }) => {
+    const DEFAULT_ROOT: string = 'body';
+
+    const [isOriginalPositionSet, setOriginalPosition] = React.useState(false);
+
     const [offset, setOffset] = React.useState(undefined);
     const [lastPosition, setLastPosition] = React.useState<IDropUIPosition>({
         position: positions?.[0] ?? DropPodPosition.bottom,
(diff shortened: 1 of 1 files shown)


toofff (approved):

packages/react/src/components/drop/DropPod.tsx
```diff
@@ -93,20 +96,20 @@ const RDropPod: React.FunctionComponent<IRDropPodProps> = ({
 
     React.useEffect(() => {
         const element = document.createElement('div');
-        const portalRoot = document.querySelector(selector ?? Defaults.DROP_ROOT);
+        const portalRoot = document.querySelector(selector ?? DEFAULT_ROOT);
```
gdostie, line 99: Why remove the usage of `Defaults.DROP_ROOT`? I'm not sure this is a good idea

GermainBergeron: Defaults.DROP_ROOT can be set, let's say in ADUI to change the default container of the drop element. With your new way it's not possible anymore

aiheon: I see :thinking:

packages/react/src/components/drop/DropPod.tsx
```diff
@@ -134,10 +137,13 @@ const RDropPod: React.FunctionComponent<IRDropPodProps> = ({
                 height: Math.max(dropOffset?.height ?? 0, minHeight),
             };
 
-            while (_.isEmpty(newDomPosition) && index < positions.length) {
-                const validator = DomPositionCalculator[positions[index]];
-                newDomPosition = (validator && validator(buttonOffset, dropOffsetPrime, boundingLimit, {})) || {};
-                index += 1;
+            if (!isOriginalPositionSet) {
```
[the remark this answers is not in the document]
aiheon: on render, the lastPosition is bottom by default so the drop will always open on the bottom :thinking: it's a bit weird effect :eyes: 


https://user-images.githubusercontent.com/63734941/160905871-ecb241d8-5cc2-4f87-ab6e-5e90ccd739c6.mp4

packages/react/src/components/drop/DropPod.tsx
gdostie: What Germain said, for example as you can see [here](https://github.com/coveo/admin-ui/blob/master/app/src/CoveoAdmin.ts#L211) in the admin-ui we want all the drops to render in the `#admin-drop-container` div and not the body.

GermainBergeron (approved):

PR #2614, opened 2022-04-01
feat(card): added props to iconcard

nathanlb:
### Proposed Changes
On the IconCard component:

- Modified the title prop type from `string` to `ReactNode`
- Added props
```typescript
/**
 * Whether to place the badges above the title.
 */
placeBadgesAbove?: boolean;
/**
 * Classes to apply to the card button element.
 */
cardClasses?: string[];
```
- Added `mod-fixed-size` class in css for use with the cardClasses prop.

| Goal | Result (using the new prop update) |
|---|---|
|Screen Shot 2022-04-01 at 4 38 06 PM|Screen Shot 2022-04-01 at 5 29 19 PM|

There shouldn't be any if I did it right.

### Acceptance Criteria

### DIFF
--- a/packages/react/src/components/iconCard/IconCard.tsx
+++ b/packages/react/src/components/iconCard/IconCard.tsx
@@ -64,7 +64,7 @@ export interface IconCardProps {
     /**
      * Classes to apply to the card button element.
      */
-    cardClasses?: string[];
+    cardClassName?: string[];
 }
 
 export const IconCard: React.FunctionComponent<Override<
@@ -83,7 +83,7 @@ export const IconCard: React.FunctionComponent<Override<
     choices,
     animateOnHover,
     placeBadgesAbove,
-    cardClasses,
+    cardClassName,
     className,
     small,
     children,
@@ -133,7 +133,7 @@ export const IconCard: React.FunctionComponent<Override<
                             'cursor-pointer': (!!onClick || hasChoices) && !disabled && !isOpen,
                             'mod-small': !!small,
                         },
-                        cardClasses
+                        cardClassName
                     )}
                     onClick={handleCardClick}
                     aria-expanded={isOpen}
--- a/packages/react/src/components/iconCard/IconCard.tsx
+++ b/packages/react/src/components/iconCard/IconCard.tsx
@@ -4,7 +4,7 @@ import * as React from 'react';
 import {slugify} from 'underscore.string';
 
 import {SlideY} from '../../animations';
-import {TooltipPlacement} from '../../utils';
+import {Override, TooltipPlacement} from '../../utils';
 import {Badge, IBadgeProps} from '../badge/Badge';
 import {SvgChild, SvgChildProps, OptionalSvgChildProps} from '../svg/SvgChild';
 import {ITooltipProps, Tooltip} from '../tooltip';
@@ -20,7 +20,7 @@ export interface IconCardProps {
     /**
      * The main text displayed on the card
      */
-    title: string;
+    title: React.ReactNode;
     /**
      * The secondary text displayed on the card
      */
@@ -57,11 +57,20 @@ export interface IconCardProps {
      * @default false
      */
     disabled?: boolean;
+    /**
+     * Whether to place the badges above the title.
+     */
+    placeBadgesAbove?: boolean;
+    /**
+     * Whether the card should have a fixed width of `456px` and height of `216px`.
+     */
+    fixedSize?: boolean;
 }
 
-export const IconCard: React.FunctionComponent<
-    IconCardProps & SvgChildProps & Omit<React.HTMLAttributes<HTMLDivElement>, 'onClick'>
-> = ({
+export const IconCard: React.FunctionComponent<Override<
+    SvgChildProps & React.HTMLAttributes<HTMLDivElement>,
+    IconCardProps
+>> = ({
     title,
     badges = [],
     description,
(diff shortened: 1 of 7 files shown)


packages/react/src/components/iconCard/tests/IconCard.spec.tsx
```diff
@@ -267,4 +270,27 @@ describe('IconCard', () => {
         const card = container.querySelector('.icon-card');
         expect(card).not.toHaveClass('animateOnHover');
     });
+
+    it('should NOT render the badges within the main content container if placeBadgesAbove is unset', () => {
```
aiheon, line 274:
```suggestion
    it('does not render the badges within the main content container if placeBadgesAbove is unset', () => {

```

aiheon: I get what you mean by unset, but I think false is more clear :see_no_evil:

aiheon:
I'm not sure why but the new props are not showing up in the props table? :thinking:  
It should get there automatically :eyes: 
image

GermainBergeron:
> I'm not sure why but the new props are not showing up in the props table? 🤔 It should get there automatically 👀 image

They are at the bottom of the props table 🤔

packages/react/src/utils/TypeUtils.ts
```diff
@@ -0,0 +1 @@
+export type Override<T1, T2> = Omit<T1, keyof T2> & T2;
```
GermainBergeron, line 1: I think if you do it that way the order in the props table will be correct
```suggestion
export type Override<T1, T2> = T2 & Omit<T1, keyof T2>;
```

aiheon:
> > I'm not sure why but the new props are not showing up in the props table? thinking It should get there automatically eyes image
> 
> They are at the bottom of the props table thinking

That's almost weirder :see_no_evil:

packages/react/src/components/iconCard/IconCard.tsx
```diff
@@ -57,11 +57,20 @@ export interface IconCardProps {
...
      */
     disabled?: boolean;
+    /**
+     * Whether to place the badges above the title.
+     */
+    placeBadgesAbove?: boolean;
+    /**
+     * Classes to apply to the card button element.
+     */
+    cardClasses?: string[];
```
GermainBergeron, line 67: I would tend to use something closer to React naming, something like `cardClassName` and make it a string. In the admin you can use className if you need something more complex
```suggestion
    /**
     * Additional CSS class to apply to the card button element.
     */
    cardClassName?: string;
```

GermainBergeron (commented):
I have a few suggestions but overall it looks good!

packages/react/src/utils/TypeUtils.ts
nathanlb: It does. 👍  Well spotted!

packages/react/src/components/iconCard/tests/IconCard.spec.tsx
nathanlb: Yep totally fair

aiheon (approved):

packages/react/src/components/iconCard/IconCard.tsx
nathanlb: Yeah totally could. It would fit the React className typing. When initially deciding I did a word search in the project and landed on other component that define it as an array and went with that. I think it was the BlankSlate which is where I also got the original name of `cardClasses`.
I'll change it because I agree your suggestion is better.

nathanlb: I just realized you already made that change in your suggestion and I let it slip. Oups!

GermainBergeron: 👍 
While we worked on the sandbox/documenting props we saw a lot of those classes `string[]` and we think it's weird to use compared to `className`

kdang-coveo (approved):

kdang-coveo:
Works well! its just I notice that the title does not respect the `mb2` or padding classes for that matter 🤔 should we get UX to quickly skim through the padding for this component?

packages/react/src/components/iconCard/IconCard.tsx
```diff
@@ -73,6 +82,8 @@ export const IconCard: React.FunctionComponent<
     tooltip,
     choices,
     animateOnHover,
+    placeBadgesAbove,
```
kdang-coveo, line 85: Will we have a time where the badges will be at the bottom of the title?

nathanlb: by default they're on the right side of the content as the container is display: flex with standard horizontal direction.
I would be surprised if UX implemented a variant with the badges below as it would be so similar to having them above and would just be annoying for us. 😛 
Am I answering the question?

kdang-coveo:
> > Works well! its just I notice that the title does not respect the `mb2` or padding classes for that matter 🤔 should we get UX to quickly skim through the padding for this component?
> 
> which `mb2` class are you referring to? You mean if you pass in a node using that class it doesn't work? margin and padding classes don't work on a span element but they work if you pass in a div instead.


Yup! please ignore comment haha, I just tried it with a div and it works with that :) 👍

packages/react/src/components/iconCard/IconCard.tsx
kdang-coveo: ahh yes okay that makes sense, 👍 I forgot we were also on top of the `IconCard` for this that utilizes the badges on the left

PR #2653, opened 2022-04-14
UITOOL-590/mvp badge component

aiheon:
### Proposed Changes

https://coveord.atlassian.net/browse/UITOOL-590

image
image
image

Revamped Badges! 
After discussion with UX we chose to remove the `Thin` variant to keep a font-weight uniform between all variants.

There's 2 new props: `type`  and `iconPlacement`. The latest is mandatory only if the badge contain an icon.

The `mod-tag` was also removed from the css and can be replaced with the `BadgeType.Default`.

### Acceptance Criteria


GermainBergeron:
From your screenshot I think you still have alignment issues 🤔 
image

packages/react/src/components/badge/Badge.tsx
```diff
@@ -24,35 +63,39 @@ interface BadgeWithIconProps extends BadgeBasicProps {
...
-                        svgName={this.props.icon}
-                        svgClass={classNames('icon', {'mod-16': !this.isSmall, 'mod-12': this.isSmall})}
-                        className={classNames({mr1: 'label' in this.props && this.props.label})}
-                    />
+            <div className={this.className} aria-label="badge">
+                {'icon' in this.props && this.props.icon && IconPlacementMapping[this.props.iconPlacement] ? (
+                    <Svg svgName={this.props.icon} className={'mod-badge'} svgClass={'icon mod-badge'} />
+                ) : null}
+                {'label' in this.props ? <div className="badge_label">{this.props.label}</div> : null}
+                {'icon' in this.props && this.props.icon && !IconPlacementMapping[this.props.iconPlacement] ? (
```
GermainBergeron, line 95:
```suggestion
```

packages/react/src/components/badge/Badge.tsx
```diff
@@ -5,12 +5,51 @@ import {Svg} from '../svg';
...
+
+export enum BadgeIconPlacement {
+    Left,
+    Right,
+}
+
+const IconPlacementMapping: Record<BadgeIconPlacement, boolean> = {
+    [BadgeIconPlacement.Left]: true,
+    [BadgeIconPlacement.Right]: false,
+};
```
GermainBergeron, line 25: See my suggestion down bellow
```suggestion
```

packages/react/src/components/badge/Badge.tsx
```diff
@@ -24,35 +63,39 @@ interface BadgeWithIconProps extends BadgeBasicProps {
...
-                {'icon' in this.props ? (
-                    <Svg
-                        svgName={this.props.icon}
-                        svgClass={classNames('icon', {'mod-16': !this.isSmall, 'mod-12': this.isSmall})}
-                        className={classNames({mr1: 'label' in this.props && this.props.label})}
-                    />
+            <div className={this.className} aria-label="badge">
+                {'icon' in this.props && this.props.icon && IconPlacementMapping[this.props.iconPlacement] ? (
+                    <Svg svgName={this.props.icon} className={'mod-badge'} svgClass={'icon mod-badge'} />
+                ) : null}
```
GermainBergeron, line 93: I'm thinking that we should move the icon with CSS instead of having it render at a different place. Since you're using flex you can check the `order` property. (I think it would be better to always have it after the label)

```suggestion
                {'icon' in this.props ? (
                    <Svg svgName={this.props.icon} className={className('mod-badge', {'mod-right': this.props.iconPlacement === BadgeIconPlacement.Right})} svgClass={'icon mod-badge'} />
                ) : null}
```

packages/style/scss/colors.scss
```diff
@@ -108,7 +108,7 @@ $secondary: (
 );
 
 $semantic: (
-    'success-10': hsla(152, 87%, 91%, 1),
+    'success-00': hsla(152, 87%, 91%, 1),
```
GermainBergeron, line 111: Is that used in the admin? It would be considered a breaking change?

aiheon: since 3 days ago :sob: 
image

I can revert this actually it was just to have the same name than in Figma :smiling_face_with_tear:

aiheon:
> From your screenshot I think you still have alignment issues thinking image

I think it's because if I place the icon in the first column, the text is then centered on the svg and vice-versa :thinking: 
I also don't think this would be too mush of a issue if you don't have both badge with inverses icons next to each other :laughing: 

image

packages/react/src/components/badge/Badge.tsx
```diff
@@ -24,35 +58,42 @@ interface BadgeWithIconProps extends BadgeBasicProps {
      * Add an icon to the badge (Required if no label)
      */
     icon: SvgName;
+    /**
+     * Whether the icon is left or right (Required if no label)
+     */
+    iconPlacement: BadgeIconPlacement;
```
GermainBergeron, line 64: I don't think it should be required, the default is `BadgeIconPlacement.Left`

packages/react/src/components/badge/Badge.tsx
```diff
@@ -24,35 +58,42 @@ interface BadgeWithIconProps extends BadgeBasicProps {
...
 
 export class Badge extends React.Component<IBadgeProps> {
-    static defaultProps: Partial<IBadgeProps> = {
+    static defaultProps: IBadgeProps = {
         extraClasses: [],
+        type: BadgeType.Default,
+        isSmall: false,
+        icon: undefined,
+        iconPlacement: undefined,
     };
```
GermainBergeron, line 75: Why do you set the defaultProps to undefined instead of just not defining a default value?

```suggestion
    static defaultProps: Partial<IBadgeProps> = {
        extraClasses: [],
        type: BadgeType.Default,
        isSmall: false,
        iconPlacement: BadgeIconPlacement.Left,
    };
```

GermainBergeron (approved):

packages/react/src/components/badge/Badge.tsx
aiheon: I've put `label: ' '` instead. I need icon or label otherwise the code throws errors 😕

gdostie (approved):

PR #2657, opened 2022-04-15
feat(icons): lazy load react-icons

GermainBergeron:
### Proposed Changes

While debugging in another package we saw that loading the icon package could take multiple seconds in Jest (using commonJs). To avoid this we wrap the component in a Loadable component. It should be transparent when using the library.

The d.ts did change a bit, if developer were importing the component from /dist/some/path/sizeXPx it now use a default export

### Acceptance Criteria

### DIFF
--- a/packages/react-icons/bin/index.js
+++ b/packages/react-icons/bin/index.js
@@ -18,10 +18,7 @@ const getComponentName = (file) => {
 
 const convertIcon = async ([iconName, variants]) => {
     await fs.ensureDir(`./generated/${iconName}`);
-    return Promise.all([
-        ...variants.map(convertVariant),
-        fs.appendFile('./generated/index.ts', `export * from './${iconName}';\n`),
-    ]);
+    return Promise.all(variants.map(convertVariant));
 };
 
 const convertVariant = async (file) => {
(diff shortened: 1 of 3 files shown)


packages/react-icons/bin/index.js
```diff
@@ -48,7 +45,14 @@ const convertVariant = async (file) => {
...
         return Promise.all([
-            fs.appendFile(`./generated/${iconName}/index.ts`, `export * from './${variantName}';\n`),
+            fs.appendFile(
+                './generated/index.ts',
+                `export const ${componentName} = loadable(() => import('./${iconName}/${variantName}'));\n`
+            ),
+            fs.appendFile(
+                `./generated/${iconName}/index.ts`,
+                `export {default as ${componentName}} from './${variantName}';\n`
+            ),
```
gdostie, line 55: I this really needed?

gdostie (commented):
This is brilliant 🎉

packages/react-icons/bin/index.js
GermainBergeron: It was just to avoid breaking the existing imports if ever they were used

aiheon:
is it me or this seems to make the plasma website load faster also? It's feels more snappy :eyes:

aiheon (approved):

gdostie (approved):

PR #2663, opened 2022-04-21
fix(plasma-react): value prop will no longer display syntax error

mikedidomizio:
### Proposed Changes

Fixes [ADUI-7921](https://coveord.atlassian.net/browse/ADUI-7921) and accidentally fixed [UITOOL-646](https://coveord.atlassian.net/browse/UITOOL-646)

- (refactor) JSONEditorConnected converted to Function Component, doing away with `mapDispatch`/`mapState`.  I did this change as part of it as it was a simple component and I find it easier to read.
- ADUI-7921 value prop will not be overridden by Redux - the `mapState` had a `value`, because the initial value was `''` from the selector, it was evaluating on an empty string before updating again from the selector, but whether to show the error was not fired, so it would show it as invalid.
- UITOOL-646 Fixes the Plasma website where the code editor wasn't synced with what was displayed in the JSON editor.
- I moved the default `{}` from CodeEditor the JSONEditor, and made the default value an empty string.  To me it seemed more appropriate than a code editor having JSON as default code.

Some of these changes are not really necessary or a requirement of the JIRAs, I would be happy to revert any of them.

You can test that `value` was the culprit by changing `defaultValue` to `value` on the Plasma website and passing in an empty object
image

- Please verify that both value and defaultValue continue to work as expected.

### Acceptance Criteria

### DIFF
--- a/packages/react/src/components/editor/JSONEditor.tsx
+++ b/packages/react/src/components/editor/JSONEditor.tsx
@@ -1,7 +1,7 @@
 import classNames from 'classnames';
-import {FunctionComponent, useState, useEffect} from 'react';
+import {FunctionComponent, useEffect, useState} from 'react';
 
-import {Svg} from '../svg/Svg';
+import {Svg} from '../svg';
 import {CodeEditor} from './CodeEditor';
 import {CodeMirrorModes, DEFAULT_JSON_ERROR_MESSAGE} from './EditorConstants';
 import {JSONEditorUtils} from './JSONEditorUtils';
@@ -53,7 +53,10 @@ export interface JSONEditorProps {
 }
 
 export interface JSONEditorStateProps {
-    value: string;
+    /**
+     * @deprecated use defaultValue instead
+     */
+    value?: string;
 }
 
 export interface JSONEditorDispatchProps {
@@ -77,7 +80,8 @@ export const JSONEditor: FunctionComponent<
     onUnmount,
     collapsibleId,
 }) => {
-    const [isInError, setIsInError] = useState(!JSONEditorUtils.validateValue(value || defaultValue));
+    const editorValue = defaultValue || value || '{}';
+    const [isInError, setIsInError] = useState(!JSONEditorUtils.validateValue(editorValue));
 
     useEffect(() => {
         onMount?.();
@@ -85,9 +89,13 @@ export const JSONEditor: FunctionComponent<
         return onUnmount;
     }, []);
 
+    useEffect(() => {
+        const hasError = !JSONEditorUtils.validateValue(editorValue);
+        setIsInError(hasError);
+    }, [editorValue]);
+
     const handleChange = (json: string) => {
         const hasError = !JSONEditorUtils.validateValue(json);
-
         setIsInError(hasError);
         onChange?.(json, hasError);
     };
@@ -95,7 +103,7 @@ export const JSONEditor: FunctionComponent<
     return (
         <div className={classNames('form-group', {'input-validation-error': isInError}, containerClasses)}>
             <CodeEditor
-                value={value || defaultValue}
+                value={editorValue}
                 onChange={handleChange}
                 mode={CodeMirrorModes.JSON}
                 readOnly={readOnly}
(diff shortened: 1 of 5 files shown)


packages/react/src/components/editor/JSONEditorConnected.tsx
```diff
@@ -1,22 +1,27 @@
...
+            dispatch(JSONEditorActions.removeJSONEditor(props.id));
+        };
+    }, []);
+
+    const onChange = (value: string, inError: boolean) => {
+        dispatch(JSONEditorActions.updateJSONEditorValue(props.id, value));
+        props.onChange?.(value, inError);
+    };
+
+    return <JSONEditor {...props} defaultValue={defaultValue} onChange={onChange} />;
```
GermainBergeron, line 26: I actually think we should continue to send `value` here and make `defaultValue` a prop of `JSONEditorConnected`, thus removing it from JSONEditor. Since we try to avoid having too many breaking changes, you could mark it deprecated and we'll remove it in the next major version

It would be clearer that the defaultValue is only used for the first render and the `JSONEditor` would only have the value to care about.

packages/react/src/components/editor/JSONEditorConnected.tsx
```diff
@@ -1,22 +1,27 @@
...
-import {PlasmaState} from '../../PlasmaState';
 import {IDispatch} from '../../utils';
 import {JSONEditor, JSONEditorDispatchProps, JSONEditorProps, JSONEditorStateProps} from './JSONEditor';
 import {JSONEditorActions} from './JSONEditorActions';
 import {JSONEditorSelectors} from './JSONEditorSelectors';
 
-const mapStateToProps = (state: PlasmaState, ownProps: JSONEditorProps): JSONEditorStateProps => ({
-    value: JSONEditorSelectors.getValue(state, ownProps.id),
-});
+export const JSONEditorConnected = (props: JSONEditorStateProps & JSONEditorDispatchProps & JSONEditorProps) => {
```
GermainBergeron, line 9: I don't think you need `JSONEditorStateProps & JSONEditorDispatchProps`, previously those weren't exposed by the connected component

```suggestion
export const JSONEditorConnected = (props: JSONEditorProps) => {
```

GermainBergeron, line 9: I don't think you need `JSONEditorStateProps & JSONEditorDispatchProps`, previously those weren't exposed by the connected component

```suggestion
export const JSONEditorConnected = (props: JSONEditorProps) => {
```

packages/react/src/components/editor/JSONEditor.tsx
```diff
@@ -53,7 +53,10 @@ export interface JSONEditorProps {
 }
 
 export interface JSONEditorStateProps {
-    value: string;
+    /**
+     * @deprecated use defaultValue instead
+     */
+    value?: string;
```
GermainBergeron, line 59: The value is already optional and deprecated in `JSONEditorProps`. If you apply my change on the JSONEditorConnected function I don't think you need this change

```suggestion
    value: string;
```

GermainBergeron, line 59: The value is already optional and deprecated in `JSONEditorProps`. If you apply my change on the JSONEditorConnected function I don't think you need this change

```suggestion
    value: string;
```

packages/react/src/components/editor/JSONEditor.tsx
```diff
@@ -77,25 +80,30 @@ export const JSONEditor: React.FunctionComponent<
...
+        const hasError = !JSONEditorUtils.validateValue(editorValue);
+        setIsInError(hasError);
+    }, [editorValue]);
+
     const handleChange = (json: string) => {
         const hasError = !JSONEditorUtils.validateValue(json);
-
         setIsInError(hasError);
         onChange?.(json, hasError);
     };
```
toofff, line 101: You do almost the same thing in the useEffect and the handleChange method, there is no way to factor this part?

toofff (approved):
After to have fixed the comments of @GermainBergeron , it's ok for me

toofff (approved):

GermainBergeron (approved):
Code looks good and demo works well!

PR #2670, opened 2022-04-26
Update react-dnd version

gdostie:
### Proposed Changes

Previously at major version 2, react-dnd dependency was long overdue for an update. I had to refactor our 2 components that used it so that they use their new hooks instead of the old decorators pattern, but now we use version 15.

Also, because of one single require function call, nextjs considered the whole website project to use commonjs. So it was using `plasma-react/dist/cjs` distribution instead of `plasma-react/dist/esm`

DnDUtils has been skimmed down to just one reorder function

### Acceptance Criteria


packages/react/src/components/dragAndDrop/DnDContainer.tsx
```diff
@@ -1,72 +1,74 @@
...
+    draggableIconProps = {},
+    icon = <Svg svgName={svg.dragDrop.name} svgClass="icon" />,
+    children,
+    isDraggable = true,
+    onMoveOver,
+    id,
+}) => {
+    const ref = React.useRef<HTMLDivElement>();
+    const iconRef = React.useRef<HTMLDivElement>();
+    const [{}, drop] = useDrop(() => ({
```
GermainBergeron, line 36:
```suggestion
    const [, drop] = useDrop(() => ({
```

GermainBergeron (approved):
Nice cleanup! Demo works well!

It seems really easier to use than before

mikedidomizio (approved):

PR #2727, opened 2022-05-09
fix(checkbox): fix checkbox disabled tooltip

nathanlb:
### Proposed Changes

Issue we're trying to fix:

Screen Shot 2022-05-09 at 12 00 47 PM

- Added prop to Input to set classes on the Tooltip`<span></span>` element as putting it in between the container and the content would break the styling.
- Modified css rule to account for possible tooltip being inserted.
ex:
  - checkbox container has a flex display
  - span wrapping the content breaks the arrangement and clashes with some css settings which function by specifying a descendant tree.

Screen Shot 2022-05-09 at 1 41 12 PM
Screen Shot 2022-05-09 at 1 40 57 PM
Screen Shot 2022-05-09 at 1 40 50 PM

After change
Screen Shot 2022-05-09 at 1 46 15 PM

### Acceptance Criteria

### DIFF
--- a/packages/react/src/components/checkbox/Checkbox.tsx
+++ b/packages/react/src/components/checkbox/Checkbox.tsx
@@ -66,7 +66,11 @@ export class Checkbox extends Component<ICheckboxProps> {
             {disabled: !!this.props.disabled, 'checkbox-clear': this.props.clearSides},
             this.props.classes
         );
-        const innerInputClasses: string = classNames('react-vapor-checkbox', this.props.innerInputClasses);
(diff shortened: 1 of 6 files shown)


toofff (approved):
LGTM

gdostie (commented):
It seems to break the style for the input itself

On master
image

On your branch
image

nathanlb:
> 

Darn... I'll look for another way then. Thanks for checking thoroughly

gdostie (approved):
GG

PR #2731, opened 2022-05-10
feat(iconbadge): removed svg and using new icons

aiheon:
### Proposed Changes

https://coveord.atlassian.net/browse/UITOOL-675

Before: 

image

After:

image

`svgClasses` is now deprecated and using it will have no effect on the style of the icon.

### Acceptance Criteria

### DIFF
--- a/packages/react/src/components/iconBadge/IconBadge.tsx
+++ b/packages/react/src/components/iconBadge/IconBadge.tsx
@@ -1,8 +1,6 @@
-import {SvgName} from '@coveord/plasma-style';
+import {Icon} from '@coveord/plasma-react-icons';
 import classNames from 'classnames';
 import {FunctionComponent} from 'react';
-
-import {Svg} from '../svg';
 
 export enum IconBadgeSize {
     Medium,
@@ -19,7 +17,7 @@ export interface IconBadgeProps {
     /**
      * Icon to display
      */
-    svgName: SvgName;
+    icon: Icon;
     /**
      * Type of the icon badge (New - Information - Warning - Major)
      */
@@ -31,7 +29,7 @@ export interface IconBadgeProps {
      */
     size?: IconBadgeSize;
     /**
-     * Additionnal CSS class for the icon
+     * @deprecated will have no effet
      */
     svgClass?: string;
     /**
@@ -51,16 +49,12 @@ const TypeColorMapping: Record<IconBadgeType, string> = {
     [IconBadgeType.Major]: 'mod-major',
 };
 
-export const IconBadge: FunctionComponent<IconBadgeProps> = ({
-    svgName,
-    type,
-    size = IconBadgeSize.Medium,
-    svgClass,
-    className,
-}) => (
-    <Svg
-        className={classNames('icon-badge', SizeClassMapping[size], TypeColorMapping[type], className)}
-        svgName={svgName}
-        svgClass={classNames('icon align-middle', SizeClassMapping[size], svgClass)}
-    />
-);
+export const IconBadge: FunctionComponent<IconBadgeProps> = ({icon, type, size = IconBadgeSize.Medium, className}) => {
+    const IconName = icon;
+
+    return (
+        <div className={classNames('icon-badge', SizeClassMapping[size], TypeColorMapping[type], className)}>
+            <IconName height={24} className={classNames('align-middle')} />
+        </div>
+    );
+};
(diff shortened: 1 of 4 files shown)


packages/react/src/components/iconBadge/IconBadge.tsx
```diff
@@ -31,7 +29,7 @@ export interface IconBadgeProps {
      */
     size?: IconBadgeSize;
     /**
-     * Additionnal CSS class for the icon
+     * @deprecated will have no effet
      */
     svgClass?: string;
```
GermainBergeron, line 34: Just remove it, you're already making a breaking change by changing `svgName` to `icon`

packages/react/src/components/iconBadge/IconBadge.tsx
```diff
@@ -51,16 +49,12 @@ const TypeColorMapping: Record<IconBadgeType, string> = {
...
-    />
-);
+export const IconBadge: FunctionComponent<IconBadgeProps> = ({icon, type, size = IconBadgeSize.Medium, className}) => {
+    const IconName = icon;
+
+    return (
+        <div className={classNames('icon-badge', SizeClassMapping[size], TypeColorMapping[type], className)}>
+            <IconName height={24} className={classNames('align-middle')} />
+        </div>
+    );
```
GermainBergeron, line 59: I'm not sure I see the value of changing `icon` to `IconName`
```suggestion
    return (
        <div className={classNames('icon-badge', SizeClassMapping[size], TypeColorMapping[type], className)}>
            <icon height={24} className={classNames('align-middle')} />
        </div>
    );
```

GermainBergeron (commented):
GG!

packages/react/src/components/iconBadge/IconBadge.tsx
gdostie: Typescript complains if the name of the component is not starting with a capital letter

packages/react/src/components/iconBadge/IconBadge.tsx
```diff
@@ -51,16 +49,12 @@ const TypeColorMapping: Record<IconBadgeType, string> = {
...
-        className={classNames('icon-badge', SizeClassMapping[size], TypeColorMapping[type], className)}
-        svgName={svgName}
-        svgClass={classNames('icon align-middle', SizeClassMapping[size], svgClass)}
-    />
-);
+export const IconBadge: FunctionComponent<IconBadgeProps> = ({icon, type, size = IconBadgeSize.Medium, className}) => {
+    const IconName = icon;
+
+    return (
+        <div className={classNames('icon-badge', SizeClassMapping[size], TypeColorMapping[type], className)}>
```
gdostie, line 56: The size mapping should be applied in the `height` prop of the icon I think. Adding a mod-24 to this div has no effect.

aiheon: mod-24 is for the dot, not the icon actually ^^
I was also wondering if we really need to have this flexibility on the size? I though those were getting deprecated soon? 🤔

packages/react/src/components/iconBadge/IconBadge.tsx
```diff
@@ -51,16 +45,12 @@ const TypeColorMapping: Record<IconBadgeType, string> = {
...
+            <IconName height={24} className={'align-middle'} />
```
gdostie, line 53: The examples use the 16 px size variant of the icons but here we set 24px

aiheon: I can use the 24px variant if you prefer 🤷‍♀️

gdostie: I think that would be more appropriate yes, otherwise the strokes look more thick then they should be compared to the other icons

aiheon: You got it then 😄

packages/react/src/components/iconBadge/tests/IconBadge.spec.tsx
```diff
@@ -1,38 +1,46 @@
 import {render, screen} from '@test-utils';
-import {svg} from '@coveord/plasma-style';
 
+import {BellSize16Px} from '@coveord/plasma-react-icons';
 import {IconBadge, IconBadgeType} from '../IconBadge';
 
 describe('IconBadge', () => {
-    it('renders the specified icon with the type New, and size Medium by default', () => {
-        render(<IconBadge svgName={svg.bellStrokedMedium.name} type={IconBadgeType.New} />);
+    it('renders the specified icon with the type New, and size Medium by default', async () => {
```
kdang-coveo, line 7: curious about this change, how come async is required for the tests here? 🤔

GermainBergeron (approved):

PR #2736, opened 2022-05-10
fix(website): performance issue with props table

GermainBergeron:
### Proposed Changes

Create the Typescript system and environment only once and re-use it in all pages
Stop modifying the map of file, instead write the file in the environment
Remove the useTypescriptServer from the Sandbox, it wasn't useful

https://user-images.githubusercontent.com/260007/167716604-172d9600-7383-4c50-8940-16561a7ad55c.mov

None, only affects the website

### Acceptance Criteria

### DIFF
--- a/packages/website/src/Store.ts
+++ b/packages/website/src/Store.ts
@@ -12,8 +12,3 @@ export const Store = createStore(
     combineReducers(PlasmaReducers),
     composeEnhancers(applyMiddleware<IDispatch>(...middlewares))
 );
-
-if (typeof window !== 'undefined') {
-    // Required for the dynamic examples
-    (window as any).Store = Store;
-}
--- a/packages/website/src/building-blocs/PropsDoc.tsx
+++ b/packages/website/src/building-blocs/PropsDoc.tsx
@@ -1,4 +1,3 @@
-import {createSystem, createVirtualTypeScriptEnvironment} from '@typescript/vfs';
 import {FunctionComponent, useState, useEffect} from 'react';
 import * as ts from 'typescript';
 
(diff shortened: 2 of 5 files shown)


gdostie (approved):
Good job, works realllly well 🚀

toofff:
When you go through the search bar then you arrive on a page, you always have this problem of slowness I have the impression.
When you change via the navigation on the left, everything is ok

GermainBergeron:
> When you go through the search bar then you arrive on a page, you always have this problem of slowness I have the impression. When you change via the navigation on the left, everything is ok

That's really weird. I hadn't realized the fix wasn't working for the search page 🤔 I check if I can get it to apply there too

GermainBergeron:
After digging a lot in the code it seems that when we do a search we're being redirected to `/Search` which does a full page reload (instead of relying on Next navigation), thus creating loading all the types and creating new typescript instance.

1. If you do a search,
2. click on a component,
3. wait for the props to load,
4. press back to go back to the result list and click on a different (or the same) component 
5. you'll see that the props load instantly.

The issue really is with the SearchBar

aiheon:
> After digging a lot in the code it seems that when we do a search we're being redirected to `/Search` which does a full page reload (instead of relying on Next navigation), thus creating loading all the types and creating new typescript instance.
> 
> 1. If you do a search,
> 2. click on a component,
> 3. wait for the props to load,
> 4. press back to go back to the result list and click on a different (or the same) component
> 5. you'll see that the props load instantly.
> 
> The issue really is with the SearchBar

Yeah, the way Atomic is wired, navigation with the next router was not an option IIRC 🤔  
That being said, we are already a few releases late and there's clean up to do with it, it's not super clean atm 🙈

aiheon (approved):
😻 GG

PR #2738, opened 2022-05-11
Date range picker behaviour improvements

gdostie:
### Proposed Changes

Auto focusing on the start date after the end date has been selected improves the usability of the component.

https://user-images.githubusercontent.com/35579930/167891275-dfa65619-47e3-45ab-81dc-21fce221bd06.mov

### Acceptance Criteria

### DIFF
--- a/packages/website/src/Store.ts
+++ b/packages/website/src/Store.ts
@@ -26,7 +26,7 @@ const loadFromSessionStorage = () => {
 };
 
 const persistedState =
-    process.env.NODE_ENV === 'development' && (globalThis as any).loaded ? loadFromSessionStorage() : undefined;
+    process.env.NODE_ENV === 'development' && (globalThis as any)?.loaded ? loadFromSessionStorage() : undefined;
 
 export const Store = createStore(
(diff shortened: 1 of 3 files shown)


packages/website/src/Store.ts
```diff
@@ -26,7 +26,7 @@ const loadFromSessionStorage = () => {
 };
 
 const persistedState =
-    process.env.NODE_ENV === 'development' && (window as any).loaded ? loadFromSessionStorage() : undefined;
+    process.env.NODE_ENV === 'development' && (globalThis as any)?.loaded ? loadFromSessionStorage() : undefined;
```
gdostie, line 29: window is not defined on the first load because Nextjs is pre-rendering in a non-browser environment (node). However `globalThis` which later on points to `window` is defined both in node and in browser, so it fixes the issue.

gdostie: Without this fix:
image

aiheon (approved):

GermainBergeron (approved):

PR #2740, opened 2022-05-11
Remove Svg from a bunch of components

gdostie:
#### Filepicker

image
image

#### Filterbox

image
image
image
image

#### LabeledValue

image
image

#### Limit

image
image

#### Modal

image
image

#### Toast 

image
image

#### TableCollapsibleRowToggle

image
image

### Acceptance Criteria


packages/react/src/components/filepicker/tests/FilepickerSelectors.spec.ts
```diff
@@ -3,7 +3,7 @@ import {FilepickerSelectors} from '../FilepickerSelectors';
 
 describe('FilepickerSelectors', () => {
     describe('getFile', () => {
-        // This selector cannot be tested, because for security reasons the browser won't allow to select a file programmatically.
+        // tested via the FilePicker component tests
```
toofff, line 6: The 2 messages give relevant information, couldn't we merge the 2 into one?

gdostie: Well the first statement is made false by the second one 🤔  It CAN be tested, because it is tested in the component tests

toofff: Ignoring your change, can this selector now be tested here? or do we still have the same problem?

If we still have the same problem, then having the information from the first comment is relevant, otherwise we don't care and we keep your change. ;)

toofff (commented):
There is a color difference between the 2 crosses for its 2 components. is this normal?
There is also a difference in font size for its 2 components, is this normal?

image

image

packages/react/src/components/filepicker/tests/FilepickerSelectors.spec.ts
aiheon: I think it's not relevant to test selectors on itself, they don't mean anything to a user.
I'm fine that it's tested ''pas la bande'' by the component 👍

packages/style/scss/components/modal.scss
```diff
@@ -1,6 +1,6 @@
 @use "sass:math";
```
aiheon, line 1: i think this can be removed 👀 
```suggestion

```

aiheon (approved):

gdostie:
> There is a color difference between the 2 crosses for its 2 components. is this normal? There is also a difference in font size for its 2 components, is this normal?
> 
> image
> 
> image

It is far from ideal, both components haven't been restyled yet but we will get to them eventually. When we get there we will make sure that they visually fit the rest of the components. The icons take the color of the text next to it, so changing the text color would mean restyling the entire component. For that reason I will leave them be like that for now, they we already different colors anyway so they can stay like that just a little longer.

PR #2745, opened 2022-05-13
UITOOL-677/extract new spinner

aiheon:
### Proposed Changes

https://coveord.atlassian.net/browse/UITOOL-677

The new loading spinner is super nice and maybe we want to be able to use it somewhere else easily 👀  🌻  

I'm not sure about the name though 🙈 

https://user-images.githubusercontent.com/63734941/168388701-192fb98b-6302-4d1d-bd14-efc909af56bc.mov

https://user-images.githubusercontent.com/63734941/168389083-3213563c-cd8f-47b1-b901-bbbbff6a8c30.mov

None

### Acceptance Criteria

### DIFF
--- a/packages/react/src/components/loading/LoadingCircle.tsx
+++ b/packages/react/src/components/loading/LoadingCircle.tsx
@@ -1,0 +1,3 @@
+import {FunctionComponent} from 'react';
+
+export const LoadingCircle: FunctionComponent<any> = () => <div className={'loading-circle'}></div>;
--- a/packages/react/src/components/loading/LoadingCircle.tsx
+++ b/packages/react/src/components/loading/LoadingCircle.tsx
@@ -1,3 +1,0 @@
-import {FunctionComponent} from 'react';
-
-export const LoadingCircle: FunctionComponent<any> = () => <div className={'loading-circle'}></div>;
--- a/packages/react/src/components/loading/LoadingSpinner.tsx
+++ b/packages/react/src/components/loading/LoadingSpinner.tsx
@@ -1,0 +1,5 @@
+import {FunctionComponent} from 'react';
+
+export const LoadingSpinner: FunctionComponent<any> = () => (
+    <div role="alert" aria-busy="true" className={'loading-spinner'}></div>
+);
--- a/packages/react/src/components/loading/LoadingSpinner.tsx
+++ b/packages/react/src/components/loading/LoadingSpinner.tsx
@@ -1,5 +1,5 @@
 import {FunctionComponent} from 'react';
 
-export const LoadingSpinner: FunctionComponent<any> = () => (
-    <div role="alert" aria-busy="true" className={'loading-spinner'}></div>
+export const LoadingSpinner: FunctionComponent = () => (
+    <div role="alert" aria-busy="true" className="loading-spinner" />
 );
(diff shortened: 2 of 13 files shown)


packages/react/src/components/loading/LoadingCircle.tsx
```diff
@@ -0,0 +1,3 @@
+import {FunctionComponent} from 'react';
+
+export const LoadingCircle: FunctionComponent<any> = () => <div className={'loading-circle'}></div>;
```
GermainBergeron, line 3: We should add an aria role, this way we won't need to rely on the class name in the tests

GermainBergeron (commented):
Looks good!

I would change the component name and add an accessible role

packages/react/src/components/loading/LoadingSpinner.tsx
```diff
@@ -0,0 +1,5 @@
+import {FunctionComponent} from 'react';
+
+export const LoadingSpinner: FunctionComponent<any> = () => (
+    <div role="alert" aria-busy="true" className={'loading-spinner'}></div>
```
GermainBergeron, line 4:
```suggestion
export const LoadingSpinner: FunctionComponent = () => (
    <div role="alert" aria-busy="true" className="loading-spinner" />
```

packages/react/src/components/searchBar/tests/SearchBar.spec.tsx
```diff
@@ -1,124 +1,106 @@
...
-    it('should have a container div with search-bar-loading class when searching is passed as prop', () => {
-        const containerDiv = shallow(<SearchBar {...requiredProps} searching />)
-            .find('div')
-            .first();
+    it('renders a loading animation when is searching', () => {
+        const {container} = render(<SearchBar {...requiredProps} searching />);
 
-        expect(containerDiv.hasClass('search-bar-loading')).toBe(true);
+        // eslint-disable-next-line testing-library/no-container
+        expect(container.querySelector('.loading-spinner')).toBeInTheDocument();
```
GermainBergeron, line 27: Here you could find by role with the new role instead

GermainBergeron (approved):

packages/react/src/components/searchBar/tests/SearchBar.spec.tsx
aiheon: :derp:

packages/react/src/components/searchBar/tests/SearchBar.spec.tsx
```diff
@@ -1,124 +1,105 @@
...
             .first();
 
-        expect(containerDiv.hasClass('extra-class')).toBe(true);
+        expect(containerDiv.prop('className')).toContain('extra-class');
     });
 
-    it('should have an input inside the div container with the search-bar-input class by default', () => {
-        const component = shallow(<SearchBar {...requiredProps} />);
+    it('render the input of the search bar with extra classes', () => {
+        const component = shallow(<SearchBar {...requiredProps} inputClassNames="extra-class" />)
```
sallainCoveo, line 45: I might be wrong, but I think you could use RTL instead of Enzyme for this test (and the one prior). I don't know if it would work, but maybe switch to this?
```suggestion
        render(<SearchBar {...requiredProps} inputClassNames="extra-class" />)
        
        expect(screen.getByRole('textbox').toHaveClass('extra-class');
```

Is there some instances we should use Enzyme?

sallainCoveo (approved):

PR #2747, opened 2022-05-17
feat(subnav): add a description in subnav

kdang-coveo:
### Proposed Changes

- Adjust the default padding of the existing sub nav `7px` -> `16px` as per Gustavo's request
- Add an option to add `description` in the SubNavigation component using a slider to appear when selected
- When a user utilizes the `description` in the navItems, the component will style the labels if it is a `string`, otherwise it respects the `ReactNode` rule the user puts on the tab
- UX padding and specification has been tweaked with @guscoveo 

This is the initial change needed for the upcoming UX updates.

- None - as the `description` prop is completely optional

https://user-images.githubusercontent.com/66333175/168903927-a06f544f-3552-4217-a1d9-0b70d263e132.mp4

Styling of all the tabs change automatically when a description is detected

https://user-images.githubusercontent.com/66333175/168904064-e006de04-2918-4977-be41-e12d0e6c52b7.mp4

### Acceptance Criteria

### DIFF
--- a/packages/react/src/components/subNavigation/SubNavigation.tsx
+++ b/packages/react/src/components/subNavigation/SubNavigation.tsx
@@ -88,7 +88,7 @@ export const SubNavigation: FunctionComponent<ISubNavigationProps & HTMLAttribut
                 </a>
                 {description && (
                     <SlideY in={openDescription}>
-                        <div className={classNames('sub-navigation-item-description')}>{description}</div>
+                        <div className="sub-navigation-item-description body-m-book-subdued">{description}</div>
                     </SlideY>
                 )}
             </li>
--- a/packages/react/src/components/subNavigation/SubNavigation.tsx
+++ b/packages/react/src/components/subNavigation/SubNavigation.tsx
@@ -93,9 +93,7 @@ export const SubNavigation: FunctionComponent<ISubNavigationProps & HTMLAttribut
                             ? createElement(
                                   'div',
                                   {
-                                      className: description
-                                          ? 'sub-navigation-item-link-with-description-label'
-                                          : 'sub-navigation-item-link-with-description-label-solo',
+                                      className: 'sub-navigation-item-link-with-description-label',
                                   },
                                   `${label}`
                               )
--- a/packages/react/src/components/subNavigation/SubNavigation.tsx
+++ b/packages/react/src/components/subNavigation/SubNavigation.tsx
@@ -83,7 +83,11 @@ export const SubNavigation: FunctionComponent<ISubNavigationProps & HTMLAttribut
             <li key={id} className={classNames('sub-navigation-item', {'mod-selected': id === selectedItem})}>
                 <a
                     href={link}
-                    className={classNames('sub-navigation-item-link', {disabled})}
+                    className={classNames(
+                        {'sub-navigation-item-link': !description},
+                        {'sub-navigation-item-link-with-description': description},
+                        {disabled}
+                    )}
                     onClick={(e: MouseEvent<HTMLAnchorElement>) => handleItemClick(e, id)}
                 >
                     {label}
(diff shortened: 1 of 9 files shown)


packages/website/src/pages/navigation/SubNavigation.tsx
```diff
@@ -83,15 +83,15 @@ const code = `
     export default () => {
         const exampleItems = [
             {label: 'Avatar', id: 'avatar'},
-            {label: 'Titanic', id: 'titanic'},
+            {label: <div className="body-m py1">Titanic</div>, id: 'titanic', description: 'Cupcake ipsum dolor sit amet. Lemon drops croissant sesame snaps cookie jelly beans tootsie roll muffin. Liquorice liquorice fruitcake tiramisu sesame snaps sugar plum lollipop gummi bears cookie.'},
```
toofff, line 86: `<div className="body-m py1">Titanic</div>`

This is the format the label should have when using a description? Is this a UX recommendation?

kdang-coveo: it is not a UX recommendation, but since the label allows for a `React.Node`, it was just a simple example since we did not have one

toofff: I understand what you want to do and it's not stupid. On the other hand, what I would like to see in his examples is also the UX recommendations.

Because if someone from the UX team tells me to add a description, I will obviously copy your example and I risk taking the format of this label thinking that this is what the UX team wants.

Maybe you could add an example with this possibility to add a ReactNode instead of a string? For exemple for the line with `The Avengers`, a different line of your example.

aiheon: (disclaimer I did not read all this thread)

I think it should not be possible to have one tab if a description and the other without? 🤔

aiheon: Also I think this would deserve a new example since this is a variant of the default pattern 😄

aiheon:
I'm not a fan of having with and without description in the same subnav 👀 

image

packages/website/src/pages/navigation/SubNavigation.tsx
aiheon: > > I'm not a fan of having with and without description in the same subnav 👀
> 
> I spoke to @guscoveo about this, I added some demo's to show what should be used when we want to have a description along with a non description.
> 
> I also adjusted the css logic to apply the correct padding as per the Figma mockups.
> 


That's better but then we should enforce this style otherwise it will be forgotten 🙈 

And you should split the example the default one and the withDescriptions one. Poke me on slack if you need help on how to do it 😄

kdang-coveo: Will do! Thanks 🙏

packages/style/scss/variables.scss
```diff
@@ -252,7 +252,7 @@ $application-container-min-height: 400px;
 // Sub Navigation
 $sub-navigation-padding-top: 26px;
 $sub-navigation-border: $default-border;
-$sub-navigation-item-padding: 7px 40px;
+$sub-navigation-item-padding: 16px 40px;
```
kdang-coveo, line 255: Directed change by @guscoveo

guscoveo (approved):

packages/react/src/components/subNavigation/SubNavigation.tsx
```diff
@@ -71,22 +77,49 @@ export const SubNavigation: FunctionComponent<ISubNavigationProps & HTMLAttribut
...
+                    <a
+                        href={link}
+                        className={classNames('sub-navigation-item-link', {disabled})}
+                        onClick={(e: MouseEvent<HTMLAnchorElement>) => handleItemClick(e, id)}
+                    >
+                        {label}
+                    </a>
+                </li>
+            );
+        }
```
toofff, line 121: The `else` is not needed in your situation

```suggestion
        }
            return (
                <li key={id} className={classNames('sub-navigation-item', {'mod-selected': id === selectedItem})}>
                    <a
                        href={link}
                        className={classNames('sub-navigation-item-link', {disabled})}
                        onClick={(e: MouseEvent<HTMLAnchorElement>) => handleItemClick(e, id)}
                    >
                        {label}
                    </a>
                </li>
            );
```

toofff (approved):

aiheon (approved):

PR #2751, opened 2022-05-18
UITOOL-692/mvp filter box

aiheon:
### Proposed Changes

https://coveord.atlassian.net/browse/UITOOL-692

New style for the Filter box component. Please note that the current icons are not the new ones (it's coming soon 👀 ).

Before: 
image
image
image

After: 
image
image
image

New variant: Disabled

image

Should be none!

### Acceptance Criteria


aiheon:
Tests about the icons switching between the Filter and the Cross have been removed. I will put them back when merging this pr in Next (where the Svg component don't exist anymore) with the new icons.

Svg component in currently setting the aria-hidden on the span wrapping the icon instead of on the icon in itself. `getByRole('img, {name: 'filter)` always return false because technically it's the span that is hidden 😕 

I have hopes that the react Icons won't have that issue

packages/react/src/components/filterBox/FilterBox.tsx
```diff
@@ -38,6 +38,12 @@ export interface IFilterBoxOwnProps extends ClassAttributes<FilterBox> {
      * @default false
      */
     isAutoFocus?: boolean;
+    /**
+     * Wheter to disabled the filter box
```
GermainBergeron, line 42:
```suggestion
     * Whether to disable the filter box
```

packages/react/src/components/filterBox/FilterBox.tsx
```diff
@@ -82,12 +88,25 @@ export class FilterBox extends Component<IFilterBoxProps, any> {
...
+    private handleIcons = (inputValue: string) => {
+        if (inputValue.length !== 0) {
+            this.filterInput.nextElementSibling.setAttribute('class', ' flex filter-box-icon hidden'); // hide filter icon
+            this.filterInput.nextElementSibling.nextElementSibling.classList.remove('hidden'); // show clear icon
+        } else {
+            this.filterInput.nextElementSibling.classList.remove('hidden'); // show filter icon
+            this.filterInput.nextElementSibling.nextElementSibling.setAttribute(
+                'class',
+                ' flex filter-box-icon hidden'
+            ); // hide clear icon
```
GermainBergeron, line 103: Instead of relying on nextSibling would it be possible to use refs?

packages/react/src/components/filterBox/FilterBox.tsx
```diff
@@ -82,12 +88,25 @@ export class FilterBox extends Component<IFilterBoxProps, any> {
 
     static defaultProps: Partial<IFilterBoxProps> = {
         isAutoFocus: false,
+        disabled: false,
+    };
+
+    private handleIcons = (inputValue: string) => {
+        if (inputValue.length !== 0) {
+            this.filterInput.nextElementSibling.setAttribute('class', ' flex filter-box-icon hidden'); // hide filter icon
+            this.filterInput.nextElementSibling.nextElementSibling.classList.remove('hidden'); // show clear icon
```
GermainBergeron, line 97: Can you add in the classList like you do for the remove?
```suggestion
            this.filterInput.nextElementSibling.classList.add('hidden'); // hide filter icon
            this.filterInput.nextElementSibling.nextElementSibling.classList.remove('hidden'); // show clear icon
```

packages/react/src/components/filterBox/FilterBox.tsx
GermainBergeron, line 103:
```suggestion
            this.filterInput.nextElementSibling.nextElementSibling.classList.add('hidden'); // hide clear icon
```

GermainBergeron:
> Tests about the icons switching between the Filter and the Cross have been removed. I will put them back when merging this pr in Next (where the Svg component don't exist anymore) with the new icons.
> 
> Svg component in currently setting the aria-hidden on the span wrapping the icon instead of on the icon in itself. `getByRole('img, {name: 'filter)` always return false because technically it's the span that is hidden 😕
> 
> I have hopes that the react Icons won't have that issue

I think you can do `getByRole('img, {name: 'filter, hidden: true)` to get hidden elements. But that makes me think, since you're only changing the hidden class, won't the aria-hidden never update?

GermainBergeron (commented):
The icons look really big to me, can we shrink them a bit?

packages/react/src/components/filterBox/FilterBox.tsx
aiheon: Probably but since the svg is getting kicked fairly soon, I would prefer not to spend too much time on the icons but I take note for when I merge this in next! 👍

aiheon:
> > Tests about the icons switching between the Filter and the Cross have been removed. I will put them back when merging this pr in Next (where the Svg component don't exist anymore) with the new icons.
> > Svg component in currently setting the aria-hidden on the span wrapping the icon instead of on the icon in itself. `getByRole('img, {name: 'filter)` always return false because technically it's the span that is hidden 😕
> > I have hopes that the react Icons won't have that issue
> 
> I think you can do `getByRole('img, {name: 'filter, hidden: true)` to get hidden elements. But that makes me think, since you're only changing the hidden class, won't the aria-hidden never update?

It's working very well actually! It's the part I found confusing with the uts giving a hard time 👀  But i'll try your suggestion right now 👍 

https://user-images.githubusercontent.com/63734941/170105903-7e264ad8-61d3-4739-b67e-02b81b930d2b.mov

aiheon:
> > > Tests about the icons switching between the Filter and the Cross have been removed. I will put them back when merging this pr in Next (where the Svg component don't exist anymore) with the new icons.
> > > Svg component in currently setting the aria-hidden on the span wrapping the icon instead of on the icon in itself. `getByRole('img, {name: 'filter)` always return false because technically it's the span that is hidden 😕
> > > I have hopes that the react Icons won't have that issue
> > 
> > 
> > I think you can do `getByRole('img, {name: 'filter, hidden: true)` to get hidden elements. But that makes me think, since you're only changing the hidden class, won't the aria-hidden never update?
> 
> It's working very well actually! It's the part I found confusing with the uts giving a hard time 👀 But i'll try your suggestion right now 👍
> 
>  Screen.Recording.2022-05-24.at.2.23.09.PM.mov

Ok i'm dumb I did not negate my condition 😅 

`aria-hidden={this.filterInput?.value.length === 0}
`
vs 
`                        aria-hidden={!(this.filterInput?.value.length === 0)}
`

aiheon:
Tried again but the main problem I have is that aria hidden is on the span and getByRole 'img' is just pointing to the svg. :( 

I'll do the tests in Next with the react icons, it will be a lot easier I think 👍 

image

GermainBergeron (approved):

aiheon:
> Is it possible to preserve the `cursor: pointer` on the `X`?

Done!

PR #2753, opened 2022-05-18
feat: new DiffViewer Component

toofff:
### Proposed Changes

Implementation of the new DiffViewer component.

This component will just display the difference of a result of a `git diff` type command.
It will no longer do the calculation to generate the difference between two states, because this calculation can have a huge cost on the resources used by the browser. So the generation of this difference must be done on the API side.

image
image

- delete the old library used for this component there (react-diff-viewer & unidiff)
- the props used for this component are no longer the same, Yes there is a BC (I can put back the old props and depreciate them?)

### Acceptance Criteria


mikedidomizio (approved):

packages/style/scss/components/diff-viewer.scss
```diff
@@ -0,0 +1,29 @@
+.diff-viewer {
+    @import '~react-diff-view/style';
+
+    // configure the style of the library
+    --diff-background-color: var(--grey-10);
+    --diff-text-color: var(--grey-80);
+    --diff-font-family: $code-font-family;
```
gdostie, line 7: This doesn't seem to be applied

[`$code-font-family: source_code_pro_regular`](https://github.com/coveo/plasma/blob/master/packages/style/scss/variables.scss#L40)

Currently
image

With `'source_code_pro_regular'` applied

image

packages/react/src/components/diffViewer/DiffViewer.tsx
```diff
@@ -1,22 +1,55 @@
...
-    /**
-     * @deprecated Use newValue instead.
-     */
-    second: string;
+enum DIFF_VIEWER_VIEW_TYPE {
+    SPLIT = 'split',
+    UNIFIED = 'unified',
+}
+
+interface Hunk {
```
gdostie, line 11: Doesn't that cause a problem? We are defining an interface with the same name as an import 🤔

packages/react/src/components/diffViewer/DiffViewer.tsx
```diff
@@ -1,22 +1,55 @@
...
+
+interface FileProps {
+    hunks: Hunk[];
+    newRevision: string;
+    oldRevision: string;
+    type: 'add' | 'delete' | 'modify' | 'rename' | 'copy';
+    viewType: DIFF_VIEWER_VIEW_TYPE;
 }
 
+const File: FunctionComponent<FileProps> = ({oldRevision, newRevision, hunks, type, viewType}) => (
```
gdostie, line 37: The name `File` already exists in the global namespace, maybe we should give it a different name.

gdostie (commented):
The commit should be tagged as a breaking change.

packages/react/src/components/diffViewer/DiffViewer.tsx
toofff: No but it's not clean, I'm changing

packages/react/src/components/diffViewer/DiffViewer.tsx
```diff
@@ -1,22 +1,55 @@
...
+        isNormal?: boolean;
+        isInsert?: boolean;
+        isDelete?: boolean;
+        lineNumber?: number;
+        oldLineNumber?: number;
+        newLineNumber?: number;
+    }>;
+}
+
+interface DiifByFileProps {
```
gdostie, line 29:
```suggestion
interface DiffByFileProps {
```

packages/react/src/components/diffViewer/DiffViewer.tsx
```diff
@@ -1,22 +1,55 @@
...
+    hunks: HunkModel[];
+    newRevision: string;
+    oldRevision: string;
+    type: 'add' | 'delete' | 'modify' | 'rename' | 'copy';
+    viewType: DIFF_VIEWER_VIEW_TYPE;
 }
 
+const DiifByFile: FunctionComponent<DiifByFileProps> = ({oldRevision, newRevision, hunks, type, viewType}) => (
```
gdostie, line 37:
```suggestion
const DiffByFile: FunctionComponent<DiifByFileProps> = ({oldRevision, newRevision, hunks, type, viewType}) => (
```

gdostie (approved):
Good for be after there 2 little corrections

packages/react/src/components/diffViewer/DiffViewer.tsx
toofff: Thank you

PR #2756, opened 2022-05-18
docs(loadingspinner): added demo for the new loading spinner

aiheon:
### Proposed Changes

Adding a demo so dev can see it and choose to use it in their components if needed 👀 

Go in the LoadingSpinner page to see it 😛 

None, plasma website only

### Acceptance Criteria

### DIFF
--- a/packages/website/src/pages/feedback/Loading.tsx
+++ b/packages/website/src/pages/feedback/Loading.tsx
@@ -12,6 +12,12 @@ const fullContent = `
     export default () => <Loading fullContent />
 `;
 
+const loadingSpinner = `
+    import {LoadingSpinner} from '@coveord/plasma-react';
+
+    export default () => <LoadingSpinner />
+`;
+
 export default () => (
     <PageLayout
         id="Loading"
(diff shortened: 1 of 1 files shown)


aiheon:
@guscoveo if this is something UX would like to reused, it's now super easy to do for the devs 😄 
Just note that the size cannot be changed atm 🙈

kdang-coveo:
Looks awesome 🔥 🔥

packages/website/src/pages/feedback/Loading.tsx
```diff
@@ -12,6 +12,12 @@ const fullContent = `
     export default () => <Loading fullContent />
 `;
 
+const loadingSpinner = `
+    import {LoadingSpinner} from '@coveord/plasma-react';
+
+    export default () => <LoadingSpinner />
```
kdang-coveo, line 18:
```suggestion
    export default () => <div className="flex justify-center" ><LoadingSpinner /></div>
```

This should line it up, but I wonder if it would be worth it to imbed the `fullContent` prop to do this so that we can do `<loadingSpinner fullContent/>` to center

kdang-coveo (approved):

packages/website/src/pages/feedback/Loading.tsx
aiheon: I don't know where else this could be used but in it's current state, the component is really easy to position either in a new component or any place you want it really 🤔 

I also don't know if it will be ever used again BY UX outside of the Download toast (I've put it in the search bar without asking anyone 🙃 😇 )

GermainBergeron (approved):

PR #2761, opened 2022-05-20
UITOOL-700/mvp listbox

aiheon:
### Proposed Changes

https://coveord.atlassian.net/browse/UITOOL-700

The List Box component have it's new look!

Before:

image

image

After:

image

image

Empty state
image

With a footer
image

None

### Acceptance Criteria

### DIFF
--- a/packages/react/src/components/listBox/ListBox.tsx
+++ b/packages/react/src/components/listBox/ListBox.tsx
@@ -76,7 +76,7 @@ export class ListBox extends Component<IListBoxProps> {
     }
 
     private getClasses(): string {
-        return classNames('list-box relative', {'with-footer': this.props.footer}, this.props.classes);
+        return classNames('relative', this.props.classes);
     }
 
     protected getItems(): ReactNode {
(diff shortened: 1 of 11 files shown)


guscoveo (approved):
So nice!!

packages/react/src/components/listBox/ListBox.tsx
```diff
@@ -151,12 +151,12 @@ export class ListBox extends Component<IListBoxProps> {
     render() {
         const items = this.props.isLoading ? this.getLoadingItems() : this.getItems();
         return (
-            <>
+            <div className={classNames({'list-box-with-footer': this.props.footer})}>
```
GermainBergeron, line 154: Do you really need this div? You already have the with-footer class on the ul when you have a footer

aiheon: It's the way I found to change the bottoms corners from curved to flat to it's straight lines between the list and the footer 😅

GermainBergeron: In that case it would be the other way around, you should keep the div but remove the `with-footer` class on the ul since you target the element by using: `.list-box-with-footer .list-box` and style it accordingly

gdostie (commented):
IMHO loading items shouldn't have a hover effect since they cannot be selected
image

packages/style/scss/components/list-box.scss
```diff
@@ -65,9 +71,8 @@
         height: $item-box-divider-height;
         margin: $item-box-vertical-spacing 0;
         padding: 0;
-        background-color: var(--deprecated-medium-grey);
-
         cursor: default;
+        --background-color: var(--deprecated-medium-grey);
```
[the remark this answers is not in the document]
aiheon: I think what ressemble most is grey-50 so I'll use that 👍

packages/react/src/components/listBox/ListBox.tsx
aiheon: Done! Thanks for the tip! 😄

GermainBergeron (approved):

PR #2763, opened 2022-05-20
fix(react): make ActionableItem display better

gdostie:
The ActionableItem component looks a little weird, so I fixed that. Refactored its implementation
and redid all its tests while at it.

### Proposed Changes

Before
image

After
image

I removed the extend of HTMLAttributes of HTMLDivElement so if additional div props were passed down to the component, it might break. (Verified in the admin and it isn't the case)

### Acceptance Criteria

### DIFF
--- a/packages/react/src/components/actionable-item/ActionableItem.tsx
+++ b/packages/react/src/components/actionable-item/ActionableItem.tsx
@@ -25,11 +25,11 @@ export interface IActionableItemProps {
      */
     actions?: IItemBoxProps[];
     /**
-     * Additionnal CSS class to add on the outermost container
+     * Additional CSS class to add on the outermost container
      */
     className?: string;
     /**
-     * Additionnal CSS class to add on the main item
+     * Additional CSS class to add on the main item
      */
     containerClassName?: string;
     /**
--- a/packages/react/src/components/actionable-item/ActionableItem.tsx
+++ b/packages/react/src/components/actionable-item/ActionableItem.tsx
@@ -1,7 +1,6 @@
 import {DotsSize16Px} from '@coveord/plasma-react-icons';
 import classNames from 'classnames';
-import {Component, HTMLAttributes, MouseEvent as ReactMouseEvent} from 'react';
-import * as _ from 'underscore';
+import {FunctionComponent, MouseEvent as ReactMouseEvent} from 'react';
 
 import {DropPodPosition} from '../drop/DomPositionCalculator';
 import {Drop, IDropOwnProps} from '../drop/Drop';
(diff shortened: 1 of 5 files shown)


aiheon (approved):

packages/react/src/components/actionable-item/ActionableItem.tsx
```diff
@@ -35,44 +38,38 @@ export interface IActionableItemProps {
...
+    className,
+    containerClassName,
+    actions = [],
+    onItemClick,
+    dropProps,
+    children,
+}) => (
+    <div
+        className={classNames('actionable-item inline-flex mod-border mod-rounded-border-2 overflow-hidden', className)}
+    >
```
GermainBergeron, line 52: We previously allowed HTML attributes on the div, it would be a breaking change not to allow it anymore

packages/react/src/components/actionable-item/ActionableItem.tsx
```diff
@@ -35,44 +38,38 @@ export interface IActionableItemProps {
...
+        >
+            {children}
+        </div>
+        {actions && actions.length ? (
+            <Drop
+                id={id}
+                positions={[DropPodPosition.bottom, DropPodPosition.top]}
+                renderOpenButton={(onClick: () => void) => (
+                    <button onClick={onClick} className="mod-border-left p1 cursor-pointer hover-effect">
+                        <Svg svgName="moreAppend" svgClass="icon mod-12" />
```
GermainBergeron, line 66: Could you use an icon from the new iconography instead?

aiheon: IIRC we cannot since plasma react icons is not installed yet 🤔  It's all in next

packages/react/src/components/actionable-item/ActionableItem.tsx
GermainBergeron: I would target next

packages/react/src/components/actionable-item/ActionableItem.tsx
```diff
@@ -26,7 +25,11 @@ export interface IActionableItemProps {
      */
     actions?: IItemBoxProps[];
     /**
-     * Additionnal CSS class to add on the container
+     * Additionnal CSS class to add on the outermost container
```
kdang-coveo, line 28:
```suggestion
     * Additional CSS class to add on the outermost container
```

packages/react/src/components/actionable-item/ActionableItem.tsx
```diff
@@ -26,7 +25,11 @@ export interface IActionableItemProps {
...
+     */
+    className?: string;
+    /**
+     * Additionnal CSS class to add on the main item
```
kdang-coveo, line 32:
```suggestion
     * Additional CSS class to add on the main item
```

kdang-coveo (approved):

GermainBergeron (approved):

PR #2777, opened 2022-06-07
feat(react, button): add new loading option

toofff:
### Proposed Changes

Added `isLoading` props on buttons to show `LoadingSpinner` component.

image

NO

### Acceptance Criteria

### DIFF
--- a/packages/react/src/components/button/tests/Button.spec.tsx
+++ b/packages/react/src/components/button/tests/Button.spec.tsx
@@ -1,6 +1,7 @@
 import {mount, ReactWrapper, shallow} from 'enzyme';
 import * as _ from 'underscore';
-import {Tooltip} from '../../tooltip/Tooltip';
+
+import {Tooltip} from '../../tooltip';
 import {Button, IButtonProps} from '../Button';
 
 describe('Button', () => {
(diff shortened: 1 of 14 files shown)


aiheon:
Personal opinion but I fell like a ''loading'' button should never be disabled? 🙈 
It's counter-intuitive for me...

Also I'm curious of the user cases for this button? 🤔 Why the loading cannot be beside it instead of IN the button? 👀

packages/style/scss/elements/btn.scss
```diff
@@ -249,6 +249,50 @@
...
+        &.mod-link .loading-spinner,
+        &.mod-confirm .loading-spinner,
+        &.mod-confirm-error .loading-spinner,
+        &.mod-confirm-success .loading-spinner {
+            display: none;
+        }
+
+        &:disabled .loading-spinner:before,
+        &.state-disabled .loading-spinner:before {
+            background-color: var(--white);
```
aiheon, line 264: Would using a mixins or a function to change the colors of the spinner be more appropriate? 🤔 

Because I feel like changing the color of the bg and border in the :before could be needed in other cases and right now the component in not super flexible 🤔

packages/style/scss/elements/btn.scss
```diff
@@ -249,6 +249,50 @@
...
+    &.mod-loading {
+        &.mod-danger .loading-spinner,
+        &.mod-link .loading-spinner,
+        &.mod-confirm .loading-spinner,
+        &.mod-confirm-error .loading-spinner,
+        &.mod-confirm-success .loading-spinner {
+            display: none;
+        }
+
+        &:disabled .loading-spinner:before,
```
aiheon, line 262: is that necessary since the bg and border are white by default?

aiheon (approved):

packages/style/scss/elements/btn.scss
toofff: It's not really necessary, you're right

guscoveo (approved):

guscoveo:
> 

@aiheon I hear you, although I feel the other way. It is disabled precisely because it is rendered unclickable while the process in the back is loading. Having it in enabled/default might prompt the user to continue clicking in case the process takes too long (in which case I'm not sure if the process would be retriggered/restarted or not in our Covean context) having it disabled might deter the user from endless clicking.

packages/react/src/components/button/tests/Button.spec.tsx
```diff
@@ -162,5 +163,42 @@ describe('Button', () => {
                 expect(spyOnClick).toHaveBeenCalledTimes(1);
             });
         });
+
+        describe('with the isLoading props', () => {
+            it("shouldn't render the loading button", () => {
```
sallainCoveo, line 168:
```suggestion
            it("shouldn't render the loading button by default", () => {
```

sallainCoveo (approved):

packages/style/scss/tools/loading.scss
```diff
@@ -0,0 +1,25 @@
+/// @output Create a spinner
```
aiheon, line 1: Gracias 🙇‍♀️

JMathieu-UX:
Making the button inactive when it is loading is prioritized since the action is not accessible. Moreover keeping the same label on the button is preferable, the loading wheel already shows that the action is loading and it will be less harmful for the eye since the label will be the same.

Let’s kill the enabled variant and maintain just the disabled instances in that case [@Yann] please. Then I think we can merge. Thanks again for the quick turnaround

packages/react/src/components/button/tests/Button.spec.tsx
```diff
@@ -162,5 +163,42 @@ describe('Button', () => {
...
+            });
+
+            it('should render the loading button for a disabled button', () => {
+                showButton({isLoading: true, enabled: false});
+
+                expect(buttonComponent.find('button').hasClass('mod-loading')).toBe(true);
+                expect(buttonComponent.find('[role="alert"]').hasClass('loading-spinner')).toBe(true);
+            });
+
+            it("shouldn't render the loading button for a primary button", () => {
```
aiheon, line 189: no loading on primary button? 🤔

PR #2792, opened 2022-07-11
feat(react): allow using dropdowns inside popover

gdostie:
### Proposed Changes

When using a dropdown inside a popover, it would toggle closed the whole popover when selecting a dropdown value.

To test in the demo:
1. copy the following code in a sandbox
2. try selecting a value in the dropdown inside the popover (the popover shouldn't close)
3. try removing the `hasDropdowns` prop and do step 2 (popover closes)

```tsx
import { Popover, SingleSelectConnected, Button } from '@coveord/plasma-react';
import { useState } from 'react';

export default () => {
    const [isOpen, setOpen] = useState(false);
    return (
        <Popover
            attachment="top left"
            targetAttachment="bottom left"
            isOpen={isOpen}
            onToggle={setOpen}
            hasDropdowns
        >
            <Button>Toggle</Button>
            <div className='p3'>
                <SingleSelectConnected
                    id="single-select-1"
                    items={[
                        { value: 'one', displayValue: 'Option 1' },
                        { value: 'two', displayValue: 'Option 2' },
                        { value: 'three', displayValue: 'Option 3' },
                    ]}
                />
            </div>
        </Popover>
    )
};

```

None

### Acceptance Criteria

### DIFF
--- a/packages/react/src/components/popover/Popover.tsx
+++ b/packages/react/src/components/popover/Popover.tsx
@@ -24,20 +24,13 @@ export interface ITetherComponentCopiedProps {
     onRepositioned?: (...args: any[]) => void;
 }
 
-interface PopoverOwnProps {
-    /**
-     * Whether the popover has dropdowns components rendered in it. The goal of this prop is to prevent the popover from closing when selecting a dropdown value.
-     */
-    hasDropdowns?: boolean;
-}
-
 export interface IPopoverDispatchProps {
     onToggle?: (isOpen: boolean) => void;
     onMount?: (isOpen: boolean) => void;
     onUnmount?: () => void;
 }
 
-export interface IPopoverProps extends PopoverOwnProps, IPopoverDispatchProps, ITetherComponentCopiedProps {
+export interface IPopoverProps extends IPopoverDispatchProps, ITetherComponentCopiedProps {
     id?: string;
     /**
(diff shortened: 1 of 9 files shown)


aiheon (approved):

packages/react/src/components/popover/Popover.tsx
```diff
@@ -23,13 +24,20 @@ export interface ITetherComponentCopiedProps {
     onRepositioned?: (...args: any[]) => void;
 }
 
+interface PopoverOwnProps {
+    /**
+     * Whether the popover has dropdowns components rendered in it. The goal of this prop is to prevent the popover from closing when selecting a dropdown value.
+     */
+    hasDropdowns?: boolean;
```
GermainBergeron, line 31: Do we need this props? I feel like the new behaviour would be less confusing and adding a prop for a specific case goes against our recent discussions

gdostie: My concern was that `Defaults.DROP_ROOT = 'body'` by default so for someone that didn't change it, the popover would never close

gdostie: Makes sense to me! as long as it is clear

GermainBergeron (approved):

PR #2802, opened 2022-07-16
feat: 🎸 adding palette demo page to better document css class

aiheon:
### Proposed Changes

From time to time I found myself wishing to have a page in Plasma documenting the colors with the name we gave them in the scss file 🙈 

Now it exist 👀  
Next step will be adding the semantic derivates (i.e. `warning-70`) and make the page prettier? 😅 It's better than nothing but I could do better.

image

None, website only

### Acceptance Criteria


packages/website/src/SideNavigation.tsx
```diff
@@ -79,6 +79,8 @@ export const Navigation: FunctionComponent = () => {
             <CollapsibleSideSection title="Foundations">
                 <NavLink href="/foundations/Iconography" label="Iconography" />
                 <NavLink href="/foundations/Typekit" label="Typekit" />
+                <NavLink href="/foundations/Palette" label="Palette" />
+
```
aiheon, line 83:
```suggestion
```

gdostie:
Heres a few things I would change:
- Remove hsla values, mentioning those will tempt people to use those values directly instead of using the variables, which is something we want to avoid.
- All the semantics colors are missing, and I think they are probably the most important ones
    - `--success-10`
    - `--success-60`
    - `--success-100`
    - `--info-10`
    - `--info-60`
    - `--info-100`
    - `--warning-10`
    - `--warning-70`
    - `--warning-90`
    - `--critical-10`
    - `--critical-70`
    - `--critical-100`
- For each color, we have the `--rgb` variant that exists, perhaps we should mention it.

aiheon:
> Heres a few things I would change:
> 
> * Remove hsla values, mentioning those will tempt people to use those values directly instead of using the variables, which is something we want to avoid.
> * All the semantics colors are missing, and I think they are probably the most important ones
>   
>   * `--success-10`
>   * `--success-60`
>   * `--success-100`
>   * `--info-10`
>   * `--info-60`
>   * `--info-100`
>   * `--warning-10`
>   * `--warning-70`
>   * `--warning-90`
>   * `--critical-10`
>   * `--critical-70`
>   * `--critical-100`
> * For each color, we have the `--rgb` variant that exists, perhaps we should mention it.

I've added the HSLA because it happens that in Figma mockups it's the only color we have (the name is not mentionned so it's making it harder to find the var name 😕 )

I already mentionne in the description that the semantic variants are next for improvement 😋 
And I agree for the RGB but I want to think about a better way to visualize the informations than a table 🤔

aiheon:
@gdostie I'll explore lint rules to make sure dev don't use HSLA or Hex color directly 🤔 
https://github.com/stylelint-scss/stylelint-scss/issues/4

kdang-coveo:
I really like this improvement! I kinda prefer a square palette over a long one as I think it might be a bit jarring to the eyes when scrolling through the colours so that it looks something like this below:

CleanShot 2022-08-25 at 13 38 28

As for someone who might be viewing this fullscreen, it might be a bit harsh on the eyes coming from the monitor with lots of bands of colour as they scroll, only a suggestion though

aiheon:
@kdang-coveo i'll look into that 👀 It's does look better indeed 😄

PR #2811, opened 2022-07-29
[MLX-170] fix(infotoken): allow passing classes to override svg properties

lprovost-coveo:
### Proposed Changes

Added the possibility to pass classes directly to the svg. So that the look of it can be overriden.
(Need to override color on svg)

:no_entry: 

### Acceptance Criteria

### DIFF
--- a/packages/react/src/components/info-token/InfoToken.tsx
+++ b/packages/react/src/components/info-token/InfoToken.tsx
@@ -28,6 +28,7 @@ export interface InfoTokenProps {
     type: InfoTokenType;
     size: InfoTokenSize;
     mode: InfoTokenMode;
+    svgClassName?: string;
     className?: string;
 }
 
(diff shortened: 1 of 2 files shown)


packages/react/src/components/info-token/InfoToken.tsx
```diff
@@ -84,10 +85,10 @@ const ModeClassMapping: Record<InfoTokenMode, string> = {
     [InfoTokenMode.Filled]: 'filled',
 };
 
-export const InfoToken: FunctionComponent<InfoTokenProps> = ({mode, size, type, className}) => (
+export const InfoToken: FunctionComponent<InfoTokenProps> = ({mode, size, type, className, svgClassName}) => (
     <Svg
         className={classNames('info-token', ModeClassMapping[mode], SizeClassMapping[size], className)}
         svgName={SvgMapping[type][size]}
-        svgClass={classNames('icon mod-stroke', SizeClassMapping[size], TypeColorMapping[type])}
+        svgClass={classNames('icon mod-stroke', SizeClassMapping[size], TypeColorMapping[type], svgClassName)}
```
dmgauthier, line 92: I dont know if its just me but I feel this component is basically mapping most of the props 1:1.  
I wonder if using the <Svg /> component strait away wouldn't be more simple.

lprovost-coveo: I'm not 100% sure what you mean but there are two styling classes being passed down here, one to the span (className) and one to the actual real SVG (svgClass)

toofff: @dmgauthier Would you like to pass the `Svg` component in the props of the `InfoToken` component?

If so, it goes against the work currently being done to get rid of the `Svg` component and it would have a pretty monstrous impact on the Plasma library but also on the AdminUI project.

I think @lprovost-coveo's solution is the right one and less time consuming

toofff (approved):

PR #2816, opened 2022-08-11
fix: 🐛 adding a fallback when maxWidth is a NaN

aiheon:
### Proposed Changes

https://coveord.atlassian.net/browse/ADUI-8247

We have an issue in the admin where the Single select gets an error when trying to set the maxWidth value.
To patch this issue, i'm checking if the calculated value is a NaN and if so, I force a fallback value (I picked 850px but it could be bigger if needed ? 🙈 )

image

I'm prenneuse of better idea also 😅 

None

### Acceptance Criteria

### DIFF
--- a/packages/react/src/components/drop/DropPod.tsx
+++ b/packages/react/src/components/drop/DropPod.tsx
@@ -213,9 +213,12 @@ const RDropPod: FunctionComponent<IRDropPodProps> = ({
             // Restrict the max-width to the inner width of the closest relatively positioned ancestor
             if (relativeParent) {
                 const {paddingLeft, paddingRight} = getComputedStyle(relativeParent) ?? {};
+                const fallbackMaxWidth = 850;
+                const calculatedMaxWidth = parentOffset.width - (parseFloat(paddingLeft) + parseFloat(paddingRight));
+                const maxWidth = !!isNaN(calculatedMaxWidth) ? fallbackMaxWidth : calculatedMaxWidth;
                 newDomPosition.style = {
                     ...newDomPosition.style,
-                    maxWidth: parentOffset.width - (parseFloat(paddingLeft) + parseFloat(paddingRight)),
+                    maxWidth: maxWidth,
                 };
 
                 // Don't show if no space to render the drop target inside the window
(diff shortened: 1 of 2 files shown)


packages/react/src/components/drop/DropPod.tsx
```diff
@@ -213,9 +213,12 @@ const RDropPod: FunctionComponent<IRDropPodProps> = ({
             // Restrict the max-width to the inner width of the closest relatively positioned ancestor
             if (relativeParent) {
                 const {paddingLeft, paddingRight} = getComputedStyle(relativeParent) ?? {};
+                const fallbackMaxWidth = 850;
+                const calculatedMaxWidth = parentOffset.width - (parseFloat(paddingLeft) + parseFloat(paddingRight));
+                const maxWidth = !!isNaN(calculatedMaxWidth) ? fallbackMaxWidth : calculatedMaxWidth;
```
GermainBergeron, line 218: isNaN is already a boolean, no need for the `!!`
```suggestion
                const maxWidth = isNaN(calculatedMaxWidth) ? fallbackMaxWidth : calculatedMaxWidth;
```

packages/react/src/components/drop/DropPod.tsx
```diff
@@ -213,9 +213,12 @@ const RDropPod: FunctionComponent<IRDropPodProps> = ({
             // Restrict the max-width to the inner width of the closest relatively positioned ancestor
             if (relativeParent) {
                 const {paddingLeft, paddingRight} = getComputedStyle(relativeParent) ?? {};
```
GermainBergeron, line 215: Maybe we should set paddingLeft and paddingRight to 0 by default?

GermainBergeron (approved):

gdostie (approved):

gprovostqa (approved):

packages/react/src/components/drop/DropPod.tsx
aiheon: I tried it and I would get a maxWidth of 0 in the admin because the parent offset is 0 in my bug which is not nice 😆 🙈

PR #2820, opened 2022-08-17
fix: 🐛 (select) select id are added only once in redux state

aiheon:
### Proposed Changes

If the ID is already in the redux state, we don't add a duplicate. Select id should be in the state only once.

Should be none

### Acceptance Criteria


packages/react/src/components/select/tests/SelectReducers.spec.ts
```diff
@@ -69,6 +69,15 @@ describe('Select', () => {
...
                 expect(oldState).toEqual(oldStateBefore);
             });
+
+            it('does not add a duplicate ID in the state', () => {
+                const oldState: ISelectState[] = [selectInitialState];
+                const newState: ISelectState[] = selectCompositeReducer(oldState, addSelect(id));
+                const newNewState: ISelectState[] = selectCompositeReducer(newState, addSelect(id)); // trying to add the same id againß
+
+                expect(newState[1].id).toBe(id);
+                expect(newNewState.length).toBe(2); // oldState lenght is 1.
```
GermainBergeron, line 79: I would put some expect after the first add just to make the behaviour a bit more clear

```suggestion
                const oldState: ISelectState[] = [selectInitialState];
                const newState: ISelectState[] = selectCompositeReducer(oldState, addSelect(id));

                expect(newState.length).toBe(2); // oldState length is 1.                
                expect(newState[1].id).toBe(id);
                                
                const newNewState: ISelectState[] = selectCompositeReducer(newState, addSelect(id)); // trying to add the same id again

                expect(newNewState.length).toBe(2); // length didn't change
```

packages/react/src/components/select/SelectReducers.ts
```diff
@@ -31,7 +31,11 @@ export const selectCompositeReducer = (
 ): ISelectState[] => {
     switch (action.type) {
         case SelectActions.add:
-            return [...state, selectReducer(undefined, action)];
+            const index = state.findIndex((element) => element.id === action.payload.id);
+            if (index === -1) {
```
GermainBergeron, line 35:
```suggestion
            // do not add to the state if it is already present
            if (index === -1) {
```

gdostie (approved):

GermainBergeron (approved):

PR #2833, opened 2022-09-13
[plasma-react] handle null limits in Limits component

gdostie:
### Proposed Changes

Null or undefined limit and usage values should not be rounded out to 0 and the label should be hidden when that occurs

image

image

image

image

I also took the time to improve the dev setup 🤖 

None expected

### Acceptance Criteria

### DIFF
--- a/packages/react/src/components/limit/Limit.tsx
+++ b/packages/react/src/components/limit/Limit.tsx
@@ -54,6 +54,8 @@ export interface LimitOwnProps {
     onHistoryIconClick?: () => void;
 }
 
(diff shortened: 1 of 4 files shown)


larosek:
Like I said to Gael, bot usage and limit value null should not happen with what the backend is returning.
We should see issues in dev quickly if it happens.

Definitely not a UX person by I feel like the limit label and value should stay on the right even when usage is null 🤔  
If I try to image that card in our settings panel, it would be better to keep all limit values align.

Screen Shot 2022-09-13 at 23 13 35

gdostie:
Good point, I added a min-height

image

aiheon (approved):

PR #2842, opened 2022-09-23
UITOOL-810/prepare react 18 bump part 1

aiheon:
### Proposed Changes

https://coveord.atlassian.net/browse/UITOOL-810

This pr is bumping deps for swc, jest and bumping to latest all deps in the package.json at root.

I need to bump react-dnd and d3 also bump i'll do it in another pr because it's huge and complicated 😅 

I also removed tests that I think were useless (they were breaking with the swc bump).

should be none

### Acceptance Criteria


gdostie (approved):

GermainBergeron:
I feel like we could target master with that PR. I see a bunch of changes in the diff which I'm not sure about (e.g., package versions, tokens, colors, etc)

aiheon:
> I feel like we could target master with that PR. I see a bunch of changes in the diff which I'm not sure about (e.g., package versions, tokens, colors, etc)

yeah I pulled master again when I shouldn't 🙈

PR #2862, opened 2022-10-20
fix(xyaxis): when few data points, the x axis would not display any l…

FelixBlaisThon:
### Proposed Changes

On low data point, the tick labels dont show up so I changed the logic to accept the whole array of data when the value 
`const ticks = _.chunk(newXScale.domain(), Math.floor(newXScale.domain().length / xNumberOfTicks))` returns nothing (hence the no labels)

Go in plasma -> Chart and change the number of point in graph data from 25 to 1-2-3-4-5. it should still displays instead of not (as of now)
### Acceptance Criteria


gdostie (approved):
:shipit:

PR #2864, opened 2022-10-24
style(facets): removed border-right to facets column, will be applied…

SimonMilord:
… to table now

### Proposed Changes

Removed the border-right on the facets column for the activity table. The border will be applied to the table instead.
image

### Acceptance Criteria


gdostie:
That will have an impact everywhere the facet-column class is used, not just the activity browser. I guess you will make the appropriate changes to all those places? (like the log-browser)

SimonMilord:
> 

I will proceed differently and add a selector for the activity facets instead so that it does not affect the rest of the code such as log-browser

gdostie (commented):
Not sure I like adding css rules that apply only for a use case that isn't part of plasma. How is this relevant for the projects other than the admin-ui that use plasma?

PR #2868, opened 2022-10-25
ci(github): add CD workflow for master branch

gdostie:
### Proposed Changes

Implementing an initial CD workflow for the master branch

Deployment might be broken after merging this, but I'll take care to adjust if it is the case.

### Acceptance Criteria


.github/workflows/cd.yml
```diff
@@ -0,0 +1,40 @@
+name: CD
+
+on:
+    push:
+        branches:
+            - master
```
GermainBergeron, line 6: Should we have something here or a bit down below that skips if it was a version bump?

.github/actions/deploy/action.yml
```diff
@@ -0,0 +1,20 @@
...
+        description: The AWS S3 bucket in which to deploy the website.
+runs:
+    using: composite
+    steps:
+        - name: Configure AWS Credentials
+          uses: aws-actions/configure-aws-credentials@v1
+          with:
+              role-to-assume: ${{ inputs.AWS_ROLE }}
+              aws-region: ${{ inputs.AWS_REGION }}
+        - name: Push distribution code to s3 demo folder
```
aiheon, line 18: This make me think that we should maybe add the `clean up demo` workflow in Plasma also? 🤔

gdostie: Good idea, I will create a separate task for it and do it in a different PR

aiheon (approved):
I like it 🎉

PR #2930, opened 2022-11-15
Mantine code editor demo page

gdostie:
### Proposed Changes

Added a demo page for the CodeEditor from plasma-mantine package. It uses the new demo-loader which is a static demo for now, but will eventually have a "Open in CodeSandbox" feature.

image

Removing the live sandbox allows the page to load instantly compared to the other pages, and also fixes the conflict with the monaco instance.

None

### Acceptance Criteria

### DIFF
--- a/packages/website/src/SideNavigation.tsx
+++ b/packages/website/src/SideNavigation.tsx
@@ -87,6 +87,7 @@ export const Navigation: FunctionComponent = () => {
             <CollapsibleSideSection title="Mantine">
                 <NavLink href="/mantine/Collection" label="Collection" />
                 <NavLink href="/mantine/Header" label="Page Header" />
+                <NavLink href="/mantine/CodeEditor" label="Code Editor" />
             </CollapsibleSideSection>
             <CollapsibleSideSection title="Layout">
                 <NavLink href="/layout/Banner" label="Banner" />
--- a/packages/website/src/building-blocs/Demo.tsx
+++ b/packages/website/src/building-blocs/Demo.tsx
@@ -1,0 +1,66 @@
+import {ActionIcon, createStyles, ScrollArea, SimpleGrid, Stack, Tooltip, useClipboard} from '@coveord/plasma-mantine';
+import {CheckSize16Px, CopySize16Px} from '@coveord/plasma-react-icons';
+import {Prism} from '@mantine/prism';
+import vsDark from 'prism-react-renderer/themes/vsDark';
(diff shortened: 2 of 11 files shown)


packages/website/src/building-blocs/Demo.tsx
```diff
@@ -0,0 +1,66 @@
...
+            </Prism>
+            <Stack className={classes.actions} spacing="xs">
+                <Tooltip label={clipboard.copied ? 'Copied' : 'Copy'} position="left">
+                    <ActionIcon radius="sm" onClick={() => clipboard.copy(snippet)}>
+                        {clipboard.copied ? <CheckSize16Px height={16} /> : <CopySize16Px height={16} />}
+                    </ActionIcon>
+                </Tooltip>
+                {/* 
+                    // TODO
+                    <Tooltip label="Open in CodeSanbox" position="left">
```
aiheon, line 55: Do you plan to get the sandbox open directly in the Sandbox website or you will open a new page > modal? 👀

aiheon:
It's hard to get on video but on my first render the initial value did not load, I refreshed the page a couple of time and once I got the values without colors then no value again, then the values with the colors 🤔 

Also for fun I went back to the header page and got this  😅 
image

packages/website/src/building-blocs/Demo.tsx
gdostie: The plan is to open the sandbox in a new tab (target blank)

gdostie:
> It's hard to get on video but on my first render the initial value did not load, I refreshed the page a couple of time and once I got the values without colors then no value again, then the values with the colors 🤔
> 
> Also for fun I went back to the header page and got this 😅

This is an issue with monaco loading its file through cdn, sometimes they just don't want to load. But that was happening with the previous sandbox as you can see from your screenshot in the page header demo page.

packages/website/src/SideNavigation.tsx
```diff
@@ -87,6 +87,7 @@ export const Navigation: FunctionComponent = () => {
             <CollapsibleSideSection title="Mantine">
                 <NavLink href="/mantine/Collection" label="Collection" />
```
GermainBergeron, line 88: 🤔 I don't see the collection link in the demo

GermainBergeron (approved):

packages/website/src/SideNavigation.tsx
gdostie: Demo updated, now you should. If not, probably a cache issue on your side

PR #2961, opened 2022-11-29
fix(react): make plasma react usable from code sandbox

gdostie:
### Proposed Changes

Right now, plasma-react cannot be used from within a code sandbox: [example](https://codesandbox.io/s/82y3hv?module=/src/Demo.tsx&file=/src/index.tsx)

The "previewError" svg from plasma-style is preventing plasma-react from being able to load in CodeSandbox, because of this bug https://github.com/facebook/create-react-app/issues/11770. CodeSandbox uses create-react-app which expects svgs to have attributes in camel case instead of standard xml markup. For example `xmlns:xlink="http://www.w3.org/1999/xlink"`  would need to be `xmlnsXlink="http://www.w3.org/1999/xlink"`. 

By not importing the svg file directly and using the svgString from the js object instead, it won't happen. But doing so requires the `xmlns` attributes to be set on the svg string.

### How to test

1. Go in the BrowserPreview demo page
2. This svg should still be able to load 
image

None

### Acceptance Criteria

### DIFF
--- a/packages/react/jest/setup.ts
+++ b/packages/react/jest/setup.ts
@@ -15,3 +15,5 @@ document.createRange = () => {
 
     return range;
 };
+
+global.URL.createObjectURL = jest.fn();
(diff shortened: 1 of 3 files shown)


SimonMilord (approved):
Good job Gael!

PR #2964, opened 2022-12-01
feat(table): added support for doubleclick on a row

SimonMilord:
### Proposed Changes

Added prop in Table to support doubleClick action on a Table Row.
We opted to add the prop in the table in case there is something wanted out of doubleclick that is not an action on the table.

None
### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/table/Table.tsx
+++ b/packages/mantine/src/components/table/Table.tsx
@@ -124,6 +124,10 @@ interface TableProps<T> {
      * Initial state of the table
      */
     initialState?: InitialTableState & Partial<TableFormType>;
+    /**
+     * Action passed when user doubleClicks on a row
+     */
+    doubleClickAction?: (datum: T) => void;
 }
 
 interface TableType {
@@ -149,6 +153,7 @@ export const Table: TableType = <T,>({
     onChange,
     children,
     loading = false,
+    doubleClickAction,
 }: TableProps<T>) => {
     const convertedChildren = Children.toArray(children) as ReactElement[];
     const header = convertedChildren.find((child) => child.type === TableHeader);
@@ -224,6 +229,7 @@ export const Table: TableType = <T,>({
             <Fragment key={row.id}>
                 <tr
                     onClick={() => toggleRowSelection(row)}
+                    onDoubleClick={() => doubleClickAction(row.original)}
                     className={cx(classes.row, {[classes.rowSelected]: row.getIsSelected()})}
                 >
                     {row.getVisibleCells().map((cell) => {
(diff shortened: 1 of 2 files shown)


packages/mantine/src/components/table/Table.tsx
```diff
@@ -224,6 +229,7 @@ export const Table: TableType = <T,>({
             <Fragment key={row.id}>
                 <tr
                     onClick={() => toggleRowSelection(row)}
+                    onDoubleClick={() => doubleClickAction(row.original)}
```
gdostie, line 232: If the doubleClickAction prop is undefined this will throw an error

```suggestion
                    onDoubleClick={() => doubleClickAction?.(row.original)}
```

This makes me realize that you forget to write a unit test for this new feature 😄

packages/mantine/src/components/table/Table.tsx
```diff
@@ -124,6 +124,10 @@ interface TableProps<T> {
      * Initial state of the table
      */
     initialState?: InitialTableState & Partial<TableFormType>;
+    /**
+     * Action passed when user doubleClicks on a row
```
gdostie, line 128:
```suggestion
     * Action passed when user double clicks on a row
     *
     * @param datum The datum of the row that was double clicked
```

packages/mantine/src/components/table/Table.tsx
SimonMilord: I realized! I was just working on this now ;)

packages/mantine/src/components/table/__tests__/Table.spec.tsx
```diff
@@ -127,6 +127,20 @@ describe('Table', () => {
...
+        render(
+            <Table<RowData>
+                data={[{firstName: 'Mario'}, {firstName: 'Luigi'}]}
+                columns={columns}
+                doubleClickAction={() => alert(`alert`)}
+            ></Table>
+        );
+        const alertMock = jest.spyOn(window, 'alert').mockImplementation();
+        await user.dblClick(screen.getByRole('cell', {name: 'Mario'}));
+        expect(alertMock).toHaveBeenCalledTimes(1);
```
gdostie, line 141:
```suggestion
        const user = userEvent.setup();
        const doubleClickSpy = jest.fn();
        render(
            <Table<RowData>
                data={[{firstName: 'Mario'}, {firstName: 'Luigi'}]}
                columns={columns}
                doubleClickAction={doubleClickSpy}
            ></Table>
        );
        await user.dblClick(screen.getByRole('cell', {name: 'Mario'}));
        expect(doubleClickSpy).toHaveBeenCalledTimes(1);
        expect(doubleClickSpy).toHaveBeenCalledWith({firstName: 'Mario'});
```

gdostie (approved):

jxu-coveo (approved):

packages/mantine/src/components/table/__tests__/Table.spec.tsx
SimonMilord: cleaner!

PR #2968, opened 2022-12-06
Feat UI tool1027 add label description props collection

SimonMilord:
### Proposed Changes

- Added props support for label, description and error validation in the collection component.
- Modified the demo to display the use of those props and validation
- Also edited Collection description to make it more representative of what it is. The old description was out of date.
- Also fixed a typo in files names 

image

none

### Acceptance Criteria


packages/mantine/src/components/collection/Collection.tsx
```diff
@@ -99,15 +100,32 @@ interface CollectionProps<T> extends DefaultProps<Selectors<typeof useStyles>> {
...
 
 const defaultProps: Partial<CollectionProps<unknown>> = {
     draggable: false,
     addLabel: 'Add item',
     addDisabledTooltip: 'There is already an empty item',
     disabled: false,
-    spacing: 'xs',
+    spacing: 'xl',
     required: false,
+    error: 'Invalid Input',
```
aiheon, line 126:
```suggestion
    error: 'Invalid input',
```

I'm pretty sure someone in the Doc team would point that out 😆

packages/mantine/src/components/collection/Collection.tsx
```diff
@@ -99,15 +100,32 @@ interface CollectionProps<T> extends DefaultProps<Selectors<typeof useStyles>> {
...
+    label: '',
+    description: '',
```
aiheon, line 128: Are does necessary? 🤔 
I feel like default props should be used only if there's a default text 👀

packages/mantine/src/components/collection/Collection.tsx
```diff
@@ -190,9 +212,17 @@ export const Collection = <T,>(props: CollectionProps<T>) => {
                         {...others}
                     >
                         <Stack spacing={spacing}>
-                            {items}
-                            {provided.placeholder}
-                            {_addButton}
+                            <Stack spacing={'xs'}>
+                                <Title order={3}>{label}</Title>
```
aiheon, line 216: to avoid exploding code if you remove the label and description from the default props you can do something like this? (i'm not sur of the syntax by heart 🙈 )

This way also you avoid to add useless divs in the DOM on render 👍 

```suggestion
                           title ??  <Title order={3}>{label}</Title>
```

packages/mantine/src/components/collection/CollectionItem.tsx
```diff
@@ -3,11 +3,13 @@ import {ActionIcon, DefaultProps, Group, Selectors} from '@mantine/core';
...
 import {Draggable} from 'react-beautiful-dnd';
 
-import useStyles from './Colllection.styles';
+import useStyles from './Collection.styles';
 
 interface CollectionItemProps extends CollectionItemSharedProps {
     draggable?: boolean;
     disabled: boolean;
+    label?: string;
+    error?: string;
```
gdostie, line 12: These don't seem to be used

packages/mantine/src/components/collection/Collection.tsx
```diff
@@ -190,9 +212,17 @@ export const Collection = <T,>(props: CollectionProps<T>) => {
...
+                                <Title order={6} color="dimmed" weight={200}>
+                                    {description}
+                                </Title>
+                            </Stack>
```
gdostie, line 220: We probably should be using the exact same style as the other input components. You can follow the CodeEditor as an example https://github.com/coveo/plasma/blob/master/packages/mantine/src/components/code-editor/CodeEditor.tsx#L129-L147

packages/mantine/src/components/collection/Collection.tsx
```diff
@@ -99,15 +100,32 @@ interface CollectionProps<T> extends DefaultProps<Selectors<typeof useStyles>> {
...
+    /**
+     * Description for the collection (under the label)
+     */
+    description?: string;
+    /**
+     * Error message if input is invalid
+     *
+     * @default 'Invalid Input'
+     */
+    error?: string;
```
gdostie, line 116: By extending `Omit<InputWrapperBaseProps, 'inputContainer' | 'inputWrapperOrder'>`, we wouldn't have to manually declare those. You can have a look at the [CodeEditor](https://github.com/coveo/plasma/blob/master/packages/mantine/src/components/code-editor/CodeEditor.tsx#L35-L36) to see how its done

packages/mantine/src/components/collection/Collection.tsx
SimonMilord: ok will fix

packages/mantine/src/components/collection/Collection.tsx
SimonMilord: touché

packages/mantine/src/components/collection/Collection.tsx
gdostie, line 126: I don't think we wanna have an error message by default.

packages/mantine/src/components/collection/Collection.tsx
```diff
@@ -99,15 +100,32 @@ interface CollectionProps<T> extends DefaultProps<Selectors<typeof useStyles>> {
...
 }
 
 const defaultProps: Partial<CollectionProps<unknown>> = {
     draggable: false,
     addLabel: 'Add item',
     addDisabledTooltip: 'There is already an empty item',
     disabled: false,
-    spacing: 'xs',
+    spacing: 'xl',
```
gdostie, line 124: I think we can make it work without changing the default spacing.

packages/mantine/src/components/collection/Collection.tsx
```diff
@@ -149,6 +170,7 @@ export const Collection = <T,>(props: CollectionProps<T>) => {
             key={index}
             disabled={disabled}
             draggable={draggable}
+            error={error}
```
gdostie, line 173: The error should probably be rendered underneath the whole collection. We don't need to manage the individual item's errors here.

packages/mantine/src/components/collection/Collection.tsx
SimonMilord: noted

packages/mantine/src/components/collection/Collection.tsx
SimonMilord: noted!

packages/mantine/src/components/collection/Collection.tsx
SimonMilord: ok!

packages/mantine/src/components/collection/Collection.tsx
SimonMilord: hmmm ok I thought for the experience it would be better to put it under the faulty input no?

gdostie: Right now your code doesn't do anything with the error 😅

gdostie (approved):

PR #2989, opened 2022-12-13
fix(mantine form): override getinputprops and use in collection

dmgauthier:
### Description

Override of UseForm to front a `getInputProps` addapted to arrays

The controlled Collection component will not leverage the onChange prop anymore. Thus Collection components controlled manually with the `onChange` prop will break

### Acceptance Criteria

[UITOOL-1030]: https://coveord.atlassian.net/browse/UITOOL-1030?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/mantine/src/components/collection/Collection.tsx
+++ b/packages/mantine/src/components/collection/Collection.tsx
@@ -12,12 +12,13 @@ import {
     Tooltip,
     useComponentDefaultProps,
 } from '@mantine/core';
-import {useId} from '@mantine/hooks';
+import {ReorderPayload} from '@mantine/form/lib/types';
+import {useDidUpdate, useId} from '@mantine/hooks';
 import {ReactNode} from 'react';
 import {DragDropContext, Droppable} from 'react-beautiful-dnd';
-import {useControlledList} from '../../hooks';
-import {CollectionItem} from './CollectionItem';
+
 import useStyles from './Collection.styles';
+import {CollectionItem} from './CollectionItem';
 
 interface CollectionProps<T>
     extends Omit<InputWrapperBaseProps, 'inputContainer' | 'inputWrapperOrder'>,
@@ -40,10 +41,6 @@ interface CollectionProps<T>
      */
     value?: T[];
     /**
-     * The initial items of the collection (for uncontrolled usage only)
-     */
-    defaultValue?: T[];
-    /**
      * Unused, has no effect
      */
     onFocus?: () => void;
@@ -59,6 +56,19 @@ interface CollectionProps<T>
      * @param itemIndex The index of the item that was removed
      */
     onRemoveItem?: (itemIndex: number) => void;
+    /**
+     * Function that gets called whenever a collection item needs to be reordered
+     *
+     * @param payload The origin and destination index of the item to reorder
+     */
+    onReorderItem?: (payload: ReorderPayload) => void;
+    /**
+     * Function that gets called when a new item needs to be added to the collection
+     *
+     * @param value The the value of the item to insert
+     * @param index The index of the new item to insert
+     */
+    onInsertItem?: (value: T, index: number) => void;
     /**
      * Whether the collection should have drag and drop behavior enabled
      *
@@ -116,9 +126,10 @@ const defaultProps: Partial<CollectionProps<unknown>> = {
 export const Collection = <T,>(props: CollectionProps<T>) => {
     const {
         value,
-        defaultValue,
         onChange,
         onRemoveItem,
+        onReorderItem,
+        onInsertItem,
         disabled,
         draggable,
         children,
@@ -146,12 +157,14 @@ export const Collection = <T,>(props: CollectionProps<T>) => {
     const {classes, cx} = useStyles(null, {classNames, name: 'Collection', styles, unstyled});
     const collectionID = useId('dnd-droppable');
 
-    const [values, {append, remove, reorder}] = useControlledList({value, onChange, defaultValue});
-    const hasOnlyOneItem = values.length === 1;
-    const removeItem = (index: number) => () => {
-        remove(index);
-        onRemoveItem?.(index);
-    };
+    const hasOnlyOneItem = value.length === 1;
+
+    /**
+     * Enforcing onChange when the value is modified will make sure the errors are carried through.
+     */
+    useDidUpdate(() => {
+        onChange?.(value);
+    }, [JSON.stringify(value)]);
 
     const _label = label ? (
         <Input.Label required={required} {...labelProps}>
@@ -171,13 +184,13 @@ export const Collection = <T,>(props: CollectionProps<T>) => {
             </Box>
         ) : null;
 
-    const items = values.map((item, index) => (
+    const items = value.map((item, index) => (
         <CollectionItem
             key={index}
             disabled={disabled}
             draggable={draggable}
             index={index}
-            onRemove={removeItem(index)}
+            onRemove={() => onRemoveItem?.(index)}
             styles={styles}
             removable={!(required && hasOnlyOneItem)}
         >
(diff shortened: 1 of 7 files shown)


packages/mantine/src/components/collection/Collection.tsx
```diff
@@ -102,6 +98,25 @@ interface CollectionProps<T>
      * @default false
      */
     required?: boolean;
+    /**
+     * Function that changes the list of values order.
+     *
+     * @param payload The origin and destination index of the item to reorder
+     */
+    reorderListItem?: (payload: Record<'from' | 'to', number>) => void;
```
gdostie, line 106: Can we import the payload type directly from mantine?

```suggestion
    reorderListItem?: (payload: ReorderPayload) => void;
```

packages/mantine/src/components/collection/Collection.tsx
```diff
@@ -102,6 +98,25 @@ interface CollectionProps<T>
...
+     * @param index The index of the item to remove
+     */
+    removeListItem?: (index: number) => void;
+    /**
+     * Function that removes a specific item from the list of values
+     *
+     * @param value The the value of the item to insert
+     * @param index The index of the new item to insert
+     */
+    insertListItem?: (value: unknown, index: number) => void;
```
gdostie, line 119:
```suggestion
    insertListItem?: (value: T, index: number) => void;
```

packages/mantine/src/components/collection/Collection.tsx
```diff
@@ -102,6 +98,25 @@ interface CollectionProps<T>
      * @default false
      */
     required?: boolean;
+    /**
+     * Function that changes the list of values order.
```
gdostie, line 102:
```suggestion
     * Function that gets called whenever a collection item needs to be reordered
```

packages/mantine/src/components/collection/Collection.tsx
```diff
@@ -102,6 +98,25 @@ interface CollectionProps<T>
...
+     */
+    reorderListItem?: (payload: Record<'from' | 'to', number>) => void;
+    /**
+     * Function that removes a specific item from the list of values
+     *
+     * @param index The index of the item to remove
+     */
+    removeListItem?: (index: number) => void;
+    /**
+     * Function that removes a specific item from the list of values
```
gdostie, line 114:
```suggestion
     * Function that gets called when a new item needs to be added to the collection
```

packages/mantine/src/components/collection/Collection.tsx
```diff
@@ -146,13 +163,19 @@ export const Collection = <T,>(props: CollectionProps<T>) => {
     const {classes, cx} = useStyles(null, {classNames, name: 'Collection', styles, unstyled});
     const collectionID = useId('dnd-droppable');
 
-    const [values, {append, remove, reorder}] = useControlledList({value, onChange, defaultValue});
-    const hasOnlyOneItem = values.length === 1;
+    const hasOnlyOneItem = value.length === 1;
     const removeItem = (index: number) => () => {
-        remove(index);
+        removeListItem(index);
         onRemoveItem?.(index);
```
gdostie, line 169: Since this prop already exist, maybe we should reuse it instead of having `onRemoveItem` and `removeListItem`. I think we should use either 
- `reorderListItem`
- `removeListItem`
- `insertListItem`
or
- `onRemoveItem`
- `onInsertItem`
- `onReorderItem`
but not a mix of the 2 naming patterns

Thoughts? I personally have a preference for the `onSomething` pattern. It sounds more like the `onChange` that we had.

packages/mantine/src/components/collection/Collection.tsx
```diff
@@ -146,13 +163,19 @@ export const Collection = <T,>(props: CollectionProps<T>) => {
...
     };
 
+    /**
+     * Enforcing onChange when the value is modified will make sure the errors are carried through.
+     */
+    useEffect(() => {
+        onChange?.(value);
+    }, [JSON.stringify(value)]);
```
gdostie, line 177: Lets `useDidUpdate` for that

> Call function in useEffect when value changes, but not when component mounts

packages/mantine/src/form/use-form.ts
```diff
@@ -0,0 +1,31 @@
+import {useForm as useMantineForm} from '@mantine/form';
+import {_TransformValues, UseFormInput, GetInputProps} from '@mantine/form/lib/types';
+
+export const useForm = <
+    Values = Record<string, unknown>,
+    TransformValues extends _TransformValues<Values> = (values: Values) => Values
+>(
+    options: UseFormInput<Values, TransformValues> = {}
+) => {
```
gdostie, line 9: Does the following works?

```suggestion
export const useForm: typeof useMantineForm = (options) => {
```

packages/mantine/src/form/use-form.ts
```diff
@@ -0,0 +1,31 @@
+import {useForm as useMantineForm} from '@mantine/form';
+import {_TransformValues, UseFormInput, GetInputProps} from '@mantine/form/lib/types';
+
+export const useForm = <
```
toofff, line 4: Is this task still relevant?
If so, we would also need to override the [FormProvider](https://github.com/mantinedev/mantine/blob/master/src/mantine-form/src/FormProvider/FormProvider.tsx) in this pull request so that it can use the useForm overload.

packages/mantine/src/components/collection/Collection.tsx
dmgauthier: agreed. I'll leverage the already existing `onRemoveItem`

packages/mantine/src/form/use-form.ts
dmgauthier: With a small tweek, it does :+1:

packages/mantine/src/form/use-form.ts
```diff
@@ -0,0 +1,31 @@
...
+    options: UseFormInput<Values, TransformValues> = {}
+) => {
+    const form = useMantineForm(options);
+
+    const getInputProps: GetInputProps<Values> = (
+        path,
+        {type = 'input', withError = type === 'input', withFocus = true} = {}
+    ) => {
+        const originalPayload = form.getInputProps(path, {type, withError, withFocus});
+        if (Array.isArray(originalPayload.value)) {
```
toofff, line 17: This can be dangerous because all form components that have an array as their value will automatically be overloaded.

dmgauthier: Yes they will be automatically overloaded. this is right. 

But in the end this should not change anything. Because if the component is an original one from Mantine. It will most certainly leverage the already implemented`onChange` method that has never been removed or changed.

Since the useForm getInputProps only passes new props to the component, nothing should be changed in actual implementations.

packages/mantine/src/form/FormProvider.tsx
```diff
@@ -0,0 +1,40 @@
+/* eslint-disable prefer-arrow/prefer-arrow-functions */
+import {_TransformValues, UseForm, UseFormReturnType} from '@mantine/form/lib/types';
+import React, {createContext, useContext} from 'react';
+
+import {useForm} from './useForm';
+
+// import {UseFormReturnType, UseForm, _TransformValues} from '../types';
```
toofff, line 7: Why this line is commented?

dmgauthier: :see_no_evil:

gdostie (approved):
Don't forget to tag the commit as breaking

toofff (approved):
LGTM

PR #2991, opened 2022-12-15
feat(mantine): implement row multi selection on table component

gdostie:
### Proposed Changes

Implemented multi selection of table rows which enables the user to perform actions in bulk.

The user only needs to set the `multiRowSelectionEnabled` prop on his table component to see this new behaviour activated on his table, and adjust his Table.Actions render function to accept an array of selected rows.

None, existing tables with single selection of row shouldn't be affected by this change.

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/table/Table.tsx
+++ b/packages/mantine/src/components/table/Table.tsx
@@ -10,7 +10,7 @@ import {
     TableState,
     useReactTable,
 } from '@tanstack/react-table';
-import {InitialTableState} from '@tanstack/table-core';
+import {CoreOptions, InitialTableState} from '@tanstack/table-core';
 import defaultsDeep from 'lodash.defaultsdeep';
 import {Children, Fragment, ReactElement, ReactNode, useCallback, useEffect, useState} from 'react';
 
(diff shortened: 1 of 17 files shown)


aiheon (approved):
lgtm!

packages/mantine/src/components/table/Table.tsx
```diff
@@ -24,58 +24,74 @@ import {TableHeader} from './TableHeader';
...
 
 interface TableProps<T> {
     /**
      * Data to display in the table
      */
     data: T[];
+    /**
+     * Defines how each row is uniquely identified. It is highly recommended that you specify this prop to an ID that makes sense.
+     */
+    getRowId?: CoreOptions<T>['getRowId'];
```
sallainCoveo, line 94: I was wondering, what happen when you do set this prop with the multiselect? Do we have a default return value (or we assume that if you enable the multi select, it should be mandatory)?

sallainCoveo (approved):
Demo is pretty nice and code looks good!

packages/mantine/src/components/table/Table.tsx
gdostie: > If you need to identify individual rows that are originating from any server-side operations, it's suggested you use this function to return an ID that makes sense regardless of network IO/ambiguity eg. a userId, taskId, database ID field, etc.

https://tanstack.com/table/v8/docs/api/core/table#getrowid

PR #2998, opened 2022-12-19
fix(style): changed mantine textinput error colors and added styles t…

FelixBlaisThon:
…o collection disabled items

### Proposed Changes

`DisabledCollectionItem` now uses the useStyles and takes the `styles` object to add custom style to it (in my case, a background color)

I have also changed the error colours on the TextInput component from mantine so it reflects our colour 

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/theme/Theme.tsx
+++ b/packages/mantine/src/theme/Theme.tsx
@@ -106,12 +106,6 @@ export const plasmaTheme: MantineThemeOverride = {
             defaultProps: {
                 radius: 8,
             },
-            styles: (theme) => ({
-                description: {
-                    fontSize: 'inherit',
-                    paddingBottom: theme.spacing.xs,
-                },
-            }),
         },
         Tooltip: {
             defaultProps: {
--- a/packages/mantine/src/theme/Theme.tsx
+++ b/packages/mantine/src/theme/Theme.tsx
@@ -99,6 +99,19 @@ export const plasmaTheme: MantineThemeOverride = {
             defaultProps: {
                 radius: 8,
             },
+            styles: (theme) => ({
+                description: {
+                    fontSize: 'inherit',
+                    paddingBottom: theme.spacing.xs,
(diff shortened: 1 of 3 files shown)


packages/mantine/src/theme/Theme.tsx
```diff
@@ -99,6 +99,19 @@ export const plasmaTheme: MantineThemeOverride = {
...
+                    fontSize: 'inherit',
+                    paddingBottom: theme.spacing.xs,
+                },
+                invalid: {
+                    color: theme.colors.red[9],
+                    borderColor: theme.colors.red[6],
+                },
+                error: {
+                    color: theme.colors.red[9],
+                },
```
toofff, line 113: Surely this style should also be applied to other form elements? I'm thinking of the Select for example, I'm certainly forgetting others.

FelixBlaisThon: I can add the same style logic to the Select component since it's still in the scope. If we find any other components later, we can just change the style the same way

gdostie: Have a look at InputWrapper, its the thing that renders the errors for every form component. If you modify its style it will apply to each of them

packages/mantine/src/theme/Theme.tsx
```diff
@@ -93,12 +93,25 @@ export const plasmaTheme: MantineThemeOverride = {
...
         TextInput: {
             defaultProps: {
                 radius: 8,
             },
+            styles: (theme) => ({
+                description: {
+                    fontSize: 'inherit',
+                    paddingBottom: theme.spacing.xs,
+                },
+            }),
```
toofff, line 114: This change should not be in `InputWrapper` component to have consistency across all form components.

gdostie (approved):
IMHO it looked better how mantine did it, but if this is what UX wants I'm onboard

PR #3001, opened 2022-12-20
fix(plasma-mantine): fix ModalWizard doclink tooltip shows a black icon on hover

JTangCoveo:
### Proposed Changes

[MLX-307](https://coveord.atlassian.net/browse/MLX-307)

Screen Shot 2022-12-20 at 11 34 34 AM

- fix ModalWizard doclink tooltip shows a black icon on hover by passing in `docLinkTooltipLabel`
- Update demo page for ModalWizard
- Make `docLinkTooltipLabel` a required prop in Header component

### Acceptance Criteria

[MLX-307]: https://coveord.atlassian.net/browse/MLX-307?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/mantine/src/components/header/Header.tsx
```diff
@@ -23,7 +23,7 @@ export interface HeaderProps extends DefaultProps {
     /**
      * The tooltip text shown when hovering over the doc link icon
      */
-    docLinkTooltipLabel?: string;
+    docLinkTooltipLabel: string;
```
gdostie, line 26: This is a breaking change. Also, not all headers have a doc link, so making this prop required wouldn't be appropriate for everyone

packages/mantine/src/components/modal-wizard/ModalWizardStep.tsx
```diff
@@ -21,7 +21,7 @@ export interface ModalWizardStepProps {
     /**
      * A tooltip label for the docLink
      */
-    docLinkTooltipLabel?: string | ReactElement | DependsOnStep<string | ReactElement>;
+    docLinkTooltipLabel: string | ReactElement | DependsOnStep<string | ReactElement>;
```
gdostie, line 24: Same reasoning here, will all ModalWizard usages have a doc link? I am not convinced

packages/mantine/src/components/header/Header.tsx
JTangCoveo: You are right this would be a breaking change for any code that uses Header component, the reason why I introduced this here is because in Tooltip, the `label` prop is required, and if nothing is passed, it will show a black icon. I reverted my change.

packages/mantine/src/components/modal-wizard/ModalWizardStep.tsx
JTangCoveo: I think there are two ways to solve this:

- Make the `docLinkTooltipLabel` required in ModalWizard to avoid the black icon ever show up in case people forget about passing in this prop when using ModalWizard
- Keep `docLinkTooltipLabel` optional

I am currently going with option 2, let me know what you think?

gdostie: The way I see it is that the issue of the black icon should be fixed, even if no tooltip label is provided

JTangCoveo: I did some tests locally and it seems that the black icon will show up even if null/ empty is passed to the mantine tooltip component

JTangCoveo: So what will be something that I can do from my end?

aiheon: It's not a black icon, it's the tooltip rendering without any text in it 🙈 

You can make the rendering conditional depending if a label is provided or not I think 👍

aiheon: It's purple not black :p 

image

gdostie: Can't we modify the Header component implementation to avoid the issue? aka not rendering a Tooltip component if the label is empty.

JTangCoveo: I think this makes sense, no need to render a component if it is missing some information, in this case, no need to show Tooltip if the tooltip does not have a label. I will test it out on the header component.

JTangCoveo: > It's not a black icon, it's the tooltip rendering without any text in it 🙈
> 
> You can make the rendering conditional depending if a label is provided or not I think 👍

The Tooltip component requires a label, so I cannot conditionally the label 🤔

gdostie: Have you tried the `disabled` prop? 😉 

image

JTangCoveo: I decided to use `disabled` prop and it seems to be working well.

JTangCoveo: > Can't we modify the Header component implementation to avoid the issue? aka not rendering a Tooltip component if the label is empty.

I tested this and it also works. so between these two approaches, I prefer to not render `Tooltip` at all when the label is missing, it does not make sense to render something in UI where it does not have a purpose, let me know what you think!

gdostie (commented):
I don't see any change to the Header component in the diff, expected?

gdostie (approved):

PR #3013, opened 2023-01-18
feat(react-icons): add forwardRef on react icons

GermainBergeron:
### Proposed Changes

Changed the script that generates the icons to add the forwardRef. It's now possible to use multiple Mantine components like the Tooltip with the react-icons

image

The inferred interface of the icon changed but I don't think we need a breaking change for that


gdostie (commented):
We probably should adjust the `Icon` type here as well

https://github.com/coveo/plasma/blob/master/packages/react-icons/src/index.ts#L5

gdostie (approved):

PR #3022, opened 2023-01-30
fix(mantine): simplify disabled tooltip

gdostie:
### Proposed Changes

I had an issue with the way the disabled tooltip mechanism was implemented when using it in conjunction with a menu item + the inline confirm button.

For example
```tsx
<Menu.Item
    component={InlineConfirm.Button}
    id={InlineConfirmId}
    disabled={tooManyRules}
    disabledTooltip={Locales.format('table.header.action.delete.disabled')}
>
    {Locales.format('table.header.action.delete')}
</Menu.Item>
```

The `component={InlineConfirm.Button}` prop would overwrite the `component={MantineMenu.Item}` from the Menu.Item component which would render something like a button instead of a menu item.

I switched the ButtonWithDisabledTooltip implementation to just wrap the button with a div instead of adding all those data attributes. 

Anyways, long story short now it works and displays well.

image

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/button/Button.tsx
+++ b/packages/mantine/src/components/button/Button.tsx
@@ -1,4 +1,4 @@
-import {Button as MantineButton, ButtonProps as MantineButtonProps, useMantineTheme} from '@mantine/core';
+import {Button as MantineButton, ButtonProps as MantineButtonProps} from '@mantine/core';
 import {forwardRef} from 'react';
 
 import {createPolymorphicComponent} from '../../utils';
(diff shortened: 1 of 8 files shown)


packages/mantine/src/components/button/Button.tsx
```diff
@@ -1,27 +1,21 @@
...
+        >
+            <MantineButton ref={ref} disabled={disabled} {...others} />
+        </ButtonWithDisabledTooltip>
+    )
+);
 
-export const Button = createPolymorphicComponent<
-    'button',
-    ButtonProps,
-    {Group: typeof MantineButton.Group; DisabledTooltip: typeof ButtonWithDisabledTooltip}
```
gdostie, line 24: I figured the button with disabled tooltip component shouldn't be exported. We should keep it for internal use only since its hard to understand what it actually is for.

packages/mantine/src/components/button/ButtonWithDisabledTooltip.tsx
```diff
@@ -1,58 +1,37 @@
...
-     * @default 'unset'
+     * Whether the button underneath the tooltip is disabled
+     */
+    disabled?: boolean;
+    children?: ReactNode;
+    /**
+     * Additional tooltip props to set on the disabled button tooltip
      */
-    disabledHoverColor?: Property.BackgroundColor;
+    disabledTooltipProps?: Omit<TooltipProps, 'disabled' | 'label' | 'children'>;
```
gdostie, line 19: This is useful to set a specific width on long tooltips so that they wrap on multiple lines. I had a few cases where I had to put this in the admin:

`disabledTooltipProps={{width: 300}}`

FelixBlaisThon (approved):

FelixBlaisThon:
This is *sublime*

PR #3037, opened 2023-02-14
fix(mantine): hide table on unfiltered empty state

gdostie:
### Proposed Changes

When a table has no data and is unfiltered, show the `noDataChildren` but hide the header and footer. If the table is filtered, either by a regular filter, a predicate or a date range, show the `noDataChildren` with the header and footer.

Added a demo for it in the table section.

None

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/table/Table.tsx
+++ b/packages/mantine/src/components/table/Table.tsx
@@ -223,6 +223,11 @@ export const Table: TableType = <T,>({
         onStateChange: setState,
     }));
     const {clearSelection, getSelectedRow, getSelectedRows} = useRowSelection(table);
+    const isFiltered =
+        !!state.globalFilter ||
+        Object.keys(form.values?.predicates ?? {}).some((predicate) => !!form.values.predicates[predicate]) ||
+        !!form.values.dateRange?.[0] ||
+        !!form.values.dateRange?.[1];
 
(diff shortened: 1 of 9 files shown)


packages/mantine/src/components/table/Table.tsx
```diff
@@ -223,6 +223,11 @@ export const Table: TableType = <T,>({
         onStateChange: setState,
     }));
     const {clearSelection, getSelectedRow, getSelectedRows} = useRowSelection(table);
+    const isFiltered =
+        !!state.globalFilter ||
+        Object.keys(form.values?.predicates ?? {}).some((predicate) => !!form.values.predicates[predicate]) ||
+        !!form.values.dateRange?.[0] ||
+        !!form.values.dateRange?.[1];
```
GermainBergeron, line 230: Should we compare the values (predicates and dateRange) to the initialState? Otherwise tables with 'all' as a predicate would always be filtered, same for table with an initial date range

gdostie: Currently, if the predicate value for "All" is falsy it will work just fine. The problem with comparing with the initial state is that the initial state might be itself in a filtered state. For example a non null date range is a filtered state so a blankslate within the table makes more sense than total empty state, otherwise the date picker disappears and you cannot change your initial range to get some values.

GermainBergeron (approved):

PR #3039, opened 2023-02-14
fix(table): +debounce on triggerChange method

FelixBlaisThon:
### Proposed Changes

https://coveord.atlassian.net/browse/UITOOL-1121

+debounce on the triggerChange method so the table does not trigger a change on every change on the filter, sort, pagination, perpage and forms (datepicker / predicate) of the mantine table

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/table/Table.tsx
+++ b/packages/mantine/src/components/table/Table.tsx
@@ -225,23 +225,20 @@ export const Table: TableType = <T,>({
     }));
     const {clearSelection, getSelectedRow, getSelectedRows} = useRowSelection(table);
 
-    const triggerChange = () => onChange?.({...state, ...form.values});
+    const triggerChange = () => debounce(() => onChange?.({...state, ...form.values}), 500);
 
     useEffect(() => {
         onMount?.({...state, ...form.values});
+        return () => {
+            triggerChange().cancel();
+        };
     }, []);
 
     useDidUpdate(() => {
-        const debounceTable = debounce(() => {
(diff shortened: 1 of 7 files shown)


packages/mantine/src/components/table/Table.tsx
```diff
@@ -231,11 +232,17 @@ export const Table: TableType = <T,>({
...
-        if (!multiRowSelectionEnabled) {
-            clearSelection();
-        }
-    }, [state.globalFilter, state.sorting, state.pagination, form.values]);
+        const debounceTable = debounce(() => {
+            triggerChange();
+            if (!multiRowSelectionEnabled) {
+                clearSelection();
+            }
+        }, 500);
```
GermainBergeron, line 240: I'm wondering if that causes issue because the debounced method is overridden each time anything changes in the state (maybe my solution is not better)

I was thinking of something like:

```typescript
const triggerChange = debounce(() => onChange?.({...state, ...form.values}), 500);

useDidUpdate(() => {
    triggerChange();
    if (!multiRowSelectionEnabled) {
        clearSelection();
    }
}, [state.globalFilter, state.sorting, state.pagination, form.values]);

useEffect(() => {
    return () => triggerChange.cancel();
}, []);
```

FelixBlaisThon: I've put the cancel in the useEffect on line `230` for less code duplication (which has the same dependency array as you suggested)

GermainBergeron (approved):

PR #3047, opened 2023-02-22
Allow Mantine table to have an initial row selection

gdostie:
### Proposed Changes

Fix the Mantine table so that it supports having rows initially selected via the `initialState` prop.

**I added an example for initially selected rows on the website, be sure to check it out!**

I had to do some type hacking to the `rowSelection` state to make it work. Normally the tanstack table `rowSelection` state is of type `Record<string, boolean>`. My fix is to highjack that state so that it becomes `Record<string, TData>`. It works fine as long as TData is not `falsy` which never occurs in reality.

The type of `initialState.rowSelection` is changed from `Record<string, boolean>` to `Record<string, TData>` which is what we would expect anyway and fixes a lot of problems, so I am not planning to mark the commit as a breaking change.

### Acceptance Criteria

### DIFF
--- a/packages/mantine/package.json
+++ b/packages/mantine/package.json
@@ -29,6 +29,7 @@
         "@tanstack/react-table": "8.7.9",
         "@tanstack/table-core": "8.7.9",
         "dayjs": "1.11.3",
+        "fast-deep-equal": "3.1.3",
         "lodash.debounce": "4.0.8",
         "lodash.defaultsdeep": "4.6.1",
         "monaco-editor": "0.34.0",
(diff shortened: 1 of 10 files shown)


GermainBergeron:
I'm not sure I understand why you need to override the rowSelection, maybe I need to play with the actual code a bit to understand

Wouldn't it be more difficult for the dev if they know that some id is selected but they don't know the content of the selected object because it is not yet fetched? Did you find the reason why it was throwing in the first place?

gdostie:
> I'm not sure I understand why you need to override the rowSelection, maybe I need to play with the actual code a bit to understand
> 
> Wouldn't it be more difficult for the dev if they know that some id is selected but they don't know the content of the selected object because it is not yet fetched? Did you find the reason why it was throwing in the first place?

We need to override the row selection, because the table row actions need to have the entire data of each selected rows in order to determine if they can render. Otherwise we would have to fetch the selected ids each time the row selection changes, which is less than ideal. 

The other solution to avoid overriding the rowSelection state and still keeping the selected row data is to have a separate state for it. This is the solution that is currently implemented before this fix, but it has issue with keeping the state in sync.

GermainBergeron (approved):

PR #3052, opened 2023-02-23
feat(mantine): add onRowSelectionChange prop to table

gdostie:
### Proposed Changes

Adding a prop on the table that allows to act whenever the row selection changes. The reason I am providing a new prop for this instead of just forcing people to reuse the onChange prop is that onChange only gets called when the table needs to refresh its data. Row selection has no effect on the data shown in the table, it just selects it.

None

### Acceptance Criteria


GermainBergeron (approved):
Really clean!

PR #3096, opened 2023-03-22
fix(mantine table): add loading state for no data children

olamothe:
### Proposed Changes

Add loading animation for `{noDataChildren}`

Let's assume this setup, and with a service that is slow to return data:
```typescript
<Table 
    noDataChildren={
          filtered ? <div>No results match your filter</div> : <div>There is no data in the backend</div>
    }
>[...]</Table>
```
1- End user types something in the filter, no results match in the service, we display "No results match your filter".

2- User clears input, a new request is sent to the backend.

3- We want to ideally display to display a loading animation while the request is getting executed, and not "There is no data in the backend".

https://coveord.atlassian.net/browse/KIT-2358


FelixBlaisThon (commented):
LGTM (would be nice to add a unit test to this)

FelixBlaisThon (approved):

olamothe:
@FelixBlaisThon : I can't merge the PR, since it requires authorized users. Could you do it for me ?

🙇 Thanks !

PR #3099, opened 2023-03-23
fix(datepicker): +maxDate prop to rangeDatePickerProps since the new …

FelixBlaisThon:
…type does include it

### Proposed Changes

maxDate is not included in the new CaledarBaseProps interface from mantine 6 (which is still needed for the table datepicker). Just a simple add-on on that prop

This will let us set a max date and disable everything after that date

image

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/date-range-picker/DateRangePickerInlineCalendar.tsx
+++ b/packages/mantine/src/components/date-range-picker/DateRangePickerInlineCalendar.tsx
@@ -37,7 +37,7 @@ export interface DateRangePickerInlineCalendarProps
     /**
      * Props for RangeCalendar displayed in the popover
      */
-    rangeCalendarProps?: Omit<
+    rangeCalendarProps?: {maxDate?: Date} & Omit<
         CalendarBaseProps,
         'value' | 'onChange' | 'isDateInRange' | 'isDateFirstInRange' | 'isDateLastInRange'
     >;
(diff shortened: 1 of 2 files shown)


elessardCoveo (approved):

packages/mantine/src/components/date-range-picker/DateRangePickerInlineCalendar.tsx
```diff
@@ -37,7 +37,7 @@ export interface DateRangePickerInlineCalendarProps
     /**
      * Props for RangeCalendar displayed in the popover
      */
-    rangeCalendarProps?: Omit<
+    rangeCalendarProps?: {maxDate?: Date} & Omit<
         CalendarBaseProps,
         'value' | 'onChange' | 'isDateInRange' | 'isDateFirstInRange' | 'isDateLastInRange'
     >;
```
GermainBergeron, line 43: Can we also use minDate? I'm thinking just adding one prop here will be a temporary fix, maybe `CalendarBaseProps ` is not the correct interface anymore and we should use `DatePickerBaseProps` instead?

dmgauthier: This type is becoming quite an eyesore. 
If we can find something else than `{customKey: type} & omit<interface, 'a' | 'bunch' | 'of | 'stuff'>` that would be nice. How many properties there is ?

dmgauthier (approved):

GermainBergeron (approved):

PR #3105, opened 2023-03-24
fix(modal): reverse to pixel in theme modal content custom size

FelixBlaisThon:
### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/theme/Theme.tsx
+++ b/packages/mantine/src/theme/Theme.tsx
@@ -1,6 +1,6 @@
 import {InfoSize24Px} from '@coveord/plasma-react-icons';
 import {color} from '@coveord/plasma-tokens';
-import {getSize, MantineThemeOverride, ModalProps, NotificationProps, rem} from '@mantine/core';
+import {getSize, MantineThemeOverride, ModalProps, NotificationProps} from '@mantine/core';
 
 import {PlasmaColors} from './PlasmaColors';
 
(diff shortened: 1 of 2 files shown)


packages/mantine/src/theme/Theme.tsx
```diff
@@ -78,12 +78,12 @@ export const plasmaTheme: MantineThemeOverride = {
             },
         },
         Modal: {
-            styles: (theme, {size, fullScreen}: ModalProps) => ({
+            styles: (theme, {fullScreen}: ModalProps, {size}) => ({
```
GermainBergeron, line 81: I think the second argument is of type `ModalRootStylesParams` and not `ModalProps`

https://github.com/mantinedev/mantine/blob/8ee2220e924faa094328ef7c6cddd5fb89f85ed5/src/mantine-core/src/Modal/ModalRoot/ModalRoot.styles.ts#L19

GermainBergeron (approved):

packages/mantine/src/theme/Theme.tsx
FelixBlaisThon: ModalRootStyleParams does not seem to be exported. should I not put a type then ? ModalProps extends fullScreen at some point (that's why I chose ModalProps in the first place)

GermainBergeron: You can omit the type then, it's better than the wrong type

PR #3139, opened 2023-04-12
feat(copy-to-clip): added copy to clipboard button

nathanlb:
### Proposed Changes

Added a component to copy content to the clipboard. The component comes in 3 variants:
- button
- text
- textarea

`None`

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/copyToClipboard/CopyToClipboard.tsx
+++ b/packages/mantine/src/components/copyToClipboard/CopyToClipboard.tsx
@@ -1,13 +1,7 @@
 import React from 'react';
 
 import {CheckSize24Px, CopySize24Px} from '@coveord/plasma-react-icons';
-import {TextInput, CopyButton, Tooltip, ActionIcon, createStyles, Textarea} from '@mantine/core';
-
-export enum CopyToClipboardVariant {
-    Button = 'button',
-    Text = 'text',
-    TextArea = 'textarea',
-}
+import {TextInput, CopyButton, Tooltip, ActionIcon, createStyles} from '@mantine/core';
 
 export interface CopyToClipboardProps {
     /**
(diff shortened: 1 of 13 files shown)


packages/mantine/src/components/copyToClipboard/CopyToClipboard.tsx
```diff
@@ -0,0 +1,77 @@
+import React from 'react';
+
+import {CheckSize24Px, CopySize24Px} from '@coveord/plasma-react-icons';
+import {TextInput, CopyButton, Tooltip, ActionIcon, createStyles, Textarea} from '@mantine/core';
+
+export enum CopyToClipboardVariant {
+    Button = 'button',
+    Text = 'text',
+    TextArea = 'textarea',
+}
```
gdostie, line 10: No need for an enum. Since the variant prop is already typed using a union of strings, it already provides type protection and autocompletion.

```suggestion
```

packages/mantine/src/components/copyToClipboard/CopyToClipboard.tsx
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * The value to be copied to the clipboard.
+     */
+    value: string;
+    /**
+     * The variant type to use when displaying the component.
+     *
+     * @default "button"
+     */
+    variant?: 'button' | 'text' | 'textarea';
```
gdostie, line 22: Although I can see some value in having text input and textarea variants pre-made like this, I am not sure if it is really worth the maintenance of it. I can already see someone needing a missing prop on the Textinput or textarea.

nathanlb: @gdostie what do you think of the compromise of only keeping a `withLabel` prop to display the value as well so that all of those occurrences can be consistent? If someone needs more props on their input or need to have it in a textArea they can do it themselves with the default "button" version of the component.

gdostie: Yup, lets go with that.

FelixBlaisThon (approved):
demo tested, :gg:

nathanlb:
Gracias! feel free to merge whenever. I don't have the rights to do it.

gdostie (approved):

PR #3141, opened 2023-04-12
feat(mantine, table): create a consumer semantic component

GermainBergeron:
### Proposed Changes

Create a new `Table.Consumer` component. This component should be used to wrap "renderless" components that triggers `onChange` on the table in some way. For instance, a component that automatically refresh every X seconds, or a component that listen for changes in the URL, etc

None

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/table/Table.tsx
+++ b/packages/mantine/src/components/table/Table.tsx
@@ -26,6 +26,7 @@ import {TableHeader} from './TableHeader';
 import {TablePagination} from './TablePagination';
 import {TablePerPage} from './TablePerPage';
 import {TablePredicate} from './TablePredicate';
+import {TableConsumer} from './TableConsumer';
 import {TableSelectableColumn} from './TableSelectableColumn';
 import {Th} from './Th';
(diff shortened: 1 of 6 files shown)


packages/mantine/src/components/table/Table.tsx
```diff
@@ -55,6 +56,7 @@ export const Table: TableType = <T,>({
     const convertedChildren = Children.toArray(children) as ReactElement[];
     const header = convertedChildren.find((child) => child.type === TableHeader);
     const footer = convertedChildren.find((child) => child.type === TableFooter);
+    const refresher = convertedChildren.find((child) => child.type === TableRefresher);
```
gdostie, line 59: I would just change the naming a bit to something more generic, like `Table.WithinContext`, `Table.ContextConsumer`, `Table.Children`, or any other idea that explains that its for components that need the table context.

gdostie (approved):

packages/mantine/src/components/table/Table.tsx
GermainBergeron: I went with Table.Consumer

PR #3234, opened 2023-04-27
feat(npm): improve esm & cjs compatibility

gdostie:
### Proposed Changes

Distribute all packages in 2 outputs (CJS & ESM) properly.

<details>
<summary><h3>Detailed explanation<h3> <h6>&nbsp;&nbsp;&nbsp;&nbsp;⬅️ interesting stuff here ( ͡° ͜ʖ ͡°)<h6></summary>

#### distributed files

Each package is now built using the shared `/scripts/build.js` script. So all `dist` folders follow the same shape:

image

Crunchy bits:

- TypeScript declarations are adjacent to their respective `.js` file. Doing this removes the need to provide a "types" field in the `package.json` to the location of the declarations.
- TypeScript declarations are generated twice: once for the `cjs` output and another time for the `esm` output. This is what the [official TypeScript documentation about ESM](https://www.typescriptlang.org/docs/handbook/esm-node.html) recommends:
  > It’s important to note that the CommonJS entrypoint and the ES module entrypoint each needs its own declaration file, even if the contents are the same between them.
- A nested `package.json` file is added to each output. This is used to specify the `"type": "module"` and `"type": "commonjs"` of each output. Doing this removes the need to either add the `.mjs` extension to every file of the esm distribution, or `.cjs` extension to every file of the commonjs distribution.

#### package.json files

image

- All packages now all have the `exports` fields, which is used by recent versions of node (14 and up) to replace "main" and "module" fields.
- "main" field is kept for compatibility with TypeScript's `"moduleResolution": "node"` which doesn't understand "exports" field.
image

</details>

None

### Acceptance Criteria


packages/react/package.json
```diff
@@ -56,6 +54,7 @@
         "d3-scale": "4.0.2",
         "d3-shape": "3.2.0",
         "dnd-core": "16.0.1",
+        "escape-string-regexp": "4.0.0",
```
GermainBergeron, line 57:
```suggestion
        "escape-string-regexp": "5.0.0",
```

GermainBergeron (approved):
Did you try to link with the admin to see if everything still resolves fine?

packages/react/package.json
gdostie: This is a deliberate change, not an error :P 

v5 is ESM only: https://github.com/sindresorhus/escape-string-regexp/releases/tag/v5.0.0

gdostie:
> Did you try to link with the admin to see if everything still resolves fine?

I did, and it seems to build somewhat fine. It fails on the type-check with some errors of type `The inferred type of 'Foo' cannot be named without a reference to 'Bar'`. This always happens to me when linking so I imagine things are ok 🤷

packages/react/package.json
GermainBergeron: It does work currently though, no?

gdostie: It works only because we are using `"moduleResolution": "node"` and this resolver doesn't properly understand the `"type": "module"` field of dependencies. As soon as we switch to something else like `node16`, we get the following error when building the declarations for the `cjs` output. 

```
src/components/partial-string-match/PartialStringMatch.tsx:1:32 - error TS1479: The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("escape-string-regexp")' call instead.
  To convert this file to an ECMAScript module, add the field `"type": "module"` to '/Users/gdostie/Developer/coveo/plasma/packages/react/package.json'.

1 import escapeStringRegexp from 'escape-string-regexp';
```

gdostie: There are 2 ways to be safe with v5:
- Stop distributing commonjs output (be an ESM only package ourselves)
- Somehow install the 2 versions (v4 and v5) and build cjs with v4 and esm with v5

PR #3253, opened 2023-05-05
UI tool 8702 modal header separator line

FelixBlaisThon:
### Proposed Changes

Changed the theme style from the Modal component so that it adds a horizontal line separator between the header and the body. Also, I've created a `prompt` variant for the Modal component, because we dont want prompt to have that separator.

### Acceptance Criteria

### DIFF
--- a/packages/components-props-analyzer/src/ComponentsList.ts
+++ b/packages/components-props-analyzer/src/ComponentsList.ts
@@ -318,6 +318,10 @@ const components: Component[] = [
         packageName: '@coveord/plasma-mantine',
         propsType: 'CopyToClipboardProps',
     },
+    {
+        name: 'Prompt',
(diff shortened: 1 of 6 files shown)


GermainBergeron (commented):
Code and demo looks good for the modal, but can you add an example of the prompt?

FelixBlaisThon:
> Code and demo looks good for the modal, but can you add an example of the prompt?

I purposely did not, since the Prompt is not updated from the design Gustavo made 🙁, but I can add one if we feel like it would be needed

GermainBergeron:
It's just weird to review a component that I cannot see. I think we should still have a demo which would help us and Gustavo when we update the style of the component

GermainBergeron (approved):

PR #3272, opened 2023-05-15
fix(mantine, table): fix a few bugs when table is disabled by fieldset

agong-coveo:
### Proposed Changes
This PR tackles a few bugs that arise when a `Table` component is wrapped inside a `fieldset`. Essentially:
* If a table's rows contain checkboxes, despite being disabled, the checkboxes are still toggable when a user clicks on the row itself:

https://github.com/coveo/plasma/assets/105073504/18479ed5-e014-4c6e-96d0-5edff75c7a86

* Also, after the changes made in this [PR](https://github.com/mantinedev/mantine/pull/4152) in Mantine, when bumping the mantine dependecies to the latest, the collapsible buttons are now disabled and its content becomes inaccessible. This is a bug since we still want users to be able to access the collapsed content.

Screenshot 2023-05-17 at 11 43 30 AM

The changes made to fix the bugs are the following:
* Modified the `onClick` method of the rows to take into account if its checkbox is disabled or not. If it is, then `onClick` is undefined;
* Modified the disabled style of the collapsible button to be clickable and to look enabled.
 
**NOTE**: Currently, there is also a bug with the Pagination and the PerPage component that is in a `Table`'s footer. More precisely, these components are disabled and unclickable. This is an issue because users should still be able to view the items inside a table and be able to manipulate these components. This [task](https://coveord.atlassian.net/browse/CTCORE-9247) was created to tackle it.

None

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/table/Table.styles.ts
+++ b/packages/mantine/src/components/table/Table.styles.ts
@@ -54,17 +54,15 @@ const useStyles = createStyles<string, TableStylesParams>((theme, {multiRowSelec
             textAlign: 'right',
             padding: `calc(${theme.spacing.xs}/2) ${theme.spacing.sm} !important`,
 
-            '&:has(button:disabled)': {
-                button: {
-                    pointerEvents: 'auto',
-                    cursor: 'pointer',
-                    backgroundColor: 'transparent',
-                    borderColor: 'transparent',
-                    color: `${theme.colors.gray[6]}`,
+            '& button:disabled': {
+                pointerEvents: 'auto',
+                cursor: 'pointer',
+                backgroundColor: 'transparent',
+                borderColor: 'transparent',
+                color: `${theme.colors.gray[6]}`,
 
-                    '&:hover': {
-                        backgroundColor: `${theme.colors.gray[0]}`,
-                    },
+                '&:hover': {
+                    backgroundColor: `${theme.colors.gray[0]}`,
                 },
             },
         },
--- a/packages/mantine/src/components/table/Table.styles.ts
+++ b/packages/mantine/src/components/table/Table.styles.ts
@@ -53,6 +53,20 @@ const useStyles = createStyles<string, TableStylesParams>((theme, {multiRowSelec
         rowCollapsibleButtonCell: {
             textAlign: 'right',
             padding: `calc(${theme.spacing.xs}/2) ${theme.spacing.sm} !important`,
+
+            '&:has(button:disabled)': {
+                button: {
+                    pointerEvents: 'auto',
+                    cursor: 'pointer',
+                    backgroundColor: 'transparent',
+                    borderColor: 'transparent',
+                    color: `${theme.colors.gray[6]}`,
+
+                    '&:hover': {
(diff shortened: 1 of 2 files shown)


packages/mantine/src/components/table/Table.styles.ts
```diff
@@ -53,6 +53,20 @@ const useStyles = createStyles<string, TableStylesParams>((theme, {multiRowSelec
...
+                    cursor: 'pointer',
+                    backgroundColor: 'transparent',
+                    borderColor: 'transparent',
+                    color: `${theme.colors.gray[6]}`,
+
+                    '&:hover': {
+                        backgroundColor: `${theme.colors.gray[0]}`,
+                    },
+                },
+            },
```
gdostie, line 69:
```suggestion
            '& button:disabled': {
                pointerEvents: 'auto',
                cursor: 'pointer',
                backgroundColor: 'transparent',
                borderColor: 'transparent',
                color: `${theme.colors.gray[6]}`,

                '&:hover': {
                    backgroundColor: `${theme.colors.gray[0]}`,
                },
            },
```

packages/mantine/src/components/table/Table.tsx
```diff
@@ -128,7 +130,12 @@ export const Table: TableType = <T,>({
         return (
             <Fragment key={row.id}>
                 <tr
-                    onClick={() => row.toggleSelected()}
+                    ref={rowRef}
+                    onClick={() =>
+                        rowRef.current.querySelectorAll('.mantine-Table-root input[type=checkbox]:disabled').length > 0
+                            ? undefined
+                            : row.toggleSelected()
+                    }
```
gdostie, line 138: Row selection should still be enabled even when inside a disabled fieldset in some cases. Some row actions are still valid even in read-only mode. 

For example, we often have a table row action that opens another panel to view the details of the resource represented by the row. Even if the table is disabled you still want to be able to view the resource's details.

gdostie:
Another concern of mine is sorting, what happens if you want to sort a table that is inside a disabled fieldset. Tables are more often than not presentational components rather than form control.

packages/mantine/src/components/table/Table.tsx
agong-coveo: I see! From the examples in [tanstack](https://tanstack.com/table/v8/docs/api/features/row-selection), I seem to have understood that row selection was for selecting and toggling the checkboxes associated to a row.

I'm not sure I understand the example completely. Are these details hidden inside a collapsible? Something like in the Sources panel?
image
If that's the case, if I'm not mistaken, this content becomes visible only when the user clicks on the arrow button and not on the row itself. The change here shouldn't affect access to the resource's details.

gdostie: You have an example of what I am referring to in your screenshot. In the table actions, we can see "Rescan", "Edit", "Activity", "... More". Those buttons still need to be activated even inside a disabled `fieldset`. Granted, not all of them, but the "Activity" one does, and the "... More" as well.

agong-coveo:
Closed because the description of the task has changed.

PR #3284, opened 2023-05-17
fix: add word break for header title

msrioux-coveo:
### Proposed Changes

This fix allows the word-break styling to visualize better long field names when a user wants to edit a Field.
(Fixes third bullet point in : https://coveord.atlassian.net/browse/UITOOL-8758)

Located in `Content -> Fields`. Click on a long named Field and then click on "edit". You should see the window in the screenshots below.

**Before :** 
Screen Shot 2023-05-17 at 2 31 35 PM

**After :** 
long Fields name break-word

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/header/Header.tsx
+++ b/packages/mantine/src/components/header/Header.tsx
@@ -51,7 +51,7 @@ export const Header: HeaderType = ({description, borderBottom, children, variant
                     <Title
                         order={variant === 'page' ? 1 : 3}
                         color={variant === 'page' ? 'gray.5' : undefined}
-                        style={{wordBreak: 'break-all'}}
+                        sx={{wordBreak: 'break-word'}}
                     >
                         {otherChildren}
                         {docAnchor}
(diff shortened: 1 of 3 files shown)


packages/mantine/src/components/header/Header.tsx
```diff
@@ -48,7 +48,11 @@ export const Header: HeaderType = ({description, borderBottom, children, variant
             >
                 <Stack spacing={0}>
                     {breadcrumbs}
-                    <Title order={variant === 'page' ? 1 : 3} color={variant === 'page' ? 'gray.5' : undefined}>
+                    <Title
+                        order={variant === 'page' ? 1 : 3}
+                        color={variant === 'page' ? 'gray.5' : undefined}
+                        style={{wordBreak: 'break-all'}}
```
gdostie, line 54:
```suggestion
                        sx={{ wordBreak: "break-word" }}
```

I think `break-word` would be more appropriate than `break-all`. It would avoid breaking actual words in two whenever we have the space for it while still breaking words that are too long.

Break all
image

Break word
image

msrioux-coveo: I see!! I will make the changes

gdostie (approved):

PR #3287, opened 2023-05-19
feat(mantine, table): add disableRowSelection property

agong-coveo:
### Proposed Changes
* Added a `disableRowSelection` property to prevent row selection.

Essentially, enabling this property will **disable row selection** and **prevent clicks on the checkboxes** present on the table's rows. The purpose of this property is to be able to present the data in the table but to disallow users from modifying the rows that have been selected.

This property could be used for a table that's part of a form and the selected row are saved in it. When the form is disabled, we do not want users to be able to change the selected row in the table. In cases like these, we want to retrieve and save the information from the table, and not perform actions on its items. Hence, table and row actions are not relevant here.

None

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/table/Table.styles.ts
+++ b/packages/mantine/src/components/table/Table.styles.ts
@@ -2,9 +2,10 @@ import {createStyles} from '@mantine/core';
 
 interface TableStylesParams {
     multiRowSelectionEnabled: boolean;
+    disableRowSelection: boolean;
 }
 
-const useStyles = createStyles<string, TableStylesParams>((theme, {multiRowSelectionEnabled}) => {
+const useStyles = createStyles<string, TableStylesParams>((theme, {multiRowSelectionEnabled, disableRowSelection}) => {
     const rowBackgroundColor =
         theme.colorScheme === 'dark'
             ? theme.fn.rgba(theme.colors[theme.primaryColor][7], 0.2)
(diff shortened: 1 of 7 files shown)


toofff (commented):
Could you add an example to the Plasma demo website, please

packages/mantine/src/components/table/Table.types.ts
```diff
@@ -97,6 +97,11 @@ export type TableContextType<TData> = {
      * Whether multi row selection is activated
      */
     multiRowSelectionEnabled: boolean;
+    /**
+     * Whether row selection is enabled or not
+     *
```
toofff, line 102:
```suggestion
```

packages/mantine/src/components/table/__tests__/Table.spec.tsx
```diff
@@ -508,5 +508,78 @@ describe('Table', () => {
...
+                        data={[
+                            {id: '🆔-1', firstName: 'first', lastName: 'last'},
+                            {id: '🆔-2', firstName: 'patate', lastName: 'king'},
+                        ]}
+                        columns={columns}
+                        multiRowSelectionEnabled
+                        disableRowSelection
+                    />
+                </div>
+            );
```
toofff, line 529: Do you really need a header for your test?

```suggestion
            render(
                <Table
                    getRowId={({id}) => id}
                    data={[
                        {id: '🆔-1', firstName: 'first', lastName: 'last'},
                        {id: '🆔-2', firstName: 'patate', lastName: 'king'},
                    ]}
                    columns={columns}
                    multiRowSelectionEnabled
                    disableRowSelection
                />
            );
```

packages/mantine/src/components/table/TableHeader.tsx
```diff
@@ -45,7 +45,7 @@ export const TableHeader: FunctionComponent<TableHeaderProps> = ({
             className={classes.root}
             {...others}
         >
-            {multiRowSelectionEnabled && selectedRows.length > 0 ? (
+            {multiRowSelectionEnabled && !disableRowSelection && selectedRows.length > 0 ? (
```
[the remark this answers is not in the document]
agong-coveo: I removed the button because if it is clicked on, it modifies and clears the selected rows. In this case, should I simply remove the clear icon and prevent the onClick method?

packages/mantine/src/components/table/__tests__/Table.spec.tsx
agong-coveo: No, I will remove it

packages/mantine/src/components/table/TableHeader.tsx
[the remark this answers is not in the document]
toofff: It should be in this state and no action should be available.

image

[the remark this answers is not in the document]
agong-coveo: That makes sense, I'll add the necessary changes

dmgauthier (approved):

packages/mantine/src/components/table/Table.styles.ts
gdostie, line 1: The hover style is still there even when row selection is disabled, do you think it would make sense to remove it?

gdostie (approved):

packages/mantine/src/components/table/Table.styles.ts
agong-coveo: IMO, the hover event is is still relevant even with row selection being disabled, as it helps users to easily follow a row of data from one column to another. What do you think?

gdostie: I can get onboard with that 🚢

PR #3314, opened 2023-06-01
feat(mantine): enable table to have multiple layouts

gdostie:
### Proposed Changes

Enabling the `<Table>` component to offer multiple layouts to the user.

In this PR, I extracted the existing logic to render rows into a `RowLayout` and exposed a new `layouts` prop on the Table component. The default value of this prop is an array containing only the row layout, so this change is not breaking.

When providing more than one layout, a layout toggle appears and let the user switch between layouts.

https://github.com/coveo/plasma/assets/35579930/97506783-989e-4901-8c0d-9c45f1697f85

None

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/table/Table.tsx
+++ b/packages/mantine/src/components/table/Table.tsx
@@ -52,7 +52,7 @@ export const Table: TableType = <T,>({
         initialValues: {
             predicates: initialState?.predicates ?? {},
             dateRange: initialState?.dateRange ?? [null, null],
-            layout: layouts[0].name,
+            layout: initialState?.layout ?? layouts[0].name,
         },
     });
     const {classes} = useStyles();
@@ -207,3 +207,4 @@ Table.AccordionColumn = TableAccordionColumn;
 Table.DateRangePicker = TableDateRangePicker;
 Table.Consumer = TableConsumer;
 Table.Loading = TableLoading;
+Table.Layouts = TableLayouts;
--- a/packages/mantine/src/components/table/Table.tsx
+++ b/packages/mantine/src/components/table/Table.tsx
@@ -1,18 +1,10 @@
-import {Box, Center, Collapse, Loader, Table as MantineTable, Skeleton, SkeletonProps} from '@mantine/core';
+import {Box, Center, Loader, Table as MantineTable} from '@mantine/core';
 import {useForm} from '@mantine/form';
 import {useDidUpdate} from '@mantine/hooks';
-import {
-    ColumnDef,
-    Row,
-    TableState as TanstackTableState,
-    defaultColumnSizing,
-    flexRender,
-    getCoreRowModel,
-    useReactTable,
-} from '@tanstack/react-table';
+import {ColumnDef, Row, TableState as TanstackTableState, getCoreRowModel, useReactTable} from '@tanstack/react-table';
 import debounce from 'lodash.debounce';
 import defaultsDeep from 'lodash.defaultsdeep';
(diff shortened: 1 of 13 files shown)


packages/mantine/src/components/table/Table.tsx
```diff
@@ -255,3 +206,4 @@ Table.CollapsibleColumn = TableCollapsibleColumn;
 Table.AccordionColumn = TableAccordionColumn;
 Table.DateRangePicker = TableDateRangePicker;
 Table.Consumer = TableConsumer;
+Table.Loading = TableLoading;
```
GermainBergeron, line 209: I think we should also re-export the Layouts here, it would be nicer to import/use Table.Layouts

packages/mantine/src/components/table/Table.tsx
```diff
@@ -61,9 +49,13 @@ export const Table: TableType = <T,>({
 
     const {predicates, dateRange, ...initialStateWithoutForm} = initialState;
     const form = useForm<TableFormType>({
-        initialValues: {predicates: initialState?.predicates ?? {}, dateRange: initialState?.dateRange ?? [null, null]},
+        initialValues: {
+            predicates: initialState?.predicates ?? {},
+            dateRange: initialState?.dateRange ?? [null, null],
+            layout: layouts[0].name,
```
GermainBergeron, line 55: Just wondering if we should let the user set the layout in the initialState

GermainBergeron (approved):
Really nice!

packages/website/src/pages/layout/Table.tsx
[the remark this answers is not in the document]
gdostie: woops yes

packages/mantine/src/components/table/Table.tsx
gdostie: Enabled it, also added a UT for it.

packages/mantine/src/components/table/layouts/RowLayout.tsx
```diff
@@ -0,0 +1,150 @@
...
+    return <>{headers}</>;
+};
+
+const RowLayoutBody = <T,>({table, doubleClickAction, getExpandChildren, loading}: TableLayoutProps<T>) => {
+    const {multiRowSelectionEnabled, disableRowSelection} = useTable();
+    const {classes, cx} = useStyles({disableRowSelection, multiRowSelectionEnabled});
+
+    const rows = table.getRowModel().rows.map((row) => {
+        const rowChildren = getExpandChildren?.(row.original) ?? null;
+        const isSelected = !!row.getIsSelected();
```
dmgauthier, line 89: out of curiosity. This function does not return a boolean ?

packages/mantine/src/theme/Theme.tsx
```diff
@@ -276,12 +276,12 @@ export const plasmaTheme: MantineThemeOverride = {
                 },
             },
         },
-        Segmented: {
-            styles: {
-                control: {
-                    zIndex: 'unset',
```
dmgauthier, line 282: you sure about removing this one ?

dmgauthier:
:raised_hands: good game !

dmgauthier (approved):

packages/mantine/src/components/table/layouts/RowLayout.tsx
gdostie: It does, good catch. I mainly just copy pasted what we already had in the table component btw.

packages/mantine/src/theme/Theme.tsx
gdostie: `Segmented` doesn't exist, so it had no effect anyway.

PR #3341, opened 2023-06-16
feat(mantine, table): create a new LastUpdated component

GermainBergeron:
### Proposed Changes

Create a new component that displays the last time the table updated. It hints the user when the table automatically refreshes

None, new component

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/table/TableLastUpdated.tsx
+++ b/packages/mantine/src/components/table/TableLastUpdated.tsx
@@ -1,0 +1,51 @@
+import {createStyles, DefaultProps, Group, Selectors, Text} from '@mantine/core';
+import {useDidUpdate} from '@mantine/hooks';
+import dayjs from 'dayjs';
+import {FunctionComponent, useState} from 'react';
+import {useTable} from './TableContext';
+
+const useStyles = createStyles((theme) => ({
+    root: {
+        minHeight: '98px',
+    },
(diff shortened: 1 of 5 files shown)


packages/mantine/src/components/table/TableLastUpdated.tsx
```diff
@@ -0,0 +1,57 @@
...
+     *
+     * @default "Last update:"
+     */
+    label?: string;
+    /**
+     * Array of dependencies that triggers an update
+     *
+     * @default []
+     */
+    dependencies?: any[];
```
gdostie, line 30: In what use cases do we need to pass this prop?

GermainBergeron: If your table is auto refreshed e.g, with an interval or when you select a facet that is outside the table you need to update the LastUpdated somehow

gdostie: Can we always pass `[data]`?

GermainBergeron: Hmm I was thinking that for a client side table it is not necessary, but it wouldn't hurt. I'm just wondering how we would do it. I'll try something with cloneElement

gdostie (approved):

PR #3361, opened 2023-06-21
fix(table): header alignment

dmgauthier:
### Proposed Changes

text-align left in the table header

**Before the Fix**
image

**After the Fix**
image

none

### Acceptance Criteria


gdostie:
The fix looks good, but this has always bothered me about the table header style: is it expected that the background and foreground color isn't the same for headers that can be sorted and those that can't?

image
image

GermainBergeron:
> The fix looks good, but this has always bothered me about the table header style: is it expected that the background and foreground color isn't the same for headers that can be sorted and those that can't?
> 
> image image

+1, we should remove the bold and make it the same colour (sorry @dmgauthier)

aboissinot-coveo (approved):
LGTM. Tested on different resolutions and it looked fine ! 👏

dmgauthier:
I changed a bit the logic that manage the `th` width + uniformized the style between the sortable and the regular header.

GermainBergeron (approved):

PR #3379, opened 2023-07-03
fix(table): column width

dmgauthier:
### Proposed Changes

**Problem** 
By using the `getSize` hook to get the size of the column and applying that size to the width css property, we create a confusing situation where setting `minSize` or `maxSize` property on a table column will not respect what we would expect using proper css.

**Solution**
Getting the default sizing or the custom sizing if used, and set those as plain CSS.

**Before the fix**
image

**After the fix**
image
image

**Disclaimer**
The solution does not change the default column layout of the table, no change will be seen unless custom width has been set on table column width. But if a custom width has been used, the new code will apply that styling to the CSS and this might change the table layout.

### Acceptance Criteria


pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.0'
+lockfileVersion: '6.1'
```
dmgauthier, line 1: Not sure we want this...  I can revert

aboissinot-coveo (approved):
LGTM

PR #3431, opened 2023-07-20
feat(table, layouts): add prop to table layouts to disable row selection toggle

aboissinot-coveo:
#### Description

Add a `keepSelection ` prop in the `TableLayout` type. This will disable the toggle behaviour the selected rows, meaning that clicking on a selected row will not deselect it.

#### Use case

The immediate use case we have for it is in the new Content Browser's Permission Details view.

permissions_details

#### How to use it?

If you want to use it with the default `RowLayout`, you can do the following:

```
const MyTableComponent = () => {
    const data: DataType = [...]
    const columns = [...]

    const layouts: TableLayout[] = [
        {
            ...TableLayouts.Rows,
            Body: (props) => <Table.Layouts.Rows.Body {...props} keepSelection />,
        },
    ];


    return (
        <Table 
            data={data} 
            columns={columns} 
            layouts={layouts}
        />
)}
```

No breaking changes since omitting this prop will enforce the current behaviour of toggling the row selection on click. 

### 🧪 Demo

**Note**: It's a bit hard to notice the clicks through a screen capture so when the text becomes highlighted it's actually me spamming the click on the row).

https://github.com/coveo/plasma/assets/133259861/90027d93-c8e6-4101-ac62-b32e366359ba

### Acceptance Criteria


packages/mantine/src/components/table/Table.types.ts
```diff
@@ -76,6 +76,10 @@ export interface TableLayoutProps<T = unknown> {
      * @param datum the row for which the children should be generated.
      */
     getExpandChildren?: (datum: T) => ReactNode;
+    /**
+     * Wether clicking on a selected row will deselect the row or not.
+     */
+    forceRowSelected?: boolean;
```
GermainBergeron, line 82: I'm not sure about the name of the props, I think I'd prefer something like `disableRowDeselection`?

packages/mantine/src/components/table/layouts/RowLayout.tsx
```diff
@@ -108,11 +114,12 @@ const RowLayoutBody = <T,>({table, doubleClickAction, getExpandChildren, loading
     const rows = table.getRowModel().rows.map((row) => {
         const rowChildren = getExpandChildren?.(row.original) ?? null;
         const isSelected = !!row.getIsSelected();
+        const shouldForceRowSelect = forceRowSelected && row.getIsSelected();
         const onClick = (event: MouseEvent<HTMLTableRowElement>) => {
             if (rowChildren) {
                 toggleCollapsible(event.currentTarget);
             }
-            if (!disableRowSelection && !multiRowSelectionEnabled) {
+            if (!disableRowSelection && !multiRowSelectionEnabled && !shouldForceRowSelect) {
```
GermainBergeron, line 122: Won't that prevent use from selecting rows that aren't the initially selected one?

packages/mantine/src/components/table/layouts/__tests__/RowLayout.spec.tsx
GermainBergeron, line 1: I'd add a test that select another row

packages/mantine/src/components/table/Table.types.ts
aboissinot-coveo: Alright no problem, changed it. ✅

packages/mantine/src/components/table/layouts/__tests__/RowLayout.spec.tsx
aboissinot-coveo: Added the test. Let me know if this wasn't what you had in mind. 🧪

packages/mantine/src/components/table/layouts/RowLayout.tsx
aboissinot-coveo: So my thought process was that if the prop is passed to the layout and we click on a selected row, it won't toggle. But if the row isn't already selected, it will toggle and thus become selected.

Here's an example of how it's currently working in my ongoing admin-ui task (it's a bit hard to notice the clicks through a screen capture so when the text becomes highlighted it's actually me spamming the click on the row).

https://github.com/coveo/plasma/assets/133259861/0c9cb784-1194-4735-af07-d653439507e6

packages/mantine/src/components/table/Table.types.ts
```diff
@@ -76,6 +76,10 @@ export interface TableLayoutProps<T = unknown> {
      * @param datum the row for which the children should be generated.
      */
     getExpandChildren?: (datum: T) => ReactNode;
+    /**
+     * Wether clicking on a selected row will deselect the row or not.
```
GermainBergeron, line 80:
```suggestion
     * Whether clicking on a selected row will deselect the row or not.
```

GermainBergeron (approved):
Looking good!

packages/mantine/src/components/table/layouts/RowLayout.tsx
```diff
@@ -108,11 +114,12 @@ const RowLayoutBody = <T,>({table, doubleClickAction, getExpandChildren, loading
     const rows = table.getRowModel().rows.map((row) => {
         const rowChildren = getExpandChildren?.(row.original) ?? null;
         const isSelected = !!row.getIsSelected();
+        const shouldDisableRowDeselection = disableRowDeselection && row.getIsSelected();
```
gdostie, line 117:
```suggestion
        const shouldDisableRowDeselection = disableRowDeselection && isSelected;
```

packages/mantine/src/components/table/Table.types.ts
```diff
@@ -76,6 +76,10 @@ export interface TableLayoutProps<T = unknown> {
...
+     */
+    disableRowDeselection?: boolean;
```
gdostie, line 82: My small brain has troubles keeping up with the double negative. Can we instead go with something like this?

```suggestion
    keepSelection?: boolean;
```

gdostie (approved):

packages/mantine/src/components/table/Table.types.ts
aboissinot-coveo: Sure no problem. ✅

PR #3449, opened 2023-07-25
fix(table, onrowselectionchange): call onRowSelectionChange callback in hook

aboissinot-coveo:
### Proposed Changes

The `Table` exposes the callback `onRowSelectionChange` to allows us to act whenever the row selection changes on the table. 

Table.types.ts

However, the callback provided to the table was called within a `setState` call. So, if the callback queues an update in the parent component, it would cause a render conflict between the `Table` component and the parent component and React would print a warning in the console.

For example, doing something like below could cause the warning to be issued by React:

```
export const ParentComponent = () => {
    const [someState, setSomeState] = useState(null);

    const changeSomeState = (rows) => {
        setSomeState(rows);
    };

    return (
        <Table data=[data] columns={columns} onRowSelectionChange={changeSomeState}/>
    )
};
```

⚠️ See this [warning](https://legacy.reactjs.org/blog/2020/02/26/react-v16.13.0.html#warnings-for-some-updates-during-render) introduced in React v16.13.0. ⚠️ 

🎥 Here's a live example of the behaviour described above:

https://github.com/coveo/plasma/assets/133259861/a8d25676-fbd3-430e-856a-7f15097df272

🧪 Here's an example to show the new hooks is working as expected:

https://github.com/coveo/plasma/assets/133259861/274ef7e3-32fb-4af7-afd9-ea22f77c1a36

Shouldn't cause any breaking changes. The `onRowSelectionChange` callback will still receive the updated selections.

### Acceptance Criteria

    - There's already a unit test covering the `onRowSelectionChange` callback (see [Table.spec.tsx](https://github.com/coveo/plasma/blob/master/packages/mantine/src/components/table/__tests__/Table.spec.tsx#L270)).


gdostie (approved):
Nice find, thanks for the investigation 👌

GermainBergeron (approved):

PR #3468, opened 2023-07-28
fix(table,defaultcolumn): set column width to auto by default

aboissinot-coveo:
### Proposed Changes

A fix was done recently to be able to enforce a `min-width` and `max-width` on table cells => https://github.com/coveo/plasma/pull/3379.

This changed the default behaviour of the `width` property table cells. Instead of being `undefined`, it's being set as the default size coming from `@tanstack` which is `150px`. 

I think it could be interesting to put back the default behaviour for cell width to `auto`. That way it's possible to leverage automatic width if nothing is specified but it's still possible to use fixed sizing if you wish so.

Another thing to consider is that the column sizing typing from `@tanstack` enforces a `number` for `size`. So we can't set it to `auto` via the `ColumnDef` without silencing typescript errors.

#### Current

`width` is set the `150px` by default. Columns with long text might get squashed.

Screenshot 2023-07-28 at 8 28 46 AM

#### New

`width` is set to auto, `min-width` is still enforced.

Screenshot 2023-07-28 at 8 29 57 AM

Screenshot 2023-07-28 at 8 29 37 AM

#### Fixed size

`width` is set to `450px`, `min-width` is still enforced

Screenshot 2023-07-28 at 8 32 14 AM

Screenshot 2023-07-28 at 8 32 39 AM

No breaking changes.

- If no `size` is specified, same behaviour as before which `auto`.
- If `size` is specified, `width` is set to that size like we did before as well.

### Acceptance Criteria


gdostie (approved):
Works well 👍 good job

PR #3495, opened 2023-08-01
feat(codeeditor): add copy and search callbacks in code editor props

aboissinot-coveo:
## [ADUI-8908](https://coveord.atlassian.net/browse/ADUI-8908)

There's a need to gather more insights on the usage of some features included with the `monaco-editor`, more specifically the search and copy features.

I added the `onCopy` and `onSearch` props in the `CodeEditor` component so we can be notified when those features are used by someone in the `admin-ui`.

⚠️ **Limitation** ⚠️ 

The `onSearch` callback will only be called when someone clicks on the search icon. It won't be called when a user uses the `CTRL/CMD + F` keyboard shortcut. 

I think ideally we would hook ourselves directly in the `editor` instance to know when the `find` search bar is opened/closed. Sadly, the `monaco-editor` API is fairly limited and doesn't expose anything to allow us to do that. 

The only way I found to achieve this is via:

`editor.getContribution('editor.contrib.findController')._state.onFindReplaceStateChange`

But since this is not exposed by the API, I found this a bit risky to use on top of having to suppress typescript warnings. Overall, I didn't think it was worth it since it the only use case we have for it at the moment is Amplitude trackers.

No breaking changes. This PR only adds new props to a component. 

### Acceptance Criteria

[ADUI-8908]: https://coveord.atlassian.net/browse/ADUI-8908?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gdostie (approved):
I think tracking only the search icon is reasonable 👍

PR #3497, opened 2023-08-02
fix(table): break long words

dmgauthier:
### Proposed Changes

Long non breaked words are breaking the table layout. I added a rule in the built in row layout.
Tell me if you think this rule should be applied elsewhere instead.

workbreak-fix-not

workbreak-fix

I do not see any.

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/table/layouts/RowLayout.tsx
+++ b/packages/mantine/src/components/table/layouts/RowLayout.tsx
@@ -158,6 +158,7 @@ const RowLayoutBody = <T,>({
                                     width: columnSizing.size ?? 'auto',
                                     minWidth: columnSizing.minSize,
                                     maxWidth: columnSizing.maxSize,
+                                    overflowWrap: 'break-word',
                                 }}
                                 className={cx(classes.cell, {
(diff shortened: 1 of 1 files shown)


gdostie (approved):

lprovost-coveo (approved):

packages/mantine/src/components/table/layouts/RowLayout.tsx
```diff
@@ -158,6 +158,7 @@ const RowLayoutBody = <T,>({
                                     width: columnSizing.size ?? 'auto',
                                     minWidth: columnSizing.minSize,
                                     maxWidth: columnSizing.maxSize,
+                                    wordBreak: 'break-word',
```
GermainBergeron, line 161: Can you use `overflowWrap: 'break-word'` instead? It's the newer version of `wordBreak: break-word`

https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap

GermainBergeron (changes requested):

packages/mantine/src/components/table/layouts/RowLayout.tsx
dmgauthier: works flawlesly

PR #3554, opened 2023-09-25
fix(mantine): performance issue in CodeEditor

gdostie:
#### Problem

While debugging a performance issue using the browser's performance profiler, I found out that the `CodeEditor` component from `@coveord/plasma-mantine` was having issues rendering very large chunks of text.

image

As we can see it in this screenshot, it would spend over 16 seconds in the `discoverBreaks` function. So I did a bit of research on monaco editor and found out about the `wrappingStrategy` option:

> wrappingStrategy: 'advanced' means that we delegate the wrap points computations to the browser. This works by us creating a dom node for each line of the editor, setting the contents of each dom node to the text on each line, and then setting a specific width. We then proceed to use browser APIs to detect the wrapping points. This is super slow (causes a forced browser layout). - https://github.com/microsoft/vscode/issues/132220#issuecomment-927856270

From my understanding, with this strategy, one dom node is created for each line of the editor. When the editor content is so large that it wraps over thousands of lines, it generates too many DOM nodes for the browser to manage efficiently.

#### Solution

The solution is actually a very simple one liner: use the default wrapping strategy called `simple`. 

None

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/code-editor/CodeEditor.tsx
+++ b/packages/mantine/src/components/code-editor/CodeEditor.tsx
@@ -188,7 +188,6 @@ export const CodeEditor: FunctionComponent<CodeEditorProps> = (props) => {
                 options={{
                     minimap: {enabled: false},
                     wordWrap: 'on',
-                    wrappingStrategy: 'advanced',
                     scrollBeyondLastLine: false,
                     formatOnPaste: true,
                     fontSize: px(theme.fontSizes.xs),


aboissinot-coveo (approved):
LGTM. Tested it and it looks the behave the same way. Still, would be curious to know why we had this in the first place considering the performance trade-off.

Anyway,  good job on the investigation and the detailed description. 🏆

PR #3580, opened 2023-10-30
feat(react, actions): add onClickCapture option to link actions

aboissinot-coveo:
### Proposed Changes

`@plasma-react`

Adding an `onClickCapture` option for table actions in the legacy table. This will allow us to add some Amplitude tracking calls on table action links and still preserve the basic link functionalities.

This is related to this [PR](https://github.com/coveo/admin-ui/pull/8571). By replacing `link` with `trigger`, it removed the ability to right-click on the action to open the link.

None. This option is not required.

### Acceptance Criteria


packages/react/src/components/actions/LinkAction.tsx
```diff
@@ -23,6 +23,15 @@ export class LinkAction extends Component<ILinkActionProps> {
...
 
-        return <a className={actionClasses} href={href} title={this.props.action.name} role="button" {...opts} />;
+        return (
+            <a
+                className={actionClasses}
+                href={href}
+                title={this.props.action.name}
+                role="button"
+                {...opts}
+                onClickCapture={this.props.action.onClickCapture}
```
GermainBergeron, line 33: I'm wondering if we should add the onClickCapture to other types of actions. E.g., a dev using the `trigger` or `subActions` could see in the autocomplete that onClickCapture is available and try to use it but it won't work.

Since it's deprecated we could also simply add a JSDoc comment to say it only works with `link`. What do you think?

aboissinot-coveo: Good point. I was also under the impression that it would be good enough to only support my use-case since the component is deprecated. I added a JSDoc on the option.

GermainBergeron (approved):

PR #3595, opened 2023-12-05
fix(wpc-154): add numeric input buttons type

tortiz-coveo:
### Proposed Changes

Adds the button types to the plus and minus numeric input buttons. This avoids premature default validations on the form when clicking on those buttons. Currently the default type of the button is submit and triggers validations on every button press. This behavior manifests itself by showing a `Please fill out this field` alert on every button press on forms where empty inputs are present in the form. Here's a [stack overflow post ](https://stackoverflow.com/a/28340579)describing the issue

image

### DIFF
--- a/packages/react/src/components/numericInput/NumericInputConnected.tsx
+++ b/packages/react/src/components/numericInput/NumericInputConnected.tsx
@@ -119,6 +119,7 @@ export class NumericInputConnected extends PureComponent<NumericInputProps & HTM
             <div className="numeric-input flex flex-column">
                 <div className="flex flex-row">
                     <button
+                        type="button"
                         className="js-decrement mr1"
                         disabled={!decrementEnabled || this.props.disabled}
                         onClick={this.onDecrement}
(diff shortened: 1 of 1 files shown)


packages/react/src/components/numericInput/NumericInputConnected.tsx
```diff
@@ -119,6 +119,7 @@ export class NumericInputConnected extends PureComponent<NumericInputProps & HTM
             <div className="numeric-input flex flex-column">
                 <div className="flex flex-row">
                     <button
+                        type="button"
```
tortiz-coveo, line 122: Do we want enzyme tests to ensure this prop is on the buttons? I feel it's pretty overkill however it could document why these types are important

tortiz-coveo: I could make them in testing library also if it's preferable and the setup is not too daunting

gdostie (approved):

GermainBergeron (approved):

PR #3602, opened 2023-12-22
Ad UI 9399 fix get input props

FelixBlaisThon:
### Proposed Changes

Added opt-in type in ```getInputProps()``` so that the ```reorder, remove and insert``` method are only eposed for ```collection``` type.

It will break all collection that uses those 3 methods, since they do not explicitly specify the type that is now required to get those methods

I point it to next since it will get added to the new mantine 7 breaking change
### Acceptance Criteria


packages/mantine/src/form/useForm.ts
```diff
@@ -1,21 +1,49 @@
...
         path,
         {type = 'input', withError = type === 'input', withFocus = true} = {},
     ) => {
-        const originalPayload = form.getInputProps(path, {type, withError, withFocus});
+        const originalPayload = form.getInputProps(path, {
+            type: type === 'collection' || 'input' ? 'input' : 'checkbox',
+            withError,
+            withFocus,
+        });
         if (Array.isArray(originalPayload.value)) {
```
gdostie, line 38: We could do the check for the type here

```suggestion
        if (type === 'collection') {
```

FelixBlaisThon: Sure, and returning only ```originalPayload``` if not ?

gdostie: exactly

gdostie (approved):

aboissinot-coveo (approved):

PR #3603, opened 2024-01-12
[MHUB-613] feat: add an extended multi-select

samuelfullerthomas:
### Proposed Changes

Jira: [MHUB-613](https://coveord.atlassian.net/browse/MHUB-613)

Before:

multiselect where the create button is at the bottom of the dropdown menu

After:
multiselect where the create button is at the top of the dropdown menu

I've added an extended multi-select that copies the `Multiselect`  from mantine 6.0.19, but changes one thing: it sorts creatable elements to the top of the list. Otherwise, the code is just from mantine.

This is to fulfil a requirement for MHUB-613. I think this is ok because mantine 6.X is in maintenance mode, and this bridges the gap while we wait for mantine 7 to get upgraded.

One thing I'd prefer not to do is write extensive unit tests for this. Since the use-case is limited, I feel confident that the tests where the component is used will be sufficient.

None. This creates a new component that will be used in one place, rather than overwriting the base `Multiselect` component.

### Acceptance Criteria

[MHUB-613]: https://coveord.atlassian.net/browse/MHUB-613?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/mantine/package.json
+++ b/packages/mantine/package.json
@@ -42,6 +42,7 @@
         "@dnd-kit/modifiers": "7.0.0",
         "@dnd-kit/sortable": "8.0.0",
         "@dnd-kit/utilities": "3.2.1",
+        "@mantine/styles": "6.0.19",
         "@mantine/utils": "6.0.19",
         "@monaco-editor/react": "4.5.1",
         "@swc/helpers": "0.5.1",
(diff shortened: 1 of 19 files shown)


GermainBergeron:
Hi Sam!

I've been thinking about your solution and I really think the combobox will make it easier to achieve with Mantine 7. In the mean time I think we could do [something like this](https://codesandbox.io/p/devbox/f43svl?file=%2Fsrc%2FDemo.tsx%3A19%2C17) with a bit of CSS to make the createLabel fixed at the bottom of the list. It feels closer to what UX want and doesn't require us to duplicate most of Mantine code

samuelfullerthomas:
sounds good

PR #3604, opened 2024-01-15
fix(plasma): fixed the test with mantine v7 bump

FelixBlaisThon:
### Proposed Changes

This is a separate PR that fixes the failing tests coming from the mantine bump to v7

If green : woohoo

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/button/Button.tsx
+++ b/packages/mantine/src/components/button/Button.tsx
@@ -45,8 +45,8 @@ const _Button = forwardRef<HTMLButtonElement, ButtonProps>(
                     loading={isLoading || loading}
                     onClick={handleClick}
                     disabled={disabled}
-                    {...others}
                     data-loading={isLoading || loading}
+                    {...others}
                 />
             </ButtonWithDisabledTooltip>
         );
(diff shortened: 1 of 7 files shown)


packages/mantine/src/components/button/Button.tsx
```diff
@@ -46,6 +46,7 @@ const _Button = forwardRef<HTMLButtonElement, ButtonProps>(
                     onClick={handleClick}
                     disabled={disabled}
                     {...others}
+                    data-loading={isLoading || loading}
```
gdostie, line 49: I am curious about the reasoning behind the order here. (why put it after others rather than before)

packages/mantine/src/components/table/layouts/__tests__/RowLayout.spec.tsx
```diff
@@ -398,13 +399,14 @@ describe.skip('RowLayout', () => {
                     disableRowSelection
                 />,
             );
-
-            expect(screen.getByRole('checkbox', {name: /select all/i})).toHaveStyle('pointerEvents: none');
+            screen.getByRole('checkbox', {name: /select all/i}).click();
```
gdostie, line 402: Shouldn't we be using `user.click(element)` instead of `element.click()` ?

FelixBlaisThon: Woops that's my bad

packages/mantine/src/components/button/Button.tsx
FelixBlaisThon: No reasonning. I added it only for unit test since the change from mantine does not allow us to us the rol ```presentation``` anymore for the loading part. I don't think it is relevant to have it before or after. I will still change it to before anyway, ```{...others}``` should be last

aboissinot-coveo (approved):

gdostie (approved):

PR #3611, opened 2024-01-25
Mantine demos

GermainBergeron:
### Proposed Changes

Added a script that fetch the demo from Mantine's GitHub and generate pages for us to showcase our theme

You should be able to see the pages in Mantine section of the menu. I had to skip over some demo because they are broken (missing import, missing prop, wrong prop, etc)

None

### Acceptance Criteria


packages/website/bin/index.ts
```diff
@@ -0,0 +1,43 @@
...
+import simpleGit, {SimpleGit} from 'simple-git';
+import {DEMO_FILENAME_PATTERN} from './constants';
+import {copyDemoFiles} from './copyDemoFiles';
+import {findTsxAndCssModule} from './findTsxAndCssModule';
+import {generatePage} from './generatePage';
+import {groupFilesByComponent} from './groupFilesByComponent';
+
+const cloneRepository = async (repoUrl: string, destinationFolder: string): Promise<void> => {
+    const git: SimpleGit = simpleGit();
+    await git.clone(repoUrl, destinationFolder);
```
gdostie, line 12: Should we add clone to [semantic-monorepo-tools](https://github.com/coveo/semantic-monorepo-tools/tree/main/src/git)?

gdostie (approved):
A good start 👍 There are stuff we could cleanup in each page like remove the empty props table or add a link to the actual mantine website page, but we can do it later on.

dmgauthier (approved):
Great stuff it works super well

packages/website/bin/index.ts
GermainBergeron: I'll create tasks so we don't forget but I wouldn't block this PR for that if you agree

PR #3616, opened 2024-02-09
feat(collection): use getStyles, fixed collection

FelixBlaisThon:
### Proposed Changes

this is a subtask of https://coveord.atlassian.net/browse/ADUI-9365 where a lot of components are using the plasma/mantine ```<Collection />``` component. When I did the migration from mantine 6 -> 7, I did not foresee that we would need to factorise this component so that devs could use it and access css selector, like so :

`<Collection styles={{root: {...}, item: {...}, otherSelector: {...}}}>{...}</Collection>`

With my migration, it was made so that the classes were directly used as an import, while now it's used with ```useStyles``` mantine hook.

With this new implementation, it is now possible to add style with styles prop and add custom style to ```'root' | 'item' | 'itemDragging' | 'dragHandle'``` selectors.

This should not break anything. In fact, it fixes a breaking behavior the first implementation did

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/collection/Collection.tsx
+++ b/packages/mantine/src/components/collection/Collection.tsx
@@ -8,7 +8,6 @@ import {
     Factory,
     Group,
     Input,
-    MantineComponent,
     MantineSpacing,
     Stack,
     StylesApiProps,
@@ -19,8 +18,9 @@ import {
 } from '@mantine/core';
 import {ReorderPayload} from '@mantine/form/lib/types';
 import {useDidUpdate} from '@mantine/hooks';
-import {ForwardedRef, ReactNode, forwardRef} from 'react';
+import {ForwardedRef, ReactNode} from 'react';
 
+import {CustomComponentThemeExtend, identity} from '../../utils';
 import {Button} from '../button';
 import classes from './Collection.module.css';
 import {CollectionProvider} from './CollectionContext';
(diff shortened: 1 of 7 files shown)


packages/mantine/src/components/collection/Collection.tsx
```diff
@@ -139,7 +175,7 @@ const defaultProps: Partial<CollectionProps<unknown>> = {
     getItemId: ({id}: any) => id,
 };
 
-export const Collection = <T,>(props: CollectionProps<T>) => {
+export const Collection = <T,>(props: CollectionProps<T> & {ref?: ForwardedRef<HTMLDivElement>}) => {
```
gdostie, line 178: I think we need to call `forwardRef()` here otherwise you cannot pass in a ref as prop

packages/mantine/src/components/collection/Collection.tsx
```diff
@@ -175,7 +176,7 @@ const defaultProps: Partial<CollectionProps<unknown>> = {
     getItemId: ({id}: any) => id,
 };
 
-export const Collection = <T,>(props: CollectionProps<T> & {ref?: ForwardedRef<HTMLDivElement>}) => {
+export const Collection = forwardRef(<T,>(props: CollectionProps<T> & {ref?: ForwardedRef<HTMLDivElement>}) => {
```
gdostie, line 179:
```suggestion
export const Collection = forwardRef<HTMLDivElement, CollectionProps>(<T,>(props: CollectionProps<T>, ref) => {
```

gdostie (approved):

packages/mantine/src/components/collection/Collection.tsx
```diff
@@ -4,30 +4,30 @@ import {restrictToParentElement, restrictToVerticalAxis} from '@dnd-kit/modifier
...
-import cx from 'clsx';
 import {Button} from '../button';
-import CollectionClasses from './Collection.module.css';
+import classes from './Collection.module.css';
+import {CollectionProvider} from './CollectionContext';
 import {CollectionItem} from './CollectionItem';
 
-interface CollectionProps<T>
-    extends Omit<
-            InputWrapperProps,
```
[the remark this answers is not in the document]
FelixBlaisThon: I've added the ```__InputWrapperProps``` from ```@mantine/core``` that contains the props I've added

packages/mantine/src/components/collection/Collection.tsx
```diff
@@ -127,8 +126,19 @@ interface CollectionProps<T>
      * @default false
      */
     required?: boolean;
+    /**
+     * Input label, displayed above the collection. If not provided, the label will not be rendered
+     */
```
gdostie, line 131:
```suggestion
```

gdostie (approved):

GermainBergeron (approved):

PR #3617, opened 2024-02-12
feat(Mantine, Table): adding edit columns visibility button

aiheon:
jira: https://coveord.atlassian.net/browse/MHUB-755

### Proposed Changes

Added a new option for the table header to have a button to edit columns visibility.

OG Figma: https://www.figma.com/file/f5M8BLV7Ganu5i4OzUB3De/Recs-Manager?type=design&node-id=1257-19353&mode=design&t=BDOCgAUkjffEOkqR-4

Demo:

https://github.com/coveo/plasma/assets/63734941/2205aa75-3e6e-45d6-a784-cde5c7907b9e

None

### Acceptance Criteria


packages/mantine/src/components/table/table-edit-columns-visibility/TableEditColumnsVisibility.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+    if (filteredColumns.length <= 0) {
+        return null;
+    }
+
+    return (
+        <Grid.Col span="content" order={TableComponentsOrder.EditColumnsVisibility} py="sm" className={classes.root}>
+            <Popover position="bottom" shadow="md">
+                <Popover.Target>
+                    <Button variant="outline">{`${label}${
```
Tooni, line 52: we might need to pass the button variant down as a prop too, since i think the designs for the commerce-hub are using the `secondary` variant rather than `outline`, but `outline` probably doesn't exist in the plasma theme

aiheon: outline is a default variant ine mantine and is used as is IIRC 🤔 But I agree with your point anyway, better safe than sorry 👍

packages/mantine/src/components/table/table-edit-columns-visibility/TableEditColumnsVisibility.tsx
```diff
@@ -0,0 +1,91 @@
...
+                                        }
+                                        onChange={column.getToggleVisibilityHandler()}
+                                    />
+                                ))}
+                        </Stack>
+                    </ScrollArea.Autosize>
+                    {maxSelectableColumns && (
+                        <>
+                            <Divider mb="xs" mt="sm" />
+                            <div>{footerText}</div>
```
GermainBergeron, line 84: You don't need the div

```suggestion
                            {footerText}
```

packages/mantine/src/components/table/table-edit-columns-visibility/TableEditColumnsVisibility.tsx
```diff
@@ -0,0 +1,91 @@
...
+        : filteredColumns.filter((column) => !column.getIsVisible() && !nonHideableColumns?.includes(column.id)).length;
+
+    let footerText;
+    if (selectedColumnsCount === filteredColumns.length) {
+        footerText = 'All available columns selected';
+    } else if (adjustedMaxSelectableColumns && selectedColumnsCount >= adjustedMaxSelectableColumns) {
+        footerText = 'Maximum columns selected';
+    } else {
+        footerText = `You can choose ${remainingColumnsCount} more`;
+    }
```
GermainBergeron, line 43: This footerText should be created using props, this way it can be overwritten

packages/mantine/src/components/table/Table.tsx
```diff
@@ -217,6 +222,7 @@ export const Table: TableType = <T,>({
     );
 };
 
+Table.EditColumnsVisibility = TableEditColumnsVisibility;
```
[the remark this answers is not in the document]
aiheon: I like shorter indeed 👍

packages/mantine/src/components/table/table-edit-columns-visibility/TableEditColumnsVisibility.tsx
aiheon: How can I deal with string from a props but still have the count here? 🤔  

`remaningColumnsCount` is calculated at line 34 and is not accessible outside the component. 👀 

image
****

aiheon: I got a new mockup for the footer so I think I will maybe juste put a `footer: reactNode` in the prop instead? 🤔 

I can also keep the current logic for the default footer if you think it can be used in the future

packages/website/src/pages/layout/Table.tsx
```diff
@@ -39,6 +40,13 @@ const DemoPage = () => (
                     title="Table with a child component using the hook to re-fetch"
                 />
             ),
+            editColumnsVisibility: (
```
GermainBergeron, line 43:
```suggestion
            columnSelector: (
```

packages/website/src/pages/layout/Table.tsx
```diff
@@ -39,6 +40,13 @@ const DemoPage = () => (
...
+                <TableColumnsSelectorDemo
+                    noPadding
+                    layout="vertical"
+                    title="Table with the ability to edit visible columns"
```
GermainBergeron, line 47:
```suggestion
                    title="Table with the ability to select columns"
```

packages/website/src/examples/layout/Table/TableColumnsSelector.demo.tsx
```diff
@@ -0,0 +1,114 @@
...
+            fullName: faker.person.fullName(),
+            jobTitle: faker.person.jobTitle(),
+            hireDate: faker.date.past(),
+            salary: faker.number.float(),
+            isFullTime: faker.datatype.boolean(),
+            email: faker.internet.email(),
+            body: faker.lorem.paragraph(),
+        }));
+
+const columnNames: Record<string, string> = {
```
GermainBergeron, line 104: We can use a stricter type here:

```suggestion
const columnNames: Partial<Record<keyof IEmployeeData, string>> = {
```

The Partial make it so you don't need to specify everythign

packages/website/src/examples/layout/Table/TableColumnsSelector.demo.tsx
```diff
@@ -0,0 +1,114 @@
...
+        >
+            <Table.Header>
+                <Table.ColumnsSelector
+                    label={'Edit columns'}
+                    nonHideableColumns={['employeeId']}
+                    maxSelectableColumns={5}
+                    showVisibleCountLabel
+                    columnNames={columnNames}
+                    footer={<Text variant="dimmed">You can display up to 5 columns</Text>}
+                    tooltip="You have reached the maximum display limit of 5 columns. Uncheck a column to display another one."
```
GermainBergeron, line 81: IMHO `tooltip` doesn't reflect the fact that it's for when the limit is reached

packages/mantine/src/components/table/__tests__/TableColumnsSelector.spec.tsx
```diff
@@ -0,0 +1,260 @@
...
+        body: 'coucou 2',
+    },
+];
+
+const columnNames: Record<string, string> = {
+    name: 'Name',
+    age: 'Age',
+    email: 'Email',
+    phone: 'Phone',
+};
```
GermainBergeron, line 260: IMHO it's better to put that at the top of the file

packages/mantine/src/components/table/table-columns-selector/TableColumnsSelector.tsx
```diff
@@ -0,0 +1,82 @@
...
+                                    />
+                                );
+
+                                return isDisabled ? (
+                                    <Tooltip label={tooltip} position="left">
+                                        <div>{checkbox}</div>
+                                    </Tooltip>
+                                ) : (
+                                    checkbox
+                                );
```
GermainBergeron, line 68: You can disable the tooltip if you don't want it to show:

```suggestion
                                return (
                                    <Tooltip label={tooltip} disabled={isDisabled} position="left">
                                        <Checkbox
                                            key={column.id}
                                            label={columnNames?.[column.id] || column.id}
                                            name={column.id}
                                            checked={column.getIsVisible()}
                                            disabled={isDisabled}
                                            onChange={column.getToggleVisibilityHandler()}
                                        />
                                    </Tooltip>
                                );
```

aiheon: but it won't show up on the disabled checkbox 😕  that's why I added the div

packages/website/src/examples/layout/Table/TableColumnsSelector.demo.tsx
aiheon: what do you suggest? 👀

packages/mantine/src/components/table/table-columns-selector/TableColumnsSelector.tsx
GermainBergeron: I edited my suggestion, but if you do `disabled={!disabled}` on the tooltip and let the checkbox being disabled does it work?

aiheon: no 😕

aiheon: I saw while fixing the tests that checkbox disabled have the `pointer-event: none`

GermainBergeron (approved):

PR #3639, opened 2024-03-15
fix(line height): 1.5

dmgauthier:
### Proposed Changes

modified the Plasma-Style reset file body class.  `line-height: 1` => `line-height: 1.5`

### Acceptance Criteria


packages/mantine/src/components/table/Table.module.css
gdostie, line 1: Are the changes in this file still needed since you reverted the removal of resete.scss?

toofff (approved):
LGTM

packages/mantine/src/components/table/Table.module.css
dmgauthier: I think I could drop the 300 font weight, but since I see it already in the predicate, I decided to leave it there. But your question make me realise I could try to **remove** both font weight in the predicate and the filter to see how it turns out.  I need the width though.

GermainBergeron (approved):

PR #3642, opened 2024-03-18
Fix ad UI 9639 popover zindex

dmgauthier:
### Proposed Changes

bumping the Zindex of the popover manually in the default props since the actual style coming from the module is not used. Modifying the style there or with the variables have no effect.

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/theme/Theme.tsx
+++ b/packages/mantine/src/theme/Theme.tsx
@@ -14,6 +14,7 @@ import {
     InputWrapper,
     List,
     Loader,
+    LoadingOverlay,
     MantineThemeOverride,
     MenuItem,
     Modal,
@@ -116,6 +117,9 @@ export const plasmaTheme: MantineThemeOverride = createTheme({
         Button: Button.extend({
             classNames: {root: ButtonClasses.root, label: ButtonClasses.label},
         }),
+        LoadingOverlay: LoadingOverlay.extend({
+            defaultProps: {zIndex: 1400},
+        }),
         Modal: Modal.extend({
             classNames: {
                 root: ModalClasses.root,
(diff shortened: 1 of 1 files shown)


gdostie (approved):

GermainBergeron (commented):
I'm wondering if it's possible to do that in the admin instead of plasma?

dmgauthier:
> I'm wondering if it's possible to do that in the admin instead of plasma?

As far as I understand it.
**Yes** if we want to repeat the process by passing the prop ourselves when required.
**No** if we want it to be generally applied without having to do it ourselves each time; The style applied is directly inline on the element, set in the strongest way possible by Mantine from the default props mechanism. Since the only way to modify the z-index is by the prop, either we do it when we use the component or we use the defaultProps. :confused: 

https://github.com/mantinedev/mantine/blob/e3e3bb834de1f2f75a27dbc757dc0a2fc6a6cba8/packages/%40mantine/core/src/components/Popover/Popover.tsx#L166-L167  :point_left:  Zindex set in the default props

https://github.com/mantinedev/mantine/blob/e3e3bb834de1f2f75a27dbc757dc0a2fc6a6cba8/packages/%40mantine/core/src/components/Popover/Popover.tsx#L304 :point_left: zindex put in the ctx

https://github.com/mantinedev/mantine/blob/e3e3bb834de1f2f75a27dbc757dc0a2fc6a6cba8/packages/%40mantine/core/src/components/Popover/PopoverDropdown/PopoverDropdown.tsx#L107-L108 :point_left: zindex used from the ctx directly in the style of the box. This Afaik will be stronger than anything else.

GermainBergeron:
I was thinking of extending the Theme to add admin-ui specifics in there. Probably with https://mantine.dev/theming/theme-object/#merge-multiple-theme-overrides. It would require us to re-declare Plasmantine inside components/mantine and import the theme object instead. Maybe it's too convoluted and not necessary

GermainBergeron (approved):

PR #3650, opened 2024-03-19
fix(mantine): add custom css variable resolver

gdostie:
### Proposed Changes

Removes transparency from default background colors. [The default css variable resolver](https://github.com/mantinedev/mantine/blob/master/packages/%40mantine/core/src/core/MantineProvider/MantineCssVariables/default-css-variables-resolver.ts#L108-L115) would set 10% opacity on the text color and use that as background color. What I did here is override that and just use the shade 0 instead as previous versions of mantine would do.

```tsx
<Alert title="asdasd" />
<Badge color="green">
    INK
</Badge>
```

Before
image

After

image

### Acceptance Criteria


packages/mantine/src/theme/CSSVariablesResolver.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const resolver: CSSVariablesResolver = (theme) => {
+    const result: ConvertCSSVariablesInput = {
+        variables: {},
+        dark: {},
+        light: {},
+    };
+    keys(theme.colors).forEach((color) => {
+        result.light[`--mantine-color-${color}-light`] = theme.colors[color][color === 'gray' ? 1 : 0];
+        result.light[`--mantine-color-${color}-light-hover`] = theme.colors[color][color === 'gray' ? 2 : 0];
```
GermainBergeron, line 11: If the color is not gray then the `light` and `light-hover` will have the same shade?

GermainBergeron (approved):

dmgauthier (approved):

PR #3660, opened 2024-03-25
fix(mantine): button loading data attribute

gdostie:
### Proposed Changes

When `data-loading="false"` is applied to the button, the loading style still applies which hides the button text. This is because Mantine uses the selector `:where([data-loading])` and not `:where([data-loading='true'])`. So when the loading is false, the attribute must be set to undefined.

Before
image

After
image

### Acceptance Criteria


toofff:
I don't see any change with the demo

image

gdostie:
> I don't see any change with the demo

We don't have any example of this in the demo, but you need a button with `loading={false}` to witness the issue.

Before
image

After
image

toofff (approved):
if everything works properly locally, I'm good to go

PR #3667, opened 2024-03-28
fix(button): prevent default onclick for link button

FelixBlaisThon:
### Proposed Changes

When using the polymorphic component of button and using it as a Link, the ```onClick``` prop was still triggered and a redirection to the ```to``` link prop specified occurred even though the button was disabled. We `preventDefault()` when ```data-disabled === true```. this should fix a lot (all) of redirection disabled button.
### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/button/Button.tsx
+++ b/packages/mantine/src/components/button/Button.tsx
@@ -1,8 +1,8 @@
 import {
-    Button as MantineButton,
     ButtonGroup,
-    ButtonProps as MantineButtonProps,
     Factory,
+    Button as MantineButton,
+    ButtonProps as MantineButtonProps,
     polymorphicFactory,
 } from '@mantine/core';
 import {ButtonCssVariables, ButtonStylesNames, ButtonVariant} from '@mantine/core/lib/components/Button/Button';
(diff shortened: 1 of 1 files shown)


gdostie (approved):

packages/mantine/src/components/button/Button.tsx
```diff
@@ -41,7 +44,7 @@ export const Button = polymorphicFactory<ButtonOverloadFactory>(
                     loaderProps={{variant: 'oval'}}
                     ref={ref}
                     loading={isLoading || loading}
-                    onClick={handleClick}
+                    onClick={dataDisabled ? (e) => e.preventDefault() : handleClick}
```
GermainBergeron, line 47: I think we should continue to forward the value to Mantine to apply the style:
```suggestion
                    onClick={dataDisabled ? (e) => e.preventDefault() : handleClick}
                    data-disabled={dataDisabled}
```

FelixBlaisThon: `Line 48` fixed the visual style of the disabled button :)

GermainBergeron (approved):

PR #3668, opened 2024-03-28
fix(mantine): table header stacking styles

gdostie:
### Proposed Changes

- Remove border between table header stacks
- Remove wrap-reverse so that actions remain closer to table
- Change filter icon color to gray-5
- Disable font smoothing
- Fix iconography cards justify space-between

Before (next)

https://github.com/coveo/plasma/assets/35579930/296c6c0c-5018-401a-aa77-cce555e2c332

After (this branch)

https://github.com/coveo/plasma/assets/35579930/efa852ea-c63e-4bc1-a883-7ba11214cbb5

### Acceptance Criteria


toofff (approved):
Nice your demo !

I don't know if you've got a handle on this or if it's a good idea.
But to avoid fast actions going to the line, couldn't we just hide the text and add this text in a tooltip attribute?

gdostie:
> Nice your demo !
> 
> I don't know if you've got a handle on this or if it's a good idea. But to avoid fast actions going to the line, couldn't we just hide the text and add this text in a tooltip attribute?

I am not sure it is realistically feasible, because you would need to know what is the width of each quick action if we render them, but without rendering them. Without this information you cannot decide if your showing the full action of the shortened version of it.

aboissinot-coveo (approved):

packages/mantine/src/components/table/Table.module.css
```diff
@@ -31,34 +31,26 @@
...
-    );
+    background-color: var(--mantine-color-gray-1);
     padding-left: var(--mantine-spacing-xl);
     padding-right: var(--mantine-spacing-lg);
     position: relative;
 }
+
 .headerGridInner {
     flex-direction: row-reverse;
-    flex-wrap: wrap-reverse;
```
GermainBergeron, line 56: Do we need to update `TableComponentsOrder` ?

GermainBergeron (approved):

FelixBlaisThon (approved):

packages/mantine/src/components/table/Table.module.css
gdostie: The order is still correct I think?, because of the `flex-direction: row-reverse;` Its just the wrap order that I inverted

PR #3676, opened 2024-04-10
feat(header): +content variant +example

FelixBlaisThon:
### Proposed Changes

https://coveord.atlassian.net/browse/ADUI-9094

This just adds another variant to the Header component. For now it is really dry, in fact the same as the modal variant, but now from within our code, we will use a relevant variant name + if we want to change the content variant visual, we could do so with module.css or other way, separating modal and content variants.

Also, nothing prevents the dev from putting any actions or breadcrumbs in that content variant, opening the possibility of creating a pretty intense header loop of stuff. (more an FYI than actual warning)

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/header/Header.tsx
+++ b/packages/mantine/src/components/header/Header.tsx
@@ -12,11 +12,11 @@ import {
     useStyles,
 } from '@mantine/core';
 import {Children, ReactElement, ReactNode} from 'react';
+import {HeaderProvider} from './Header.context';
 import classes from './Header.module.css';
 import {HeaderActions, HeaderActionsStyleNames} from './HeaderActions/HeaderActions';
 import {HeaderBreadcrumbs, HeaderBreadcrumbsStyleNames} from './HeaderBreadcrumbs/HeaderBreadcrumbs';
 import {HeaderDocAnchor, HeaderDocAnchorStyleNames} from './HeaderDocAnchor/HeaderDocAnchor';
-import {HeaderProvider} from './Header.context';
 
 export type {HeaderActionsProps} from './HeaderActions/HeaderActions';
 export type {HeaderBreadcrumbsProps} from './HeaderBreadcrumbs/HeaderBreadcrumbs';
@@ -46,7 +46,7 @@ export interface HeaderProps extends StylesApiProps<HeaderFactory>, Omit<GroupPr
      *
      * @default 'page'
      */
-    variant?: 'page' | 'modal';
+    variant?: 'page' | 'modal' | 'content';
     /**
      * The title of the header.
      */
--- a/packages/mantine/src/components/header/Header.tsx
+++ b/packages/mantine/src/components/header/Header.tsx
@@ -42,7 +42,7 @@ export interface HeaderProps extends StylesApiProps<HeaderFactory>, Omit<GroupPr
      */
     borderBottom?: boolean;
     /**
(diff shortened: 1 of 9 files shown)


packages/mantine/src/components/header/Header.tsx
```diff
@@ -46,7 +46,7 @@ export interface HeaderProps extends StylesApiProps<HeaderFactory>, Omit<GroupPr
      *
      * @default 'page'
      */
-    variant?: 'page' | 'modal';
+    variant?: 'page' | 'modal' | 'content';
```
GermainBergeron, line 49: I think we should go with primary/secondary instead of duplicating modal. Content doesn't make much sense for me

```suggestion
    variant?: 'primary' | 'secondary';
```

packages/mantine/src/components/header/Header.tsx
```diff
@@ -44,9 +44,9 @@ export interface HeaderProps extends StylesApiProps<HeaderFactory>, Omit<GroupPr
     /**
      * Use the modal variant when displaying the header inside a modal
```
gdostie, line 45: This documentation needs to be adjusted

GermainBergeron (approved):

toofff (approved):
Nice !!!

Now the `secondary` variant will have to be used in Modals but also inside pages? It's perfect and much simpler!

packages/website/src/pages/layout/Header.tsx
```diff
@@ -15,7 +14,7 @@ export default () => (
         propsMetadata={HeaderMetadata}
         demo={<HeaderDemo />}
         examples={{
-            modalVariant: <HeaderModalDemo grow title="Modal variant" />,
+            secondaryVariant: <HeaderSecondaryDemo grow title="Secondary variant" />,
```
toofff, line 17: I have the impression that the import of the `HeaderSecondaryDemo` component is missing.

GermainBergeron:
> Nice !!!
> 
> Now the `secondary` variant will have to be used in Modals but also inside pages? It's perfect and much simpler!

Yes, primary for the title of the page, secondary for modal and content

packages/website/src/pages/layout/Header.tsx
FelixBlaisThon: yes i am dumb today

packages/mantine/src/components/header/Header.tsx
```diff
@@ -42,11 +42,11 @@ export interface HeaderProps extends StylesApiProps<HeaderFactory>, Omit<GroupPr
      */
     borderBottom?: boolean;
     /**
-     * Use the modal variant when displaying the header inside a modal
+     * Use the secondary variant when displaying the header not at the top of the page
```
GermainBergeron, line 45: I think the description is a bit hard to understand
```suggestion
     * Use the primary variant for page header and secondary variant elsewhere
```

packages/mantine/src/components/header/Header.tsx
```diff
@@ -12,17 +12,17 @@ import {
     useStyles,
 } from '@mantine/core';
 import {Children, ReactElement, ReactNode} from 'react';
+import {HeaderProvider} from './Header.context';
 import classes from './Header.module.css';
```
GermainBergeron, line 16: There is style to adjust in this file

gdostie (approved):

PR #3677, opened 2024-04-10
fix(style): add missing reset rules to reset.css

gdostie:
### Proposed Changes

HTML tags like `<b>` or `<strong>` that inherently imply css style (`font-weight: bold` in this case) would not work anymore. This is due to the fact that Mantine version 6 would import [normalize.css](https://necolas.github.io/normalize.css/) and they stopped doing that in version 7. 

In this PR, I cherry picked some of the normalize.css that make sense for us and added them in the existing `reset.scss` file of the plasma-style package.

You can test it out in the demo, by editing the HTML somewhere to add a `<b>` tag around text. It should make it bold, while before on the branch `next` it would not. On `master` branch it does work.

### Acceptance Criteria


FelixBlaisThon (approved):
Oooh so that's what was missing. good job

GermainBergeron (approved):

PR #3686, opened 2024-04-17
fix(mantine): adjust Alert visual look

gdostie:
### Proposed Changes

Made some adjustments in the theme for the Alert component

- Icon should be 16px (was 20px)
- Space between icon and text should be 16px (was 24px)
- Padding around the alert should be 16px (was 24px)
- Content text should have color gray-7 (was gray-9)

| Before | After | Figma |
| --- | --- | --- |
| image | image | image |

### Acceptance Criteria

### DIFF
--- a/packages/website/src/building-blocs/PageHeader.tsx
+++ b/packages/website/src/building-blocs/PageHeader.tsx
@@ -8,7 +8,7 @@ export interface PageHeaderProps {
     title: string;
     thumbnail?: TileProps['thumbnail'];
     description?: ReactNode;
-    section: 'Foundations' | 'Layout' | 'Form' | 'Navigation' | 'Feedback' | 'Advanced' | 'Mantine' | 'Feedbacl';
+    section: 'Foundations' | 'Layout' | 'Form' | 'Navigation' | 'Feedback' | 'Advanced' | 'Mantine';
     /**
      * Path to a relevant source file in the repo
      *
--- a/packages/website/src/building-blocs/PageHeader.tsx
+++ b/packages/website/src/building-blocs/PageHeader.tsx
@@ -8,7 +8,7 @@ export interface PageHeaderProps {
     title: string;
     thumbnail?: TileProps['thumbnail'];
     description?: ReactNode;
-    section: 'Foundations' | 'Layout' | 'Form' | 'Navigation' | 'Feedback' | 'Advanced' | 'Mantine';
+    section: 'Foundations' | 'Layout' | 'Form' | 'Navigation' | 'Feedback' | 'Advanced' | 'Mantine' | 'Feedbacl';
     /**
      * Path to a relevant source file in the repo
      *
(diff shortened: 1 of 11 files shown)


gdostie:
> Thank you for these adjustments 🙏 they're very appreciated. While we're at it, could we add a stroke to them, the lack of them have been a source of friction. If feasible, I would propose:
> 
> * Info = Navy1
> * Tip = Grey3
> * Success = Sec. Green1
> * Warning = Yellow2
> * Error = Red1
> 
> I've updated in Figma. If not feasible please let me know so I can remove them. Thanks again @gdostie !

@guscoveo Using the direct colors that you wanted for the border paused to be problematic since the shade used is different depending on the color. I managed to achieve a visually similar result by way of dark magic (using a [color-mix](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-mix) of the title color and the background color). Let me know if this is an acceptable compromise
image

gdostie:
@guscoveo If you are wondering, I am using a mix of 15% title color, 85% background color:

```css
border-color: color-mix(in srgb, var(--alert-color), var(--alert-bg) 85%);
```

guscoveo:
> @guscoveo If you are wondering, I am using a mix of 15% title color, 85% background color:
> 
> ```css
> border-color: color-mix(in srgb, var(--alert-color), var(--alert-bg) 85%);
> ```

I think this is an interesting compromise. That said it will be the case in the future that as we evolve on our tokens, actual tokens will be used for this kind of value, so something to keep in mind.

Thanks again @gdostie

guscoveo (approved):

packages/website/src/building-blocs/PageHeader.tsx
```diff
@@ -8,7 +8,7 @@ export interface PageHeaderProps {
     title: string;
     thumbnail?: TileProps['thumbnail'];
     description?: ReactNode;
-    section: 'Foundations' | 'Layout' | 'Form' | 'Navigation' | 'Feedback' | 'Advanced' | 'Mantine';
+    section: 'Foundations' | 'Layout' | 'Form' | 'Navigation' | 'Feedback' | 'Advanced' | 'Mantine' | 'Feedbacl';
```
GermainBergeron, line 11: There's a typo in `Feedbacl` and `Feedback` is already there
```suggestion
    section: 'Foundations' | 'Layout' | 'Form' | 'Navigation' | 'Feedback' | 'Advanced' | 'Mantine';
```

packages/website/src/pages/feedback/Alert.tsx
```diff
@@ -0,0 +1,29 @@
...
+        thumbnail="placeholder"
+        description="Attract user attention with important static message."
+        demo={<AlertDemo center />}
+        examples={{
+            tip: <AlertTipDemo center title="Tip" />,
+            success: <AlertSuccessDemo center title="Success" />,
+            warning: <AlertWarningDemo center title="Warning" />,
+            critical: <AlertCriticalDemo center title="Critical" />,
+        }}
+        sourcePath="/packages/mantine/src/components/button/Button.tsx"
```
GermainBergeron, line 24: This is not the correct link

GermainBergeron (approved):

packages/website/src/pages/feedback/Alert.tsx
gdostie: updated thanks

PR #3690, opened 2024-04-23
feat(mantine): add variant modal-footer to StickyFooter

gdostie:
### Proposed Changes

Currently, without going full blown modal compound components, there is no good way of having a border above the modal footer that stretches to the edges of the modal. This is because the footer is inside the modal body and that body has padding by default.

The solution I suggest is to add a new prop on the `StickyFooter` component called `removeModalPadding`. When true, this prop removes the modal's default padding so that the footer properly hugs the bottom of the modal. Making it easy for people to achieve the desired look without introducing any breaking change.

See first Modal example in the demo website

```
<StickyFooter borderTop removeModalPadding>
    <Button variant="outline" onClick={() => setOpened(false)}>
        Cancel
    </Button>
    <Button onClick={() => setOpened(false)}>Accept</Button>
</StickyFooter>
```

image

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/sticky-footer/StickyFooter.module.css
+++ b/packages/mantine/src/components/sticky-footer/StickyFooter.module.css
@@ -9,3 +9,11 @@
 .border {
     border-top: 1px solid var(--mantine-color-gray-3);
 }
+
+.modalFooter {
+    padding-bottom: 0;
+    margin-top: var(--mb-padding);
+    margin-bottom: calc(var(--mantine-spacing-md) - var(--mb-padding));
+    margin-left: calc(-1 * var(--mb-padding));
+    margin-right: calc(-1 * var(--mb-padding));
+}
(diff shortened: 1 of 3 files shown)


GermainBergeron:
I think using a `variant="modal"` instead of a custom prop would be more in line with Mantine's way of doing things

aboissinot-coveo:
Your solution seems to work well and would achieve what we want. 👏  Although I will say that I like this suggestion:

> I think using a `variant="modal"` instead of a custom prop would be more in line with Mantine's way of doing things

gdostie:
> I think using a `variant="modal"` instead of a custom prop would be more in line with Mantine's way of doing things

> Your solution seems to work well and would achieve what we want. 👏 Although I will say that I like this suggestion

✅

GermainBergeron (approved):

PR #3713, opened 2024-05-07
Table revamp

gdostie:
**_This is a big PR for which I apologize in advance. I've tried to split this into intelligible commits to ease the review._**

### Proposed Changes

We've been using the plasma-mantine table for a while now and have noticed some quirks that generate unwanted patterns and code duplication. Most notably when using the Table in combination with react-query, it forces the implementer to duplicate the state externally in his implementation in order to build the query.

In this PR, the whole table state management has been revamped.

#### `store` prop & useTable hook

The Table component now expects to receive a `store` as prop. This store contains the table state and a few functions to interact with the state. You can obtain a store by calling the useTable() hook.

```tsx
const MyTable = () => {
    const store = useTable();
    return (
        <Table store={store} />
    );
};
```

Essentially, the table state has been extracted from the component implementation into an external hook. Since the state is now accessible by the implementer directly, it simplifies a lot of things including building queries or empty states. The hook also accepts a few options to make customizing the table easier.

#### `onMount` and `onChange` props removed

```tsx
// before
const fetchData = (state) => {...};
const MyTable = () => (
    <Table onMount={fetchData} onChange={fetchData} />
);

// after
const fetchData = (state) => {...};
const MyTable = () => {
    const store = useTable();
    useEffect(() => {
        fetchData(store.state);
    }, []); // you could add store.state.something to mimic the onChange prop if needed
    return (
        <Table store={store} />
    );
};
```

#### `Table.NoData`

This PR introduces an new compound component to replace the `noDataChildren` prop:

```tsx
// before
<Table noDataChildren={<EmptyState />} />

// after
<Table>
    <Table.NoData>
        <EmptyState />
    </Table.Nodata>
</Table>
```

#### `Table.Consumer` removed

Table consumer no longer exists. The reason for its existence was to provide a way for the table implementer to access the table state. Now that the store (which contains the state) is accessible from outside the table, there is no reason to keep it anymore.

#### `totalPages` prop removed

The `totalPages` prop was required by the Pagination to determine the number of pages to display. In almost all the use cases, you would need to compute that number yourself from the total entries number you receive from the backend. This computation is now done automatically by the table, you just need to specify the number of entries using the store.

```tsx
// before
// get totalCount from the query response
const totalCount = Number(response.headers.get('x-total-count'));
// obtain the pageSize from the state somehow and compute totalPages yourself, then pass it down as props
<Table.Pagination totalPages={Math.ceil(totalCount / state.pagination?.pageSize)} />

// after
// get totalCount from the query response 
const totalCount = Number(response.headers.get('x-total-count'));
// set it in the store
store.setTotalEntries(totalCount);
<Table.Pagination />
```

There's a lot of them I've tried to describe them in each commit message, but I will be writing a complete changelog in the github release itself.

### Acceptance Criteria


packages/website/src/examples/layout/Table/TableColumnsSelector.demo.tsx
```diff
@@ -17,13 +17,14 @@ interface IEmployeeData {
 const columnHelper = createColumnHelper<IEmployeeData>();
 
 const columns: Array<ColumnDef<IEmployeeData>> = [
-    columnHelper.accessor('fullName', {
-        header: 'Name',
-        cell: (info) => info.row.original.fullName,
-    }),
     columnHelper.accessor('employeeId', {
         header: 'Employee ID',
```
GermainBergeron, line 21:
```suggestion
        header: 'Employee ID', // the column header is used to display the name of the checkbox in the column selector
```

GermainBergeron (commented):
Really good job! I tried to look at everything and think about what we discussed and I don't have many comments

packages/website/src/examples/layout/Table/TableReactQuery.demo.tsx
```diff
@@ -0,0 +1,170 @@
...
+            getExpandChildren={(datum) => <Box py="xs">{datum.body}</Box>}
+        >
+            <Table.Header>
+                <UserPredicate />
+                <Table.Filter placeholder="Search posts by title" />
+                <Table.DateRangePicker
+                    rangeCalendarProps={{maxDate: dayjs().endOf('day').toDate()}}
+                    presets={DatePickerPresets}
+                />
+                <Table.Actions>{(datum: IExampleRowData) => <TableActions datum={datum} />}</Table.Actions>
```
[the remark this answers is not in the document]
gdostie: Yes I was planning to do it as a second step in another PR since this one was getting pretty big

packages/mantine/src/components/table/layouts/row-layout/RowLayoutIcon.tsx
```diff
@@ -1,16 +1,13 @@
 import {Icon as IconType, ListSize16Px} from '@coveord/plasma-react-icons';
-import {CompoundStylesApiProps, factory, Factory, useProps} from '@mantine/core';
-import {useRowLayout} from './RowLayoutContext';
+import {CompoundStylesApiProps, Factory, factory, useProps} from '@mantine/core';
 
-export type RowLayoutIconStylesNames = 'icon';
```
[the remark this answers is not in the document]
gdostie: The way its setup right now we can't, because when we switch to another layout, the RowLayoutContext is not wrapping the table anymore so it throws an error when trying to render the layout segmented control. I was thinking that allowing to override the icon through the default props was enough.

toofff:
Do we still need to delete the Table component when we delete the current filter?
TABLE-remove-filter

packages/website/src/examples/layout/Table/TableEmptyState.demo.tsx
```diff
@@ -1,18 +1,21 @@
...
-const NoData = () => {
-    const {isFiltered, clearFilters, state} = useTable();
-
-    return isFiltered ? (
+const EmptyState: FunctionComponent<{isFiltered: boolean; clearFilters: () => void; filter: string}> = ({
+    clearFilters,
+    isFiltered,
+    filter,
+}) =>
+    isFiltered ? (
```
toofff, line 16: I think this example points in the wrong direction.
You shouldn't go from a state with a filter that gives nothing to an empty state.

I think the example should focus on the empty state.

packages/website/src/pages/layout/Table.tsx
```diff
@@ -33,12 +31,11 @@ const DemoPage = () => (
...
-                <TableConsumerDemo
-                    noPadding
-                    layout="vertical"
-                    title="Table with a child component using the hook to re-fetch"
-                />
+            multiSelect: (
+                <TableMultiSelectionDemo noPadding layout="vertical" title="Table with bulk selection of rows" />
+            ),
+            disableRowSelection: (
+                <TableDisableRowSelection noPadding layout="vertical" title="Table with disabled row selection" />
```
toofff, line 38: Is a state disabled row planned? With CSS that directly changes the color of the font?
Or could a simple example be added?

packages/mantine/src/components/table/table-date-range-picker/TableDateRangePicker.tsx
```diff
@@ -44,26 +44,25 @@ const defaultProps: Partial<TableDateRangePickerProps> = {
 };
 
 export const TableDateRangePicker = factory<TableDateRangePickerFactory>((props, ref) => {
```
toofff, line 46: When you click outside the modal, should it close automatically?

image

packages/website/src/examples/layout/Table/TableReactQuery.demo.tsx
```diff
@@ -0,0 +1,170 @@
...
+            columns={columns}
+            getExpandChildren={(datum) => <Box py="xs">{datum.body}</Box>}
+        >
+            <Table.Header>
+                <UserPredicate />
+                <Table.Filter placeholder="Search posts by title" />
+                <Table.DateRangePicker
+                    rangeCalendarProps={{maxDate: dayjs().endOf('day').toDate()}}
+                    presets={DatePickerPresets}
+                />
```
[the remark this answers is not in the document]
gdostie: I wanted at least one example of the date picker, but the API we're using does't have any date fields. So the compromise I opted for was to keep the date picker, but don't use it in the query

packages/website/src/examples/layout/Table/TableEmptyState.demo.tsx
gdostie: This example is there to show how to differenciate between an empty state for a filtered table version no data

toofff (approved):
I think what you've done is really amazing, the Table component has become awesome.

As we've been talking about, it would be great if the code in the examples could be simpler, so that we only have the code that corresponds to what we want to present.

packages/mantine/src/components/table/table-date-range-picker/TableDateRangePicker.tsx
gdostie: I think yes, but it happens also on master branch, I will log a separate ticket for it

packages/website/src/pages/layout/Table.tsx
gdostie: I will add a prop like the getRowId function called getRowAttributes. It will spread the attributes on the table row (`<tr>`). This way you will be able to do something like

```tsx
getRowAttributes={(rowData) => ({'data-disabled': !rowData.enabled})}
```

and then target that attribute with css

gdostie: I will do that in a separate PR

gdostie:
I will fix the examples in my next PR

FelixBlaisThon:
BRRRRRRRRRRRRRRRAVIIIIIISSSSSSIIIIIIIIMOOOOOOOOOOOOO Gaelo

PR #3719, opened 2024-05-10
feat(mantine): add getRowAttributes prop on table

gdostie:
### Proposed Changes

Added a prop on the table called `getRowAttributes`. This prop is useful to provide some attributes on the `<tr>` tag that depend on the row data, for example if you need to disable one row.

None

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/table/Table.types.ts
+++ b/packages/mantine/src/components/table/Table.types.ts
@@ -19,6 +19,10 @@ export interface TableLayoutProps<TData = unknown> {
      * @param datum the row for which the children should be generated.
      */
     getExpandChildren?: (datum: TData) => ReactNode;
+    /**
+     * Function that can be used to add additional attributes on rows
+     */
+    getRowAttributes?: (datum: TData) => Record<string, unknown>;
 }
 
 export interface TableLayout {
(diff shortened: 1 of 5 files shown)


dmgauthier (approved):
Should we have some kind of quick README on how to use this ?

gdostie:
> Should we have some kind of quick README on how to use this ?

You will, in the changelogs of the release. For now we are merging in the branch next

packages/mantine/src/components/table/Table.types.ts
```diff
@@ -19,6 +19,10 @@ export interface TableLayoutProps<TData = unknown> {
      * @param datum the row for which the children should be generated.
      */
     getExpandChildren?: (datum: TData) => ReactNode;
+    /**
+     * Function that can be used to add additional attributes on rows
+     */
+    getRowAttributes?: (datum: TData) => Record<string, unknown>;
```
FelixBlaisThon, line 25: I think the phrasing might be confusing, because the method says `getRowAttribute` where the definition says `add`. Could it be `setRowAttributes` ? Or were you thinking of it as getting the attribute passed from `datum: TData` and then we do logic with it ? (which I would be fine with)

GermainBergeron (approved):

FelixBlaisThon (approved):

packages/mantine/src/components/table/Table.types.ts
gdostie: I named it `getRowAttributes` because from the table implementer standpoint you get the attributes from the row data. We already have the prop `getRowId` and `getExpandChildren` so it made sense to me to use the same naming pattern for the same kind of pattern:

```ts
getRowId: (row: TData) => string;
getRowAttributes: (row: TData) => Record<string, unknown>;
getExpandChildren: (row: TData) => ReactNode;
```

FelixBlaisThon: Understandable, have a nice day

gdostie: ink

PR #3720, opened 2024-05-13
refactor(mantine, table): decouple table actions from table header

GermainBergeron:
### Proposed Changes

Rework the Table.Actions
- Developer need to move it outside of the `<TableHeader>` , it will be automatically added inside the header if the prop `showActions` is true.
- Created a new `Table.ActionsColumn` that can be used the same way as the `Table.CollapsibleColumn`, `Table.AccordionColumn` or `Table.SelectableColumn`. I wasn't able to find any UX mockup so take it with a grain of salt, but it's a good base.
- children must now return Table.ActionItem
- Updated the examples and fixed the tests

See above!

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/table/Table.tsx
+++ b/packages/mantine/src/components/table/Table.tsx
@@ -9,7 +9,7 @@ import {
     useReactTable,
 } from '@tanstack/react-table';
 import isEqual from 'fast-deep-equal';
-import {Children, cloneElement, ForwardedRef, ReactElement, useRef} from 'react';
+import {Children, ForwardedRef, ReactElement, useRef} from 'react';
 import {CustomComponentThemeExtend, identity} from '../../utils';
 import {TableLayouts} from './layouts/TableLayouts';
 import {
@@ -17,7 +17,6 @@ import {
     TableActionItemStylesNames,
     TableActions,
     TableActionsStylesNames,
-    TableHeaderActions,
     TableHeaderActionsStylesNames,
 } from './table-actions';
 import {TableActionsColumn} from './table-column/TableActionsColumn';
@@ -139,19 +138,6 @@ export const Table = <T,>(props: TableProps<T> & {ref?: ForwardedRef<HTMLDivElem
     const noData = convertedChildren.find((child) => child.type === TableNoData);
     const actions = convertedChildren.find((child) => child.type === TableActions);
 
-    // clone the header and add the actions to it
-    const headerWithActions = !!header
-        ? cloneElement(header, {
-              ...header.props,
-              children: (
-                  <>
-                      {header.props.children}
-                      <TableHeaderActions>{actions}</TableHeaderActions>
-                  </>
-              ),
-          })
-        : null;
-
     const table = useReactTable({
         data,
         state: {
@@ -253,7 +239,7 @@ export const Table = <T,>(props: TableProps<T> & {ref?: ForwardedRef<HTMLDivElem
                                     {!!header ? (
                                         <tr>
                                             <th style={{padding: 0}} colSpan={table.getAllColumns().length}>
-                                                {headerWithActions}
+                                                {header}
                                             </th>
                                         </tr>
                                     ) : null}
--- a/packages/mantine/src/components/table/Table.tsx
+++ b/packages/mantine/src/components/table/Table.tsx
@@ -1,21 +1,26 @@
-import {Box, Center, Factory, Loader, useProps, useStyles} from '@mantine/core';
+import {Box, Center, Factory, Loader, Menu, useProps, useStyles} from '@mantine/core';
 import {useClickOutside, useMergedRef} from '@mantine/hooks';
 import {
     ColumnDef,
-    Row,
-    RowSelectionState,
     defaultColumnSizing,
     getCoreRowModel,
+    Row,
+    RowSelectionState,
     useReactTable,
 } from '@tanstack/react-table';
 import isEqual from 'fast-deep-equal';
-import {Children, ForwardedRef, ReactElement, useRef} from 'react';
+import {Children, cloneElement, ForwardedRef, ReactElement, useRef} from 'react';
 import {CustomComponentThemeExtend, identity} from '../../utils';
-import classes from './Table.module.css';
-import {TableLayout, TableProps} from './Table.types';
-import {TableProvider} from './TableContext';
 import {TableLayouts} from './layouts/TableLayouts';
-import {TableActions, TableActionsStylesNames} from './table-actions/TableActions';
+import {
+    TableActionItem,
+    TableActionItemStylesNames,
+    TableActions,
+    TableActionsStylesNames,
+    TableHeaderActions,
+    TableHeaderActionsStylesNames,
+} from './table-actions';
+import {TableActionsColumn} from './table-column/TableActionsColumn';
 import {
     TableAccordionColumn,
     TableCollapsibleColumn,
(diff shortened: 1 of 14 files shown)


packages/mantine/src/components/table/table-actions/TableActions.tsx
```diff
@@ -1,11 +1,14 @@
...
 
 import {CustomComponentThemeExtend, identity} from '../../../utils';
-import {TableComponentsOrder} from '../Table';
+import {Table} from '../Table';
 import {useTableContext} from '../TableContext';
 
-export type TableActionsStylesNames = 'actionsRoot' | 'actionsGroup';
+export type TableActionsStylesNames = 'actionsTarget' | 'actionsDropdown' | 'actionsTooltip';
+
+export type TableActionsItems = typeof Table.ActionItem | typeof Table.ActionLabel | typeof Table.ActionDivider;
```
gdostie, line 11: We could change the type here like this

```ts
export type TableActionsItem = typeof Table.ActionItem | typeof Table.ActionLabel | typeof Table.ActionDivider;
export type TableActionsItems = TableActionsItem | Iterable<TableActionsItem>;

children: ((datum: T) => TableActionsItems) | ((data: T[]) => TableActionsItems);
```

This way it would allow setting actions using a fragment instead of the array.

```tsx
{(selectedRow: Person) => (
    <>
        <Table.ActionItem
            primary
            onClick={() => alert(`Action 1 triggered for row: ${selectedRow.id}`)}
            leftSection={<EditSize16Px height={16} />}
        >
            Action 1
        </Table.ActionItem>
        <Table.ActionItem
            primary
            onClick={() => alert(`Action 2 triggered for row: ${selectedRow.id}`)}
            leftSection={<AbTestingSize16Px height={16} />}
        >
            Action 2
        </Table.ActionItem>
    </>
)}
```

The type also accepts a single action without a fragment.

```tsx
{(selectedRow: Person) => (
    <Table.ActionItem
        primary
        onClick={() => alert(`Action 1 triggered for row: ${selectedRow.id}`)}
        leftSection={<EditSize16Px height={16} />}
    >
        Action 1
    </Table.ActionItem>
)}
```

I don't know how complicated it is to support in the table itself thought. I think Children.toArray() already supports this.

packages/mantine/src/components/table/table-actions/TableActionItem.tsx
```diff
@@ -0,0 +1,81 @@
...
+        const {getStyles} = useTableContext();
+        const {classNames, className, style, styles, vars, children, primary, component, ...others} = useProps(
+            'PlasmaTableActionItem',
+            defaultProps,
+            props,
+        );
+
+        if (primary) {
+            return (
+                <Button
```
gdostie, line 58: In the case of the row actions column, it won't be a Button for primaries I think, but more an ActionIcon, and in the case of a context menu it will be perhaps something else. I wonder if we should let the parent decide which component to use (for example TableActionsHeader)

packages/mantine/src/components/table/table-actions/TableActions.tsx
GermainBergeron: That would be really nice, I'm gonna try it out! (I didn't really like the array, it felt against React's way of doing things)

FelixBlaisThon (approved):
Apart from some unused variables (and Gael's comments), this looks pretty good to me :) Demo was tested also

packages/mantine/src/components/table/Table.tsx
```diff
@@ -123,6 +137,23 @@ export const Table = <T,>(props: TableProps<T> & {ref?: ForwardedRef<HTMLDivElem
...
+              children: !!actions ? (
+                  <>
+                      {header.props.children}
+                      <TableHeaderActions>{actionsForHeader}</TableHeaderActions>
+                  </>
+              ) : (
+                  header.props.children
+              ),
+          })
+        : null;
```
gdostie, line 156: We could enhance the logic here so that we automatically add the `<Table.Header />` if it is missing. I am thinking of something like this (naming to review perhaps).

```suggestion
    let headerWithActions = null;
    if (!!header && !!actionsForHeader) {
        headerWithActions = cloneElement(header, {
            ...header.props,
            children: (
                <>
                    {header.props.children}
                    <TableHeaderActions>{actionsForHeader}</TableHeaderActions>
                </>
            ),
        });
    } else if (!header && !!actionsForHeader) {
        headerWithActions = (
            <Table.Header>
                <TableHeaderActions>{actionsForHeader}</TableHeaderActions>
            </Table.Header>
        );
    } else if (!!header && !actionsForHeader) {
        headerWithActions = header;
    }
```

packages/mantine/src/components/table/table-actions/TableActions.tsx
```diff
@@ -1,33 +1,40 @@
...
+export interface TableActionsProps<T> {
     /**
      * Function that return components for the selected row or selected rows when multi row selection is enabled
      *
      * @param datum the data of the selected row(s)
      * @example
      * <Table.Actions<MyType>>
-     *     {(datum: MyType) => (
-     *         <Button
+     *     {(datum: MyType) => [
```
gdostie, line 21:
```suggestion
     *     {(datum: MyType) => 
```

packages/mantine/src/components/table/table-actions/TableActions.tsx
```diff
@@ -1,33 +1,40 @@
...
      *             to={`edit/${datum.id}`}
      *             leftIcon={<EditSize24Px />}
-     *             variant="subtle"
-     *             color="navy.8"
      *         >
      *             Edit
-     *         </Button>
-     *     )}
+     *         </Table.ActionItem>
+     *     ]}
```
gdostie, line 29:
```suggestion
     *     }
```

packages/website/src/examples/layout/Table/TableLayouts.demo.tsx
```diff
@@ -83,15 +82,18 @@ const Demo = () => {
...
-                <Table.Actions>
-                    {(selected) => (
-                        <Button variant="subtle" onClick={() => alert(`Clicked ${selected.firstName}`)}>
-                            Action 1
-                        </Button>
-                    )}
-                </Table.Actions>
-            </Table.Header>
+            <Table.Actions>
+                {(selected: Person) => [
```
gdostie, line 86:
```suggestion
                {(selected: Person) => (
```

packages/website/src/examples/layout/Table/TableLayouts.demo.tsx
```diff
@@ -83,15 +82,18 @@ const Demo = () => {
...
+                    <Table.ActionItem
+                        primary
+                        onClick={() => alert(`Clicked ${selected.firstName}`)}
+                        leftSection={<EditSize16Px height={16} />}
+                    >
+                        Action 1
+                    </Table.ActionItem>,
+                ]}
```
gdostie, line 94:
```suggestion
                    </Table.ActionItem>
                )}
```

gdostie (approved):

packages/mantine/src/components/table/Table.tsx
GermainBergeron: I simplified it a bit since I can directly use `actions` instead of `actionsForHeader`. Do you think it's still worth it to automatically add the header?

gdostie: Does this mean we can avoid doing this now? 
image

GermainBergeron: We already have some tables with the actions in the rows without a `Table.Header` in the admin-ui. I think it's a legit case

gdostie: Ah yeah ok makes sense

gdostie (approved):

PR #3722, opened 2024-05-16
feat(mantine): pass down row object to getRow* props

gdostie:
### Proposed Changes

Adjusting Table's `getRowAttributes` and `getExpandChildren` props so that they all match the same arguments signature. I chose the signature of the existing `getRowId` prop, which is the following:

```ts
getRowId?: (originalRow: TData, index: number, parent?: Row<TData>) => string;
```

Doing so has a few benefits:
- It gives more flexibility to the Table implementer. For instance since you have access to the row object (3rd param) you can gain access to tanstack table's row API. So you can do stuff like 
```tsx
<Table
    getExpandChildren={(rowData, index, row) => row.getIsExpanded() 
        ? <ExpensiveContentOnlyFetchedWhenExpanded id={rowData.id} /> 
        : <Loader />
    }
/>
```
- It brings consistency to the table props `get*`

It can break custom layout implementations because the layout props now expect 3 arguments to be passed down the affected props.

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/table/Table.tsx
+++ b/packages/mantine/src/components/table/Table.tsx
@@ -156,7 +156,7 @@ export const Table = <T,>(props: TableProps<T> & {ref?: ForwardedRef<HTMLDivElem
         manualPagination: options?.getPaginationRowModel === undefined,
         enableMultiRowSelection: !!store.multiRowSelectionEnabled,
         getRowId,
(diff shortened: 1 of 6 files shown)


gdostie:
I crammed in another small fix while at it. It removes the inconsistency between non-sortable and sortable headers. Sortable headers would wrap its children into a Group and non-sortable header would not. Example:

image

GermainBergeron (approved):

PR #3728, opened 2024-05-22
fix(tableActions): changed interface so it extends button

FelixBlaisThon:
### Proposed Changes

I've extended ButtonProps for the TableActionItemsProps so that we can have access to `disabled, leftSection, rightSection, disabledTooltip, etc`. 

I've omitted some props to redefine them with its own description and also to make leftSection mandatory (which is optional in the Button implementation)

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/table/table-actions/TableActionItem.tsx
+++ b/packages/mantine/src/components/table/table-actions/TableActionItem.tsx
@@ -1,12 +1,17 @@
-import {BoxProps, CompoundStylesApiProps, Menu, PolymorphicFactory, polymorphicFactory, useProps} from '@mantine/core';
+import {CompoundStylesApiProps, Menu, PolymorphicFactory, polymorphicFactory, useProps} from '@mantine/core';
 import {ReactNode} from 'react';
-import {Button} from '../../button';
+import {Button, ButtonProps} from '../../button';
 import {useTableContext} from '../TableContext';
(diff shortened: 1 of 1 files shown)


packages/mantine/src/components/table/table-actions/TableActionItem.tsx
```diff
@@ -1,12 +1,17 @@
...
 import {useTableContext} from '../TableContext';
 import {useTableActionContext} from './TableActionContext';
 
 export type TableActionItemStylesNames = 'actionItemRoot';
 
-export interface TableActionItemProps extends BoxProps, CompoundStylesApiProps<TableActionItemFactory> {
+export interface TableActionItemProps
+    extends Omit<
+            ButtonProps,
+            'classNames' | 'styles' | 'vars' | 'variant' | 'leftSection' | 'rightSection' | 'disabled'
```
GermainBergeron, line 12: Since we now extend ButtonProps we can remove `rightSection` and `disabled` from the interface and from the Omit. It will use Mantine's type

FelixBlaisThon: I agree, I kept it because our JSDoc gave a different definition, but after I looked at it it's pretty much the same :P

GermainBergeron (approved):

PR #3764, opened 2024-06-10
fix(mantine): make sure rowSelection row data stays fresh

gdostie:
### Proposed Changes

`table.state.rowSelection` is an object whose key is the row id and the value is the row data. Before this change, the row data in this object would remain the same even if the `data` prop changed. In this PR, I am adding an effect which keeps this `rowSelection` object in sync with the latest data.

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/table/__tests__/TableActions.spec.tsx
+++ b/packages/mantine/src/components/table/__tests__/TableActions.spec.tsx
@@ -1,11 +1,11 @@
 import {ColumnDef, createColumnHelper} from '@tanstack/table-core';
 import {render, screen, userEvent, waitFor, within} from '@test-utils';
+import {useState} from 'react';
 
 import {Table} from '../Table';
 import {useTable} from '../use-table';
 
 type RowData = {name: string};
-
 const columnHelper = createColumnHelper<RowData>();
 const columns: Array<ColumnDef<RowData>> = [columnHelper.accessor('name', {enableSorting: false})];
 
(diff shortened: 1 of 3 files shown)


aveilleux (approved):

packages/mantine/src/components/table/__tests__/TableActions.spec.tsx
```diff
@@ -151,6 +143,56 @@ describe('Table.Actions', () => {
...
+                                ) : null,
+                        },
+                        {
+                            group: '$$primary',
+                            component:
+                                selected[0].status === 'eaten' ? (
+                                    <Table.ActionItem key="trash">Throw away</Table.ActionItem>
+                                ) : null,
+                        },
+                    ]}
```
GermainBergeron, line 180: Any reason to have two different actions?

```suggestion
                    getRowActions={(selected: Food[]) => [
                        {
                            group: '$$primary',
                            component:
                                selected[0].status === 'fresh' 
                                ? (
                                    <Table.ActionItem key="eat" onClick={eatFruit}>
                                        Eat
                                    </Table.ActionItem>
                                ) 
                                : <Table.ActionItem key="trash">Throw away</Table.ActionItem>,
                        },
                    ]}
```

GermainBergeron (approved):

toofff (approved):
LGTM

packages/mantine/src/components/table/__tests__/TableActions.spec.tsx
gdostie: It was to better mimic the reality of the admin 👾 You will often have both actions but protected by guards rather than conditionally pushing into the array. I can do what you suggest if you prefer.

GermainBergeron: I don't mind either solutions, go with what you prefer. In the admin it could be a Guard with a fallback which would be closer to what I propose

PR #3791, opened 2024-07-02
feat(mantine, prompt): update look of prompt component

GermainBergeron:
### Proposed Changes

Updated the look of the prompt component

| Before | After | Mockup |
| --- | --- | --- |
| image | image | image |

The prompt now uses the compound components (Modal.Root, Modal.Header, etc) from Mantine's Modal instead of using Modal

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/prompt/Prompt.module.css
+++ b/packages/mantine/src/components/prompt/Prompt.module.css
@@ -1,45 +1,35 @@
 .root {
-    &[data-variant='success']&[data-size] {
-        .header {
-            background-color: var(--mantine-color-lime-6);
-        }
+    --prompt-icon-size: 88px;
+
+    .header {
+        border-bottom: none;
+        align-items: center;
     }
-    &[data-variant='warning']&[data-size] {
-        .header {
-            background-color: var(--mantine-color-yellow-5);
-        }
-    }
-    &[data-variant='critical']&[data-size] {
-        .header {
-            background-color: var(--mantine-color-red-6);
-        }
-    }
-    &[data-variant='info']&[data-size] {
-        .header {
-            background-color: var(--mantine-color-navy-5);
-        }
+
+    .body {
+        padding: 0 var(--mantine-spacing-lg) var(--mantine-spacing-lg)
+            calc(var(--prompt-icon-size) + (2 * var(--mantine-spacing-lg)));
     }
 }
 
-.title {
-    color: var(--mantine-color-white);
+.icon {
+    width: var(--prompt-icon-size);
+    height: var(--prompt-icon-size);
 }
 
 .header {
+    gap: var(--mantine-spacing-lg);
     width: 100%;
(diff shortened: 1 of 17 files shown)


packages/mantine/src/components/prompt/Prompt.module.css
```diff
@@ -1,45 +1,35 @@
...
-            background-color: var(--mantine-color-yellow-5);
-        }
-    }
-    &[data-variant='critical']&[data-size] {
-        .header {
-            background-color: var(--mantine-color-red-6);
-        }
+    --prompt-icon-size: 88px;
+
+    .header {
```
gdostie, line 4: Why is this style not in `.header {}` selector? (vs currently `.root .header {}`)

packages/mantine/src/components/prompt/Prompt.module.css
```diff
@@ -1,45 +1,35 @@
...
+        border-bottom: none;
+        align-items: center;
     }
-    &[data-variant='info']&[data-size] {
-        .header {
-            background-color: var(--mantine-color-navy-5);
-        }
+
+    .body {
```
gdostie, line 9: Same question here

packages/mantine/src/components/prompt/Prompt.tsx
```diff
@@ -1,43 +1,121 @@
...
+    Modal,
+    ModalRootProps,
+    ModalStylesNames,
+    StylesApiProps,
+    useProps,
+    useStyles,
+} from '@mantine/core';
 import {Children, ReactElement, ReactNode} from 'react';
-import PromptClasses from './Prompt.module.css';
+import classes from '../header/Header.module.css';
```
gdostie, line 15: I think the import is wrong here

```suggestion
import classes from './Prompt.module.css';
```

packages/mantine/src/components/prompt/icons/critical.svg
```diff
@@ -0,0 +1,29 @@
+<svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 88 90">
+  <g filter="url(#a)">
```
gdostie, line 2: With the iconography I remember we had some problems because the ids (here `a`) would clash with each other since ids are supposed to be unique in the DOM. Perhaps we should add some kind of prefix? Maybe something like `critical-img-a` would be safer.

(same comment applied to other icons)

gdostie (approved):

packages/mantine/src/components/prompt/Prompt.module.css
GermainBergeron: Without the scoping inside the `.root` the style is not applied. It's due to the fact that the prompt renders a modal and the modal style is applied after the prompt's

gdostie: ah I see, ok

gdostie (approved):

PR #3820, opened 2024-07-16
ci: eslint v9 + flat config + fix errors

gdostie:
### Proposed Changes

- Updated eslint to version 9
- Updated all eslint plugins we use
- Migrated all our configs to flat configs (see https://eslint.org/docs/latest/use/configure/migration-guide)
- Had to replace `eslint-plugin-import` by `eslint-plugin-import-x` because of https://github.com/import-js/eslint-plugin-import/issues/2948
- tried enforcing typescript-eslint recommended config and fixed a few errors but there was just too many to fix. I didn't want this PR to grow too large so I reverted back, but I think we should do it in another PR. I kept the fix I already did in this PR

Concrete differences:
- no more eslint config in each package, only one config at the root level
- no more `tsconfig.eslint.json`

### Acceptance Criteria


package.json
```diff
@@ -43,20 +43,32 @@
         "IE 11",
         "not dead"
     ],
+    "dependencies": {
+        "eslint-plugin-import-x": "3.0.1"
+    },
```
GermainBergeron, line 48: Shouldn't that be a devDependency?

GermainBergeron (approved):

PR #3857, opened 2024-08-16
fix(mantine, skeleton): increase z-index

GermainBergeron:
### Proposed Changes

This fixes the issue with buttons, indicator, etc being visible when wrapped by a Skeleton 

https://mantine.dev/styles/css-variables/#z-index-variables

The z-index changed but it shouldn't break anything ™️ 

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/styles/Skeleton.module.css
+++ b/packages/mantine/src/styles/Skeleton.module.css
@@ -1,11 +1,11 @@
 .root {
-    &[data-visible] {
+    &:where([data-visible]) {
         &::before {
-            z-index: unset;
+            z-index: var(--mantine-z-index-overlay);
         }
 
         &::after {
-            z-index: unset;
+            z-index: calc(var(--mantine-z-index-overlay) + 1);
         }
     }
 }


dmgauthier (approved):

packages/mantine/src/styles/Skeleton.module.css
```diff
@@ -1,11 +1,11 @@
 .root {
-    &[data-visible] {
+    &:where([data-visible]) {
```
dmgauthier, line 2: I am curious, what is the effect of using [:where](https://developer.mozilla.org/en-US/docs/Web/CSS/:where) here.

gdostie (approved):

packages/mantine/src/styles/Skeleton.module.css
GermainBergeron: I wanted to override Mantine's selector so I re-used the same thing https://github.com/mantinedev/mantine/blob/master/packages/%40mantine/core/src/components/Skeleton/Skeleton.module.css#L24

PR #3945, opened 2024-11-06
feat(mantine): ChildForm component

aboissinot-coveo:
### Proposed Changes

Added a new `ChildForm` component. This pattern for forms is used at a couple of places in the Admin-UI but each place has its own implementation with different styling. 

On our side, we have a need for it for two features we are currently working on, so I thought it would be a good time to add it.

I validated with UX to make sure the styling was up to their liking. 

[Figma](https://www.figma.com/design/NRBnfgMxgVs7h6qwP6ADBV/Elevation-%26-Focus-FX---PDS?node-id=804-297&node-type=frame&m=dev).🎨 

Screenshot 2024-11-07 at 10 43 51 AM

[Demo](https://plasma.coveo.com/feature/SEARCHAPI-11252-child-form/form/ChildForm). 🧪 

Screenshot 2024-11-07 at 10 48 37 AM

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/child-form/ChildForm.tsx
+++ b/packages/mantine/src/components/child-form/ChildForm.tsx
@@ -50,10 +50,10 @@ export const ChildForm = polymorphicFactory<ChildFormFactory>((props, ref) => {
 
     return (
         <Collapse ref={ref} {...others} {...getStyles('root')}>
-            <Paper bg="gray.1" shadow="sm" p="lg" {...getStyles('paper')}>
+            <Paper bg="gray.1" p="md" {...getStyles('paper')}>
                 {(title || description) && (
                     <Stack gap={0} mb="md">
-                        {title && <Title order={4}>{title}</Title>}
+                        {title && <Title order={5}>{title}</Title>}
                         {description && <Text c="gray.7">{description}</Text>}
                     </Stack>
                 )}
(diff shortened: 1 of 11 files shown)


packages/mantine/src/components/child-form/ChildForm.tsx
```diff
@@ -0,0 +1,70 @@
...
+    return (
+        <Collapse ref={ref} {...others} {...getStyles('root')}>
+            <Paper bg="gray.0" p="md" {...getStyles('paper')}>
+                {(title || description) && (
+                    <Stack gap={0} mb="md">
+                        {title && <Title order={5}>{title}</Title>}
+                        {description && <Text c="gray.7">{description}</Text>}
+                    </Stack>
+                )}
+                <Stack gap="md">{children}</Stack>
```
toofff, line 66: Should we make this gap configurable from outside the component?

toofff (approved):
Nice component!

packages/mantine/src/components/child-form/ChildForm.tsx
aboissinot-coveo: Thought about it but was under the impression that it could reduce the consistency across the panels that use it. But I'm not entirely against it if some use-cases require it to be configurable. 🤷

toofff: Okay, let's go with that for now.

PR #3963, opened 2024-11-22
fix(WEB-7276): avoid sliver of content being viewable below the `StickyFooter`

jkatofskyCoveo:
### Proposed Changes
Put down the sticky footer down by one pixel to avoid the visual issue reported [here](https://github.com/coveo-platform/admin-ui/pull/13077#pullrequestreview-2430781237)

### DIFF
--- a/packages/mantine/src/components/sticky-footer/StickyFooter.module.css
+++ b/packages/mantine/src/components/sticky-footer/StickyFooter.module.css
@@ -1,6 +1,6 @@
 .root {
     position: sticky;
-    bottom: 0;
+    bottom: -1px;
     z-index: 10;
     background-color: white;
     padding: var(--mantine-spacing-md) var(--mantine-spacing-lg);


FelixBlaisThon:
I couldn't see the sliver in your previous PR. was it on a 4k screen ?

jkatofskyCoveo:
@FelixBlaisThon, hmm, it's finnicky to replicate, but I don't think it has to do with 4K, rather I think it's a rounding error with how CSS calculates the bottom of a sticky element or something. Here is the modal in dev on my laptop screen (1920 X 1200). You can see the sliver
image

jkatofskyCoveo:
Haha no you're not annoying, it's this bug that is annoying @FelixBlaisThon 😅

Windows 11, Chrome 131.

The flickering footer is indeed another issue, but for those of us who can replicate the sliver of content underneath, the flickering happens with or without the `bottom: -1px` 😕

FelixBlaisThon:
> Haha no you're not annoying, it's this bug that is annoying @FelixBlaisThon 😅
> 
> Windows 11, Chrome 131.
> 
> The flickering footer is indeed another issue, but for those of us who can replicate the sliver of content underneath, the flickering happens with or without the `bottom: -1px` 😕

Ok. I think we have to maybe check that before merging any fix (at least remove any flickering, with or without the `bottom: -1px`.. I will create a JIRA so that we can investigate it. From the time being, I would suggest to close this PR and link the JIRA to this one https://coveord.atlassian.net/browse/ADUI-10401

PR #3985, opened 2024-12-16
feat(mantine): add syncWithUrl option to table hook

gdostie:
### Proposed Changes

Add `syncWithUrl` option to `useTable` hook. When true, the table will sync its state with the url using search parameters.

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/table/use-table.ts
+++ b/packages/mantine/src/components/table/use-table.ts
@@ -3,13 +3,12 @@ import {type ExpandedState, type PaginationState, type SortingState} from '@tans
 import defaultsDeep from 'lodash.defaultsdeep';
 import {Dispatch, SetStateAction, useCallback, useMemo, useState} from 'react';
 import {type DateRangePickerValue} from '../date-range-picker';
+import {useUrlSyncedState} from './use-url-synced-state';
 
 // Create a deeply optional version of another type
-type PartialDeep<T> = T extends object
-    ? {
-          [P in keyof T]?: PartialDeep<T[P]>;
-      }
-    : T;
+type DeepPartial<T> = {
+    [P in keyof T]?: T[P] extends object ? DeepPartial<T[P]> : T[P];
+};
 
 export interface TableState<TData = unknown> {
     /**
@@ -166,7 +165,7 @@ export interface UseTableOptions<TData = unknown> {
     /**
      * Initial state of the table.
      */
-    initialState?: PartialDeep<TableState<TData>>;
+    initialState?: DeepPartial<TableState<TData>>;
     /**
      * Whether rows can be selected.
      *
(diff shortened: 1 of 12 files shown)


packages/website/src/examples/layout/Table/TableClientSide.demo.tsx
```diff
@@ -67,6 +67,7 @@ const Demo = () => {
             pagination: {pageSize: 5},
             totalEntries: data.length,
         },
+        syncWithUrl: true,
```
gdostie, line 70: Go try the url sync with the client side table in the demo!

packages/mantine/src/components/table/use-table.ts
```diff
@@ -213,21 +219,111 @@ export const useTable = <TData>(userOptions: UseTableOptions<TData> = {}): Table
     const options = defaultsDeep({}, userOptions, defaultOptions) as UseTableOptions<TData>;
     const initialState = defaultsDeep({}, options.initialState, defaultState) as TableState<TData>;
 
-    const [pagination, setPagination] = useState<TableState<TData>['pagination']>(initialState.pagination);
+    // synced with url
+    const [pagination, setPagination] = useUrlSyncedState<TableState<TData>['pagination']>({
+        initialState: initialState.pagination,
+        serializer: ({pageIndex, pageSize}) => [
+            ['page', (pageIndex + 1).toString()],
```
gdostie, line 226: Converting to 1-based indexing

packages/mantine/src/components/table/use-table.ts
```diff
@@ -213,21 +219,120 @@ export const useTable = <TData>(userOptions: UseTableOptions<TData> = {}): Table
     const options = defaultsDeep({}, userOptions, defaultOptions) as UseTableOptions<TData>;
     const initialState = defaultsDeep({}, options.initialState, defaultState) as TableState<TData>;
 
-    const [pagination, setPagination] = useState<TableState<TData>['pagination']>(initialState.pagination);
+    // synced with url
+    const [pagination, setPagination] = useUrlSyncedState<TableState<TData>['pagination']>({
```
GermainBergeron, line 223: It looked a bit daunting at first with all those `useUrlSyncedState` but I like that each property is scoped to a single call.

Have you check to make sure it doesn't re-render 100x?

packages/mantine/src/components/table/use-url-synced-state.ts
```diff
@@ -0,0 +1,70 @@
...
+    } else {
+        url.searchParams.set(key, value);
+    }
+    window.history.replaceState(null, '', url.toString());
+};
+
+const getSearchParams = (): URLSearchParams => {
+    const url = new URL(window.location.href);
+    return url.searchParams;
+};
```
GermainBergeron, line 16: Really simple, I like it

GermainBergeron (approved):

packages/mantine/src/components/table/use-table.ts
gdostie: Yeah the fact that each property of the state serializes to a different shape does not help here for the conciseness of the code. What I like with this implementation is that the logic for one state property is all in the same place and does not influence the other properties.

gdostie: It does not influence the number of rendering loops, because it does not set the state anymore often than a regular setState. The only time the url can influence the state is on initial render before calling `useState` so it does not create any additional rendering loops there either 🎉

FelixBlaisThon (approved):
very wow, gg

FelixBlaisThon:
Demo works too from what I've tested

PR #3991, opened 2025-01-07
feat(mantine): allow table url sync custom functions

gdostie:
### Proposed Changes

In applications where a custom router is used, for instance a hash router from react-router-dom, you need the ability to customize how search params are set in the url when syncing the table state with the url.

In this PR, I added a possibility of passing down custom functions to the useTable hook to control how search params are set in the url.

### Acceptance Criteria


gdostie:
Better solution incoming

PR #4001, opened 2025-01-21
feat(modal): implement "@coveord/plasma-mantine/Modal"

sfeng-coveo:
### Proposed Changes

Implement a new feature for the library: `@coveord/plasma-mantine/Modal`.
The feature mostly inherits from the modal feature in `@mantine/core`, except that it supports a `Modal.Footer`, which uses our `StickyFooter` implementation.
Given we have two ways of using modal footers (i.e. [in a `<Modal />` vs in a `<Modal.Root />`](http://localhost:3000/layout/Modal)) and that the two require different styling, there is a boolean value `sticky` on the `Modal.Footer` component to indicate whether the footer should be pushed down ("stick") to the bottom or not.

Removed the unused doc for `ModalWizard` along the way.

- The `variant` prop of `StickyFooter` is marked as deprecated, aiming to eventually be removed.
  - From now on, users should use `Modal.Footer` for all "modal footers", and the default `StickyFooter` without variant for anything else.

### Acceptance Criteria


gdostie (approved):

packages/mantine/src/components/modal/ModalFooter.tsx
```diff
@@ -0,0 +1,45 @@
...
+import clsx from 'clsx';
+import {Factory, factory} from '@mantine/core';
+import {StickyFooter, StickyFooterProps, StickyFooterStylesNames} from '../sticky-footer';
+import classes from './Modal.module.css';
+
+export interface ModalFooterProps extends Omit<StickyFooterProps, 'variant'> {
+    /**
+     * If the footer is sticky, its margin will be adjusted to counteract the padding of the Modal.Body, ensuring the footer visually sticks to the bottom of the modal.
+     */
+    sticky?: boolean;
```
toofff, line 11: This prop is not tested, is this correct?

sfeng-coveo: It is not, you're right. I'll cover it.

packages/mantine/src/components/modal/ModalFooter.tsx
```diff
@@ -0,0 +1,45 @@
...
+    props: ModalFooterProps;
+    ref: HTMLDivElement;
+    stylesNames: ModalFooterStylesNames;
+}>;
+
+export const ModalFooter = factory<ModalFooterFactory>(({sticky, ...props}, ref) => {
+    // ADUI-10401: when the footer's height is an odd number, the footer would align towards the top and not fill the bottom completely
+    // the following workaround essentially ensures that the footer's height is always an even number, on the basis of best-effort (i.e. when ref is available)
+    if (ref === null) {
+        ref = useRef<HTMLDivElement>();
```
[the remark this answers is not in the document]
sfeng-coveo: After reading into it, I'm surprised too.
I tried some other conditions and hooks, it appears that the rule is simply not being triggered in all cases, although looking at the config we should have everything set up correctly.
Without digging deeper into it, I moved the ref to the top level.

packages/mantine/src/components/modal/ModalFooter.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export type ModalFooterFactory = Factory<{
+    props: ModalFooterProps;
+    ref: HTMLDivElement;
+    stylesNames: ModalFooterStylesNames;
+}>;
+
+export const ModalFooter = factory<ModalFooterFactory>(({sticky, ...props}, ref) => {
+    // ADUI-10401: when the footer's height is an odd number, the footer would align towards the top and not fill the bottom completely
+    // the following workaround essentially ensures that the footer's height is always an even number, on the basis of best-effort (i.e. when ref is available)
```
jkatofskyCoveo, line 24: Like my other comment about comments, I think we can do without these

maybe `adjustHeight`  could be more descriptively named, so it makes clear that this is a workaround regarding odd numbers

jkatofskyCoveo (commented):
Some minor things. But _great_ job with this!

packages/mantine/src/components/modal/__tests__/ModalFooter.spec.tsx
```diff
@@ -0,0 +1,37 @@
...
+                <div>im the children</div>
+            </Modal.Footer>,
+        );
+
+        const footer = screen.getByText('im the children').parentElement;
+        expect(footer.className).contains('modalFooterSticky');
+    });
+    // ADUI-10401: when the footer's height is an odd number, the footer would align towards the top and not fill the bottom completely
+    // this tests an workaround we've implemented to solve the issue
+    it('has an even height value', () => {
```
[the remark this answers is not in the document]
sfeng-coveo: I can remove the Jira number.

I however don't think we can really do without a comment explaining the issue. For us who've investigated into the issue, we know that the cause of this issue is quite intricate, and the effect on UI is easy to miss when you don't pay attention (but annoying when you do).

Let's say we have a more generalized test name and no comment. What I'm worried about is that, weeks or even months later, other developers start working on these files and fails to understand why this is implemented:

1. They look at the test name and understand that it's a workaround to stick the modal to the bottom, but still don't know why this is necessary;
2. They remove the piece of fix and try to observe what's different (this is what I would do if I'm not sure about the utility of a piece of code). On a harsh morning and because they don't know what specifically to look at, they didn't catch the 1px issue and believes everything is the same. Or even their modal just happens to naturally have an even-height footer, then they'd not see the problem at all.
3. The developer thinks this piece of code is useless or deprecated and remove it.

For this reason, I think we should be very explicit about why we're implementing this fix and how it solves the issue, so that any developer looking at this can replicate the issue. Given this, I don't think any reasonable variable naming could be descriptive enough, and the test naming would come out a bit too long.

packages/mantine/src/components/modal/ModalFooter.tsx
sfeng-coveo: Similar comment as https://github.com/coveo/plasma/pull/4001#discussion_r1926028413, and here I don't think any naming within reason would be descriptive enough to explain the issue.

packages/mantine/src/components/modal/ModalFooter.tsx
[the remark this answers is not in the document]
gdostie: Thanks for investigating 🙏

packages/mantine/src/components/modal/ModalFooter.tsx
jkatofskyCoveo: That may well be the case, and I'll let you use your judgement after my response in the other thread 😄 

But I do think there could be something better than `adjustHeight`; just the height _in which way_, and for _what purpose_.?

sfeng-coveo: I changed the function's name to `ensuresFooterHasEvenHeight` to answer the "what" part of the question.

As for the "why" part, my personal philosophy is that a function explains whatever it does, and callers of the function decides how to use it, and the function itself shouldn't dictate how others use it. For example, the famous Fourier transform algorithm was originally invented for solve heat transfer problem, but is later widely used for image denoising and compression, which the original creators wouldn't think of (a bit unrelated to this case, but just to explain what I mean).

Given we talked about not having too much comments for the height issue, I think we can leave the function's name as proposed since there's definitely enough visibility to the issue within this PR.

PR #4002, opened 2025-01-21
fix(prompt): filter footer children in one pass

sfeng-coveo:
While working on #4001 something in `Prompt` slightly bothered me so I decided to make another fix.

### Proposed Changes

The `Prompt` component looks for children of type footer and ensures it is rendered outside of the `Modal.Body`. It currently does it in three passes of the `children` array:
- One pass to convert the children element to plain array;
- One pass to filter everything not footer;
- One pass (worst case) to find the first footer component;

This seems unnecessary as the same thing can be done with only one pass and the same amount of code, as suggested by this Pull Request.

Additionally, since the current code looks for the **first** footer within all children, other footers will be lost if users have used more than one `<Prompt.Footer />` component (they *shouldn't*, that I agree).

- Users who used more than one `<Prompt.Footer />` component in the same `<Prompt />` would originally find only one of their footers rendering in the DOM, will now see all their footers rendering vertically.
  - While we're definitely not expecting them to use `<Prompt.Footers />` like this, we don't explicitly forbid it neither. If anyone does this for whatever reason, I find rendering all components make more sense than just discard the elements.
  - Since we're not encouraging the use of footers in such way, this path is not covered by tests.

### Acceptance Criteria


packages/mantine/src/components/prompt/Prompt.tsx
```diff
@@ -119,7 +121,7 @@ export const Prompt = factory<PromptFactory>((_props, ref) => {
                     <Modal.Body {...getStyles('body', stylesApiProps)}>
                         <Box {...getStyles('inner', stylesApiProps)}>{otherChildren}</Box>
                     </Modal.Body>
-                    {footer}
+                    {...footers}
```
gdostie, line 124: React handles rendering `ReactElement[]` on its own

```suggestion
                    {footers}
```

gdostie (approved):

PR #4011, opened 2025-02-07
fix(table): correct useUrlSyncedState serialization logic

rphilippen:
### Proposed Changes

I've added a feature to the `useUrlSyncedState` hook for the table, that some parameters will always be added to the URL. In short, the idea behind this is to do this for `from` and `to`, since the default/initial values for those fields are often relative to now. By always serializing, sharing and opening the URL leads to seeing the same range, even if days later. (For more details, see #3996).

I **thought** I did two things:
  - Manually test this in Admin-UI.
  - Add proper unit test coverage.

For the manual test part, I think my hacked "local link" set-up wasn't working as well as I thought it did, and I probably tested with some weird mixed old-version and new-version states. For the unit test coverage, I honestly don't know why I completely missed that. The only excuse I have is that I've been context switching a bit too much. I could have sworn I wrote tests, but reality says no. So I'm sorry for delivering sloppy work (again 😞).

There was an observed issue with the hook in Data Health (currently the only UI that uses this new feature, to my knowledge). If you'd open a URL, it would just go to the default range ("last week"), and actually also update the URL to reflect that. Or, in other words, deserializing from the URL of the initial state didn't work as it should. It would write the default value to the URL before deserializing.

I set out to fix the deserialization, and noticed another gap I had: I completely ignored the `sync` flag in that logic, which should toggle off any param synchronization. This has been fixed, and there is a test that will fail if you remove the `sync &&` part from the `if (sync && alwaysEmit && value)` (e.g. this behavior is now covered by a test).

While writing the tests, I also observed the hash route tests were lacking. I decided to refactor the tests to use a `describe.each`, to basically parameterize the window.location. This makes the test code more complex (it needs to repeat some of the logic the code under test does, like detecting hash router locations), but it also gives a lot more confidence that the URL handling is correct. Of course I found out my substring logic was flawed for non-hash-router URLS: it would drop the hash value from the URL. This has also been fixed, and breaking this behavior should fail tests.

These changes only fix the behavior of the `useUrlSyncedState` hook, no breaking changes. People may have observed from and to being added to the URL, even if they didn't use the `syncWithUrl` option of the table. That will be corrected by this PR.

### Manual Testing - not done

I've tried very hard to test this in Admin-UI using local linking and other voodoo magic. I can't get it to work. If you have any feedback on how I can manually test this before merging this, I'd appreciate that. I'm confident this makes the hook behave a whole lot better, but I'd like to validate it actually behaves completely as expected in practice too.

### Acceptance Criteria

-   [X] [README.md](https://github.com/coveo/plasma/blob/master/README.md) is adjusted to reflect the proposed changes (if relevant)

### DIFF
--- a/packages/mantine/src/components/table/__tests__/use-url-synced-state.unit.spec.ts
+++ b/packages/mantine/src/components/table/__tests__/use-url-synced-state.unit.spec.ts
@@ -1,117 +1,148 @@
 import {act, renderHook} from '@test-utils';
 import {useUrlSyncedState} from '../use-url-synced-state';
 
-describe('useUrlSyncedState', () => {
-    afterEach(() => {
-        window.history.replaceState(null, '', '/');
-    });
+/**
+ * Split a url into its parts.
+ * Note that this method is purposefully different from the implementation,
+ * and it splits to an object instead of an Array.
+ *
+ * @param href The url to extract the parts from.
+ * @returns The separate parts, all are an empty string if not present.
+ */
+const extractParts = (href: string) =>
+    /^(?<pathname>[^?#]*)(?<search>\?[^#]*|)(?<hash>#[^?]*|)(?<hashSearch>\?.*|)$/.exec(href).groups as {
+        pathname: string;
+        search: string;
+        hash: string;
+        hashSearch: string;
+    };
 
-    it('serializes the state value as url parameter when the state changes', () => {
-        const {result} = renderHook(() =>
-            useUrlSyncedState({
-                initialState: '',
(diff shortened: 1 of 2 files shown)


FelixBlaisThon (approved):
It looks ok to me (sorry for plasma that doesn't want to cooperate on the linking process). I think that the unit test covers a big part of it. I would be ok merging that and see how it goes

rphilippen:
I will need someone to merge this for me, I don't have write access to this repository.

GermainBergeron (approved):

PR #4028, opened 2025-03-17
fix(CodeEditor): format text only on valid input

sfeng-coveo:
### Proposed Changes

Currently, the `CodeEditor` uses the default formatting provided by the Monaco Editor. For `json` language, the editor inherently relies on `vscode-json-languageservice`, which does a best-effort formatting of the text regardless whether the JSON is valid or not.

However, in the scenario where the JSON is invalid, we've observed UI issues caused by the formatting, where the mouse pointer is relocated to the end of the text:

https://github.com/user-attachments/assets/d5766c42-2259-4f30-af3c-32519328475f

Thus we want to disable the best-effort formatting. This is done by overriding the default formatter with a simple `JSON.stringify(JSON.parse(model.getValue()), null, tabSize)`, which will fail in invalid scenarios and thus does nothing.

None since change is behavioral.

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/code-editor/CodeEditor.tsx
+++ b/packages/mantine/src/components/code-editor/CodeEditor.tsx
@@ -249,9 +249,12 @@ export const CodeEditor: FunctionComponent<CodeEditorProps> = (props) => {
                     registerThemes(monaco);
                     editor.onDidFocusEditorText(() => onFocus?.());
                     editor.onDidBlurEditorText(async () => {
-                        if (!hasMonacoErrorRef.current) {
-                            await editor.getAction('editor.action.formatDocument').run();
-                        }
+                        // monaco editor has a timeout of 500ms populating errors, we want to ensure that checking errors happen after that
+                        setTimeout(async () => {
+                            if (!hasMonacoErrorRef.current) {
+                                await editor.getAction('editor.action.formatDocument').run();
+                            }
+                        }, 550);
                     });
                 }}
             />
--- a/packages/mantine/src/components/code-editor/CodeEditor.tsx
+++ b/packages/mantine/src/components/code-editor/CodeEditor.tsx
@@ -249,6 +249,17 @@ export const CodeEditor: FunctionComponent<CodeEditorProps> = (props) => {
                     editor.onDidBlurEditorText(async () => {
                         await editor.getAction('editor.action.formatDocument').run();
                     });
+
+                    if (language === 'json') {
+                        monaco.languages.registerDocumentFormattingEditProvider('json', {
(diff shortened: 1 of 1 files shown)


packages/mantine/src/components/code-editor/CodeEditor.tsx
```diff
@@ -249,6 +249,17 @@ export const CodeEditor: FunctionComponent<CodeEditorProps> = (props) => {
                     editor.onDidBlurEditorText(async () => {
                         await editor.getAction('editor.action.formatDocument').run();
                     });
```
GermainBergeron, line 251: I was thinking about the solution and I think we should maybe go another route. What do you think of formatting only if there are no errors? I was able to achieve it this way:

```suggestion
                    editor.onDidBlurEditorText(async () => {
                        const markers = monaco.editor.getModelMarkers({resource: editor.getModel().uri});
                        const hasErrors = markers.some(
                            (marker) => marker.severity === loader.__getMonacoInstance().MarkerSeverity.Error,
                        );
                        if (!hasErrors) {
                            await editor.getAction('editor.action.formatDocument').run();
                        }
                    });
```

sfeng-coveo: Yeah I like this solution.
I'll quickly test it to see if there are no other unintended behaviors, then I'll switch implementation.

GermainBergeron: I feel like it will be a better UX, assuming the same thing can happen in other langages. Let's say you're editing a file of 200 lines and have an error at line 10, you don't want your cursor to jump at the end of the file.

I think prettier does the same thing and only format when your code is valid

sfeng-coveo: I've put the condition to rely on `hasMonacoError` instead, since those are the language syntax errors and any text can still be formatted properly with non-syntax errors.

packages/mantine/src/components/code-editor/CodeEditor.tsx
```diff
@@ -247,7 +245,9 @@ export const CodeEditor: FunctionComponent<CodeEditorProps> = (props) => {
                     registerThemes(monaco);
                     editor.onDidFocusEditorText(() => onFocus?.());
                     editor.onDidBlurEditorText(async () => {
-                        await editor.getAction('editor.action.formatDocument').run();
+                        if (!hasMonacoError) {
+                            await editor.getAction('editor.action.formatDocument').run();
+                        }
```
GermainBergeron, line 250: This didn't work for me when I tried locally. It also formats in the demo when there are errors

PR #4037, opened 2025-04-04
feat(header): refactor of the Header.Actions compound to become Header.Right

dmgauthier:
#### Context
We need to achieve something like :point_down: in the Coveo Knowledge Hub
image

I think I could achieve this throught the `Header.Action` compound, since its location is proper and it can take any react node.
But it wouldn't feel right and make the implementation code confusing.

#### The code

Renamed `Header.Actions` to Header.Right in order to reflect that the compound component can render more than just actions.
Header.Actions can still be used in order to keep retro compatibility. But its future usage will be discouraged in favor of the renamed version.

None of the actually implemented header should be impacted.

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/header/Header.tsx
+++ b/packages/mantine/src/components/header/Header.tsx
@@ -14,11 +14,11 @@ import {
 import {Children, ReactElement, ReactNode} from 'react';
 import {HeaderProvider} from './Header.context';
 import classes from './Header.module.css';
-import {HeaderActions, HeaderActionsStyleNames} from './HeaderActions/HeaderActions';
+import {HeaderRight, HeaderRightStyleNames} from './HeaderRight/HeaderRight';
 import {HeaderBreadcrumbs, HeaderBreadcrumbsStyleNames} from './HeaderBreadcrumbs/HeaderBreadcrumbs';
 import {HeaderDocAnchor, HeaderDocAnchorStyleNames} from './HeaderDocAnchor/HeaderDocAnchor';
 
-export type {HeaderActionsProps} from './HeaderActions/HeaderActions';
+export type {HeaderRightProps} from './HeaderRight/HeaderRight';
 export type {HeaderBreadcrumbsProps} from './HeaderBreadcrumbs/HeaderBreadcrumbs';
 export type {HeaderDocAnchorProps} from './HeaderDocAnchor/HeaderDocAnchor';
 
@@ -28,9 +28,10 @@ export type HeaderStyleNames =
     | 'title'
     | 'description'
     | 'divider'
+    | 'body'
     | HeaderDocAnchorStyleNames
     | HeaderBreadcrumbsStyleNames
-    | HeaderActionsStyleNames;
+    | HeaderRightStyleNames;
 
 export interface HeaderProps extends StylesApiProps<HeaderFactory>, Omit<GroupProps, 'classNames' | 'styles' | 'vars'> {
     /**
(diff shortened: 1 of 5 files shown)


FelixBlaisThon (commented):
Overall I don't think it's a bad idea, of course we would have to make a guideline for it with UX (so we have the same experience and also so that UX are all on board).

I would add a unit test tho :trollface:

dmgauthier:
> I would add a unit test tho :trollface:

The logic changed for a renaming of the Header.Actions compound component!  But IIRC I had added a UT before changing the approach :stuck_out_tongue_closed_eyes:

packages/mantine/src/components/header/Header.tsx
```diff
@@ -123,13 +125,17 @@ export const Header = factory<HeaderFactory>((_props, ref) => {
                         {description}
                     </Text>
                 </Stack>
-                {actions}
+                <Group wrap="nowrap">{right}</Group>
```
gdostie, line 128:
```suggestion
                {right}
```

packages/website/src/examples/layout/Header/Header.demo.tsx
```diff
@@ -9,9 +9,15 @@ const Demo = () => (
...
-            <Button variant="outline">Secondary</Button>
+        <Header.Actions wrap="nowrap">
+            <Stack gap="sm" bg="gray.3" p="sm">
+                <Text>Some content</Text>
+                <Text>Some more content</Text>
+            </Stack>
+            <Group gap="sm">
+                <Button>Primary</Button>
+                <Button variant="outline">Secondary</Button>
+            </Group>
```
gdostie, line 20: I have the feeling this is a pretty specific use case, the most common happy path is to have only buttons in the header's right section

```suggestion
            <Button>Primary</Button>
            <Button variant="outline">Secondary</Button>
```

gdostie (approved):

dmgauthier:
I'll let a one of the owner of the house merge this when this is ready. 
Meanwhile I'll stay tuned for any suggestion or improvement that could be made in the present code.

GermainBergeron:
I'm pretty sure this could be considered breaking because there's a lot of interfaces and type that changed name. If you don't mind we could wait until we merge the `next` branch in master to merge this PR

packages/mantine/src/components/header/__tests__/Header.spec.tsx
```diff
@@ -71,7 +71,22 @@ describe('Header', () => {
         expect(screen.getByText('description')).toBeInTheDocument();
     });
 
-    it('renders provided actions in the header', () => {
+    it('renders provided components in Table.Right', () => {
```
GermainBergeron, line 74:
```suggestion
    it('renders provided components in Header.Right', () => {
```

packages/mantine/src/components/header/__tests__/Header.spec.tsx
```diff
@@ -71,7 +71,22 @@ describe('Header', () => {
...
+                    <span>action 2</span>
+                </Header.Right>
+            </Header>,
+        );
+
+        expect(screen.getByText('action 1')).toBeInTheDocument();
+        expect(screen.getByText('action 2')).toBeInTheDocument();
+    });
+
+    it('renders provided components in Table.Right when called as Table.Actions', () => {
```
GermainBergeron, line 89:
```suggestion
    it('renders provided components in Header.Right when called as Header.Actions', () => {
```

GermainBergeron (approved):

PR #4043, opened 2025-04-09
feat(codeEditor): front the editor instance

hamamou:
#### Context
We had an issue where the switch tabs and code editor doesn't play well together when doing `ctrl-z`, since we change the content of the editor based on the  selected tab. Doing an "undo(ctrl-z)" after switching tab would set the content to the content of the previous tab. We added the `clearUndoStack` method to be able to clear the stack when switching tabs and prevent this behaviour
Here's an example of an issue we have.

web-7521

Here I pressed `ctrl-z` after switching tab and it sets the value to the old tab content.

#### Code
- Added a new optional `editorRef` prop to the `CodeEditor` component, allowing external access to the editor instance. Specifically, the `editorRef` exposes a `clearUndoStack` method that, when called, resets the editor's content to its current value, effectively clearing the undo stack. This functionality is set up using a `useEffect` hook to ensure the `editorRef` is properly updated with the `clearUndoStack` method.

Potential Breaking Changes
None

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/code-editor/CodeEditor.tsx
+++ b/packages/mantine/src/components/code-editor/CodeEditor.tsx
@@ -15,7 +15,7 @@ import {
 } from '@mantine/core';
 import {useUncontrolled} from '@mantine/hooks';
 import Editor, {Monaco, loader} from '@monaco-editor/react';
-import {editor as monacoEditor, MarkerSeverity} from 'monaco-editor';
+import {MarkerSeverity, editor as monacoEditor} from 'monaco-editor';
 import {FunctionComponent, useEffect, useRef, useState} from 'react';
 
 import cx from 'clsx';
@@ -48,6 +48,8 @@ interface CodeEditorProps
     onCopy?(): void;
     /** Called whenever the code editor gets the focus */
     onFocus?(): void;
+    /** Ref object that provides access to the editor's functionality */
+    editorHandle?: React.MutableRefObject<monacoEditor.IStandaloneCodeEditor | null>;
     /**
      * The minimal height of the CodeEditor (label and description included)
      *
@@ -108,6 +110,7 @@ export const CodeEditor: FunctionComponent<CodeEditorProps> = (props) => {
         disabled,
         monacoLoader,
         options: {tabSize} = {tabSize: 2},
+        editorHandle,
         ...others
     } = useProps('CodeEditor', defaultProps, props);
     const [loaded, setLoaded] = useState(false);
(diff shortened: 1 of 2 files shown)


packages/mantine/src/components/code-editor/CodeEditor.tsx
```diff
@@ -48,6 +48,10 @@ interface CodeEditorProps
     onCopy?(): void;
     /** Called whenever the code editor gets the focus */
     onFocus?(): void;
+    /** Ref object to access the editor instance */
+    editorRef?: React.MutableRefObject<{
```
cdroulers-coveo, line 52: The name confuses me (and it might be because I don't know this codebase!)
But it's named just like the inner ref `editorRef` but they have no common methods. I don't know if this library has a convention in terms of how to expose explicit methods to client code?

I think `useImperativeHandle` might be the better way to explicitely expose methods for parent component code to call?

packages/mantine/src/components/code-editor/CodeEditor.tsx
```diff
@@ -247,6 +265,7 @@ export const CodeEditor: FunctionComponent<CodeEditorProps> = (props) => {
                     editorRef.current = editor;
                     registerLanguages(monaco);
                     registerThemes(monaco);
+                    editor.getModel().setValue('');
```
cdroulers-coveo, line 268: Not sure why this is here? Does it just ensure that on mounting th ecomponent, any previous undo stack / values are emptied (e.g. if you load the editor with two different values in two modals, it still "remembers" the undo stack from the first one?)

cdroulers-coveo (commented):
I made two comments before realizing this was draft, sorry. If they are not relevant, feel free to ignore them!

packages/mantine/src/components/code-editor/CodeEditor.tsx
hamamou: renamed to `editorHandle` what do you think?

packages/mantine/src/components/code-editor/CodeEditor.tsx
```diff
@@ -48,6 +48,13 @@ interface CodeEditorProps
     onCopy?(): void;
     /** Called whenever the code editor gets the focus */
     onFocus?(): void;
+    /** Ref object that provides access to the editor's functionality.
```
gdostie, line 51: I wonder if instead we should just front the monaco instance directly. This way the users of the component can implement any additional behaviour they need without bloating the shared implementation

```
    onMount={(editor, monaco) => {
        editorRef.current = editor;
        if (editorHandle) {
            editorHandle.current = editor;
        }
        ...
```

hamamou: done

gdostie (approved):

PR #4046, opened 2025-04-11
chore(mantine, input): update the style of the input and input wrapper

GermainBergeron:
### Proposed Changes

Update the style of the Input and Input.Wrapper

None

### Acceptance Criteria


GermainBergeron:
> The difference in height of the input is smaller than before which makes it look misaligned in the table, can we do anything for that?

I reverted that change, Pretine was updated so the Inputs and Buttons are both 36px tall

gdostie (approved):

PR #4056, opened 2025-05-05
feat(pagination): +active-bg +active-color

FelixBlaisThon:
### Proposed Changes

https://coveord.atlassian.net/browse/ADUI-10590

Added root variables override for active bg-color and active-color with respect to UX recommendations.

[figma](https://www.figma.com/design/lFjxQoLHYzdLObC0g4HdWd/Pretine-7-for-Coveo?node-id=8613-333163&m=dev&focus-id=10073-410423)

Even though UX requested changes for Hover & disabled, we want to make adjustments only through css variables and mantine does not provide any for these two use case. So as a first iteration, we will keep it like this, as close to mantine as possible.

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/styles/Pagination.module.css
+++ b/packages/mantine/src/styles/Pagination.module.css
@@ -1,6 +1,7 @@
 .root {
     --pagination-active-bg: var(--mantine-primary-color-light);
     --pagination-active-color: var(--mantine-primary-color-filled);
+    --pagination-control-fz: var(--mantine-font-size-sm);
 }
 
 .control {
--- a/packages/mantine/src/styles/Pagination.module.css
+++ b/packages/mantine/src/styles/Pagination.module.css
@@ -1,7 +1,6 @@
 .root {
     --pagination-active-bg: var(--mantine-primary-color-light);
     --pagination-active-color: var(--mantine-primary-color-filled);
-    --pagination-control-fz: var(--mantine-font-size-sm);
 }
 
 .control {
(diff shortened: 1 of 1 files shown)


packages/mantine/src/styles/Pagination.module.css
```diff
@@ -1,3 +1,8 @@
+.root {
+    --pagination-active-bg: var(--mantine-primary-color-light);
+    --pagination-active-color: var(--mantine-primary-color-5);
```
GermainBergeron, line 3: Since the 5 can change:
```suggestion
    --pagination-active-color: var(--mantine-primary-color-filled);
```

FelixBlaisThon: 👍

packages/mantine/src/styles/Pagination.module.css
```diff
@@ -1,3 +1,8 @@
+.root {
+    --pagination-active-bg: var(--mantine-primary-color-light);
+    --pagination-active-color: var(--mantine-primary-color-filled);
```
gdostie, line 3: What figma has:
image
What the demo has:
image

Does that mean we should change `--pagination-control-fz` to 14px, or it means we should be setting the prop `size: 'sm'` by default in the theme?

FelixBlaisThon: I'm reverting this after Gustavo correction

gdostie:
This
image
should not have a border
image

GermainBergeron (approved):

PR #4057, opened 2025-05-05
feat(mantine): update style of checkbox

gdostie:
### Proposed Changes

|Before|After|
|---|---|
|image|image|

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/theme/Theme.tsx
+++ b/packages/mantine/src/theme/Theme.tsx
@@ -46,8 +46,7 @@ import {
     Text,
     Tooltip,
 } from '@mantine/core';
-import {InfoToken} from '../components';
-import {CircleLoader} from '../components/loader/CircleLoader';
+import {CheckboxIcon, CircleLoader, InfoToken} from '../components';
 import ActionIconClasses from '../styles/ActionIcon.module.css';
 import AlertClasses from '../styles/Alert.module.css';
 import BadgeClasses from '../styles/Badge.module.css';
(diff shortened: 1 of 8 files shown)


packages/mantine/src/theme/Theme.tsx
```diff
@@ -181,8 +182,8 @@ export const plasmaTheme: MantineThemeOverride = createTheme({
                 if (props.readOnly && !props.disabled) {
                     return {
                         root: {
-                            '--checkbox-icon-color': theme.colors.gray[7],
-                            '--checkbox-color': theme.colors.gray[2],
+                            '--checkbox-icon-color': 'var(--mantine-color-black)',
+                            '--checkbox-color': theme.colors.gray[1],
```
FelixBlaisThon, line 186: This variable could be in the `plasmaCSSVariablesResolver.ts` in the `// custom color` section. We could name it `'--coveo-color-input-bg-readonly'` so we could reuse it in every other inputs ?

gdostie: added the variables to the variable resolver and custom style to achieve the desired look, I think we should do the same for the other components until we have a cleaner way to do this

packages/mantine/src/theme/plasmaCSSVariablesResolver.ts
```diff
@@ -9,6 +9,9 @@ export const plasmaCSSVariablesResolver: CSSVariablesResolver = (theme) => {
             '--coveo-color-input-border': theme.colors.gray[3],
             '--coveo-color-title': theme.colors.gray[8],
             '--coveo-color-text-disabled': theme.colors.gray[3],
+            '--coveo-color-bg-disabled': theme.colors.gray[1],
```
FelixBlaisThon, line 12:
```suggestion
            '--coveo-color-bg-disabled': alpha(theme.colors.gray[4], 0.1),
```

GermainBergeron (approved):

PR #4058, opened 2025-05-06
Ad UI 10593 apply style radio

FelixBlaisThon:
### Proposed Changes

This removes some previous custom style that would remove the builtin visual behavior of the error radio button to reflect mantine better. The ux request cannot be fulfilled completely because of lack of css variable accessible from mantine to cover their demands.

[discussion](https://www.figma.com/design/lFjxQoLHYzdLObC0g4HdWd?node-id=9038-234373&m=dev#1244897930)

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/styles/Radio.module.css
+++ b/packages/mantine/src/styles/Radio.module.css
@@ -16,7 +16,7 @@
 
 .radio {
     @mixin light {
-        &[data-error='false'] {
+        &:not([data-disabled]) {
             border-color: var(--coveo-color-input-border);
         }
 
--- a/packages/mantine/src/styles/Radio.module.css
+++ b/packages/mantine/src/styles/Radio.module.css
@@ -1,17 +1,38 @@
 .labelWrapper {
     display: flex;
     align-items: flex-start;
+
+    @mixin light {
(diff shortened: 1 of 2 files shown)


packages/mantine/src/styles/Radio.module.css
```diff
@@ -1,10 +1,38 @@
...
+    }
 }
 
 .radio {
-    @mixin where-light {
-        border-color: var(--coveo-color-input-border);
+    @mixin light {
+        &[data-error='false'] {
+            border-color: var(--coveo-color-input-border);
+        }
```
GermainBergeron, line 21: I don't see the `--coveo-color-input-border` when there's no error. I think you should use the same syntax as Mantine instead:

```suggestion
        &:not([data-error]) {
            border-color: var(--coveo-color-input-border);
        }
```

GermainBergeron (approved):

packages/mantine/src/styles/Radio.module.css
FelixBlaisThon: fixed it. thanks

GermainBergeron (approved):

PR #4061, opened 2025-05-08
feat(collection): add-data-testid-param-to-collection

WilliamBlanchetLafreniere:
### Proposed Changes

Added data-testid attribute to remove buttons that has the matching CollectionItem item ids. This way we can differentiate between remove buttons of the same collection.

[Jira Link](https://coveord.atlassian.net/browse/ADUI-10483?atlOrigin=eyJpIjoiMWEzNTUxOTY4ZThkNGYyODgzZDJjMjBiNWNiMTdlZWEiLCJwIjoiaiJ9)

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/collection/Collection.tsx
+++ b/packages/mantine/src/components/collection/Collection.tsx
@@ -243,7 +243,7 @@ export const Collection = <T,>(props: CollectionProps<T> & {ref?: ForwardedRef<H
             draggable={draggable}
             onRemove={() => onRemoveItem?.(index)}
             removable={!(required && hasOnlyOneItem)}
-            data_index={index}
+            testId={`item-${item.id}`}
         >
             {children(item.data, index)}
         </CollectionItem>
--- a/packages/mantine/src/components/collection/Collection.tsx
+++ b/packages/mantine/src/components/collection/Collection.tsx
@@ -243,6 +243,7 @@ export const Collection = <T,>(props: CollectionProps<T> & {ref?: ForwardedRef<H
             draggable={draggable}
             onRemove={() => onRemoveItem?.(index)}
             removable={!(required && hasOnlyOneItem)}
+            data_index={index}
         >
             {children(item.data, index)}
         </CollectionItem>
--- a/packages/mantine/src/components/collection/Collection.tsx
+++ b/packages/mantine/src/components/collection/Collection.tsx
@@ -243,7 +243,6 @@ export const Collection = <T,>(props: CollectionProps<T> & {ref?: ForwardedRef<H
             draggable={draggable}
             onRemove={() => onRemoveItem?.(index)}
             removable={!(required && hasOnlyOneItem)}
-            testId={`item-${item.id}`}
         >
             {children(item.data, index)}
         </CollectionItem>
--- a/packages/mantine/src/components/collection/CollectionItem.tsx
+++ b/packages/mantine/src/components/collection/CollectionItem.tsx
@@ -8,23 +8,21 @@ import {ActionIcon} from '../action-icon';
 import {useCollectionContext} from './CollectionContext';
 
 interface CollectionItemProps extends CollectionItemSharedProps {
-    data_index: number;
     draggable?: boolean;
     disabled: boolean;
 }
 
 interface CollectionItemSharedProps extends GroupProps {
     id: string;
-    data_index: number;
+    testId: string;
     onRemove?: React.MouseEventHandler<HTMLButtonElement>;
     removable?: boolean;
 }
 
 const RemoveButton: FunctionComponent<{
-    data_index: number;
     onClick: React.MouseEventHandler<HTMLButtonElement>;
-}> = ({onClick, data_index}) => (
-    <ActionIcon.Quaternary data-testid={`remove-${data_index}`} style={{alignSelf: 'center'}} onClick={onClick}>
+}> = ({onClick}) => (
+    <ActionIcon.Quaternary style={{alignSelf: 'center'}} onClick={onClick}>
         <IconCircleMinus aria-label="Remove" size={20} />
     </ActionIcon.Quaternary>
 );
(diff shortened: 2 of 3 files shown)


FelixBlaisThon (approved):
Very good!

Here a few take away to go even further beyond (as goku would say when he turned SS3)

- Add the JIRA link in your PR description so we can refer to it if needed
- A little bit more context (even if obvious) in your PR description. i.e. why add data-testid ? -it is in fact stated in the slack thread in the JIRA, but having it here would be even better imo- But it's to give better accessibility to devs while testing with collections
- That's the cherry on the sunday : link the `admin-ui` repo with your branch in plasma and show that the data-testid is there ([documentation](https://docs.admin-ui.coveo.com/getting-started/07-external-dependencies#running-admin-ui-with-other-packages-that-we-maintain)). It's something I always do with feature that changes a component. In your case it's for testing purpose, but it will give you the habbit and also familiarize you with our process

Other than that, I think it's a solid start!

packages/mantine/src/components/collection/CollectionItem.tsx
```diff
@@ -8,20 +8,23 @@ import {ActionIcon} from '../action-icon';
 import {useCollectionContext} from './CollectionContext';
 
 interface CollectionItemProps extends CollectionItemSharedProps {
+    data_index: number;
```
FelixBlaisThon, line 11: Just noticed that this prop is duplicated. CollectionItemProps extends CollectionItemSharedProps, so I would remove this line and keep only in the CollectionItemSharedProps interface

packages/mantine/src/components/collection/CollectionItem.tsx
```diff
@@ -8,20 +8,23 @@ import {ActionIcon} from '../action-icon';
...
     removable?: boolean;
 }
 
 const RemoveButton: FunctionComponent<{
+    data_index: number;
     onClick: React.MouseEventHandler<HTMLButtonElement>;
-}> = ({onClick}) => (
-    <ActionIcon.Quaternary style={{alignSelf: 'center'}} onClick={onClick}>
+}> = ({onClick, data_index}) => (
+    <ActionIcon.Quaternary data-testid={`remove-${data_index}`} style={{alignSelf: 'center'}} onClick={onClick}>
```
gdostie, line 27: Instead of adding the test id to the remove button directly, I am wondering if instead it would not be better to simply add the test id on the CollectionItem.

This way, devs could easily do something like this in their tests which follows more closely the real user behaviour (you spot the item you want to delete, then you click on its remove button):

```tsx
await user.click(within(screen.getByTestId('item-1')).getByRole('button', {name: /remove/i}));
```

Doing this also removed the need for devs to add themselves the data-testid on the item.

packages/mantine/src/components/collection/Collection.tsx
```diff
@@ -243,6 +243,7 @@ export const Collection = <T,>(props: CollectionProps<T> & {ref?: ForwardedRef<H
             draggable={draggable}
             onRemove={() => onRemoveItem?.(index)}
             removable={!(required && hasOnlyOneItem)}
+            data_index={index}
```
gdostie, line 246: Instead of setting the index of the item as test id, we could probably use the `id`.

gdostie (commented):
Good first PR! 👏

packages/mantine/src/components/collection/Collection.tsx
WilliamBlanchetLafreniere: Good idea, this will need further testing.

packages/mantine/src/components/collection/CollectionItem.tsx
gdostie: Suppose I have two items in my collection, with the solution I propose I can distinguish between the two remove buttons of each item like this (see snippet) so it satisfies the initial demand.

```tsx
const item1 = screen.getByTestId('item-1');
const item2 = screen.getByTestId('item-2');
const removeItem1Button = within(item1).getByRole('button', {name: /remove/i});
const removeItem2Button = within(item2).getByRole('button', {name: /remove/i});
```

packages/mantine/src/components/collection/CollectionItem.tsx
```diff
@@ -30,11 +33,17 @@ const RemoveButtonPlaceholder = () => <div style={{width: 28}} />;
 
 const StaticCollectionItem: FunctionComponent<PropsWithChildren<CollectionItemSharedProps>> = ({
     onRemove,
+    data_index,
```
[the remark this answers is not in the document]
WilliamBlanchetLafreniere: I haven't found any exemple of usage of collections in any pages of the admin-ui project.

packages/mantine/src/components/collection/Collection.tsx
```diff
@@ -243,6 +243,7 @@ export const Collection = <T,>(props: CollectionProps<T> & {ref?: ForwardedRef<H
             draggable={draggable}
             onRemove={() => onRemoveItem?.(index)}
             removable={!(required && hasOnlyOneItem)}
+            testId={`item-${item.id}`}
```
GermainBergeron, line 246: Since we're already sending the `id` on line 241 I would suggest to remove this prop and simply construct the `data-testid` in the `CollectionItem`

```tsx
data-testid={`item-${id}`}
```

My reasoning is the `Collection` doesn't need to care about the implementation of the `CollectionItem`

packages/mantine/src/components/collection/__tests__/Collection.spec.tsx
```diff
@@ -138,6 +204,38 @@ describe('Collection', () => {
         expect(items[0]).toHaveTextContent('banana');
         expect(items[1]).toHaveTextContent('orange');
         expect(items[2]).toHaveTextContent('new');
+        expect(screen.getByTestId('item-2')).toBeInTheDocument();
+        expect(screen.getByTestId('form-state')).toHaveTextContent('{"fruits":["banana","orange","new"]}');
+    });
+
+    it('adds a new item when clicking on the add button with custom id', async () => {
+        const user = userEvent.setup({delay: null});
```
GermainBergeron, line 212:
```suggestion
        const user = userEvent.setup();
```

packages/mantine/src/components/collection/__tests__/Collection.spec.tsx
```diff
@@ -13,7 +13,32 @@ describe('Collection', () => {
...
+        const items = screen.getAllByTestId('item');
+        expect(items).toHaveLength(2);
+        expect(screen.queryByTestId('item-0')).toHaveTextContent('banana');
+        expect(screen.queryByTestId('item-1')).toHaveTextContent('orange');
+        expect(screen.getByTestId('item-0')).toBeInTheDocument();
+        expect(screen.getByTestId('item-1')).toBeInTheDocument();
+    });
+
+    it('renders one item for each initial values in the form with custom id', () => {
+        const getItemId = vi.fn((item) => `id-${item}`);
```
GermainBergeron, line 32:
```suggestion
        const getItemId = (item) => `id-${item}`;
```

packages/mantine/src/components/collection/__tests__/Collection.spec.tsx
```diff
@@ -29,7 +29,7 @@ describe('Collection', () => {
     });
 
     it('renders one item for each initial values in the form with custom id', () => {
-        const getItemId = vi.fn((item) => `id-${item}`);
+        const getItemId = (item: any) => `id-${item}`;
```
gdostie, line 32:
```suggestion
        const getItemId = (item: string) => `id-${item}`;
```

packages/mantine/src/components/collection/__tests__/Collection.spec.tsx
```diff
@@ -38,7 +38,7 @@ describe('Collection', () => {
             });
             return (
                 <Collection<string> getItemId={getItemId} newItem="" {...form.getInputProps('fruits')}>
-                    {(name) => <span data-testid={'item'}>{name}</span>}
+                    {(name) => <span data-testid="item">{name}</span>}
```
gdostie, line 41: I think we don't those anymore since we now add the test id automatically

```suggestion
                    {(name) => <span>{name}</span>}
```

FelixBlaisThon (approved):
First of many! :gg:

PR #4069, opened 2025-05-27
Fix/ua 10229 fix get item id on collection component

aperron-coveo:
### Proposed Changes

Add an `index: number` parameter to `getItemId` in Collection component.

In cases where the Collection type is a primitive type (`number`, `boolean`, `string`), every change of the value will call `getItemId` and potentially get a different result. When using a textbox, this causes a loss of focus and forces the user to click on the box each time a char is entered.
demo-collection-focus-loss

Using the item index fixes this issue by providing a value that wont change with the item.


packages/mantine/src/components/collection/Collection.tsx
```diff
@@ -148,7 +149,7 @@ const defaultProps: Partial<CollectionProps<unknown>> = {
     readOnly: false,
     gap: 'xs',
     required: false,
-    getItemId: ({id}: any) => id,
+    getItemId: ({id}: any, index: number) => id,
```
gdostie, line 152:
```suggestion
    getItemId: ({id}: any) => id,
```

gdostie (approved):

packages/mantine/src/components/collection/Collection.tsx
gdostie: It isn't used in this particular function so we can remove it, as ESLint suggests
image

PR #4088, opened 2025-07-03
feat(chip): adjust style of chip, export prop restricted component, c…

FelixBlaisThon:
…reate demo in the website

### Proposed Changes
This PR updates the style of the mantine [Chip](https://mantine.dev/core/chip/) component with UX pretine mockups.

I also added an entry in the website to have a demo of the chip, since the mockup seemed to have removed any variant of the chip. the only variant is `light`. I re-exported the component, but without this prop so people cannot use other things (less custom)

[Figma](https://www.figma.com/design/lFjxQoLHYzdLObC0g4HdWd/Opal-Components--Pretine-source-library-?node-id=17201-41100&m=dev)

### Before:
image

### After:

https://github.com/user-attachments/assets/8f6c1112-f103-4b83-8763-7097df67b6f4

#### disabled:
image

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/chip/Chip.tsx
+++ b/packages/mantine/src/components/chip/Chip.tsx
@@ -1,0 +1,5 @@
+import {ChipFactory, ChipGroup, factory, Chip as MantineChip} from '@mantine/core';
+
+export const Chip = factory<ChipFactory>(({variant: _variant, ...props}, ref) => <MantineChip {...props} ref={ref} />);
+
+Chip.Group = ChipGroup;
--- a/packages/mantine/src/styles/Chip.module.css
+++ b/packages/mantine/src/styles/Chip.module.css
@@ -1,0 +1,41 @@
+.root {
+    --chip-color: var(--coveo-color-text-primary);
+    --chip-bg: var(--mantine-primary-color-light);
+    --chip-hover: var(--mantine-primary-color-filled-hover);
+    --chip-radius: var(--mantine-radius-lg);
+    --chip-padding: var(--mantine-spacing-sm);
+    --chip-spacing: var(--mantine-spacing-sm);
+
+    > .label {
+        &:not([data-checked]) {
+            border: 1px solid var(--mantine-color-default-border);
+            border-radius: var(--mantine-radius-lg);
+            background-color: var(--mantine-color-white);
+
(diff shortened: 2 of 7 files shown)


packages/mantine/src/styles/Chip.module.css
```diff
@@ -0,0 +1,41 @@
...
+
+            @mixin hover {
+                &[data-disabled] {
+                    border: none;
+                    color: var(--coveo-color-text-disabled);
+                    background-color: var(--coveo-color-bg-disabled);
+                }
+
+                background-color: var(--mantine-color-gray-light);
+                border: 1px solid var(--coveo-color-input-border);
```
gdostie, line 23: Can we change just the border-color here?

```suggestion
                border-color: var(--coveo-color-input-border);
```

packages/mantine/src/components/chip/Chip.tsx
```diff
@@ -0,0 +1,6 @@
+import {Chip as MantineChip, type ChipProps as MantineChipProps} from '@mantine/core';
+import React from 'react';
+
+export type ChipProps = Omit<MantineChipProps, 'variant'>;
+
+export const Chip: React.FC<ChipProps> = (props) => <MantineChip {...props} />;
```
gdostie, line 6: I think this will need to be a forward ref, otherwise we cannot set a tooltip on it. Also, we only use named imports of react stuff.

```suggestion
import {Chip as MantineChip, type ChipProps as MantineChipProps} from '@mantine/core';
import {forwardRef} from 'react';

export type ChipProps = Omit<MantineChipProps, 'variant'>;

export const Chip = forwardRef<HTMLInputElement, ChipProps>((props, ref) => <MantineChip ref={ref} {...props} />);
```

FelixBlaisThon: Yes I forgot about ref thanks!

packages/website/src/pages/form/Chip.tsx
gdostie, line 1: There is already a Chip demo page in the Mantine section no?

Second question: I noticed that the "ChipDeselect" example is busted in your PR demo

Next
image


Demo

image

gdostie (approved):

PR #4128, opened 2025-08-07
feat(mantine): new LastUpdated component

aboissinot-coveo:
### Proposed Changes

[PSE-334](https://coveord.atlassian.net/browse/PSE-334)

Introduced a new `LastUpdated` component. The goal is to provide a generic component that can be used anywhere, but that is also used under the hood by the `Table.LastUpdated` so we have consistent styling everywhere.

I went with a stateless approach for the `LastUpdated` component to leave the logic of specifying the correct timestamp to the parent as I thought it would be more flexible. If you have a different view on the matter, let me know.

Screenshot 2025-08-07 at 12 13 25 PM

Screenshot 2025-08-07 at 12 13 37 PM

There shouldn't be any breaking changes to `Table.LastUpdated`. If you identify a potential source of breaking change with this existing component, let me know.

### Acceptance Criteria

[PSE-334]: https://coveord.atlassian.net/browse/PSE-334?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/mantine/src/components/last-updated/LastUpdated.tsx
+++ b/packages/mantine/src/components/last-updated/LastUpdated.tsx
@@ -6,9 +6,16 @@ export type LastUpdatedStylesNames = 'root' | 'label';
 
 export interface LastUpdatedProps extends BoxProps, Pick<GroupProps, 'justify'>, StylesApiProps<LastUpdatedFactory> {
     /**
-     * The unformatted time to display. Either a Date object, a number timestamp (milliseconds since epoch) or string that can be parsed by dayjs.
+     * Optional formatter function to format the time value.
+     * Receives the time prop and should return a string.
+     * @default (time) => dayjs(time).format('h:mm:ss A')
      */
-    time: Date | string | number;
+    formatter?: (time: dayjs.ConfigType) => string;
+    /**
+     * The unformatted time to display that can be parsed by dayjs.
+     * @default The current time via dayjs().valueOf()
+     */
+    time?: dayjs.ConfigType;
     /**
      * Label to contextualize the time.
      *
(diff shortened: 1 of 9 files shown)


packages/mantine/src/components/last-updated/LastUpdated.tsx
```diff
@@ -0,0 +1,64 @@
...
+        return (
+            <Group
+                justify={props.justify}
+                ref={ref}
+                {...getStyles('root', {className, style, ...stylesApiProps})}
+                {...others}
+            >
+                <Text size="xs" {...getStyles('label', {className, style, ...stylesApiProps})}>
+                    {label}
+                    <span role="timer">{dayjs(time).format('h:mm:ss A')}</span>
```
gdostie, line 59: In the admin we will want to use the core/format package to have the proper date format. Ideally the component would allow this level of flexibility

packages/mantine/src/components/last-updated/LastUpdated.tsx
```diff
@@ -0,0 +1,64 @@
...
+import dayjs from 'dayjs';
+import React from 'react';
+
+export type LastUpdatedStylesNames = 'root' | 'label';
+
+export interface LastUpdatedProps extends BoxProps, Pick<GroupProps, 'justify'>, StylesApiProps<LastUpdatedFactory> {
+    /**
+     * The unformatted time to display. Either a Date object, a number timestamp (milliseconds since epoch) or string that can be parsed by dayjs.
+     */
+    time: Date | string | number;
```
gdostie, line 11: **Thinking out loud**:
Should we make it optional? When time is undefined, it equals to `dayjs()` (now) so it would simply display the time of last render.

```suggestion
    time?: Date | string | number;
```

packages/mantine/src/components/last-updated/LastUpdated.tsx
aboissinot-coveo: Yes good point. I added a `formatter` prop which allows to pass a custom formatter. Otherwise it defaults to this one.

PR #4152, opened 2025-09-24
fix: rendering of badges with `circle` prop

malaporte:
### Proposed Changes

After the Commerce Health dashboards were refactored to use the "semantic badges", I realized that when we used them with the `circle` prop the rendering was broken.

None.

### Acceptance Criteria

-   [X] [README.md](https://github.com/coveo/plasma/blob/master/README.md) is adjusted to reflect the proposed changes (if relevant)

### DIFF
--- a/packages/mantine/src/components/badge/Badge.tsx
+++ b/packages/mantine/src/components/badge/Badge.tsx
@@ -66,10 +66,8 @@ const enhanceBadge = (
             <Component
                 ref={ref}
                 {...props}
-                py={2}
-                px={12}
+                {...(!props.circle ? {py: 2, px: 12, h: props.size === 'large' ? 22 : 20} : {})}
                 size={props.size === 'large' ? 'lg' : 'md'}
-                h={props.size === 'large' ? 22 : 20}
             />
         );
     });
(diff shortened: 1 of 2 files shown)


gdostie:
> @gdostie The icon we use in the [notification](https://plasma.coveo.com/mantine/Notification) component could be used as an alternative in this case. As for the size of the icon, I'm not sure it matches what's in the DS, at least for the notification component. I see const `checkIcon = <IconCheck size={20}` in [FE](https://plasma.coveo.com/mantine/Notification), and `24px` in [Figma](https://www.figma.com/design/lFjxQoLHYzdLObC0g4HdWd/Opal-Components--Pretine-source-library-?node-id=13070-161980&t=4L3WVU5JDFmOiBlT-4). Note: we have `20px` as an available value for the prop size, but it's set to `24px` from what I can see. Screenshot 2025-09-24 at 11 44 48 AM
> 
> Screenshot 2025-09-24 at 11 38 39 AM

Ok good, lets use the icon instead of the badge then 👍 

As to the size I think it was a mistake on our end. Its easy to get mixed up if you click on the wrong thingy. For example:

image

PR #4395, opened 2026-04-09
feat: upgrade monorepo to strict type check

gdostie:
### Proposed Changes

This PR enables TypeScript `strict` mode in the shared monorepo base config and fixes the resulting type errors in packages that inherit from it.

- Shared config
  - Add `"strict": true` to `packages/tsconfig-base.json`.
  - Keep `packages/components-props-analyzer` opted out for now.
- `packages/figma`
  - Add explicit generics for `figma.boolean(...)` / `figma.instance(...)` so optional text and instance swaps resolve correctly under strict mode.
- `packages/mantine`
  - Fix strict-mode issues around nullability, default props, optional callbacks/refs, and helper/store typing.
  - Clean up `defaultProps` declarations by switching them to `satisfies Partial<...>` and removing the old cast-based pattern.
  - Make a few behaviors explicit where strict mode surfaced ambiguity, mainly in `CodeEditor`, date-range pickers, `LastUpdated`, and `Table*`.
- `packages/storybook` and build helpers
  - Tighten story arg and plugin return types where previous inference was too loose for strict mode.

No intentional breaking changes.

The only expected behavior changes are low-risk null/undefined handling fixes made explicit during the strict-mode migration.

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/CodeEditor/CodeEditor.tsx
+++ b/packages/mantine/src/components/CodeEditor/CodeEditor.tsx
@@ -80,12 +80,14 @@ interface CodeEditorProps
     options?: Pick<monacoEditor.IStandaloneEditorConstructionOptions, 'tabSize'>;
 }
 
-const defaultProps: Partial<CodeEditorProps> = {
+const defaultProps = {
     language: 'plaintext',
     monacoLoader: 'local',
     defaultValue: '',
     minHeight: 300,
-};
+    justify: 'flex-start',
+    gap: 'sm',
+} satisfies Partial<CodeEditorProps>;
 
 export const CodeEditor: FunctionComponent<CodeEditorProps> = (props) => {
     const {
@@ -109,6 +111,8 @@ export const CodeEditor: FunctionComponent<CodeEditorProps> = (props) => {
         monacoLoader,
         options: {tabSize} = {tabSize: 2},
         editorHandle,
+        h,
+        mah,
         ...others
     } = useProps('CodeEditor', defaultProps, props);
     const [loaded, setLoaded] = useState(false);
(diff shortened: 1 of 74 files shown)


packages/mantine/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -80,12 +80,14 @@ interface CodeEditorProps
...
 
-const defaultProps: Partial<CodeEditorProps> = {
+const defaultProps = {
     language: 'plaintext',
     monacoLoader: 'local',
     defaultValue: '',
     minHeight: 300,
-};
+    justify: 'flex-start',
+    gap: 'sm',
```
GermainBergeron, line 89: New props?

packages/mantine/src/components/Table/Table.tsx
```diff
@@ -232,12 +232,23 @@ export const Table = <T,>(props: TableProps<T> & {ref?: ForwardedRef<HTMLDivElem
...
+            <TableProvider<T>
+                value={{
+                    getStyles,
+                    getRowActions: getRowActions,
+                    store,
+                    table,
+                    layouts: layouts,
+                    containerRef,
+                }}
+            >
```
GermainBergeron, line 251:
```suggestion
            <TableProvider<T> value={{getStyles, getRowActions, store, table, layouts, containerRef}}>
```

packages/mantine/src/components/Prompt/Prompt.tsx
```diff
@@ -94,17 +96,22 @@ const _Prompt = factory<PromptFactory>((_props, ref) => {
...
-        (child.type === PromptFooter ? footers : otherChildren).push(child);
+        otherChildren.push(child);
     });
 
     const IconComponent = PromptVariantIconsMapping[variant];
 
     return (
-        <PromptContextProvider value={{variant, getStyles}}>
-            <Modal.Root ref={ref} variant="prompt" size="sm" {...others} {...getStyles('root')}>
+        <PromptContextProvider value={{variant: variant, getStyles}}>
```
GermainBergeron: I don't know why we start having those `var: var`, but IMHO EsLint/prettier should fix those for us

```suggestion
        <PromptContextProvider value={{variant, getStyles}}>
```

GermainBergeron (approved):

packages/mantine/src/components/CodeEditor/CodeEditor.tsx
gdostie: no, only new defaults

PR #4401, opened 2026-04-14
ci: switch to changesets for publishing new versions

GermainBergeron:
### Proposed Changes

As discussed in team, this PR modifies the deployment to use Changesets instead of relying on semantic commits for version bumping.

The CI/CD might be a bit flaky but the business code didn't change

### Acceptance Criteria


scripts/releasePreview.mjs
```diff
@@ -0,0 +1,186 @@
...
+
+const execFile = promisify(execFileCallback);
+const ROOT = process.cwd();
+const PREVIEW_MARKER = '<!-- changesets-release-preview -->';
+
+const exec = async (file, args, options = {}) => {
+    const {stdout, stderr} = await execFile(file, args, {
+        cwd: ROOT,
+        encoding: 'utf8',
+        maxBuffer: 10 * 1024 * 1024,
```
gdostie, line 15: Is specifying this buffer length needed? Could we use node's default max buffer instead?

scripts/releasePreview.mjs
```diff
@@ -0,0 +1,186 @@
...
+        .split('\n')
+        .map((line) => line.trim())
+        .filter(Boolean);
+};
+
+const hasPackageCodeChanges = (changedFiles) =>
+    changedFiles.some(
+        (filePath) =>
+            filePath.startsWith('packages/') &&
+            !filePath.endsWith('.md') &&
```
gdostie, line 72: If we publish a skill with instructions in .md file, that condition could become problematic

scripts/releasePreview.mjs
```diff
@@ -0,0 +1,186 @@
...
+            !filePath.startsWith('packages/website/src/docs/'),
+    );
+
+const createPreviewDir = async () => {
+    const tempDir = await mkdtemp(path.join(os.tmpdir(), 'plasma-release-preview-'));
+    await cp(ROOT, tempDir, {
```
gdostie, line 78: Why do we need to double up the repo in a temp directory? Heres a couple other options I am thinking of that could be less flaky:
- create a worktree with git
- use a git command to find the version value before in the index before the preview is run
- save the current version first in a variable, run the preview after

scripts/releasePreview.mjs
```diff
@@ -0,0 +1,186 @@
...
+    return tempDir;
+};
+
+const removePreviewDir = async (previewDir) => rm(previewDir, {recursive: true, force: true});
+
+const runVersionPreview = async (worktreePath) => {
+    await execFile('pnpm', ['exec', 'changeset', 'version'], {
+        cwd: worktreePath,
+        encoding: 'utf8',
+        maxBuffer: 10 * 1024 * 1024,
```
gdostie, line 100: same question about the buffer here. In fact, we could reuse the function defined where I left the first comment

scripts/releasePreview.mjs
GermainBergeron: Fixed, I updated the whole releasePreview logic to leverage the `changeset status` command

gdostie (approved):

PR #4411, opened 2026-04-20
DS-292 override Mantine combobox chevron

gdostie:
## What changed
- override Mantine's default `ComboboxChevron` globally with a custom `PlasmaComboboxChevron`
- render the Plasma `IconChevronDown` instead of Mantine's built-in chevron
- rotate the chevron when combobox-backed inputs are expanded
- keep read-only `Select` from rotating by tagging the read-only path with `data-readonly`
- preserve right-section color behavior for default, disabled, error, and read-only states, including Mantine's combined clear-section wrapper

## Why it changed
Mantine's default combobox chevron does not match the Plasma iconography we want across `Select`-style controls. 

## Preview

https://github.com/user-attachments/assets/44503ab4-1375-49b5-b646-f5ec028909fb

### DIFF
--- a/packages/mantine/src/components/Combobox/PlasmaComboboxChevron.module.css
+++ b/packages/mantine/src/components/Combobox/PlasmaComboboxChevron.module.css
@@ -1,5 +1,7 @@
 .root {
     display: block;
     transform: rotate(0deg);
-    transition: transform 150ms ease;
+    transition-property: transform;
+    transition-timing-function: var(--coveo-transition-function);
+    transition-duration: var(--coveo-transition-duration);
 }
--- a/packages/mantine/src/components/Combobox/PlasmaComboboxChevron.module.css
+++ b/packages/mantine/src/components/Combobox/PlasmaComboboxChevron.module.css
@@ -1,0 +1,5 @@
+.root {
+    display: block;
+    transform: rotate(0deg);
+    transition: transform 150ms ease;
+}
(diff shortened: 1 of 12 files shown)


packages/mantine/src/components/Combobox/PlasmaComboboxChevron.module.css
```diff
@@ -0,0 +1,5 @@
+.root {
+    display: block;
+    transform: rotate(0deg);
+    transition: transform 150ms ease;
```
scharron-coveo: Would try to save easing function into a global variable to be potentially used in other UI motion later and allow batch changes as needed.

Would be curious to see a list of available easing function. Using ease() usually animate with default easing if non specified.

scharron-coveo (approved):
Looks good to me

Would try to save the easing function into a global variable to be potentially used in other UI motion later and allow batch changes as needed.

Would be curious to see a list of available easing function. Using ease() usually animate with default easing if non specified.

FelixBlaisThon (approved):
Demo tested, banco

FelixBlaisThon:
Question, whats with the reset file ?

gdostie:
> Question, whats with the reset file ?

Somehow I had a visually different result locally vs in the demo. When building for production it seems like Vite or Storybook is somehow changing the import order of CSS files sometimes. 

|local|demo|
|---|---|
|image|image|


I added CSS layers to everything in the storybook config to force the right CSS import order.

packages/mantine/src/components/Combobox/PlasmaComboboxChevron.module.css
gdostie: Done ✅ 
Created two new variables, one for the timing function, the other for the duration:

image

FelixBlaisThon (approved):

PR #4420, opened 2026-04-29
fix(inputlabelinfo): make pointer-event auto

FelixBlaisThon:
### Proposed Changes

The InputLabelInfo component was not displaying the tooltip if the input was in read-only. By explicitly set `pointer-event: auto`, it overrides the `pointer-events: none` coming from the read only parent state

### Acceptance Criteria

### DIFF
--- a/packages/mantine/src/components/Input/InputLabelInfo.module.css
+++ b/packages/mantine/src/components/Input/InputLabelInfo.module.css
@@ -2,4 +2,5 @@
     position: relative; /* Required to make it work within a Switch label */
     margin-left: var(--mantine-spacing-xxs);
     vertical-align: text-bottom;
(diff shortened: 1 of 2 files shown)


packages/mantine/src/components/Input/InputLabelInfo.module.css
```diff
@@ -2,4 +2,5 @@
     position: relative; /* Required to make it work within a Switch label */
     margin-left: var(--mantine-spacing-xxs);
     vertical-align: text-bottom;
+    pointer-events: auto;
```
gdostie, line 5: Doing this allows to click on the info token to check or uncheck the checkbox even while in readonly 😬

gdostie (approved):

PR #4451, opened 2026-05-28
chore: migrate from eslint to oxlint

GermainBergeron:
### Proposed Changes

Replaced all the eslint config to a simpler oxlint setup.
The second commit is to fix the warnings/errors reported by oxlint

None for library consumer


package.json
```diff
@@ -58,32 +59,23 @@
         "@commitlint/config-conventional": "20.5.0",
         "@types/node": "24.12.2",
         "@typescript/native-preview": "7.0.0-dev.20260420.1",
-        "@vitest/eslint-plugin": "1.6.16",
         "aws-sdk": "2.1693.0",
         "commitizen": "4.3.1",
         "conventional-changelog-angular": "8.3.1",
         "cz-conventional-changelog": "3.3.0",
         "eslint": "9.39.4",
```
gdostie, line 66: 👁️ do we still need it?

.github/workflows/cd.yml
```diff
@@ -36,9 +36,7 @@ jobs:
             - name: Setup
               uses: ./.github/actions/setup
             - name: Lint
-              uses: ./.github/actions/lint
-              with:
-                  WORKFLOW: 'cd'
+              run: pnpm oxlint
```
gdostie: Should we use the same thing as in the CI workflow?

```suggestion
              run: pnpm lint
              env:
                  # Without this, oxlint fails with a cryptic error (Error: Invalid tsconfig)
                  OXLINT_TSGOLINT_DANGEROUSLY_SUPPRESS_PROGRAM_DIAGNOSTICS: true
```

packages/mantine/src/components/Facet/FacetScrollArea.tsx
```diff
@@ -3,7 +3,7 @@ import {forwardRef} from 'react';
 
 export const FacetScrollArea = forwardRef<HTMLDivElement, ScrollAreaProps>(
     ({style, children, ...others}: ScrollAreaProps, ref) => (
-        <ScrollArea {...others} style={{width: '100%', ...style}} viewportRef={ref}>
+        <ScrollArea {...others} style={[{width: '100%'}, style]} viewportRef={ref}>
```
gdostie, line 6: style as an array?

packages/storybook/package.json
```diff
@@ -6,14 +6,9 @@
     "scripts": {
         "build": "storybook build",
         "chromatic": "chromatic --build-script-name build",
-        "lint": "../../node_modules/.bin/eslint \"**/*.{ts,tsx}\"",
+        "lint": "../../node_modules/.bin/oxlint",
```
gdostie: do we still need to use the relative path like that?


```suggestion
        "lint": "oxlint",
```

package.json
GermainBergeron: We still need it because it's a peer of the js plugins (storybook and testing library)

.github/workflows/cd.yml
GermainBergeron: Yes, good catch

packages/mantine/src/components/Facet/FacetScrollArea.tsx
GermainBergeron: I was surprised, but Mantine supports that. `style` can be a function, an array and a few other things. Using the array leverage Mantine's handling of the style prop

packages/storybook/package.json
GermainBergeron: No, we don't need it. Fixed

gdostie (approved):

PR #4459, opened 2026-05-29
feat(storybook): add iconography page

GermainBergeron:
### Proposed Changes

Add the iconography page in the Storybook foundation's

https://plasma.coveo.com/feature/feat-ADUI-11460-add-iconography-to-storybook/?path=/docs/foundation-iconography--docs

None

### DIFF
--- a/packages/storybook/src/foundation/iconography/Iconography.stories.tsx
+++ b/packages/storybook/src/foundation/iconography/Iconography.stories.tsx
@@ -1,0 +1,247 @@
+import {
+    Alert,
+    Anchor,
+    BlankSlate,
+    Button,
+    Code,
+    type ColumnDef,
+    createColumnHelper,
+    FilterFn,
+    getFilteredRowModel,
+    getPaginationRowModel,
+    Header,
+    Stack,
+    Table,
+    Text,
+    Title,
+    useTable,
+    useTableContext,
+} from '@coveord/plasma-mantine';
+import * as PlasmaReactIcons from '@coveord/plasma-react-icons';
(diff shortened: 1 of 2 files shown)


packages/storybook/src/foundation/iconography/Iconography.stories.tsx
```diff
@@ -0,0 +1,247 @@
...
+            <TablerIconsTable />
+
+            <Stack>
+                <Header variant="secondary">Custom</Header>
+                <Alert.Warning title="Deprecated">
+                    The icons in this list are deprecated in favor of Tabler Icons. Only use in-house icons if no Tabler
+                    icon fits your needs.
+                </Alert.Warning>
+                <PlasmaIconsTable />
+            </Stack>
```
gdostie: There are two stacks here and TablerIconsTable is outside both of them while PlasmaIconsTable is inside one of them. Can we consolidate the two stacks into one with both tables inside?

gdostie (dismissed):

packages/storybook/src/foundation/iconography/Iconography.stories.tsx
```diff
@@ -0,0 +1,247 @@
...
+
+const tablerColumnHelper = createColumnHelper<TablerIconRow>();
+const tablerColumns = [
+    tablerColumnHelper.accessor('name', {
+        header: 'Name',
+        enableSorting: false,
+    }),
+    tablerColumnHelper.accessor('componentName', {
+        id: 'import',
+        header: 'JSX',
```
gdostie: Here the column is called `JSX` but in the other table its written `Code`. I think "Code" is easier to understand for everyone

gdostie (approved):

PR #4476, opened 2026-06-17
fix(mantine): only apply inline middlewares in test environment

samuelfullerthomas:
## Summary

Remove `middlewares: {inline: true}` from Combobox and Popover defaultProps in the base plasma theme. This fixes NaN warnings at runtime when these components are inside a hidden `Tabs.Panel` (see [mantine#8084](https://github.com/mantinedev/mantine/issues/8084)).

Instead, `Plasmantine` now automatically injects the inline middlewares only when `env="test"` is passed, keeping JSDOM tests working without polluting the production theme.

## Changes

- **Theme.tsx**: Removed `middlewares: {inline: true}` from Combobox and Popover defaultProps
- **Plasmantine.tsx**: Detects `env === "test"` and merges inline middlewares into the theme
- **Utils.tsx**: Simplified test wrapper to use `<Plasmantine env="test">` instead of manual testTheme
- **Plasmantine.spec.tsx** (new): Verifies middlewares are absent in production and present in test env
- **DateRangePicker tests**: Fixed assertions using `.not.toBeVisible()` on unmounted elements

## Testing

- All 50 test files pass (386 tests)
- New regression test confirms production theme has no middlewares

MHUB-2938

### DIFF
--- a/packages/mantine/src/components/DateRangePicker/__tests__/DateRangePicker.spec.tsx
+++ b/packages/mantine/src/components/DateRangePicker/__tests__/DateRangePicker.spec.tsx
@@ -70,7 +70,7 @@ describe('DateRangePicker', () => {
 
         expect(onChange).toHaveBeenCalledTimes(1);
         expect(onChange).toHaveBeenCalledWith(['2022-01-02T00:00:00.000Z', '2022-01-08T23:59:59.999Z']);
-        expect(screen.queryByRole('dialog')).not.toBeVisible();
+        expect(screen.queryByRole('dialog')).not.toBeInTheDocument();
     });
 
     it('closes popover when Cancel is clicked', async () => {
@@ -81,7 +81,7 @@ describe('DateRangePicker', () => {
         await screen.findByRole('dialog');
         await user.click(screen.getByRole('button', {name: 'Cancel'}));
 
-        await waitFor(() => expect(screen.queryByRole('dialog')).not.toBeVisible());
+        await waitFor(() => expect(screen.queryByRole('dialog')).not.toBeInTheDocument());
     });
 
     it('calls onCancel when Cancel button is clicked', async () => {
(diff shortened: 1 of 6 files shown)


packages/mantine/src/components/DateRangePicker/__tests__/DateRangePicker.spec.tsx
```diff
@@ -70,7 +70,7 @@ describe('DateRangePicker', () => {
 
         expect(onChange).toHaveBeenCalledTimes(1);
         expect(onChange).toHaveBeenCalledWith(['2022-01-02T00:00:00.000Z', '2022-01-08T23:59:59.999Z']);
-        expect(screen.queryByRole('dialog')).not.toBeVisible();
+        expect(screen.queryByRole('dialog')).not.toBeInTheDocument();
```
samuelfullerthomas, line 73: The old test Utils.tsx only added `middlewares: {inline: true}` — it
  didn't pass `env="test"` to `MantineProvider`. So transitions and
  portals were still active in tests.

  Our new `Utils.tsx` uses `<Plasmantine env="test">`, which tells
  Mantine to disable transitions and portals. With
  portals/transitions disabled, when a popover closes, Mantine
  completely removes the dialog element from the DOM instead of
  hiding it with CSS.

  The 3 failing tests asserted:

`  expect(screen.queryByRole('dialog')).not.toBeVisible();`

  This expects the element to still exist in the DOM but be hidden.
  With env="test", the element is gone entirely — `queryByRole`
  returns null, and `.not.toBeVisible()` throws because it requires an
  actual DOM element.

  The fix:

`  expect(screen.queryByRole('dialog')).not.toBeInTheDocument();`

  This correctly asserts "the dialog is no longer present"
  regardless of whether it was hidden or removed.

  These tests were subtly fragile before — they only worked because
  transitions delayed removal. The new assertions are actually more
  correct for what the tests intend to verify ("popover is closed").

samuelfullerthomas:
can't decide if this is a breaking change? I guess it will cause tests run inside of a non `<Plasmantine env="test">` wrapper  to fail so it is a breaking change

gdostie (approved):

FelixBlaisThon (approved):
Thanks!

FelixBlaisThon (approved):
Thanks!

PR #4481, opened 2026-06-19
feat(mantine, table): toolbar to display components above the Table

GermainBergeron:
### Proposed Changes

Adds a new `Table.Toolbar` sub-component that allows `Table.Filter`, `Table.Predicate`, and `Table.DateRangePicker` to be rendered outside of the `Table.Header` grid layout. This is opt-in and fully backward compatible. Existing usage with `Table.Header` is unchanged.

#### Motivation

`Table.Header` comes with some style and renders its children inside a `Grid`. Due to this Grid, the components we normaly see in the header (`Filter`, `Predicate`, `DateRangePicker`) are wrapped in a `Grid.Col`. In some implementation people had to duplicate a lot of logic to achieve a similar behaviour while matching UX mockups 

image

#### Usage

```tsx
<Table store={store} columns={columns} data={data}>
    <Table.Toolbar>
        <Table.Filter />
        <Table.DateRangePicker />
        <Table.Predicate id="status" data={[...]} label="Status" />
    </Table.Toolbar>
...
</Table>
```

Components inside `Table.Toolbar` skip the `Grid.Col` wrapper and render as plain flex items. The toolbar renders above the table layout. It can be customized via `renderRoot`, `className`, or standard Box props.

#### Open question

> The name `Toolbar` was chosen to distinguish it from `Header` (which has specific grid/actions/layout-control behaviour). Open to renaming if there's a better fit.

No breaking changes. Existing `Table.Header` usage is unaffected.

### DIFF
--- a/.changeset/gentle-otters-jump.md
+++ b/.changeset/gentle-otters-jump.md
@@ -1,0 +1,10 @@
+---
+'@coveord/plasma-mantine': minor
+'@coveord/plasma-llms': minor
+---
+
+Add `Table.Toolbar` sub-component for rendering filters outside the header grid
+
+`Table.Toolbar` is a new opt-in container that renders above the table. Place `Table.Filter`, `Table.Predicate`, or `Table.DateRangePicker` inside it to display them outside the `Table.Header`. Components inside the toolbar skip the `Grid.Col` wrapper and render as plain flex items instead.
+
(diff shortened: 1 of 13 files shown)


gdostie (commented):
The task was about making the table components usable outside the context of the table, but here the toolbar still has to be rendered within the table to have its context.

aiheon:
Found an issue in the storybook 👀 

image

aiheon (approved):

PR #4496, opened 2026-07-08
chore: trigger release PR workflow

gdostie:
### Proposed Changes

When a workflow file is introduced in the same commit (merge) that should trigger it, that first push event does not trigger the newly-added workflow. GitHub only runs workflows that already existed on the default branch before the push event. When next was merged in master with the new release-pr workflow, it did not execute as it should have.

### Acceptance Criteria


GermainBergeron (approved):
Good PR, really simple!

PR #4499, opened 2026-07-08
feat(table):  TableCell component

FelixBlaisThon:
### Proposed Changes

This PR creates a cell component that handles:

- wrap
- ellipsis
- show more
- classes for width and other style.

+ unit tests and storybook adjustment

I split the expandable part on a seperate commit since it's not something we provide as of today (we rather use the ellipsis component when content is too big to render. 

Knowing that we will revisit the table component as a whole, we can just take out the commit and add it back at a later time, after UX asses it properly. From the time being it's a nice add-on to play with. Could even be merged as is as still revisit it later.

### Acceptance Criteria


gdostie (approved):

packages/mantine/src/components/Table/table-cell/TableCell.tsx
```diff
@@ -0,0 +1,149 @@
...
+}
+
+export type TableCellFactory = Factory<{
+    props: TableCellProps;
+    defaultRef: HTMLDivElement;
+    defaultComponent: 'div';
+    stylesNames: TableCellStylesNames;
+}>;
+
+const defaultProps: Partial<TableCellProps> = {};
```
gdostie:
```suggestion
const defaultProps = {} satisfies Partial<TableCellProps>;
```

gdostie (commented):
y manque le fichier changeset

gdostie (approved):

PR #4527, opened 2026-07-17
fix table missing a border top when no header is present

gdostie:
### Proposed Changes

Before

image

After

image

Also I fixed a bug with the local setup where a .module.css files would not trigger an update to the storybook website.

### Acceptance Criteria

### DIFF
--- a/.changeset/thin-lions-fly.md
+++ b/.changeset/thin-lions-fly.md
@@ -1,0 +1,5 @@
+---
+'@coveord/plasma-mantine': major
+---
+
+Remove the `borderTop` prop on `Table.Header`. The table border is now handled automatically.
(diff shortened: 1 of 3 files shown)


aiheon:
In StoryBook the borderTop does not look the same as your screenshot in the description 👀 

To be fair I think it might be because there's no space between the top of the table and the storybook header. Hard to say event when zooming in. I was expecting to see both the border top + the border bottom of storybook header 🤔 

image

gdostie:
> In StoryBook the borderTop does not look the same as your screenshot in the description 👀
> 
> To be fair I think it might be because there's no space between the top of the table and the storybook header. Hard to say event when zooming in. I was expecting to see both the border top + the border bottom of storybook header 🤔
> 
> image

Yeah the screenshot in the description is taken with an extra padding manually added through the browser debug console just to make the border visible, otherwise you don't see it

aiheon (approved):

PR #4536, opened 2026-07-24
fix(changesets): only suppress changelog entries in true prerelease mode

gdostie:
### Proposed Changes

`.changeset/changelog.cjs`'s `isPrerelease()` only checked whether `.changeset/pre.json` existed on disk, not its content. `changeset pre exit` sets `pre.json`'s `mode` field to `"exit"` rather than deleting the file immediately — the file is only removed after the following `changeset version` run completes. This meant that the first `changeset version` run after exiting prerelease mode (i.e. the real, final release) still saw `pre.json` on disk and incorrectly suppressed every changelog entry.

This is what happened in #4519: despite consuming 9 changesets (including the Mantine 9 bump and two breaking changes), every package's `CHANGELOG.md` only got an empty `## x.y.z` heading with no content.

Fix: `isPrerelease()` now reads and parses `pre.json` and returns `true` only when `mode === "pre"`. Missing file, `mode === "exit"`, or malformed JSON all correctly return `false` (fail open — don't suppress).

None. Internal changelog tooling only.

### Acceptance Criteria

- [ ] README.md is adjusted to reflect the proposed changes (if relevant)

### DIFF
--- a/.changeset/changelog.cjs
+++ b/.changeset/changelog.cjs
@@ -1,14 +1,20 @@
 // Custom changelog generator for Plasma.
-// - Suppresses all changelog entries during prereleases.
+// - Suppresses all changelog entries while .changeset/pre.json mode is "pre".
 // - Suppresses dependency update lines entirely.
 // - Simple entries: "- Summary [#PR](url)"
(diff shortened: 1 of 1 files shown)


toofff (approved):
LGTM

PR #4437, opened 2026-05-15
Expand Plasma LLM component docs with UX guidance

johnnyCoveo:
## Summary
- expand the  skill to generate richer UX and decision guidance
- extend the component doc format with guidance sections beyond props and usage
- add  as a first test component for the new format

## Why
This makes the LLM-facing component docs more useful for humans and agents by documenting not only API details, but also when to use components, when not to use them, and what an AI agent should infer from them.

## Notes
- this PR is intentionally based on  to make review easier
- only the delta from that branch should be reviewed

PR #4540, opened 2026-07-27
docs(llms): UX guidance for A–B components

johnnyCoveo:
## What

Carves the **A** and **B** component usage-guidance docs out of #4437 into a smaller, independently reviewable PR — the first of the planned splits of that larger PR. #4437 stays open as the tracking PR.

## Components in this slice

**A** — Accordion, ActionIcon, Alert, Anchor, AppShell
**B** — Badge, Breadcrumbs, BrowserPreview, Button

(`Anchor`, `Breadcrumbs` are new files; the rest are modifications.)

## What changed vs. a plain carve-out

These files also include a usage-guidance review pass (cross-checked against each component's Mantine `llms` doc and its own API reference):

- Explicit "ignore Mantine's variant/color model, use the Plasma sub-components" overrides where Plasma renames the variant model (ActionIcon, Alert, Badge, Button).
- Per-variant "when to use" coverage for destructive sub-components (ActionIcon, Button).
- Fixed an internal secondary/tertiary contradiction (Button).
- Removed unverified/incorrect claims (e.g. the `unstyled` accordion "variant", a false chevron-removal behavior, `Button.Group` guidance since it isn't in Figma/Storybook).

## Not included (belongs in later slices)

Global/shared parts of #4437 — the aggregate build files (`llms-full-txt.ts`, `llms-txt.ts`, `skill.md`, `tsconfig.json`) and `src/content/*` changes — are intentionally left out of this component-scoped slice.

## Follow-ups for the API-reference owner (out of scope here)

- **ActionIcon**: async-`onClick` auto-loading claim in the API reference is not real (confirmed).
- **Anchor** / **CloseButton**: API reference says "no additional Plasma props" but `size` is restricted — a real delta to note.
- **Button**: `Button.Group` listed in the API reference/sub-components but absent from Figma/Storybook.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/llms/src/components/Accordion.md
+++ b/packages/llms/src/components/Accordion.md
@@ -1,7 +1,96 @@
 ---
 name: Accordion
-description: Collapsible content panels with an additional disabled control variant.
+description: Collapsible content group for progressively revealing related sections without leaving the current page.
 ---
+
+# Usage guidance
+
+## What problem does it solve?
+
+The `Accordion` lets users scan a set of related section headings and open the content they need without showing every detail at once.
+
+It is useful when content is secondary to the main task, grouped into clear sections, and worth keeping on the same page.
+
+## When to use it
+
+Use `Accordion` when:
+
+- related content can be organized into short, meaningful section labels
+- users benefit from scanning section headings before reading details
+- the content is useful in context but does not need to be visible all at once
+- one or many sections MAY be open depending on the task
+
+Use `multiple` when users commonly compare information across several sections or need more than one section open while working.
+
+## When not to use it
+
+Do not use `Accordion` when:
+
+- the content is required for completing the current task and should be visible without disclosure
+- the sections represent navigation destinations rather than expandable content
+- users need to compare large amounts of content side by side
+- there are only one or two very short details that could be shown inline
+- hiding the content would make errors, warnings, or required decisions easier to miss
+
+## Decision-making guidance
+
+- Use `Accordion` for progressive disclosure inside a page, not for moving between pages or major views.
+- Use `Tabs` when each section is a peer view and users choose one view at a time.
+- Use inline helper text or an `Alert` when the information is important enough to remain visible.
+- Use a table, list, or cards when users need to compare many items rather than open panels one by one.
+- Use `Accordion.ControlDisabled` when a section heading should appear in the accordion structure but should not be interactive.
+
+## Variants
+
+- Use `default` for the standard accordion treatment.
+- Use `contained` when the accordion needs a clearer boundary from surrounding page content.
+- Use `separated` when each item should read as an independent expandable section.
+- Use `filled` when the accordion needs stronger visual weight.
+
+## States
+
+Important states include:
+
+- collapsed
+- expanded
+- multiple expanded items, when `multiple` is enabled
+- disabled control, through `Accordion.ControlDisabled`
+
+## Interaction notes
+
+- Each section needs a unique, consistent identifier so the accordion can track which panels are open or closed.
+- Icons next to section labels can help users recognize content faster, but the label text must always be clear enough on its own.
+- A section heading can be made non-interactive — it appears in the accordion structure but cannot be expanded or collapsed.
+
+## Accessibility expectations
+
+- Each control MUST have clear text that describes the panel content.
+- Panel content SHOULD remain logically related to its control.
+- Interactive content inside panels MUST remain keyboard reachable when the panel is open.
+
+## Content guidance
+
+- Control labels SHOULD be short and specific.
+- Labels SHOULD describe the content inside the panel, not the action of expanding it.
+- Keep panel content focused on one section topic.
+- If every panel label needs a long sentence to be understandable, the content likely needs another structure.
+
+## Common anti-patterns
+
+- Using accordions as primary navigation.
+- Hiding required form fields or critical warnings inside collapsed panels.
+- Creating many vague sections such as "Details" or "More information."
+- Using an accordion to avoid designing a clearer content hierarchy.
+- Nesting accordions unless the relationship between levels is unmistakable.
+
+## What an AI agent should understand
+
+- `Accordion` is for progressive disclosure of related in-page content.
+- It is a poor choice when the content is required, critical, or needs direct comparison.
+- Use `Accordion.Item`, `Accordion.Control`, and `Accordion.Panel` together to compose each section.
+- `Accordion.ControlDisabled` is Plasma-specific and should be used for non-interactive controls inside an accordion structure.
+
+# API reference
 
 ## Props
 
@@ -9,14 +98,12 @@ _No additional props beyond the Mantine base component._
 
 ## Sub-components
 
-Plasma provides pre-configured sub-components as convenience wrappers. You SHOULD use these over setting props manually.
+Plasma exposes Mantine's compound accordion API and adds a disabled control convenience wrapper.
 
-- `Accordion.Control`
-- `Accordion.ControlDisabled`
-- `Accordion.Panel`
 - `Accordion.Item`
-
-`Accordion.ControlDisabled` renders a control with no chevron and disabled pointer events. You SHOULD use it for accordion items that are always expanded and cannot be collapsed by the user.
+- `Accordion.Control`
+- `Accordion.Panel`
+- `Accordion.ControlDisabled`
 
 ## Usage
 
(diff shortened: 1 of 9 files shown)


aiheon (approved):

packages/llms/src/components/Accordion.md
```diff
@@ -1,22 +1,109 @@
...
+- Use inline helper text or an `Alert` when the information is important enough to remain visible.
+- Use a table, list, or cards when users need to compare many items rather than open panels one by one.
+- Use `Accordion.ControlDisabled` when a section heading should appear in the accordion structure but should not be interactive.
+
+## Variants
+
+- Use `default` for the standard accordion treatment.
+- Use `contained` when the accordion needs a clearer boundary from surrounding page content.
+- Use `separated` when each item should read as an independent expandable section.
+- Use `filled` when the accordion needs stronger visual weight.
```
GermainBergeron, line 48: Do we want to support all those variants? There's no variant in Figma

packages/llms/src/components/Anchor.md
```diff
@@ -0,0 +1,99 @@
...
+
+- Navigating to an external URL or a different route within the application.
+- Embedding a navigation action inline within a sentence or paragraph.
+- Providing secondary navigation options such as "Learn more" or "See all results" beside descriptive text.
+
+## When not to use it
+
+- When the action triggers a mutation, submission, or workflow step — use `Button` instead.
+- When the link stands alone as a primary call-to-action with enough prominence to warrant a button shape.
+- When breadcrumb-style navigation is needed — use `Breadcrumbs` with `Anchor` children rather than bare `Anchor` elements.
```
GermainBergeron: We usually use `Header.Breadcrumbs` with `Header.BreadcrumbAnchor` instead of directly using `Breadcrumbs` and `Anchor`. I'm not sure if we should mention it here

```tsx
<Header.Breadcrumbs>
    <Header.BreadcrumbAnchor>Parent</Header.BreadcrumbAnchor>
</Header.Breadcrumbs>
```

packages/llms/src/components/Anchor.md
```diff
@@ -0,0 +1,99 @@
...
+## Variants
+
+Plasma exposes two size options:
+
+- `sm` (default) — standard body-text size.
+- `xs` — smaller, suited to dense UI areas.
+
+## Accessibility expectations
+
+- Link text MUST describe the destination or action, not just "click here" or "read more".
```
GermainBergeron, line 40: Not a hard requirement, but on line 16 we say `Learn more`, which is kind of close to `read more`. I'm wondering if we should use a different example

packages/llms/src/components/Anchor.md
```diff
@@ -0,0 +1,99 @@
...
+## Common anti-patterns
+
+- Using `Anchor` to trigger an `onClick` action with no `href` — use `Button.Tertiary` or an `ActionIcon` instead.
+- Wrapping large blocks of content in an `Anchor`; links SHOULD target meaningful, concise labels.
+- Omitting `inherit` when placing an `Anchor` inside a typographic component, causing a size mismatch.
+
+## What an AI agent should understand
+
+- `Anchor` is a thin Plasma wrapper around Mantine's `Anchor`; all Mantine `AnchorProps` are available.
+- Use `inherit` when nesting inside `Text` or `Breadcrumbs` to avoid font-size conflicts.
```
GermainBergeron: I would add a mention about font-weight
```suggestion
- Use `inherit` when nesting inside `Text` or `Breadcrumbs` to avoid font-size and font-weight conflicts.
```

And again would use a different component than Breadcrumbs to avoid confusing the agent

packages/llms/src/components/Anchor.md
```diff
@@ -0,0 +1,99 @@
...
+# API reference
+
+## Props
+
+> Extends: `AnchorProps` from `@mantine/core`. No additional Plasma-specific props beyond the Mantine base component.
+
+## Usage
+
+```tsx
+import {Anchor, Breadcrumbs} from '@coveord/plasma-mantine';
```
GermainBergeron:
```suggestion
import {Anchor} from '@coveord/plasma-mantine';
```

packages/llms/src/components/Anchor.md
```diff
@@ -0,0 +1,99 @@
...
+        <Breadcrumbs>
+            <Anchor href="/home" inherit>
+                Home
+            </Anchor>
+            <Anchor href="/catalog" inherit>
+                Catalog
+            </Anchor>
+        </Breadcrumbs>
+    );
+}
```
GermainBergeron:
```suggestion
// Inside an Alert description — inherit parent font size
import {Alert, Anchor} from '@coveord/plasma-mantine';

function SomeAlert() {
    return (
        <Alert.Information title="Alert title">
            Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et <Anchor inherit href="#dolore-magna-aliqua">dolore magna aliqua</Anchor>.
        </Alert.Information>
    );
}
```

packages/llms/src/components/Accordion.md
```diff
@@ -1,22 +1,109 @@
...
+
+## Common anti-patterns
+
+- Using accordions as primary navigation.
+- Hiding required form fields or critical warnings inside collapsed panels.
+- Creating many vague sections such as "Details" or "More information."
+- Using an accordion to avoid designing a clearer content hierarchy.
+- Nesting accordions unless the relationship between levels is unmistakable.
+
+## What an AI agent should understand
```
gdostie: What is this section for? I feel like this is just duplicating the information above. The whole file is supposed to be describing what an AI agent should understand 😕

packages/llms/src/components/Badge.md
```diff
@@ -3,6 +3,67 @@ name: Badge
...
+- the value is textual and needs more context than an icon alone
+- a list, card, or header needs compact status information
+
+## When not to use it
+
+Do not use `Badge` when:
+
+- the message needs explanation or action; use `Alert` or inline text
+- only a small semantic marker is needed; use `StatusToken`
+- the value is interactive; use a button, chip, or link pattern
```
GermainBergeron: `Chip` was previously misused in the commerce-hub as some kind of badge (I guess they preferred the styling). I'm not sure I would mention it here. A chip is akin to a checkbox

packages/llms/src/components/Breadcrumbs.md
GermainBergeron, line 1: I'm not sure we need this file, or at least not in this state. We could refer to the Header component as we always put the breadcrumbs in the Header

packages/llms/src/components/BrowserPreview.md
```diff
@@ -3,6 +3,56 @@ name: BrowserPreview
...
+- the interface needs to distinguish editable settings from preview output
+
+## When not to use it
+
+Do not use `BrowserPreview` when:
+
+- the content is the actual page, not a preview
+- the frame would add decoration without clarifying context
+- users need to interact with a full application shell
+- the preview content is too small or abstract to benefit from browser chrome
```
GermainBergeron: I don't understand this one (`from browser chrome`)

GermainBergeron (commented):
I got a couple of comments. I think I would remove the `What an AI agent should understand` section of all the files since the file is targeted at the agents anyway

packages/llms/src/components/Accordion.md
johnnyCoveo: I'm not sure yet — based on our discussion, I feel like we need to review this with the designers to understand whether there's a need for some variants. I believe there might be a need for the `contained` variant along with `default`. For now, since they're still using the components and not custom ones, I think it's fine.

packages/llms/src/components/Breadcrumbs.md
johnnyCoveo: I agree, but it's also shown in our Storybook by itself — if we remove this file, I'd remove it from Storybook too. I don't see it being used anywhere other than in headers, except maybe the "sidebar header" Esther showed us the other day. For now, I think we can keep it like this.

packages/llms/src/components/Anchor.md
johnnyCoveo: Done — repointed the reference to `Header.Breadcrumbs` / `Header.BreadcrumbAnchor`.

packages/llms/src/components/Accordion.md
johnnyCoveo: Done — removed the "What an AI agent should understand" section across all the files.

GermainBergeron (approved):

GermainBergeron (approved):

PR #4548, opened 2026-07-27
docs(llms): UX guidance for Q–S components

johnnyCoveo:
## What

Carves the **Q through S** component usage-guidance docs out of #4437 into a smaller, independently reviewable PR — the fourth split (after #4540 A–B, #4542 C–H, #4547 I–P). #4437 stays open as the tracking PR.

## Components in this slice (11)

Radio, RadioCard, ScrollArea, SegmentedControl, Select, Skeleton, Slider, StatusToken, Stepper, StickyFooter, Switch

## Review-pass refinement

- **StatusToken**: added the missing default `info` variant to the usage guidance (the API reference lists it as the default, but the Variants section omitted it).

The other ten were reviewed and found consistent with Mantine and their API references — notably RadioCard's `disabledTooltip` delta, Select's read-only visual state, SegmentedControl's Tabs-vs-mode distinction, and StickyFooter's deprecated `modal-footer` variant guidance.

## Not included (belongs in later slices)

Global/shared parts of #4437 — aggregate build files (`llms-full-txt.ts`, `llms-txt.ts`, `skill.md`, `tsconfig.json`) and `src/content/*` changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4547, opened 2026-07-27
docs(llms): UX guidance for I–P components

johnnyCoveo:
## What

Carves the **I through P** component usage-guidance docs out of #4437 into a smaller, independently reviewable PR — the third split (after #4540 for A–B and #4542 for C–H). #4437 stays open as the tracking PR.

## Components in this slice (18)

**I–L** — Image, InfoToken, Input, Kbd, LastUpdated, Loader
**M–P** — Menu, Modal, MonthPickerInput, MultiSelect, NavLink, Notification, NumberInput, Pagination, PasswordInput, Pill, Progress, Prompt

## Review-pass refinements included

Cross-checked against each component's Mantine `llms` doc / type definitions (where a Mantine base exists); custom Coveo components (InfoToken, LastUpdated) reviewed on internal consistency.

- **Loader**: removed an inaccurate claim that Mantine renders `role="presentation"` by default (its source sets no ARIA role); kept the correct "not announced to screen readers" point. Simplified a Storybook-internal default-size sentence.
- **Progress** & **Notification**: both are pure Mantine re-exports with **no `variant` prop** — reframed their "Variants" sections as the real `color`-prop conventions (verified `color` exists in Mantine's type defs; Storybook just doesn't surface it).
- **NumberInput**: corrected "read-only not built-in" — Mantine supports `readOnly` natively.
- **Image**: added `fallbackSrc` guidance for broken-image handling.
- **MonthPickerInput** / **MultiSelect**: trimmed Storybook-internal `withLabelInfoProps` references.
- Verified in-repo that referenced components (`BackgroundImage`, `Code`, `CircleLoader`, `RingProgress`, `SemiCircleProgress`, `TagsInput`, `PillsInput`) all exist.

## Not included (belongs in later slices)

Global/shared parts of #4437 — aggregate build files (`llms-full-txt.ts`, `llms-txt.ts`, `skill.md`, `tsconfig.json`) and `src/content/*` changes.

## Follow-up for the API-reference owner (out of scope here)

- **Pill**: API reference lists `size` as `sm|md`, but it's a pure re-export accepting Mantine's `xs`–`xl`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4542, opened 2026-07-27
docs(llms): UX guidance for C–H components

johnnyCoveo:
## What

Carves the **C through H** component usage-guidance docs out of #4437 into a smaller, independently reviewable PR — the second split of that larger PR (after #4540 for A–B). #4437 stays open as the tracking PR.

## Components in this slice

**C** — Card, Checkbox, ChildForm, Chip, CloseButton, CodeEditor, Collection, CopyToClipboard
**D** — DatePickerInput
**E–H** — EllipsisText, Facet, Header

## Review-pass refinements included

Cross-checked against each component's Mantine `llms` doc (where one exists) and its own API reference; the custom Coveo components (ChildForm, CodeEditor, Collection, CopyToClipboard, Facet, Header, EllipsisText) were reviewed on internal consistency.

- **CloseButton**: corrected a false claim that Mantine renders a default `aria-label="Close"` — it doesn't; an explicit `aria-label` is required.
- **DatePickerInput**: removed the incorrect "presets only meaningful in range mode" anti-pattern and fixed the `presets` value shape (verified against source: Plasma re-exports Mantine's component verbatim, so `presets` works for single-date too).
- **Card**: fixed the redesign-banner grammar and reworded the "not for messages/notifications → use `Alert`" guidance.
- **Checkbox**: removed a stray "Open questions for our system" section.
- **Header / Facet / EllipsisText**: added missing coverage (breadcrumb sub-components; the 7-value search auto-hide threshold; single-line vs `lineClamp` truncation).

## Not included (belongs in later slices)

Global/shared parts of #4437 — aggregate build files (`llms-full-txt.ts`, `llms-txt.ts`, `skill.md`, `tsconfig.json`) and `src/content/*` changes — are left out of this component-scoped slice.

## Follow-ups for the API-reference owner (out of scope here)

- **CloseButton**: API reference says "no additional Plasma props" but `size` is restricted to `sm`/`md` — a real delta to note.
- **DatePickerInput**: API-reference line for `presets` repeats the "only meaningful with type=range" error; correct it there too.
- **ChildForm**: API reference duplicates the intro sentence.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4549, opened 2026-07-27
docs(llms): UX guidance for T–Z components

johnnyCoveo:
## What

Carves the final **T through Z** component usage-guidance docs out of #4437 into a smaller, independently reviewable PR — the fifth and last component slice (after #4540 A–B, #4542 C–H, #4547 I–P, #4548 Q–S). #4437 stays open as the tracking PR.

## Components in this slice (7)

Table, Tabs, TextInput, Textarea, TimePicker, Tooltip, YearPickerInput

## Review-pass refinements

- **TextInput**: fixed a broken "when not to use" reference — `DateInput`/`DateTimePicker` don't exist in Plasma (verified in-repo); now points to `DatePickerInput`/`TimePicker`. (`Autocomplete` reference confirmed valid.)
- **Textarea**: corrected imprecise alternatives — `Input` → `TextInput` for single-line, and "a `Select` with multi-select" → `MultiSelect`.
- **Tooltip**: removed a Storybook-only `freezeOpen` story-prop mention; kept the real `opened` guidance.
- **Tabs / TimePicker / YearPickerInput**: reviewed, consistent with Mantine.
- **Table**: left as-is — it carries an "under active redesign" banner, so deep edits are deferred (same handling as Card).

## Not included (belongs in later slices)

Global/shared parts of #4437 — aggregate build files (`llms-full-txt.ts`, `llms-txt.ts`, `skill.md`, `tsconfig.json`) and `src/content/*` changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/llms/src/components/Table.md
+++ b/packages/llms/src/components/Table.md
@@ -16,7 +16,7 @@ It is useful for operational views where users need to inspect many records and 
 
 Use `Table` when:
 
-- data is naturally structured as rows and columns
+- each record has many fields which the user may need to consult
 - users need to compare values across records
 - filtering, pagination, row actions, selection, or expandable row content are needed
 - the view needs table-specific empty, loading, footer, or pagination patterns
--- a/packages/llms/src/components/Table.md
+++ b/packages/llms/src/components/Table.md
@@ -27,6 +27,7 @@ Do not use `Table` when:
 - the data shows a progression over time, use chart components instead
 - users are editing a repeatable form list; use `Collection`
 - there are only a few simple values that can be shown as text or cards
+- the number of entries is static and pre-determined — a full table is overkill when all items are already known
 - the layout would require hiding the most important information in many columns
 
 ## Decision-making guidance
@@ -50,11 +51,7 @@ Important states include:
 
 ## Interaction notes
 
-- `store` MUST come from `useTable`.
-- `getRowId` SHOULD be provided when records have stable identifiers.
-- `getRowActions` SHOULD return only actions relevant to the selected row or rows.
 - Expandable row content SHOULD add detail without replacing the row's core scannable information.
-- Additional root nodes MAY be provided when related overlays should not clear row selection.
 
 ## Content guidance
 
@@ -143,7 +140,7 @@ export function Example() {
     const store = useTable<TData>({
         initialState: {
             totalEntries: data.length,
-            pagination: {pageIndex: 0, pageSize: 10},
+            pagination: {page: 0, perPage: 10},
         },
     });
 
--- a/packages/llms/src/components/Table.md
+++ b/packages/llms/src/components/Table.md
@@ -3,6 +3,83 @@ name: Table
 description: Table for displaying tabular data with optional filtering, pagination, row actions, and layout switching.
 ---
 
+> **Note:** This component is currently under active redesign. Usage guidance will be updated once the initiative concludes. In the meantime, use this as a general reference only.
+
+# Usage guidance
+
+## What problem does it solve?
+
+The `Table` lets users scan, compare, filter, paginate, select, and act on structured data in rows and columns.
+
+It is useful for operational views where users need to inspect many records and perform actions on one or more of them.
+
+## When to use it
+
+Use `Table` when:
+
+- data is naturally structured as rows and columns
+- users need to compare values across records
+- filtering, pagination, row actions, selection, or expandable row content are needed
+- the view needs table-specific empty, loading, footer, or pagination patterns
+
+## When not to use it
+
+Do not use `Table` when:
+
+- the content is primarily visual or card-like
+- users are editing a repeatable form list; use `Collection`
+- there are only a few simple values that can be shown as text or cards
+- the layout would require hiding the most important information in many columns
+
+## Decision-making guidance
+
+- Use `Table` for data display and operational record management.
+- Use `Collection` for editable repeated form items.
+- Use `Facet` or table filter sub-components when users need to narrow data.
+- Use row actions for item-specific commands and header/footer actions for table-level commands.
+
+## States
+
+Important states include:
+
+- initially loading with `data={null}`
+- loaded with rows
+- empty or no matching data
+- selected rows
+- expanded rows
+- paginated data
+- filtered data
+
+## Interaction notes
+
+- `store` MUST come from `useTable`.
+- `getRowId` SHOULD be provided when records have stable identifiers.
+- `getRowActions` SHOULD return only actions relevant to the selected row or rows.
+- Expandable row content SHOULD add detail without replacing the row's core scannable information.
+- Additional root nodes MAY be provided when related overlays should not clear row selection.
+
+## Content guidance
+
+- Column headers SHOULD be short and describe the data clearly.
+- Empty states SHOULD explain whether there is no data or no matching data.
+- Row actions SHOULD use clear verbs and destructive labels when appropriate.
+- Filter labels SHOULD match the terms users use to understand the data.
+
+## Common anti-patterns
+
+- Using a table for layout instead of data.
+- Putting editable repeated form fields in a table when `Collection` is the intended pattern.
+- Omitting stable row IDs for selectable or reorderable data.
+- Hiding key record identity in expandable content instead of visible columns.
+
+## What an AI agent should understand
+
+- `Table` is for structured data display and operational record actions.
+- Use it with `useTable`, stable columns, and row IDs when available.
+- Use `Collection` for editable repeated form inputs and simpler layouts for non-tabular content.
+
+# API reference
+
 ## Props
 
 > Extends: `BoxProps & StylesApiProps<PlasmaTableFactory>`. Only Plasma-specific props are listed below; refer to Mantine documentation for inherited props.
@@ -74,7 +151,7 @@ export function Example() {
     const store = useTable<TData>({
         initialState: {
             totalEntries: data.length,
-            pagination: {page: 0, perPage: 10},
+            pagination: {pageIndex: 0, pageSize: 10},
         },
     });
 
--- a/packages/llms/src/components/Table.md
+++ b/packages/llms/src/components/Table.md
@@ -25,7 +25,7 @@ Use `Table` when:
 
 Do not use `Table` when:
 
-- the content is primarily visual or card-like
+- the data shows a progression over time, use chart components instead
 - users are editing a repeatable form list; use `Collection`
 - there are only a few simple values that can be shown as text or cards
 - the layout would require hiding the most important information in many columns
--- a/packages/llms/src/components/Table.md
+++ b/packages/llms/src/components/Table.md
@@ -3,7 +3,6 @@ name: Table
 description: Table for displaying tabular data with optional filtering, pagination, row actions, and layout switching.
 ---
 
-> **Note:** This component is currently under active redesign. Usage guidance will be updated once the initiative concludes. In the meantime, use this as a general reference only.
 
 # Usage guidance
 
--- a/packages/llms/src/components/Table.md
+++ b/packages/llms/src/components/Table.md
@@ -3,7 +3,6 @@ name: Table
 description: Table for displaying tabular data with optional filtering, pagination, row actions, and layout switching.
 ---
 
-
 # Usage guidance
 
 ## What problem does it solve?
@@ -70,12 +69,6 @@ Important states include:
 - Putting editable repeated form fields in a table when `Collection` is the intended pattern.
 - Omitting stable row IDs for selectable or reorderable data.
 - Hiding key record identity in expandable content instead of visible columns.
-
-## What an AI agent should understand
-
-- `Table` is for structured data display and operational record actions.
-- Use it with `useTable`, stable columns, and row IDs when available.
-- Use `Collection` for editable repeated form inputs and simpler layouts for non-tabular content.
 
 # API reference
 
--- a/packages/llms/src/components/Tabs.md
+++ b/packages/llms/src/components/Tabs.md
@@ -1,0 +1,98 @@
+---
+name: Tabs
+description: Navigation pattern that organises related content into labelled panels, showing one at a time.
+---
+
+# Usage guidance
+
+## What problem does it solve?
+
+`Tabs` divides a page section into peer content categories so users can switch between them without leaving the page or losing scroll position in the rest of the layout.
+
+## When to use it
+
+- Content belongs to parallel, peer categories of roughly equal importance (e.g., Overview / Settings / History).
+- All panels are part of the same conceptual entity (e.g., a resource's detail view).
+- The number of tabs is between two and roughly six.
+
+## When not to use it
+
+- The categories are sequential steps; use `Stepper`.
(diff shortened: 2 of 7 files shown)


packages/llms/src/components/Table.md
```diff
@@ -3,6 +3,83 @@ name: Table
 description: Table for displaying tabular data with optional filtering, pagination, row actions, and layout switching.
 ---
 
+> **Note:** This component is currently under active redesign. Usage guidance will be updated once the initiative concludes. In the meantime, use this as a general reference only.
```
gdostie: This note kind of undermines the importance of what is written in this file. I would rather not have the note and simply update this file iteratively with the code when the table gets updated.


```suggestion
```

packages/llms/src/components/Table.md
```diff
@@ -3,6 +3,83 @@ name: Table
...
+
+The `Table` lets users scan, compare, filter, paginate, select, and act on structured data in rows and columns.
+
+It is useful for operational views where users need to inspect many records and perform actions on one or more of them.
+
+## When to use it
+
+Use `Table` when:
+
+- data is naturally structured as rows and columns
```
gdostie: What does this mean?


```suggestion
- each record has many fields which the user may need to consult
```

packages/llms/src/components/Table.md
```diff
@@ -3,6 +3,83 @@ name: Table
...
+- users need to compare values across records
+- filtering, pagination, row actions, selection, or expandable row content are needed
```
gdostie, line 20: filtering, pagination, etc are just means to an end. The real need for the user is what? Being able to dig down into a large dataset perhaps? Manage records?

packages/llms/src/components/Table.md
```diff
@@ -3,6 +3,83 @@ name: Table
...
+- the view needs table-specific empty, loading, footer, or pagination patterns
+
+## When not to use it
+
+Do not use `Table` when:
+
+- the content is primarily visual or card-like
```
gdostie: We do have a card layout available for the table so this is kind of misleading.

Image


```suggestion
- the data shows a progression over time, use chart components instead
```

packages/llms/src/components/Table.md
```diff
@@ -3,6 +3,83 @@ name: Table
...
+- users are editing a repeatable form list; use `Collection`
+- there are only a few simple values that can be shown as text or cards
+- the layout would require hiding the most important information in many columns
```
gdostie, line 31: I would also add:

```md
- the number of entries is static (pre-determined)
```

We usually don't want to display a full fledged table when all the items are already known (static)

packages/llms/src/components/Table.md
```diff
@@ -3,6 +3,83 @@ name: Table
...
+- selected rows
+- expanded rows
+- paginated data
+- filtered data
+
+## Interaction notes
+
+- `store` MUST come from `useTable`.
+- `getRowId` SHOULD be provided when records have stable identifiers.
+- `getRowActions` SHOULD return only actions relevant to the selected row or rows.
```
gdostie: These are more implementation focused no? 🤔 They should rather appear in the API reference portion of the spec.

packages/llms/src/components/Table.md
```diff
@@ -74,7 +151,7 @@ export function Example() {
     const store = useTable<TData>({
         initialState: {
             totalEntries: data.length,
-            pagination: {page: 0, perPage: 10},
+            pagination: {pageIndex: 0, pageSize: 10},
         },
```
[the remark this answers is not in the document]
gdostie: Copilot is right.

packages/llms/src/components/Tabs.md
```diff
@@ -0,0 +1,98 @@
...
+- Content belongs to parallel, peer categories of roughly equal importance (e.g., Overview / Settings / History).
+- All panels are part of the same conceptual entity (e.g., a resource's detail view).
+- The number of tabs is between two and roughly six.
+
+## When not to use it
+
+- The categories are sequential steps; use `Stepper`.
+- Only one panel exists; the tab bar adds noise without value.
+- The panels represent separate destinations or pages; use router-level navigation (`Link`, `NavLink`) instead.
+- More than six or seven tabs are needed; consider a secondary sidebar navigation or a `Select`-driven panel switcher.
```
gdostie: What is a "Select-driven panel switcher"? I don't think we have that pattern anywhere.

```suggestion
- More than six or seven tabs are needed; consider a secondary sidebar navigation.
```

packages/llms/src/components/Tabs.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Variants
+
+- **Horizontal** (default): tab list appears above the panels.
+- **Vertical**: tab list appears to the left of the panels; set `orientation="vertical"`.
+
+## Interaction notes
+
+- Keyboard: arrow keys move focus between tabs; `Enter` or `Space` activates the focused tab.
+- Tabs activate on focus by default in Mantine; set `activateTabWithKeyboard` to `false` if panels have expensive loading cost.
```
gdostie: Perhaps we should have guidance around disabled tabs. Should we display disabled tabs or not display them at all?

packages/llms/src/components/Tabs.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Accessibility expectations
+
+- The component renders a `role="tablist"` with `role="tab"` items and associated `role="tabpanel"` elements, satisfying the ARIA Tabs pattern automatically.
```
gdostie, line 42: This is done transparently by the Tab component's implementation, users of plasma don't need to think about it, so does agents.

```suggestion
```

packages/llms/src/components/Tabs.md
```diff
@@ -0,0 +1,98 @@
...
+- Tab labels MUST be unique within the tab list.
+- Panels MUST not be empty; if a panel has no content yet, show a loading indicator or a placeholder message.
+
+## Common anti-patterns
+
+- Using tabs for navigation that changes the URL without wiring the `value` to the router, causing the active tab to reset on refresh.
+- Mixing tab-level and page-level navigation in the same `Tabs.List`.
+- Nesting tab sets inside tab panels (deeply nested tabs are disorienting).
```
gdostie: This is not really true, if one tab set is in orientation horizontal and the nested one is orientation vertical (or the other way around). The problem is true when both parent and nested tab sets are of the same orientation.

packages/llms/src/components/Textarea.md
```diff
@@ -0,0 +1,138 @@
...
+- Default `rows` is sufficient for most use cases; only increase it when the expected content is substantially longer (e.g. a legal disclaimer or a full message body).
+- Enable `autosize` when the amount of text is unpredictable and you want the field to grow with the user's input rather than scroll internally; pair with `minRows` and `maxRows` to keep the layout bounded.
+- Set `resize="none"` in constrained layouts (e.g. inside cards or narrow panels) where user resizing would break the surrounding design.
+- `clearable` is not a prop on `Textarea`; to allow clearing, provide a separate clear button or use a controlled value.
+
+## States
+
+- **Default** — editable, showing `placeholder` when empty.
+- **Disabled** — visually dimmed; interaction is blocked. Use when the field is not applicable in the current context.
+- **Read-only** — displays the value without allowing edits. Use when the value must be visible but not changeable (e.g. a generated summary).
```
gdostie:
```suggestion
- **Read-only** — displays the value without allowing edits. Use when the value must be visible but not changeable (e.g. the user lacks permissions to edit).
```

packages/llms/src/components/TextInput.md
```diff
@@ -0,0 +1,117 @@
...
+## Interaction notes
+
+- `leftSection` and `rightSection` accept any `ReactNode`; icons or small interactive elements (e.g., a clear button) work well here.
+- Section elements do not receive focus automatically; interactive elements inside sections MUST be individually focusable.
+
+## Accessibility expectations
+
+- The `label` prop renders a `<label>` associated with the `<input>` via `id`/`htmlFor`. You MUST always pair a visible label with the field.
+- `description` and `error` are linked to the input via `aria-describedby` automatically by Mantine.
+- When using only a `placeholder` and no `label`, add `aria-label` manually so screen readers announce the field's purpose.
```
gdostie: > When using only a `placeholder` and no `label`

This is a pattern we want to avoid

packages/llms/src/components/TextInput.md
```diff
@@ -0,0 +1,117 @@
...
+- **Read-only**: non-editable but the value can be selected and copied.
+- **Error**: red border and inline error message below the field.
+- **Required**: label marked with a required indicator.
+
+## Interaction notes
+
+- `leftSection` and `rightSection` accept any `ReactNode`; icons or small interactive elements (e.g., a clear button) work well here.
+- Section elements do not receive focus automatically; interactive elements inside sections MUST be individually focusable.
+
+## Accessibility expectations
```
gdostie, line 47: We should add something about the `type` prop. It tells the browser about which type of information will be entered. It is part of the html standard.

See https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input#input_types

For instance, a text input for entering an email address should look like:

```tsx
<TextInput
  label="Email address"
  placeholder="example@email.com"
  type="email"
/>
```

packages/llms/src/components/Tooltip.md
```diff
@@ -0,0 +1,126 @@
...
+- Providing additional contextual information about a UI element without cluttering the interface.
+- Providing a short elaboration on a truncated label or an abbreviation.
+- Showing a full value when an `EllipsisText` component has truncated content.
+
+## When not to use it
+
+- When the information is critical for the user to complete their task — critical guidance MUST be visible at all times; use `description` on the form field or inline text instead.
+- When the content requires interaction (links, buttons, form elements) — a tooltip dismisses on mouse-out, making interactive content inside it inaccessible. Use a `Popover` instead.
+- When the content is longer than two short sentences — tooltips are for brief labels, not paragraphs. Use a `Popover` or a drawer for richer content.
+- On mobile touch targets where hover is unavailable — consider `Popover` with a tap trigger.
```
gdostie: We don't support mobile


```suggestion
```

packages/llms/src/components/Tooltip.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Decision-making guidance
+
+- Keep `label` to a single short phrase or sentence; if you need more space, reconsider whether a `Popover` or adjacent text is more appropriate.
+- Use `position` to avoid clipping against viewport edges or overlapping nearby UI; prefer `"top"` by default and override as needed.
```
gdostie: It is best to let the tooltip manage its position itself, by not specifying any. Use position only when needed to overcome a specific positioning issue.

packages/llms/src/components/Tooltip.md
```diff
@@ -0,0 +1,126 @@
...
+
+```tsx
+import {Tooltip, ActionIcon} from '@coveord/plasma-mantine';
+
+// Labelling an icon-only button
+function DeleteButton() {
+    return (
+        <Tooltip label="Delete item">
+            <ActionIcon aria-label="Delete item" color="red">
+                <DeleteIcon />
```
gdostie:
```suggestion
                <IconTrash />
```
(and add the import)
```tsx
import {IconTrash} from '@coveord/plasma-react-icons';
```

packages/llms/src/components/Table.md
johnnyCoveo: Done — banner removed.

packages/llms/src/components/Table.md
johnnyCoveo: Applied ✅ (the "each record has many fields" bullet).

packages/llms/src/components/Table.md
johnnyCoveo: The intro now frames it around inspecting/managing many records, and the "each record has many fields" bullet captures the real need — I kept the capability list as a secondary signal. Happy to sharpen it further if you'd like.

packages/llms/src/components/Table.md
johnnyCoveo: Applied ✅ — replaced with the "data shows a progression over time → use chart components" bullet.

packages/llms/src/components/Table.md
johnnyCoveo: Added ✅ — "the number of entries is static and pre-determined".

packages/llms/src/components/Table.md
johnnyCoveo: Removed those implementation-focused bullets from Interaction notes — that detail belongs in the API reference.

packages/llms/src/components/Table.md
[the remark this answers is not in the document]
johnnyCoveo: Fixed — the example now uses `{page, perPage}` (zero-based), matching `useTable`.

packages/llms/src/components/Tabs.md
johnnyCoveo: Applied ✅ — dropped the "Select-driven panel switcher" mention.

packages/llms/src/components/Tabs.md
johnnyCoveo: Good question — deferring disabled-tabs guidance to a follow-up; it needs a design call on whether to show or hide disabled tabs.

packages/llms/src/components/Tabs.md
johnnyCoveo: Removed ✅ — it's transparent to consumers/agents.

packages/llms/src/components/Tabs.md
johnnyCoveo: Reworded to scope it to same-orientation nesting (and to suggest differing the orientation if nesting is unavoidable).

packages/llms/src/components/Textarea.md
johnnyCoveo: Applied ✅

packages/llms/src/components/TextInput.md
johnnyCoveo: Reworded — the guidance now requires an accessible name, preferring a visible label, and the placeholder-only wording is gone.

packages/llms/src/components/TextInput.md
johnnyCoveo: Added guidance on the `type` prop (`email`, `tel`, `url`, `number`, …) ✅

packages/llms/src/components/Tooltip.md
johnnyCoveo: Removed ✅

packages/llms/src/components/Tooltip.md
johnnyCoveo: Reworded — let the tooltip auto-position by default; only set `position` to fix a specific placement issue.

packages/llms/src/components/Tooltip.md
johnnyCoveo: Applied — using `IconTrash` and added the import.

GermainBergeron (approved):

PR #4550, opened 2026-07-27
Fix CodeEditor crash when hidden and shown again by React Activity

gdostie:
### Proposed Changes

Following the Mantine 9 upgrade, placing a `CodeEditor` inside a container that hides inactive content with React 19's `<Activity>` (Tabs, Collapse, Accordion with `keepMountedMode="activity"`) would crash when switching away from and back to the panel.

Root cause: `<Activity mode="hidden">` tears down effects while preserving state. `@monaco-editor/react` disposes the underlying Monaco editor in its mount-effect cleanup but keeps its internal "editor ready" flag, so on re-show it never recreates the editor and operates on a disposed instance.

Changes:

- **`CodeEditor`**: subscribe to `editor.onDidDispose` and bump a `key` on the `<Editor>` so a fresh instance is created when the panel becomes visible again. The value is preserved via the controlled `_value` state.
- **Test util** (`src/__tests__/Utils.tsx`): added an optional `env` to the shared `render` helper (defaults to `'test'`) so tests can opt into the real runtime behavior where React `<Activity>` is active.
- **Regression test** (`CodeEditor.activity.spec.tsx`): exercises the real `@monaco-editor/react` mount/dispose lifecycle inside real Mantine `Tabs` with `env="default"`, faking only the Monaco engine via `window.monaco`. Asserts the editor is recreated after a hide/show cycle.
- **Vitest config**: inline the `@monaco-editor/*` packages so their default-import interop resolves correctly under Vitest.
- **Mock**: added `onDidDispose` to the `@monaco-editor/react` test mock.

The existing `Tabs` theme default (`keepMountedMode: 'display-none'`) is intentionally kept as a cheap belt-and-suspenders default; the `CodeEditor` fix covers every other Activity source.

> Note: a manual verification story (`CodeEditor` within `Tabs`) was implemented locally but intentionally not committed.

None. No public API changes.

### Acceptance Criteria

### DIFF
--- a/.changeset/happy-tigers-dream.md
+++ b/.changeset/happy-tigers-dream.md
@@ -4,7 +4,4 @@
 
 Fix `CodeEditor` crashing when hidden and shown again by React's `Activity`
 
-When a `CodeEditor` was placed inside a component that hides inactive content with React's `Activity`
-(such as `Tabs`, `Collapse`, or `Accordion` with `keepMountedMode="activity"`), switching away and back
-disposed the underlying Monaco editor and left it in a broken state. The editor now recreates itself when
-it is disposed while still mounted, so it keeps working across visibility changes.
+When a `CodeEditor` was placed inside a component that hides inactive content with React's `Activity` (such as `Tabs`, `Collapse`, or `Accordion` with `keepMountedMode="activity"`), switching away and back disposed the underlying Monaco editor and left it in a broken state. The editor now recreates itself when it is disposed while still mounted, so it keeps working across visibility changes.
--- a/.changeset/happy-tigers-dream.md
+++ b/.changeset/happy-tigers-dream.md
@@ -1,0 +1,10 @@
+---
+'@coveord/plasma-mantine': patch
+---
+
+Fix `CodeEditor` crashing when hidden and shown again by React's `Activity`
+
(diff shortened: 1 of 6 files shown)


.changeset/happy-tigers-dream.md
```diff
@@ -0,0 +1,10 @@
+---
+'@coveord/plasma-mantine': patch
+---
+
+Fix `CodeEditor` crashing when hidden and shown again by React's `Activity`
+
+When a `CodeEditor` was placed inside a component that hides inactive content with React's `Activity`
+(such as `Tabs`, `Collapse`, or `Accordion` with `keepMountedMode="activity"`), switching away and back
+disposed the underlying Monaco editor and left it in a broken state. The editor now recreates itself when
+it is disposed while still mounted, so it keeps working across visibility changes.
```
GermainBergeron: Do we need the manual line wrapping?

```suggestion
When a `CodeEditor` was placed inside a component that hides inactive content with React's `Activity` (such as `Tabs`, `Collapse`, or `Accordion` with `keepMountedMode="activity"`), switching away and back disposed the underlying Monaco editor and left it in a broken state. The editor now recreates itself when it is disposed while still mounted, so it keeps working across visibility changes.
```

GermainBergeron (approved):

GermainBergeron (approved):

PR #4565, opened 2026-08-10
chore: add missing exports for the Table component

GermainBergeron:
### Proposed Changes

Add missing type exports for the Table component.
Add Mantine style namespace exports so consumer can leverage `.Props`, e.g., `Table.Header.Props` instead of importing `TableHeaderProps`

None, adding exports

### DIFF
--- a/packages/mantine/src/components/Table/Table.tsx
+++ b/packages/mantine/src/components/Table/Table.tsx
@@ -1,4 +1,4 @@
-import {Box, Center, Factory, Loader, useProps, useStyles} from '@mantine/core';
+import {Box, Center, Factory, Loader, type SkeletonProps, useProps, useStyles} from '@mantine/core';
 import {useClickOutside, useMergedRef} from '@mantine/hooks';
 import {
     ColumnDef,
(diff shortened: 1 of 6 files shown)


packages/mantine/src/components/Table/Table.tsx
```diff
@@ -399,3 +443,77 @@ Table.Predicate = TablePredicate;
 Table.Toolbar = TableToolbar;
 
 Table.extend = identity as CustomComponentThemeExtend<PlasmaTableFactory>;
+
+export namespace Table {
```
toofff, line 447: This is a great idea, but since it seems to be a new feature that not all components support at this time, could we add some JSDoc with the `@experimental` tag to warn that this might change in the future?

toofff (approved):

packages/mantine/src/components/Table/Table.tsx
GermainBergeron: I'm working on adding the rest of them to follow Mantine's code standard

toofff: OK, if it's added soon and in the same version of Plasma, then GO :)

toofff (approved):

PR #4568, opened 2026-08-11
feat(mantine): add namespace type alias to components

GermainBergeron:
### Proposed Changes

Add namespace exports to custom components to follow Mantine conventions

None, only new exports

### DIFF
--- a/packages/mantine/src/components/BrowserPreview/BrowserPreview.tsx
+++ b/packages/mantine/src/components/BrowserPreview/BrowserPreview.tsx
@@ -47,3 +47,7 @@ export const BrowserPreview = ({
 );
 
 BrowserPreview.displayName = 'BrowserPreview';
+
+export namespace BrowserPreview {
+    export type Props = BrowserPreviewProps;
+}
--- a/packages/mantine/src/components/ChildForm/ChildForm.tsx
+++ b/packages/mantine/src/components/ChildForm/ChildForm.tsx
@@ -78,4 +78,5 @@ ChildForm.displayName = 'ChildForm';
 export namespace ChildForm {
     export type Props = ChildFormProps;
     export type StylesNames = ChildFormStylesNames;
+    export type Factory = ChildFormFactory;
 }
--- a/packages/mantine/src/components/ChildForm/ChildForm.tsx
+++ b/packages/mantine/src/components/ChildForm/ChildForm.tsx
@@ -74,3 +74,8 @@ export const ChildForm = polymorphicFactory<ChildFormFactory>((props) => {
 });
 
 ChildForm.displayName = 'ChildForm';
+
+export namespace ChildForm {
+    export type Props = ChildFormProps;
+    export type StylesNames = ChildFormStylesNames;
+}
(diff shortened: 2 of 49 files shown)


packages/mantine/src/components/BrowserPreview/BrowserPreview.tsx
```diff
@@ -47,3 +47,7 @@ export const BrowserPreview = ({
 );
 
 BrowserPreview.displayName = 'BrowserPreview';
+
+export namespace BrowserPreview {
+    export type Props = BrowserPreviewProps;
+}
```
toofff, line 53: @GermainBergeron—maybe we should create a ticket to use the Mantine factory here?

packages/mantine/src/components/ChildForm/ChildForm.tsx
```diff
@@ -74,3 +74,8 @@ export const ChildForm = polymorphicFactory<ChildFormFactory>((props) => {
 });
 
 ChildForm.displayName = 'ChildForm';
+
+export namespace ChildForm {
+    export type Props = ChildFormProps;
+    export type StylesNames = ChildFormStylesNames;
```
toofff, line 80: Shouldn't the factory be exported?

```suggestion
    export type StylesNames = ChildFormStylesNames;
    export type Factory = ChildFormFactory;
```

packages/mantine/src/components/EllipsisText/EllipsisText.tsx
```diff
@@ -86,3 +86,8 @@ export const EllipsisText = polymorphicFactory<EllipsisTextFactory>((props) => {
 });
 
 const isOverflowing = (h: HTMLDivElement) => h.scrollWidth > h.clientWidth || h.scrollHeight > h.clientHeight;
+
+export namespace EllipsisText {
+    export type Props = EllipsisTextProps;
+    export type StylesNames = EllipsisTextStylesNames;
```
toofff, line 92:
```suggestion
    export type StylesNames = EllipsisTextStylesNames;
    export type Factory = EllipsisTextFactory;
```

packages/mantine/src/components/Header/Header.tsx
```diff
@@ -165,3 +169,20 @@ Header.BreadcrumbAnchor = HeaderBreadcrumbAnchor;
 Header.BreadcrumbText = HeaderBreadcrumbText;
 Header.Right = HeaderRight;
 Header.DocAnchor = HeaderDocAnchor;
+
+export namespace Header {
+    export type Props = HeaderProps;
+    export type Factory = HeaderFactory;
```
toofff, line 179:
```suggestion
    export type StylesNames = HeaderStyleNames;
    export type Factory = HeaderFactory;
```

And you could rename `HeaderStyleNames` to `HeaderStylesNames` to be consistent with the rest.

toofff (approved):

packages/mantine/src/components/Header/Header.tsx
GermainBergeron: I did the change, but kept `HeaderStyleNames` too to avoid a breaking change

packages/mantine/src/components/BrowserPreview/BrowserPreview.tsx
GermainBergeron: I don't think we need the factory everywhere, it adds a lot of boiller plate and make our components a bit TOO flexible (which goes against what the Design system team wants)

toofff (approved):

PR #4571, opened 2026-08-11
Small change to skill.md files

cmcinnis-coveo:
### Purpose

This pull request makes small changes to the Storybook MDX conversion and component guidelines skills: 
- No using em-dashes at all. (This was already a rule in the SKILL.md files, but it was occasionally ignored by the agent in execution.)
- No semicolons except when quoting existing text. 
- The validation checklists are updated to clarify these rules, which should add consistency in the generated text from the conversion workflow.


.github/skills/storybook-component-guidelines/SKILL.md
```diff
@@ -208,13 +208,19 @@ Exceptions. Do not apply this rule inside:
...
 
-If you feel the need for an em-dash, rewrite the sentence. Use a period, a comma, or split into two sentences instead.
+When you feel the need for either, rewrite the sentence using the appropriate substitute:
 
-The only exception is when quoting or reproducing existing text that already contains an em-dash.
+- **Listing or explaining something:** use a colon. Example: "There are two options: light and dark."
+- **Separating two independent clauses:** use a period. Split into two sentences.
+- **Separating two dependent clauses:** use a comma.
+
+The only exception is when quoting or directly reproducing existing text that already contains an em-dash or semicolon.
```
GinoVin: I think we need to settle on one rule here. The PR purpose says no em dashes at all, and the punctuation/self-check sections enforce that, but this section and the Step 2 validation checklist allow them when quoting or reproducing existing text.

It looks like the Copilot autofix may have carried the semicolon exception over to em dashes. If the intent is truly “no em dashes at all,” I’d remove the exception here and from the validation checklist so the agent gets one unambiguous rule.

.github/skills/converting-md-to-storybook-mdx/SKILL.md
```diff
@@ -163,6 +163,17 @@ Do not create or modify any `.stories.tsx` file. Do not modify the original `.md
 
 ---
 
+## Writing rules for prose you generate
+
+This skill preserves existing content exactly, but you still write prose in a few places: comments, build reports, and messages to the user. Apply these rules to all prose you write.
+
+- **No em-dashes (`—`).** Not as clause separators, not with spaces, not at all. Use a colon when listing something, a period when separating independent clauses, or a coordinating conjunction in between dependent clauses.
+- **No semicolons (`;`).** Use a period instead. The only exceptions are code examples or code syntax that purposefully contains semicolons.
```
GinoVin: This no longer matches either the PR purpose or the validation checklist. Both say semicolons are allowed when quoting/reproducing existing text, while this now only makes an exception for code.

I’d keep the exception consistent across the skill and checklist, especially since these instructions need to be deterministic for the agent.

.github/skills/storybook-component-guidelines/SKILL.md
```diff
@@ -288,6 +295,11 @@ Do not use: it'll, would've, could've, should've, you'd, it'd, there'd, there'll
 - American English spelling throughout. Use "customize" not "customise", "color" not "colour", "canceled" not "cancelled", "behavior" not "behaviour", "organize" not "organise".
 - Exception: proper names always use their official spelling regardless of locale.
 
+### Numbers
+
+- Use numerals rather than spelling numbers out, including ranges and counts: "1-3 words" not "one to three words", "3 words or fewer" not "three words or fewer".
+- The only exception is naming a single number in running prose, which stays spelled out: "add one component", "two actions per area".
```
GinoVin: I’m not sure an agent can apply this rule consistently as written. We say counts use numerals, but “two actions per area” is also a count, while the self-check uses “4 states” as a numeral example.

Could we define the actual boundary instead of “a single number”? Right now the examples allow either treatment for essentially the same type of case.

.github/skills/storybook-component-guidelines/SKILL.md
```diff
@@ -288,6 +295,11 @@ Do not use: it'll, would've, could've, should've, you'd, it'd, there'd, there'll
 - American English spelling throughout. Use "customize" not "customise", "color" not "colour", "canceled" not "cancelled", "behavior" not "behaviour", "organize" not "organise".
 - Exception: proper names always use their official spelling regardless of locale.
 
+### Numbers
```
GinoVin, line 277: Since we identify packages/llms/src/content/ above as the source of truth for writing rules, should this live in WritingMechanics.md instead?

Adding it only to these Storybook skills gives us two places defining writing behavior and makes it easier for the rules to drift. If this is intentionally specific to Storybook output, I think we should make that distinction explicit.

.github/skills/storybook-component-guidelines/SKILL.md
```diff
@@ -208,13 +208,19 @@ Exceptions. Do not apply this rule inside:
...
+This means no clause separators, no spaced usage (`word — word`), and no em-dashes at all.
+It also means no semicolons joining independent clauses or separating list items.
 
-If you feel the need for an em-dash, rewrite the sentence. Use a period, a comma, or split into two sentences instead.
+When you feel the need for either, rewrite the sentence using the appropriate substitute:
 
-The only exception is when quoting or reproducing existing text that already contains an em-dash.
+- **Listing or explaining something:** use a colon. Example: "There are two options: light and dark."
+- **Separating two independent clauses:** use a period. Split into two sentences.
+- **Separating two dependent clauses:** use a comma.
```
GinoVin: Could we avoid prescribing punctuation based on “dependent clauses” here? A comma isn’t necessarily the correct replacement, and the Step 1 skill gives different guidance for the same case. Something like “rewrite with a colon, period, comma, or conjunction as grammatically appropriate” would be safer and keep the two skills aligned.

GinoVin (commented):
Great work Christina, just a few comments worth checking :)

.github/skills/storybook-component-guidelines/SKILL.md
cmcinnis-coveo: > If this is intentionally specific to Storybook output, I think we should make that distinction explicit.

Yes, this is the case. I added clarification for this and tweaked all writing instructions in the skill files to align with doc writing rather than UX copy, like WritingMechanics.md is designed for.

scouture-coveo (approved):
looks good!

PR #4592, opened 2026-08-26
fix(radio): fix spacing styling between inputWrapper and radioGroup

FelixBlaisThon:
This PR handles the spacing needed for the Radio.Group component to reflect figma mock ups. No need to explicitly set it when developping.

### DIFF
--- a/packages/mantine/src/styles/Radio.module.css
+++ b/packages/mantine/src/styles/Radio.module.css
@@ -35,3 +35,8 @@
         }
     }
 }
+
+div[role='radiogroup'] {
+    padding-top: var(--mantine-spacing-xs);
+    padding-bottom: var(--mantine-spacing-xs);
+}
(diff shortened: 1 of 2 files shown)


packages/mantine/src/styles/Radio.module.css
```diff
@@ -35,3 +35,8 @@
         }
     }
 }
+
+div[role='radiogroup'] {
+    padding-top: var(--mantine-spacing-xs);
+    padding-bottom: var(--mantine-spacing-xs);
```
gdostie, line 41: Should we align on the design specs or let Sam know so he can change the spec from xxs to xs

Image

gdostie (approved):

packages/mantine/src/styles/Radio.module.css
FelixBlaisThon: explained with our mouths and ears

gdostie: 👄👂

PR #4611, opened 2026-09-18
Fix CodeEditor find widget buttons unclickable due to tooltip flickering

gdostie:
### Proposed Changes

Fixes [ADUI-11560](https://coveord.atlassian.net/browse/ADUI-11560): opening the `CodeEditor` find widget (search button / Cmd+F) left its close (X) button and adjacent buttons unclickable, with the tooltip visibly flickering (reproduced in Storybook).

**Root cause (DOM-confirmed).** The find widget's buttons show a pointer-style hover tooltip that monaco renders in a `.workbench-hover-container` on `document.body`. The tooltip repositions on layout, fires a `mouseout` on the button it describes, disposes, then re-triggers when the pointer is "back on" the button, in a rapid add/remove loop (a MutationObserver showed ~12 add/remove cycles while hovering the X button). The loop swallows the click. This is the upstream bug [microsoft/monaco-editor#5208](https://github.com/microsoft/monaco-editor/issues/5208) / [#5139](https://github.com/microsoft/monaco-editor/issues/5139).

**Fix.** A scoped global CSS rule hides only the pointer-style hover container (`.workbench-hover-container:has(.workbench-hover-pointer)`), stopping the loop while leaving normal editor content hovers (code symbols, markers) working. The container is appended to `document.body`, outside the component subtree, so the rule must be global. Buttons keep their `aria-label`/`title`, so functionality and accessibility are preserved.

**Why this is a library fix, not a Storybook-only one.** The flicker was verified to be independent of the things that would make it Storybook-specific: it reproduces on monaco **0.55.1 and 0.56.0**, and with **both** the `cdn` and `local` loaders, and it was not resolved by changing the demo's layout/size. It is a latent bug in the shipped `CodeEditor` that any consumer can hit when using the find widget; admin-ui happens not to trigger it today, but scoping the fix to the Storybook demo would leave the bug live for real consumers. Fixing it in `@coveord/plasma-mantine` protects everyone.

_Note: upgrading `monaco-editor` (0.55.1 → 0.56.0) was investigated and confirmed **not** to fix this, so no version bump is included._

None. Only the pointer-style widget hover tooltips (find widget buttons) are suppressed where the `CodeEditor` is used; normal content hovers are unaffected and button labels remain available via `aria-label`/`title`.

### Acceptance Criteria

- [x] [README.md](https://github.com/coveo/plasma/blob/master/README.md) is adjusted to reflect the proposed changes (if relevant)

<sub>Note: no unit test added — the fix targets a browser-rendered monaco hover loop not observable under jsdom (monaco is mocked in tests). Verified manually in local Storybook; content hovers confirmed still working.</sub>

[ADUI-11560]: https://coveord.atlassian.net/browse/ADUI-11560?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


GermainBergeron:
If the issue is only in the storybook can we scope the fix to the storybook?

FelixBlaisThon (approved):

gdostie:
> If the issue is only in the storybook can we scope the fix to the storybook?

Investigated thoroughly and added an summary of my explorations in the PR description, but tl;dr: the issue is likely not only in storybook

GermainBergeron (approved):

PR #4612, opened 2026-09-18
Fix Table row opacity when enableRowSelection is false

gdostie:
### Proposed Changes

When `enableRowSelection` was set to the boolean `false`, every row rendered at 50% opacity. That dimming is only meant for rows rejected by an `enableRowSelection` predicate.

The dimming was keyed off `data-selectable='false'`, but `data-selectable` reflects `row.getCanSelect()`, which is `false` whenever selection is disabled entirely (boolean `false`), during loading, or when a predicate rejects the row. Only the predicate case should dim.

The fix decouples dimming from `data-selectable`. Both layouts already computed `isRowSelectionRejected` via the existing `isRowSelectionPredicateRejected` helper (true only when `enableRowSelection` is a function that rejects the row). It is now exposed as a new `data-selection-rejected` attribute, and the `opacity: 0.5` rule targets `[data-selection-rejected='true']` in both `RowLayout` and `CardLayout`. `data-selectable` continues to drive cursor/hover/checkbox styling.

Tests were added in both layout specs covering the boolean-`false` case (rows not marked rejected) and the predicate case (rejected rows marked). A patch changeset for `@coveord/plasma-mantine` is included.

None. The predicate behavior and public API are unchanged; only the boolean-`false` dimming regression is fixed.

### Acceptance Criteria


FelixBlaisThon (approved):
demo tested

